PR #11, opened 2018-02-17
Add support for subscriptions

vektah:
```graphql
type Subscription {
    messageAdded(roomName: String!): Message!
}

```
will generate resolver methods that return a channel:
```
Subscription_messageAdded(ctx context.Context, roomName string) (<-chan Message, error)
```
these resolvers are expected to stop returning results when context is done.

I've also included an example apollo app that uses subscriptions:
subscriptions

Closes [Subscriptions](https://github.com/vektah/gqlgen/issues/7)


DenisNeustroev:
You are gem! Thank you for your amazing work! 🎉

PR #45, opened 2018-03-10
Retain orignal resolver error and support overriding error message

### LINKED ISSUE #38: ResolverError field in QueryError not set and public errors in general
I see that the `ResolverError` field on `QueryError` is not set. What is the plan for this? The `Builder` for errors currently only takes the error message and creates a `QueryError` from that.

I was a bit surprised to see that by default the generated execution context drops the original error and exposes the raw message error in the GraphQL response. What are your thoughts for this?

I can imagine one wants to be able to control how errors thrown from the resolvers are rendered. I'm using a custom written HTTP handler and would like to be able to log the original errors. Do you  think it's better to do logging etc. of the errors inside the resolvers and return a custom 'public' error?

Thanks for the awesome lib! Please let us know if there are any areas where you could use some help!

dvic:
Fixes #38.


test/resolvers_test.go
```diff
@@ -2,6 +2,103 @@
...
+func TestErrorConverter(t *testing.T) {
+	t.Run("with", func(t *testing.T) {
+		testConvErr := func(e error) string {
+			if _, ok := errors.Cause(e).(*specialErr); ok {
+				return "override special error message"
+			}
+			return e.Error()
+		}
+		{
+			// return special error
```
vektah: rather then use a comment and anonymous blocks, you can nest t.Runs, eg
```go
t.Run("return special error", func(t *testing.T) {
```

neelance/errors/errors.go
```diff
@@ -42,12 +51,23 @@ var _ error = &QueryError{}
...
 }
 
 func (c *Builder) Errorf(format string, args ...interface{}) {
 	c.Errors = append(c.Errors, Errorf(format, args...))
 }
 
 func (c *Builder) Error(err error) {
-	c.Errors = append(c.Errors, Errorf("%s", err.Error()))
+	var gqlErrMessage string
+	if c.ErrorMessageFn != nil {
```
vektah: you could eliminate this branch by setting the default error message function to `func(err error)  string { return err.Error() }`

codegen/templates/generated.gotpl
```diff
@@ -8,8 +8,12 @@ import (
 {{ end }}
 )
 
-func MakeExecutableSchema(resolvers Resolvers) graphql.ExecutableSchema {
-	return &executableSchema{resolvers}
+func MakeExecutableSchema(resolvers Resolvers, opts ...ExecutableOption) graphql.ExecutableSchema {
```
vektah: :+1:

dvic:
I've updated the branch by addressing your comments, haven't gotten to splitting the errors into `ResolverError` and `QueryError` though.

PR #75, opened 2018-04-04
Replace Invalid Characters in Package Name with an Underscore

mathewbyrne:
Came across this while trying to map a schema type to an import from a package with a dash in the name.  gqlgen will currently use the last part of the import path as the local import identifier.  This leads to a syntax error in the generated code.

This PR ensures a valid identifier by sanitising local import names to a valid go identifier by replacing any non-word characters with an underscore.

### DIFF
--- a/codegen/import_build.go
+++ b/codegen/import_build.go
@@ -2,6 +2,7 @@ package codegen
 
 import (
 	"path/filepath"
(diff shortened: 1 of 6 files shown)


vektah:
Thanks for the PR :ok_hand:

PR #94, opened 2018-04-26
updated custom error func

### LINKED ISSUE #90: Complex Errors
How would we represent complex error types using this library. We currently only have a single error message for each error found performing a request. However how can a UI component utilise this effectively. For example say we have a query which requires some kind of authorisation. How can we return this in the UI. More importantly how we give some context of the type of error? i.e in the authorisation problem, the UI should react differently to session timeout errors over not authorised to access a resource. So I think that:
```json
{ "data": null, "errors": [ { "message": "token expired" } ] }
```
is not enough context or detail to be able to react appropriately as there will be a lot of string parsing. Could we not add an additional field to the errors given with some extra metadata about that error, i.e at least an error code or something more static around a particular error in the system. 

Even making the returned error an interface we can implement ourselves would be better. The apollo client `GraphQLError` returned also seems to have a field called `extraInfo` even if we could populate that with additional context.

jacklaaa89:
updated custom error func to allow for extra data to be returned and rendered as JSON. Fixes #90

### DIFF
--- a/graphql/jsonw.go
+++ b/graphql/jsonw.go
@@ -1,6 +1,7 @@
 package graphql
 
 import (
+	"encoding/json"
 	"io"
 	"strconv"
 )
@@ -79,5 +80,12 @@ func (a Array) MarshalGQL(writer io.Writer) {
 func lit(b []byte) Marshaler {
 	return WriterFunc(func(w io.Writer) {
 		w.Write(b)
+	})
+}
+
+// MarshalJSON marshals an interface into JSON.
+func MarshalJSON(i interface{}) Marshaler {
+	return WriterFunc(func(w io.Writer) {
+		json.NewEncoder(w).Encode(i)
 	})
 }
--- a/neelance/errors/errors.go
+++ b/neelance/errors/errors.go
@@ -10,6 +10,7 @@ type QueryError struct {
 	Path          []interface{} `json:"path,omitempty"`
 	Rule          string        `json:"-"`
 	ResolverError error         `json:"-"`
+	ExtraInfo     interface{}   `json:"extraInfo,omitempty"`
 }
 
 type Location struct {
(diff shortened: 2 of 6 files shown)


graphql/jsonw.go
```diff
@@ -81,3 +82,10 @@ func lit(b []byte) Marshaler {
 		w.Write(b)
 	})
 }
+
+// MarshalJSON marshals an interface into JSON.
+func MarshalJSON(i interface{}) Marshaler {
+	return WriterFunc(func(w io.Writer) {
+		json.NewEncoder(w).Encode(i)
```
vektah: There's an error being ignored here, I wonder why gometalinter isn't complaining.

neelance/errors/errors.go
```diff
@@ -49,25 +51,35 @@ func (err *QueryError) Error() string {
...
 
+// FormattedError a formatted error which includes the error message and extra information
+// which is JSON encoded and passed with the error.
+type FormattedError struct {
+	Message string
+	Extra   interface{}
+}
+
+// ErrorMessageFunc a func which given an error returns a formatted error.
+type ErrorMessageFunc func(err error) FormattedError
```
vektah: I don't think this is quite right. It only lets the user change whats in this "extra" key.

Maybe it could return a graphql.Marshaler?

neelance/errors/errors.go
```diff
@@ -29,10 +30,11 @@ func Errorf(format string, a ...interface{}) *QueryError {
 
 // WithMessagef is the same as Errorf, except it will store the err inside
 // the ResolverError field.
-func WithMessagef(err error, format string, a ...interface{}) *QueryError {
+func WithMessagef(err error, extra interface{}, format string, a ...interface{}) *QueryError {
```
vektah: This is the worst kind of BC break. 

`WithMessagef(err, "hello %s", name)` will continue to compile but the format string will dumped into extra and the name into the format string.

test/resolvers_test.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/vektah/gqlgen/test/models"
 )
 
+type fErr = gqlerrors.FormattedError
```
vektah: I'm not a fan of abusing type aliases like this.

PR #104, opened 2018-05-03
Add a NewRequestContext method

vektah:
Hows this look @jon-walton?

Fixes https://github.com/vektah/gqlgen/issues/102


jon-walton:
perfect, that'll work fine with me if you're happy with it

PR #105, opened 2018-05-03
Automatically add a _ suffix to reserved words

### LINKED ISSUE #101: Generated code ends up being invalid when using keywords as parameters
### Expected Behaviour

Running `gqlgen -schema schema.graphql` (below) should correctly generate models and scaffolding to use this library

### Actual Behavior

Right after generating code, `gofmt` chokes pretty hard on the generated code, specifically this line:

```
Mutation_tokenExchange(ctx context.Context, token string, type *TokenType) (*Token, error)
```

The error here is `type *TokenType` not being valid go, throwing `generated.go:<line>:<col>: expected 'IDENT', found 'type'`.

Is there any way defined for dealing with top-level go keywords, aside from "don't use them"?

```
type Mutation {
  tokenExchange(
    # The token to exchange.
    token: String!

    # The type of token provided.
    type: TokenType
  ): Token
}

# An auth token
type Token {
  # The token ID
  id: String!
}

# A token type
enum TokenType {
  FOO
  BAR
}
```

vektah:
Hows this look @ameliaikeda?

Fixes #101

### DIFF
--- a/codegen/input_test.go
+++ b/codegen/input_test.go
@@ -90,3 +90,76 @@ func TestComplexInputTypes(t *testing.T) {
 
 	require.NoError(t, err)
 }
+
+func TestKeywordInputFields(t *testing.T) {
+	err := generate("input_keywords_fields", `
+		input Object {
+			break:       String!
+			default:     String!
+			func:        String!
+			interface:   String!
+			select:      String!
+			case:        String!
+			defer:       String!
+			go:          String!
+			map:         String!
+			struct:      String!
+			chan:        String!
+			else:        String!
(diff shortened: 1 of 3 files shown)


vektah:
the neelance lib is only used for parsing the schema and parsing and validating the request. The entire execution environment is in the generated code.

eg:
https://github.com/vektah/gqlgen/blob/26ac13ffec364baf5d6db0a3d6bb613c5fba25ea/example/scalars/generated.go#L121-L149

ameliaikeda:
Nice, thanks!

PR #108, opened 2018-05-08
generate json tag to model field  by gql name.

imiskolee:
1. add json tag in models_gen.go
2. use gqlname to model filed json tag.

why we need it？

use the json tag, we can easy to `go test` used the generated struct,and don't need write origin graphQL


imiskolee:
@vektah  i can't understand the CI error.

vektah:
you need to run `go generate ./...`

vektah:
Thanks for the PR :+1:

Issue #110, opened 2018-05-09
[bug] can't accept null to array input type(nil pointer panic)


imiskolee:
define a simple input struct

```graphql
input A {
  slices:[Int]
}
mutation {
  create(input:A)
}
```

its will panic of nil pointer.

```graphql
mutation {
  A(slices:null) {slices}
}
```

question 1:

1. Did graphQL need support null -> array ?

PR #111, opened 2018-05-09
1. fix bug: #110

imiskolee:


imiskolee:
@vektah  updated commit, can you merge it ?

PR #120, opened 2018-05-22
Fixing panic when non array value is passed to array type

andrewmunro:
Encountered a panic similar to issue https://github.com/vektah/gqlgen/issues/110 when passing an object to an array type.

### DIFF
--- a/codegen/type.go
+++ b/codegen/type.go
@@ -107,7 +107,9 @@ func (t Type) unmarshal(result, raw string, remainingMods []string, depth int) s
 
 		return tpl(`var {{.rawSlice}} []interface{}
 			if {{.raw}} != nil {
-				{{.rawSlice}} = {{.raw}}.([]interface{})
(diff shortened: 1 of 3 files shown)


imiskolee:
@andrewmunro

 can you share an example?

vektah:
Thabks for the pr, a test would be the best way to give an example :)

andrewmunro:
👋 Added a test to the dataloader example to demonstrate this. 

Essentially if you pass an object instead of an array in your query/mutation, gqlgen panics. Handling is not ideal as the field is now ignored with no errors, I imagine this will be fixed properly when @vektah introduces the new parser 🎉! 

For now I hope this is sufficient to prevent app's from panic-ing with malformed input, without having some kind of middleware to recover the application...

vektah:
Yeah that definitely should be caught by validation and give a good message to the user, but the runtime shouldn't panic.

:+1:

PR #125, opened 2018-06-05
Subscriptions Enhancements

wyattjoh:
# What does this do?

This PR adds some features to ensure better support and parity with the existing websocket subscriptions proposal found in:

https://github.com/apollographql/subscriptions-transport-ws/blob/e1b10f9d94cb9c0736a8d67585ae315427b024c7/PROTOCOL.md

Of course, open to feedback, loving the library so far!

### WebsocketOnConnectMiddleware

New `WebsocketOnConnectMiddleware` that lets you handle when the websocket connection is initialized, and you want to inspect the `connectionParams` to provide authentication into the context. Default does nothing, but can be enhanced using the new `handler.WebsocketOnConnectMiddleware(middleware graphql.OnConnectMiddleware) Option` `handler.Option`.

### WebsocketOnOperationMiddleware

New `WebsocketOnOperationMiddleware` that lets you inject new context parameters per graphql resolve operation. This is _critical_ when operating with a DataLoader, as the cache should not be maintained for the duration of the subscription request as it is now, and instead should be flushed every time you have a new subscription result to resolve. This ensures that when you have a situation:

```graphql
subscription {
  commentCreated {
    id
    body
    author {
      id
      username
  }
}
```

When a new `comment` is created, and the author's `username` changes in-between you getting the initial subscription request and the comment, that you get the updated `username` instead of the cached value.

This also ensures that every time that you may do something like copying a Redis connection, that you can release resources when no activity is occurring.

## Keep Alive

Added initial implementation of a keep-alive timer that will ensure the connection won't be killed. Default disabled, but can be enabled with the `handler.Option` function `handler.WebsocketKeepAliveDuration(duration time.Duration) Option`.


vektah:
Its good to see subscriptions getting some love. I'm not thrilled with doubling the surface area of these hooks but I can see the use case for both, and the OnConnect one is limited to the handler namespace so that's a plus.

Should `OnOperation` be called from the non streaming operations too?

wyattjoh:
It is certainly a concern, but without it, you can't perform things like authentication over websocket via the onConnect payload, or generate operation specific dataloaders.

I'd expect that the websocket style requests are the only ones that require that the `onOperation` occurs in, as that would give a good point to inject for a "on request" middleware as you have it for HTTP/1.1 style graph calls now.

handler/websocket.go
```diff
@@ -57,56 +58,117 @@ func connectWs(exec graphql.ExecutableSchema, w http.ResponseWriter, r *http.Req
...
+			return
+		}
+	}
+
+	next := func(ctx context.Context, params map[string]interface{}) error {
+		return conn.run(ctx)
+	}
+
+	if err := cfg.onConnectHook(r.Context(), connectionParams, next); err != nil {
+		// TODO: handle the error somehow?
```
vektah: please handle this, you can see an example a few lines above of returning an error to the client.

vektah:
I'm closing this as its massively out of date.

If you want to reopen it, please split break up the two different changes in here (keepalives vs hooks)

wyattjoh:
Apologies!! Got buried in my notifications :( I'll look into a separate PR to cover things!

PR #127, opened 2018-06-06
convert windows input path separators to slash

jon-walton:
I found recently that generating from windows was causing circular imports within `generated.go`. looks like we had another case of windows file path issues.

this feels a bit hacky to me but thought i would start the discussion.

windows is able to use \ or / as a path separator. I'm not aware of any other (modern and supported ;) ) OS using \

one use case we might come across is the source files being on a UNC share which might be due to people using virtualbox etc and sharing the host's filesystem; not sure how likely that is, but anyway


vektah:
Cheers :+1:

PR #130, opened 2018-06-11
Added support for @include/@skip directives.

### LINKED ISSUE #126: @skip directive is not working
```
{
user @skip(if: true){ id }
}
```

### Expected Behaviour
It will not display user data in result
### Actual Behavior
It shows user data in result

sergeyvlakh:
# Overview
Resolves #126
Added support for builtin `@skip`/`@include` directives. They only work with `$if` argument (!).
This supports syntax with inline variables and default ones.

**_It is very-very-very WIP PR. I'm really looking into ways of improving this code to actually bring in support for directives!_**
## Default variables case
**Query**:
```graphql
query($hero:Boolean = true){
  hero @include (if:$hero) {
    name
  }
}
```
**Result**:
```graphql
{
  "data": {}
}
```
As expected.
## Inline variables case
**Query**:
```graphql
{
  hero @include (if:false) {
    name
  }
}
```
**Result**:
```graphql
{
  "data": {}
}
```
As expected.
## Passed variables with default values
**Query**:
```graphql
query($hero:Boolean = false){
  hero @include (if:$hero) {
    name
  }
}
```
**Variables**:
```graphql
{
"hero": true
}
```
**Result**:
```graphql
{
  "data": {
    "hero": {
      "name": "R2-D2"
    }
  }
}
```
As expected.


vektah:
Neat! 

I think I'm happy to add a hacky solution temporarily until the new parser lands. Once thats in we will need a more generic user pluggable way to define directives, and there will need to be a bit of planning around that to support various use cases (authentication, api fetch resolvers, caching).

graphql/exec.go
```diff
@@ -21,12 +21,75 @@ func CollectFields(doc *query.Document, selSet []query.Selection, satisfies []st
...
+	}
+	return true
+}
+
+func runSingleDirective(directive *common.Directive, variables map[string]interface{}) bool {
+	arg := directive.Args.MustGet("if")
+	fieldAllowed := true
+	switch arg.(type) {
+	case *common.Variable:
+		{
```
vektah: I'm surprised none of the linters fire on this, you don't need curlies here.

graphql/exec.go
```diff
@@ -21,12 +21,75 @@ func CollectFields(doc *query.Document, selSet []query.Selection, satisfies []st
...
+			return !fieldAllowed
+		}
+	case "include":
+		{
+			return fieldAllowed
+		}
+	default:
+		// Custom directives are not supported yet
+		return fieldAllowed
```
vektah: probably should `return true`. Do nothing rather then do something unexpected. Or error.

graphql/exec.go
```diff
@@ -21,12 +21,75 @@ func CollectFields(doc *query.Document, selSet []query.Selection, satisfies []st
...
+			variables[variable.Name.Name] = variable.Default.Value(nil)
+		}
+	}
+	return variables
+}
+
 func collectFields(doc *query.Document, selSet []query.Selection, satisfies []string, variables map[string]interface{}, visited map[string]bool) []CollectedField {
 	var groupedFields []CollectedField
-
+	oop, err := doc.GetOperation("")
```
vektah: :worried: 

Maybe these methods should be refactored to take the request context instead of document so they can get the correct operation

graphql/exec.go
```diff
@@ -21,12 +21,75 @@ func CollectFields(doc *query.Document, selSet []query.Selection, satisfies []st
...
+		// Custom directives are not supported yet
+		return fieldAllowed
+	}
+}
+
+func populateVariables(operation *query.Operation, variables map[string]interface{}) map[string]interface{} {
+	if variables == nil {
+		variables = make(map[string]interface{})
+	}
+	for _, variable := range operation.Vars {
```
vektah: This seems like a lot of work to do on every fieldset, I think this goes away if we pass context in?

sergeyvlakh: If there will be one specific place which will have both variables and their default values, this will be super good!

sergeyvlakh:
Thanks for addressing some comments. I will try to update PR ASAP.

sergeyvlakh:
@vektah could you please do review once again? Thanks in advance :}

graphql/exec.go
```diff
@@ -21,12 +21,66 @@ func CollectFields(doc *query.Document, selSet []query.Selection, satisfies []st
...
+	}
+	return true
+}
+
+func runSingleDirective(directive *common.Directive, variables map[string]interface{}) bool {
+	arg := directive.Args.MustGet("if")
+	fieldAllowed := true
+	switch arg.(type) {
+	case *common.Variable:
+		value := arg.Value(variables)
```
vektah: Why can't you pass the operation in here and do the variable lookup only when its actually referenced, instead of every resolver?

`populateVariables` is a big smell to me, its going to run on every resolver regardless of if there is even a directive.

graphql/exec.go
vektah: This comment sill applies, everywhere that calls CollectFields is grabbing the doc out of the RequestContext embedded in execution context, they could pretty easily be updated to do
```go
fields := graphql.CollectFields(ec.RequestContext, sel, implementors)
```
instead of
```go
fields := graphql.CollectFields(ec.Doc, sel, implementors, ec.Variables)
```

and then you have access to the right operation instead of assuming it was an anonymous operation.

graphql/exec.go
sergeyvlakh: Will fix ASAP!

sergeyvlakh:
@vektah thanks for comments! Will try to fix ASAP!

sergeyvlakh:
@vektah 
I've moved `populateVariables` to block only if directive present. Regarding your comment about request ctx I'm not sure. Should I change gotpl to achieve this one? Because we call `fields := graphql.CollectFields(ec.Doc, sel, implementors, ec.Variables)` in there.
And If it's not necessary, maybe merge this PR and land `CollectFieldsWithCtx` in another PR ?

vektah:
Yeah, the template changes are going to be required. A graphql query can have a few forms:
 - a single, unnamed query, eg `{ hello { world }`. 
 - a single, named query, eg `query Hello { hello { world } }`
 - multiple named queries, and the name of the one you want to run.

The only thing this works for is the first one, in order to get the operation name (you are using "", but thats 100% a bug). All this resolution has already been done though, when the validation happened, and it was put in context so the rest of the runtime knows whats going on.

graphql/exec.go
vektah: You can do this check inline, the one place you need the value from a variable, this function shouldn't exist.

And it really shouldn't mutate `variables`, they should be static after the initial validation pass.

sergeyvlakh:
Actually, as I understood, we are only taking one operation. And doesn't matter if it was named or not according to this code:
```golang
if operationName == "" {
		if len(d.Operations) > 1 {
			return nil, fmt.Errorf("more than one operation in query document and no operation name given")
		}
		for _, op := range d.Operations {
			return op, nil // return the one and only operation
		}
	}
```

vektah:
Thats only `if operationName == ""`

sergeyvlakh:
@vektah 
Updated. I've added `OperationName` to requestCtx. It wasn't in in previously. Now I'm calling 
`getOperationName(reqctx.OperationName)`

vektah:
The new parser is nearly done now, full directive support should only be a few weeks away.

sergeyvlakh:
Great. I'll close second one then. Thanks

Issue #82, opened 2018-04-06
Override type mismatch


JonLundy:
Not really a bug.. but more an annoyance for working with the database/sql Scanner/Valuer interfaces.

I am working with a custom type that is defined in graphql as 

```
type Thing {
  listOfEnums: [FooEnum]!
}
enum FooEnum {
   bin, bar, baz
}
```

the type generated in go is 
```
type FooEnum string
type Thing struct{
  ListOfEnums: []FooEnum
}
```

the trouble i am running into is this. The value for ListOfEnums is stored in a database as a string `bin,bar,baz`

for the Valuer interface i can't use []FooEnum as the type but need to define a list type like: 
```
type ListFooEnum []FooEnum
func (f ListFooEnum) Value() (driver.Value, error) {/* logic to convert the list to a string... */}
func (f *ListFooEnum) Scan(value interface{}) (error) {/* logic to convert the string into struct... */}

```

Currently how i am working around this is by changing the type on the model file to the []FooEnum when running gqlgen and then changing it back to ListFooEnum after. 

I understand i could probably use an intermediary step of scanning to the ListFooEnum and casting to the []FooEnum. I am wondering there would be a possibility of having an annotation in the future to set the generated model value to the List version over the [] version.  

:)

vektah:
I haven't tested this, but you might be able to define the parent (Thing) yourself with the type you need, and point to it in your typemap.

The other way would be to define a [custom scalar type](https://gqlgen.com/reference/scalars/) for ListFooEnum, and your schema would become:

```
scalar CommaList
type Thing struct{
  ListOfEnums: CommaList
}
```

mastercactapus:
I've ran into this as well. Certain cases make it really painful when the field names match, but have to be exposed through graphql by a different type.

A couple examples:
- struct with an ID field that is an `int` but needs to be a `string` (for graphql)
- a field is of type `time.Time` but it must be exposed to graphql as a `string` in a certain format
- a field is of a custom enum type that needs to be "mapped" (e.g. `iota`), but the parent struct would ideally be used directly

In cases where there is a type mismatch during code generation, it would be helpful if there was an option to behave as though there were no matching field at all (thus resolving the type mismatch with use of a resolver method).

The resolver code would then be "best-effort" where possible to auto-map fields. Anything that can't be done automatically would require a method for the Resolver interface, just like if there was no name match.

[...]

PR #133, opened 2018-06-15
skip struct fields with incompatible types

mastercactapus:
This PR changes the behavior of `bindObject` to ignore fields that cannot be automatically mapped from a struct. In the case of a type mismatch, the generated resolver will now require a method for the field, as if there were no matching field on the struct to begin with. This allows the logic for mapping field values to be provided as-needed, making cases like #82 hopefully easier to work with.

Previously this case would result in a fatal error (code gen would fail with a type mismatch error).

---

Example scenario (with this change):

Schema:
```graphql
type User {
  createdAt String
  listOfEnums [FooEnum]
  name String
}
enum FooEnum {
  bin, bar, baz
}
```

Application struct (mapped with `types.json`):
```go
type ListOfEnum []FooEnum
type User struct {
  Name string
  CreatedAt time.Time
  Foo ListOfEnum
}
```

Result:
```go
type Resolvers interface {
  User_createdAt(ctx context.Context, obj *user.User) (*string, error)
  User_listOfEnums(ctx context.Context, obj *user.User) ([]FooEnum, error)
}
```

One could then simply define `User_createdAt` to map `time.Time -> *string` and `User_listOfEnums` to map `ListOfEnum -> []FooEnum`.

### DIFF
--- a/codegen/util.go
+++ b/codegen/util.go
@@ -174,6 +174,7 @@ func bindObject(t types.Type, object *Object, imports Imports) error {
 		}
 
 		if structField := findField(underlying, field.GQLName); structField != nil {
+			prevModifiers := field.Type.Modifiers
 			field.Type.Modifiers = modifiersFromGoType(structField.Type())
 			field.GoVarName = structField.Name()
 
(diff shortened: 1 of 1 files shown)


vektah:
I think this makes sense, if gqlgen cant map directly between two types asking the user to implement a resolver seems like the correct thing to do.

Issue #106, opened 2018-05-04
Make generated code pass golint


flicaflow:
golint is not very happy with the generated code. Main offenders are missing function and type commends and method names with underscores. The first problem is contained in generated code and might be ok since users are not supposed look at it (it stil clutters the output of a go lint run).However the second is bad since users have to implement the resolver interface. I care about this because my editor screams at me and I cannot do anything about it.

I think it would be good style to follow the established code conventions. Adding code comments should be an easy fix. Changing the naming scheme of the resolver methods is not so simple since it would break resolver implementations. However I would say better now then later.

Btw. Thanks for the wonderful work!

vektah:
The reason the underscores are there is to split the type from the method in a way that isn't likely to be used normally. Foo.barBaz -> Foo_barBaz. Without a separator Foo.barBaz and FooBar.baz are ambiguous.

The go lint rule is to stop people using snake case, which I appreciate, but this is camel case with underscores as a separator. I don't think the limiting rule should apply. Gomatalinter lets you add comments that opt out of rules, but go lint by itself doesn't. 

It was a considered decision, How do others feel about this?

flicaflow:
I get your point and I don't have a better solution to clarify the ambiguity right now. Still I think it would be a big plus if the generated code would be 100% standard tooling compliant. There might be projects out there which enforce a build/test/lint pre-commit hook. And the nice thing about go is that there is just one default linter (looking at you ESLint). 

However I can certainly life with this..

Cheers

rodcorsi:
I think a good start point for this topic is golint pass at least in client resolvers.
Change the resolver to:

```go
type Mutation interface {
	createTodo(ctx context.Context, text string) (Todo, error)
}

type Query interface {
	todos(ctx context.Context) ([]Todo, error)
}

type Resolvers interface {
	Mutation() Mutation
	Query() Query
}
```

everdev:
What about adding a "gen" or "gql" prefix like:
```go
func (a *App) GenQueryUser(ctx context.Context, id string) (*user.User, error) {
  ...
}
func (a *App) GenUserEmail(ctx context.Context, u *user.User) (string, error) {
  ...
}
func (a *App) GenMutationSetEmail(ctx context.Context, id, email string) (*user.User, error) {
  ...
}
```
or
```go
func (a *App) GqlQueryUser(ctx context.Context, id string) (*user.User, error) {
  ...
}
[...]

PR #134, opened 2018-06-15
add lint-friendly small interfaces option for resolvers

mastercactapus:
This PR adds a `ShortResolvers` interface to `generated.go` allowing lint-friendly resolver methods. There is also a utility method that can take a `ShortResolvers` and return a `Resolvers` interface.

The new interface is a possible solution to the linting issues raised in #106. It provides a lint-friendly path to implementing resolvers via smaller type-specific interfaces for each object type.

The implementation defines methods that return resolvers for each object type from the base. Each type (if it has resolver fields) additionally has a `TypeNameResolver` interface where `TypeName` would be the object type (e.g. `Query`).

No changes were made to the existing interface/implementation. To use the new one, a `FromShort` method will map a `ShortResolvers` to the existing `Resolvers` interface. That way the existing `Resolvers` interface remains the only necessary input type. This also makes it opt-in without breaking backwards compatibility, or adding additional maintenance to the existing code.

The "smaller interfaces" example from https://github.com/vektah/gqlgen/issues/106#issuecomment-394782033 could then be used as follows:

```go
// graph is the package name where `generated.go` lives
graph.MakeExecutableSchema(graph.FromShort(&App{}))
```

`FromShort` takes in the `App` (which fulfills the small interfaces needed by `ShortResolvers`) and returns the present-day `Resolvers` interface.

----

## Generated Code Example

Existing interface (unchanged):
```go
type Resolvers interface {
	Query_user(ctx context.Context, id string) (*user.User, error)

	User_avatarURL(ctx context.Context, obj *user.User) (*string, error)
	User_contactMethods(ctx context.Context, obj *user.User) ([]contactmethod.ContactMethod, error)
}
```

Net new code (example) in `generated.go`
```go
type ShortResolvers interface {
	Query() QueryResolver
	User() UserResolver
}
type QueryResolver interface {
	User(ctx context.Context, id string) (*user.User, error)
}
type UserResolver interface {
	AvatarURL(ctx context.Context, obj *user.User) (*string, error)
	ContactMethods(ctx context.Context, obj *user.User) ([]contactmethod.ContactMethod, error)
}

func FromShort(r ShortResolvers) Resolvers { return shortMapper{r: r} }

type shortMapper struct {
	r ShortResolvers
}

func (s shortMapper) Query_user(ctx context.Context, id string) (*user.User, error) {
	return s.r.Query().User(ctx, id)
}

func (s shortMapper) User_avatarURL(ctx context.Context, obj *user.User) (*string, error) {
	return s.r.User().AvatarURL(ctx, obj)
}

func (s shortMapper) User_contactMethods(ctx context.Context, obj *user.User) ([]contactmethod.ContactMethod, error) {
	return s.r.User().ContactMethods(ctx, obj)
}
```
----

I'm not particularly attached to the name(s), just tried to keep it distinct and descriptive and picked something I could come up with in a reasonable amount of time. Suggestions welcome :)

*EDIT:* Switched from `ShortResolver` to `ShortResolvers` to prevent possible collision with an object named `Short`


mastercactapus:
@vektah What do you think of this approach overall, worth pursuing?

Also, I kept the implementation simple (just upper-cased the first letter of the field resolver) but I didn't see the `ToCamel` helper when I did -- that might be a better choice for mapping the names.

mastercactapus:
Fair point. If it is introduced as a breaking change we should make sure it plays well with dep and such so things don't break for people updating though. I think we should be careful/thoughtful of the migration path, so as to not hurt adoption and set precedence for future changes as they may happen.

And as far as that goes (and thus the changes to user code) I'm thinking:
- Offer a `go fix` equivalent that will refactor existing code (would require some research, but maybe the most painless for users)
- Deprecate `MakeExecutableSchema`, make a new method, and offer a way to generate a "shim" (inverse of this PR, old interface in - new interface out), projects can upgrade on their time
- Release as a new major version, and if no tooling, offer suggestions on how to restructure/refactor  existing code

Maybe we can try looking at usage around github. If people have a single struct with all methods, for example, something not too unlike what's in this PR would work. If there are projects that break the resolver into pieces, using embedded structs to put them together, it's a little more complex, but still possible.

It could be a fun problem to solve (not sarcastic) :)

eric-am:
Fwiw, I think this looks kind of exciting.

I recently migrated a pretty good sized project from the neelance libraries
to this one.  (It was a really excellent change and lots of people were
very happy :))  It was also a project that had several different teams
operating in the code base, which made for an interesting example.

The main challenge in the migration process was trying to keep things
incremental.  I ended up basically making a big dang struct with a bunch of
`panic("TODO")` methods to get the compiler to hush enough to let me focus
on getting other pieces in place.  Smaller interfaces seem like they
might've been a helpful guiderail.

We also have several different packages (roughly aligning with the
different teams, per Conway's Law) in this project.  And giving each
package all of its own interfaces would neat.  (We have something like the
embedded structs situation already, but making the connection between the
generated Resolver interfaces and those structs still sounds like it would
be an improvement in clarity.)

(P.S. Hi Nate, fancy running into you here!  :wave:)

vektah:
I think breaks are fine while gqlgen is still 0.x provided that:
 - they fail at compile time, eg dont swap the meaning of args with the same types.
 - there is a good description of the change + migration notes in the release notes

> Deprecate MakeExecutableSchema, make a new method, and offer a way to generate a "shim" (inverse of this PR, old interface in - new interface out), projects can upgrade on their time

This sounds good, lets land this so people can start opting into the new behaviour and work on the rest as a separate piece.

> Offer a go fix equivalent that will refactor existing code (would require some research, but maybe the most painless for users)

First thought is this is probably overkill, but it does feed really well into https://github.com/vektah/gqlgen/issues/9.

There are a few codemods that would make working with gqlgen nicer:
 - new project = generate all resolver stubs
 - modified schema = add stubs / remove resolvers
 - breaking changes = convert resolvers between formats

mastercactapus:
>This sounds good, lets land this so people can start opting into the new behaviour and work on the rest as a separate piece.

@vektah Just to make sure I understand: for this PR:
-  I'll drop the `WithShort` wrapper function
- add a new function (How about `NewExecutableSchema(ShortResolvers) gqlgen.ExecutableSchema` ?)
- from there, for now, call `MakeExecutableSchema` with a private wrapper  (inverse of `WithShort`) .

Sound right?

Also:
- Any names we should change? (e.g. `SchemaResolvers` instead of `ShortResolvers` or something?)
- Should I add a deprecation comment to `MakeExecutableSchema`, or wait?

vektah:
> add a new function (How about NewExecutableSchema(ShortResolvers) gqlgen.ExecutableSchema ?)
> from there, for now, call MakeExecutableSchema with a private wrapper (make WithShort private) .

:ok_hand: 

> Any names we should change? (e.g. SchemaResolvers instead of ShortResolvers or something?)

`SchemaResolvers` sounds like something that returns schema. What about `ResolverRoot`? It's not a resolver itself, its an interface returning each of the resolvers?

> Should I add a deprecation comment to MakeExecutableSchema, or wait?
Yeah, lets mark it as deprecated and maybe a link pointing back here or to the issue.

There are some doc updates here too, I wonder if they should land as part of this PR or in a follow up?

mastercactapus:
Updated, and went with `ResolverRoot`. I'll play with it a bit through the day today and make sure nothing else crops up. I'll post findings by EOD.

For the doc updates, a case could be made either way. One reason for follow up work would be to give some time to determine what the "recommended" structure/conventions should be with the new interface.

vektah:
> One reason for follow up work would be to give some time to determine what the "recommended" structure/conventions should be with the new interface.

Yeah, I think that makes sense. Lets leave it a bit and give people a chance to try it.

It probably makes sense to do the deprecation/docs/examples all together. Lets pull the deprecation out for now so this can land cleanly on its own.

example/chat/chat_test.go
```diff
@@ -13,40 +12,70 @@ import (
...
 
 		sub := c.Websocket(`subscription { messageAdded(roomName:"#gophers") { text createdBy } }`)
 		defer sub.Close()
 
-		wg.Done()
-		var resp struct {
+		postErrCh := make(chan error)
+		go func() {
+			var resp interface{}
+			// can't call t.Fatal from separate goroutine, so we return to the err chan for later
```
vektah: You can't call fatal (because it relies on panic handlers...) but you can call t.Error and friends (assert.Equal, instead of require.Equal).

I think this can go back to the way it was, except using `go func`, and `assert` instead of `t.Run` and `require.`

mastercactapus: Ah, I think that comment might be a little misleading (see: oops forgot to update).

The issue we saw locally (after moving out of `t.Run`) was that `sub.Next` would hang indefinitely (or until timeout) if the mutation call failed (we did a negative test by making the mutation invalid to see) because it never gets any message.

The real reason it's there (and what that comment probably *should* say) is so we can use `select` to wait for either `sub.Next` *or* the `Post` call to fail and then exit to avoid the deadlock.

jayp:
Nice one @mastercactapus

PR #136, opened 2018-06-17
fix GOPATH case mismatch issue

tianhai82:
fixes https://github.com/vektah/gqlgen/issues/135


vektah:
Cheers, I don't do any testing on windows. I wonder if there any of the free CI hosts offer windows as a build target?

tianhai82:
I doubt so, but because of this issue, gqlgen fails on my windows machine.

PR #146, opened 2018-06-22
fix template error on generated defaults

### LINKED ISSUE #144: map literal missing comma for query/edge attributes
I've been trying out gqlgen with our project and been running into a small issue with the generated code. Included a stripped down version of our schema that will cause the error.

### Expected Behaviour
Generate: `var tmp interface{} = map[string]interface{}{"page":0, "size":25}`

### Actual Behavior
Generate: `var tmp interface{} = map[string]interface{}{"page":0"size":25}`

graphql.schema
```graphql
schema {
  query: Query
}

type Query {
  formGroups(pagination: Pagination = {page: 0, size: 25}): FormGroupConnection
}

type FormGroup {
  id: ID!
  forms(pagination: Pagination = {page: 0, size: 25}): FormConnection
}

type FormGroupConnection {
  pages: Int!
}

type FormConnection {
  pages: Int!
}

input Pagination {
  page: Int!
  size: Int!
}
```

JonLundy:
See: https://github.com/vektah/gqlgen/issues/143#issuecomment-399567902

The generated code doesn't separate default values with a comma. This adds a comma.

before: 
```
		var tmp interface{} = map[string]interface{}{"offset": 0"limit": 25}
```

after:
```
		var tmp interface{} = map[string]interface{}{"offset": 0, "limit": 25}
```

fixes #144


vektah:
Thanks, you're cleaning the issues :)

This could use a test, maybe just a second arg in https://github.com/vektah/gqlgen/blob/master/example/scalars/schema.graphql#L3?

example/scalars/generated.go
```diff
@@ -203,7 +203,7 @@ func (ec *executionContext) _Query_search(ctx context.Context, field graphql.Col
 			return graphql.Null
 		}
 	} else {
-		var tmp interface{} = map[string]interface{}{"location": "37,144", "isBanned": false}
+		var tmp interface{} = map[string]interface{}{"isBanned": false, "location": "37,144"}
```
vektah: Oh-oh! I think the map is making this non deterministic. Probably needs a sort first?

vektah:
Nice work

Issue #143, opened 2018-06-21
Generated code imports own package (import cycle)


dabfleming:
There may be a way to avoid this using the `package` and `modelpackage` flags, but it isn't obvious to me.

My situation is I'm generating in the `foo/bar/model` package, and types.json includes `"Form": "foo/bar/model.Form"`

### Expected Behaviour

Generator recognizes this is the destination package, does not attempt import.

### Actual Behavior

generated.go includes import line: ` "foo/bar/model"` resulting in import cycle.

z0mbie42:
a simple schema to reproduce the bug:
```graphql
input Filter {
  offset: Int!
  limit: Int!
}

type Todo {
  id: String!
  text: String!
  author: User!
  events(filter: Filter = {offset: 0, limit: 25}): [Event!]!
}
```

JonLundy:
@z0mbie42 i filled in some of the missing items i see a interesting bug with Filter. 

```go
input Filter {
  offset: Int!
  limit: Int!
}

type Todo {
  id: String!
  text: String!
  author: User!
  # the default value cause a bug
  events(filter: Filter = {offset: 0, limit: 25}): [Event!]!
}

type User {
  name: String!
}

type Event {
  name: String!
}

type Query {
  todos: [Todo!]!
} 
```
generates the following:
```
			var tmp interface{} = map[string]interface{}{"offset":0"limit":25}
```

there is a missing comma between the offset and limit.

vektah:
Thanks for the defaults fix.

The import loop is less obvious, most of the examples work the way you're describing. eg https://github.com/vektah/gqlgen/blob/master/example/scalars/types.json#L2

I wonder if this is another case sensitivity issue, what OS are you running? Have you tried since https://github.com/vektah/gqlgen/pull/136?

A more complete example would be handy here.

z0mbie42:
The simpler way to reproduce the error is using the following project:
https://github.com/z0mbie42/goes/blob/master/_examples/api
 and replace https://github.com/z0mbie42/goes/blob/master/_examples/api/schema.graphql#L14
```graphql
 events(filter: Filter): [Event!]!
```
by
```graphql
events(filter: Filter = {offset: 0, limit: 25}): [Event!]!
```

But, after applying the patch #146 the code generate without error (no missing comma and no import cycle)

PR #147, opened 2018-06-23
Improve import handling

### LINKED ISSUE #114: deterministic import names
### Expected Behaviour

Generated code should be deterministic.

Even when the types specified by typemap.json are in several packages with the same name or same final dir of their import path.

### Actual Behavior

gqlgen does detect this situation :+1: and reacts by giving each package with the same name a unique name when imported by suffixing a number.  So if we have "foobar/types" with package name "types" and "bazfoo/types" with package name "types", gqlgen will make imports roughly like:

```
import (
    // other stdlib stuff

    types "foobar/types"
    types1 "bazfoo/types"
```

The trouble is... it seems to do this semirandomly :-1:  So when using gqlgen repeatedly, the `git diff` of the results sometimes results in a bunch of 

```
-   types "foobar/types"
-   types1 "bazfoo/types"
+   types "bazfoo/types"
+   types1 "foobar/types"
```

as well as every method on the Resolvers interface and so on also having that unhelpful diff.

If one changes the names of the packages internally, but leaves them in the same paths -- e.g. files under "foobar/types" path now declare "package foobartypes" -- gqlgen doesn't change behavior; looks like it bases all the names on the last segment of the import path.  I think it should probably use the declared package name instead.

This isn't a correctness issue, exactly, because it's all internally consistent and doesn't affect anything outside of that generated file, but it definitely makes for some uggo diffs, and makes the Resolvers interface a lot harder to read.

Ideally, gqlgen would use the declared package names, and then if names still have uniqueness issues, use sort on the imports before doing the name suffixing.

vektah:
Fixes for a few bugs in import handling:
 - imports will now use the package name defined in the package, instead of generating one
 - dont do name decollisioning until all imports have been resolved and sorted
 - use a case insensitive compare for determining if a package refers to itself

Fixes #114 and maybe #143

### DIFF
--- a/codegen/build.go
+++ b/codegen/build.go
@@ -49,7 +49,7 @@ func (cfg *Config) models() (*ModelBuild, error) {
 		PackageName: cfg.ModelPackageName,
 		Models:      models,
 		Enums:       cfg.buildEnums(namedTypes),
-		Imports:     imports.imports,
+		Imports:     imports.finalize(),
 	}, nil
 }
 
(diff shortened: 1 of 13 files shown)


eric-am:
:heart_eyes:

PR #150, opened 2018-06-27
Regenerate files for dataloader example for latest naming convention

dvic:


vektah:
the Gopkg.lock needs to be updated

`dep ensure -update github.com/vektah/dataloaden` should do the trick.

dvic:
You're right, and I also understand that vektah/dataloaden#7 needs a fix so that it becomes backwards compatible.

PR #151, opened 2018-06-27
Fix bug with multiple GOPATH full package name resolving

dvic:
This commit fixes the bug where GOPATH values that are longer than the input package name cause 'slice bounds out of range'  errors.


vektah:
Mind throwing a test on it?

dvic:
Sure no problem, do you have some advice on how to mock/control `build.Default.GOPATH`? Or do I need to change the method signature or read the GOPATH environment variable directly?

vektah:
It looks like its public, can you set it? 😈

vektah:
If you're about pop into https://gitter.im/gqlgen/Lobby

PR #157, opened 2018-06-29
Updated mutation to take correct argument

johncurley:
I've just started using the library and worked through the tutorial today and I had to make these changes to get it working.

I'm not sure if the UserID is the correct argument for the mutation however from my knowledge this looks correct and it compiles.


vektah:
Thanks, looks good 👍

PR #160, opened 2018-07-01
Pregenerate interfaces

### LINKED ISSUE #152: gqlgen on new RootResolver return declared name XxxResolver
### Expected Behaviour
run `gqlgen -typemap ./types.json -schema ./schema_generated.graphql` command on provided graphql schema, schema.go and types.json should not return compile error undeclared: QueryResolver

### Actual Behaviour
    gqlgen return undeclared QueryResolver

https://gist.github.com/pistachiology/181b54d7e913a61fff5a7eeee5c966f0

vektah:
Currently when using the new resolver syntax its impossible to put your resolvers in the same package as the generated code.

This is because the user written resolver code depends on the generated interfaces which is not present during generation. This means the package cant be compiled, which codegen will try to do if there are any models (generated or user defined) in the package.

This PR creates empty interface stubs for all possible resolvers, eg:
```go
// If you are see this file it means code generation has failed.
//
// It is created to satisfy all interfaces that might be generated, before codegen runs. This ensures that any code
// relying on these interfaces will still compile without the generated code, so that we can load the packages during
// codegen.

package todo

import "github.com/vektah/gqlgen/graphql"

func MakeExecutableSchema(resolvers Resolvers) graphql.ExecutableSchema {
	return nil
}

func NewExecutableSchema(resolvers ResolverRoot) graphql.ExecutableSchema {
	return nil
}

type (
	Resolvers    interface{}
	ResolverRoot interface{}

	MyMutationResolver interface{}
	TodoResolver       interface{}
	MyQueryResolver    interface{}
)
```
when codegen finishes the file is replaced with the real interfaces.

Fixes #152 

Thoughts @mastercactapus?


mastercactapus:
Hmm, I hadn't thought of that; we had problems even with the old interface and having them share a package. So we opted to keep them separate early on without too much thought. (namely, type assertion `var _ Resolvers = &App{}`, as well as regenerating while still implementing the resolver methods).

With the stubs, I think we could still get an error when we do more than just return types (i.e. if any of the user provided code calls methods on the interfaces). It also still depends on the current package compiling when the only dependency is actually the schema and stuff referenced by `types.json`. While it would solve most cases in all likelihood, there are still edge cases that won't have a good workaround (other than commenting/moving code temporarily).

We also have the issue that if the interface is changed in any way, we would have the same problem (since the types would no longer fulfil all methods of the interface).

---

In another approach, some tools (like `gofmt` or `gorename`) seem to follow a: "don't change anything unless it's going to work"-principal. I noticed `gqlgen` deletes the old files (thus introducing a compile break) before actually doing the generation work.

Commenting out lines 41-42 in `codegen.go` get's us partly there -- assuming the code *currently* compiles, it will continue to do so until we gen & replace.

```
	// _ = syscall.Unlink(cfg.ExecFilename)
	// _ = syscall.Unlink(cfg.ModelFilename)
```

We still need to cleanup `models_gen.go` if it is no longer needed, but we could move that to AFTER code generation. BUT, this still has the problem of: If you add a method, first generate works, but second fails (until you add the new resolver method), because it would no longer compile.

---

We could also try generating code in a separate/temporary package and then move the result to make things more-or-less atomic. That would make it possible to generate & re-generate code even if the user is not finished implementing all resolvers in the current package. It should work with `types.json` since that references the full package name (and thus not dependant on where this "temporary" package exists) -- only that the type-packages compile and are thus readable.

It gives the benefit of keeping the generated interface & friends in an isolated package, but without explicitly requiring users to do so.

Having said all that, there's still somewhat of a circular dependency when the generated code depends on stuff in the package it's meant to output too. The only real way to avoid that would be to put all generated code in it's own package after all. That, and to require users to NOT implement types used by the generated code in the same package that uses the generated code, as Go does not like circular dependencies.

In that case, users would either use the generated models (for graphql layer), or types from the application, keeping the graphql/glue stuff on it's own.

Maybe relevant:
> **Don't use a single package for all your APIs.**
> Many well-intentioned programmers put all the interfaces exposed by their program into a single package named api, types, or interfaces, thinking it makes it easier to find the entry points to their code base. This is a mistake. Such packages suffer from the same problems as those named util or common, growing without bound, providing no guidance to users, accumulating dependencies, and colliding with other imports. Break them up, perhaps using directories to separate public packages from implementation.
> -- https://blog.golang.org/package-names

---

Overall, personally, I'd advocate for putting the generated code in a separate package (and thus only exposing the public methods) so that generation & re-generation is durable without any circular dependencies. But there are also things we can do to preserve the current behavior, and yet make it less prone to compile errors.

Really, I see two main use-cases:
- single-package apps; everything (server, API, types & gen) in a single package
- multi-package apps; separate packages for things like users, graphql, the main package, etc..

For the all-in-one, fixing regeneration (e.g. don't delete until the end) might be enough.
For multi-package, keeping the gen. code independent would probably be best.

---

TL;DR; What about:
- Having generated code output to its own package
- Using a temporary package while generating
- Don't touch/remove/update any existing code (generated or otherwise) until generation is complete & succeeds

vektah:
> Overall, personally, I'd advocate for putting the generated code in a separate package (and thus only exposing the public methods)

Yeah, I agree, but single package apps are useful. This probably gets better with a generated skeleton that sets up a multi package layout.

Ok, lets give not deleting a try. I think it makes working on gqlgen a bit harder because you cant start from a clean slate easily, but thats probably just a sign that the examples should be using a mutli-package layout.

vektah:
Ended up going a third way, just delay showing errors until after the code has been generated - https://github.com/vektah/gqlgen/pull/166

PR #169, opened 2018-07-05
Support interface unexported methods

dvic:
I like to use unexported methods with interfaces to create union types in Go. This PR fixes the "missing method" warning. It replaces the original calls to `findMethod` with `findExportedMethod ` (which has the same logic).


dvic:
Build was failing due to linting false positive, I had to add `// nolint: megacheck` to the violating line in the new example.

dvic:
@vektah sorry it took so long, I rebased and pushed again but this time without the example and with the testing approach you suggested. In order to properly test this I had to replace the `fmt.FPrintf` with `stdErrLog.Printf`, where `stdErrLog` is a logger instance from the standard library. Is this okay?

dvic:
@vektah I updated the PR for 0.4.x. It seems that the interface types tests were removed? (while the test models were still defined). Not sure what you want to do with them, so for now I just added the tests back (with the updated `codegen/gen` path).

vektah:
I think `onlyExported` is the wrong approach here too. This is opening up codegen to write invalid code if the interface is defined in a different package to the generator. Can we check the packages are the same instead of a boolean?

codegen/codegen_test.go
```diff
@@ -41,3 +41,26 @@ func TestGenerateServer(t *testing.T) {
 	_, err = conf.Load()
 	require.NoError(t, err)
 }
+
+func generate(name string, schema string, typemap ...TypeMap) error {
```
vektah: why is this moving?

codegen/interface_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestShapes(t *testing.T) {
```
vektah: doubled up with https://github.com/99designs/gqlgen/blob/master/codegen/testserver/schema.graphql#L112

codegen/interface_test.go
```diff
@@ -0,0 +1,72 @@
...
+			    id: ID!
+				y: String!
+			}
+	`, TypeMap{
+		"Bar":    {Model: "github.com/99designs/gqlgen/codegen/testserver.IfaceBar"},
+		"Entity": {Model: "github.com/99designs/gqlgen/codegen/testserver.IfaceEntity"},
+		"Foo":    {Model: "github.com/99designs/gqlgen/codegen/testserver.IfaceFoo"},
+	})
+	require.NoError(t, err)
+	assert.Equal(t, "", buf.String(), "using an interface with unexported methods should not emit warnings")
```
vektah: rather than test for the absence of a warning, could you look to see if a resolver was generated?

eg https://github.com/99designs/gqlgen/blob/master/codegen/testserver/generated_test.go#L24-L28

vektah:
This change has been rolled into #335

PR #173, opened 2018-07-06
add resolver option support to field

### LINKED ISSUE #161: request: process arbitrary fields by resolver
I want to process `id: ID!` field by resolver.

[Relay Global Object Identification Specification](https://facebook.github.io/relay/graphql/objectidentification.htm) required `This id should be a globally unique identifier for this object`.
but almost Database is not support that.
GitHub v4 uses something like `echo 010:Repository52265602 | base64` for id.

My app already have `ID` field and It is difficult to change.
so, I want to modify id value by resolver.

I think that it is good to write in types.json like this.

```
{
  "Event": {
    "entity": "github.com/vvakame/foobar.Event",
    "fields": {
      "id": {
        "resolver": true
      }
    }
  }
}
```

I saw #156 and `@resolver` directive.
but I want to use `schema.graphql` to communication between server-side engineer and frontend engineer.
I don't want to expose server-side issue to frontend.

What do you think about this idea?

vvakame:
closes #161

### DIFF
--- a/codegen/config.go
+++ b/codegen/config.go
@@ -29,6 +29,7 @@ type TypeMapEntry struct {
 }
 
 type TypeMapField struct {
+	Resolver bool `yaml:"resolver"`
 }
 
 func (c *PackageConfig) normalize() error {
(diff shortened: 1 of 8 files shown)


vektah:
Looks good, thanks!

PR #177, opened 2018-07-08
Add docs for error handling

### LINKED ISSUE #174: Feature: error extensions
It would be nice if we could set the `extensions` field on errors. Content is app-specific arbitrary JSON. Maybe something like if the `error` returned from a resolver implements `interface { GraphQLExtensions() interface{} }`, add that as the `extensions` field?

http://facebook.github.io/graphql/June2018/#sec-Errors

```
{
  "errors": [
    {
      "message": "Name for character with ID 1002 could not be fetched.",
      "locations": [ { "line": 6, "column": 7 } ],
      "path": [ "hero", "heroFriends", 1, "name" ],
      "extensions": {
        "code": "CAN_NOT_FETCH_BY_ID",
        "timestamp": "Fri Feb 9 14:33:09 UTC 2018"
      }
    }
  ]
}
```

vektah:
I cleaned up the return type from the Error presenter to try and better document the intent behind that function. Its really not returning an `error`, its returning an `interface{}` to be json marshaled.

Closes #174

### DIFF
--- a/docs/content/reference/errors.md
+++ b/docs/content/reference/errors.md
@@ -1,0 +1,33 @@
+---
+linkTitle: Handling Errors
+title: Sending custom error data in the graphql response
+description: Customising graphql error types to send custom error data back to the client using gqlgen.
+menu: main
+---
+
+All errors raised by gqlgen pass through a hook before being displayed to the user. This hook gives you the ability to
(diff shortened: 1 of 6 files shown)


vektah:
Closing in favor of #186

PR #179, opened 2018-07-09
Improve Output Filename and Package Handling

mathewbyrne:
Improves some of the errors encountered in https://github.com/vektah/gqlgen/issues/170

- If a package is specified for a model then the final `.` should come after the last `/` character in the package name.  This is performed before build is started.

- Generated `filename` config options must end in `.go` or an error will be generated.

- If the output package name is not specified, then load the package at the output file path and read the name of the package there.  Use the package name from the import if one is resolved, otherwise fallback to old behaviour of using the last directory name.


codegen/config.go
```diff
@@ -79,25 +81,45 @@ func (c *PackageConfig) normalize() error {
...
+		return fmt.Errorf("package should be the output package name only, do not include the output filename")
+	}
+	return nil
+}
+
 func (cfg *Config) Check() error {
-	err := cfg.Models.Check()
-	if err != nil {
+	if err := cfg.Models.Check(); err != nil {
+		return fmt.Errorf("config: %s", err.Error())
```
vektah: errors.Wrap(err, "config.models")

codegen/config_test.go
```diff
@@ -52,3 +54,27 @@ func TestLoadDefaultConfig(t *testing.T) {
...
+		ok := assert.NotPanics(t, func() { got = p.ImportPath() })
+		if ok {
+			assert.Equal(t, "/b/src/y/foo/bar", got)
+		}
+	})
+	t.Run("stop searching on first hit", func(t *testing.T) {
+		p := PackageConfig{Filename: "/a/src/x/foo/bar/baz.go"}
+		build.Default.GOPATH = "/a/src/x:/b/src/y"
+		var got string
+		ok := assert.NotPanics(t, func() { got = p.ImportPath() })
```
vektah: this can all probably simplify down to
```go
require.Equal(t, "/a/src/x/foo/bar", p.ImportPath)
```

right?

PR #183, opened 2018-07-11
Allowing for resolvers to return multiple errors to errorbuilder

### LINKED ISSUE #141: Return multiple errors for a resolver
The interface for passing your own `ErrorPresenter` is great, but has one downfall; we are unable to return multiple errors for a single resolver due to the `ErrorBuilder` simply appending the errors.

https://github.com/vektah/gqlgen/blob/master/graphql/error.go#L41

This would be particularly useful for input validation as you could return all validation errors at once rather than one at a time... something like:

```
{
  "data": {
    "createHero": null
  },
  "errors": [
    {
      "type": "Validation Error",
      "message": "Episode not found",
      "path": [
        "createHero",
        "heroInput",
        "episode"
      ]
    },
    {
      "type": "Validation Error",
      "message": "Name already exists",
      "path": [
        "createHero",
        "heroInput",
        "name"
      ]
    }
  ]
}
```

I'm thinking if we could check if it confirms to some kind of interface, similar to pkg/errors `Cause()` interface, that returns a slice of errors to append?

andrewmunro:
Attempting to fix #141 by allowing you to pass an error that includes an `Errors()` function.

Maybe not the nicest solution as it requires a lot of work from the consumer's point of view, but this gets people unblocked in the meantime.

In the future, I'd want to see the `ErrorBuilder`or even the `graphql.Response` opened up to some kind of formatter somehow.

### DIFF
--- a/graphql/error.go
+++ b/graphql/error.go
@@ -26,6 +26,11 @@ func (r *ResolverError) Error() string {
 	return r.Message
 }
 
+type Errors interface {
+	Errors() []error
+	Error() string
+}
+
 type ErrorBuilder struct {
 	Errors []error
 	// ErrorPresenter will be used to generate the error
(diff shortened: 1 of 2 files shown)


vektah:
Closing in favour of #186. Let me know if I've missed anything.

PR #186, opened 2018-07-12
Error redux

### LINKED ISSUE #141: Return multiple errors for a resolver

### LINKED ISSUE #174: Feature: error extensions

vektah:
I set out to document the current state of errors, ended up rewriting them instead. :worried: 

The old ErrorPresenter took an `error` and returned an `error`, but really it was returning the json encodable response, from that point backward its never treated as an error, just a part of the response.

I've also added support for Error extensions, which using the default presenter can be set by returning an error that meets the `ExtendedError` interface (having an `Extensions()` method).

I've also added support to return multiple errors by calling:
```go
graphql.AddError(ctx, err)
graphql.AddErrorf(ctx, "boom")
graphql.AddError(ctx, &graphql.Error{
   Message: "Totally customizable in resolver method errors",
})
```

[Check the docs](https://github.com/vektah/gqlgen/blob/error-redux/docs/content/reference/errors.md)

Supersedes #177
Fixes #174
Fixes #141


mathewbyrne:
Looks good. Not sure about the naming though, `AddGraphqlErorr` isn't immediately obvious to me that it's a customised error. Maybe even just `AddCustomError` is clearer.

andrewmunro:
Much more flexible approach! 🎉 

Might want to add a response payload or ordering of multiple errors to the documentation as it might not be clear?

PR #187, opened 2018-07-13
Added include/skip directives support.

sergeyvlakh:
@vektah 
Had to make another fork, because I dun goofed with branch. Functionality stays the same, as it was on my previous branch: we have support for `@skip` and `@include` directives.
`OperationName` is taken from Request Context.


sergeyvlakh:
Closing this, as vektah said:
>The new parser is nearly done now, full directive support should only be a few weeks away.

PR #190, opened 2018-07-14
Allow forcing resolvers on generated types

### LINKED ISSUE #189: Force creation of field resolvers for automatically generated models
It would be nice to have a way to let gqlgen generate resolver interfaces for specified fields, without  having to provide the Model implementation.
Consider following example:
```
type A {
  name String!
  id ID!
  b B!
}
type B {
  ... expensive content ...
}
```
On default this produces following model for A:
```
type A struct {
  name string
  id string
  b B
}
```
This is sub optimal if the data for B is expensive to produce and is often not requested from the client.
The solution to this is having a separate field resolver which is called only on demand.

Currently the only way of doing this is implementing the model yourself and omitting field b. This forces qglgen to create the resolver for the missing fields.

However implementing the model yourself adds the burden to keep the model in sync when the schema evolves. In many cases self implemented models do not add any additional benefit. Hence it would be much nicer if qglgen is able to create the models automatically and generate the resolver interfaces as desired.

### Solution using the configuration file
Currently the it is possible to define for which field a resolver is generated within the models section of the qglgen.yml config file. But this only works if a model is defined as well.
```
models:
  Todo:
    model: github.com/my/app/models.Todo
    fields:
      id:
        resolver: true # force a resolver to be generated
```
The desired behavior could be achieved if qglgen permits to omit the "model" field and generates the model instead, together with the requested field resolver definition.

vektah:
Fixes #189


flicaflow:
@vektah thank you for doing this. 
I'm just trying out the new feature. I noticed the the auto generated model still has the fields defined despite having a field resolver defined. Wouldn't it be more natural to have this field omitted?

Ehekatl:
@vektah I found that force resolver doesn't work when fields have already defined in model, it that intend behavior?

e.g. 

this won't work
```
type Host struct {
  id string
  status string
}

type Host implements Node {
  id: ID!
  status: String
}

models:
  Host:
    model: xxx/models.Host
    fields:
      status:
        resolver: true
```

this works
```
type Host struct {
  id string
}

type Host implements Node {
  id: ID!
  status: String
}

models:
  Host:
    model: xxx/models.Host
    fields:
      status:
        resolver: true
```

PR #205, opened 2018-07-19
Use original credential for query request in playground

creativej:
## Problem:
Currently the playground doesn't forward any credentials when making
query calls. This can cause problems if your playground requires login credentials.

## Solution:
Setting the request credential to "same-origin" fixes the issue.


hueys:
FWIW this breaks passing values into the endpoint as query params for my team. We recently upgraded from v0.3.0 to v0.7.2 and passing an endpoint value of `/graphql?foo=bar` no longer works, the query is stripped. For now we're working around this with a custom version of the playground handler that removes the changes made in this commit.

Issue #156, opened 2018-06-29
Planning for directive support


vektah:
Once the new parser lands and we get full support for directives. These things don't need to be in the core runtime, and should be implementable outside. This probably means codegen hooks.

Possible uses:

### `@resolver`
A way to force creating a resolver method for a field, eg
```graphql
type User {
    name @resolver
}
```
would generate a Name resolver even if the user struct has a Name field.

### `@rest`, `@grpc`, `@swagger`
These would implement the resolver method for you by binding to an external API.
```graphql
type Query {
    user(id: Int!): User @grpc(accounts:"foo.bar", method:"GetUsers")
}
```
### `@hasRole`, `@isAdmin`, `@isAuthenticated`
These act as middleware filters, and should be able block calling the resolver
```graphql
type Mutation {
    banUser(id: Int!): User @isAdmin
}
```

### `@resolveString` and all of the [graphql-cats](https://github.com/graphql-cats/graphql-cats) directives
They shouldn't be part of the core, but a way to implement them to test against cats is really high on my list.

### `@skip` and `@include`
these are both runtime directives, not schema but there is probably some overlap here and its worth thinking about them.

These are the ones I want. What do you want to see? Lets collect some ideas before rushing into an implementation.

### `@connection`?
Is there some codegen hook that would make dealing with relay style connections easier?

vvakame:
My team really needs this feature ASAP 🤣
I want to use `@isResourceOwner` and `@isStaff` on `QUERY | FIELD`.
so, Is there anything I can do for you? 🌹

sergeyvlakh:
`@skip/@include` :+1: :+1: :+1:

ntucker:
@skip and @include are both key for using relay-modern effectively.

vvakame:
https://github.com/vektah/gqlgen/pull/130#issuecomment-404762711
vektah said
> The new parser is nearly done now, full directive support should only be a few weeks away.

andy9775:
Would it be possible to write our own directives like in [Apollo](https://www.apollographql.com/docs/graphql-tools/schema-directives.html)?

mathewbyrne:
@andy9775 yes we are currently planning to make directives work much the same as resolvers currently do; directives found in a schema will be generated and you will be able to supply your own implementation.

vektah:
I've broken these out into smaller cards on the 1.0 board

PR #208, opened 2018-07-20
Support @skip and @include directives

mathewbyrne:
Adds support for `@skip` and `@include` directives.  These should function according to the [specification](http://facebook.github.io/graphql/June2018/#sec--skip) and skip or include fields, fragments, and inline fragments.

This change also includes passing the request context through to `CollectFields`.

See #156 

Also thanks to @sergeyvlakh for his PRs #187 and #130


graphql/exec.go
```diff
@@ -118,3 +123,33 @@ func getOrCreateField(c *[]CollectedField, name string, creator func() Collected
...
+	arg := d.Arguments.ForName("if")
+	if arg == nil {
+		panic(fmt.Sprintf("%s: argument 'if' not defined", d.Name))
+	}
+	value, err := arg.Value.Value(variables)
+	if err != nil {
+		panic(err)
+	}
+	ret, ok := value.(bool)
+	if !ok {
```
mathewbyrne: @vektah I noticed this panics when a variable is omitted from the query altogether.  "Invalid" values like strings/numbers _seem_ to get coerced into a `bool`. But there's probably somewhere better to report on missing variables?

vektah: Yeah, theres a post-validation step thats still missing. Ill add a card and make sure its covered before we :shipit:

codegen/templates/object.gotpl
```diff
@@ -25,7 +26,8 @@ func (ec *executionContext) _{{$object.GQLType}}(ctx context.Context, sel ast.Se
 }
 {{- else }}
 func (ec *executionContext) _{{$object.GQLType}}(ctx context.Context, sel ast.SelectionSet{{if not $object.Root}}, obj *{{$object.FullName}} {{end}}) graphql.Marshaler {
-	fields := graphql.CollectFields(ec.Doc, sel, {{$object.GQLType|lcFirst}}Implementors, ec.Variables)
+	reqCtx := graphql.GetRequestContext(ctx)
+	fields := graphql.CollectFields(reqCtx, sel, {{$object.GQLType|lcFirst}}Implementors)
```
vektah: maybe ctx?

graphql/exec.go
```diff
@@ -51,23 +54,25 @@ func collectFields(doc *ast.QueryDocument, selSet ast.SelectionSet, satisfies []
 
 			f.Selections = append(f.Selections, sel.SelectionSet...)
 		case *ast.InlineFragment:
-			if !instanceOf(sel.TypeCondition, satisfies) {
+			if !shouldIncludeNode(sel.Directives, reqCtx.Variables) || !instanceOf(sel.TypeCondition, satisfies) {
```
sergeyvlakh: I might be wrong, but when I did this, `reqCtx` had no default arguments (ones from document). So this query
```graphql
query MyAwesomeQuery($shouldIncludeField: Boolean = false) {
    resolver @include(if: $shouldIncludeField) {
        __typename
    }
}
```
would still include in case no variables were explicitly passed. That's why I did this:
[https://github.com/vektah/gqlgen/pull/187/files#diff-47a8f73d5b7f0dbad02ee4611bbedb10R79](https://github.com/vektah/gqlgen/pull/187/files#diff-47a8f73d5b7f0dbad02ee4611bbedb10R79)
BUT again, I might be very-very wrong.

vektah: :+1: it's probably worth a testcase

PR #221, opened 2018-07-24
Implement FieldMiddleware Stack

mathewbyrne:
Following on from #217 this change introduces a the concept of DirectiveMiddleware to generated code.

A `DirectiveRoot` struct is now created and can be passed in when creating an executable schema.  If a directive is encountered during codegen from the source schema, this `DirectiveRoot` will contain a field for each that the user can assign.  Each field has the same signature as a `ResolverMiddleware`, and is run on fields with this directive at the end of the resolver middleware chain.

Within the directive middleware, users can access the current request and resolver contexts, and can either call the next middleware in the chain, or return a value early.

## Limitations

- Currently there is no direct support for directive arguments.  This is coming soon but may depend on some refactoring around coerced values both in `gqlgen` and `gqlparser`.
- If returning a new value, middlewares need to take care to return a value of the expected type, as this will be cast back after the middleware stack has resolved.
- There is probably a refactor to happen around config now required to pass into `NewExecutableSchema`, and the config that can be provided to `GraphQL`.
- This only supports schema-directives on fields definitions.
- Currently `DirectiveRoot` is implemented as a struct, but this could easily also be an interface instead.  Current thinking is that a struct allows a user to get up and running without having to implement their directives immediately; they are noops until set.


codegen/directive.go
```diff
@@ -3,9 +3,9 @@ package codegen
 import "strings"
 
 type Directive struct {
-	name string
+	Name string
 }
 
-func (d *Directive) Name() string {
-	return strings.Title(d.name)
+func (d *Directive) GoName() string {
```
vektah: maybe this should be NameUpper? where is this used?

vektah: also, https://github.com/vektah/gqlgen/blob/master/codegen/templates/templates.go#L65 is available in templates

ridhamzuru:
@mathewbyrne can we get HTTP headers value in field middlewares as mostly auth details are passed as HTTP Headers.

mathewbyrne:
@ridhamzuru you have access to `Context` in middleware, so depending on your authentication scheme, you probably want to do authentication in some middleware before graphql and pass through related data.

rajzru:
@mathewbyrne Are we talking about graphql `handler.RequestMiddleware` as I have only specific mutation and queries on which I need auth so I can not use HTTP middleware as I will need the name of mutation or query on this.

This is what I get in the context variable so still can't access the HTTP variables
image

rajzru:
But how can I get the name of query or mutation in `authentication as a middleware` approach?

mathewbyrne:
It sounds like you might be mixing up your authentication and authorisation logic.  Authentication should not need information about the query to happen.  Once you have authenticated through whatever scheme you're using, you can make authorisation decisions at the resolver level using a directive.

rajzru:
Can we one simple example for JWT with recommended way?

rajzru:
@mathewbyrne I have implemented with context suggestion and commented on #262
Thanks for the help anyways. You guys are doing the awesome job.

PR #229, opened 2018-07-27
Fixing init command

creativej:
The init command always return "file already exists" error if there are no
configFilename specified

This is caused by `codegen.LoadDefaultConfig()` hiding the loading details
and always return the default config with no error while the init
command code expects it to tell us if config doesn't exists in default
locations.

To avoid confusion I have split out the loading config code into its own function so we can handle different cases easier.

Additionally I also moved default config into its own function so we always
generating new a config instead of passing it around and potentially
mutating the default config.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -60,8 +60,8 @@ var initCmd = &cobra.Command{
 	Short: "Generate .gqlgen.yml",
 	Long:  "",
 	Run: func(cmd *cobra.Command, args []string) {
-		initConfig()
 		initSchema()
+		initConfig()
 	},
 }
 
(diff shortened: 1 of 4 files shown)


cmd/init.go
```diff
@@ -70,26 +70,24 @@ func initConfig() {
...
 	if configFilename != "" {
 		config, err = codegen.LoadConfig(configFilename)
-	} else {
-		config, err = codegen.LoadDefaultConfig()
-	}
-	if os.IsNotExist(errors.Cause(err)) {
-		// ok
-		if configFilename == "" {
-			configFilename = ".gqlgen.yml"
+		if err != nil {
```
vektah: Shouldn't this also check if it already exists?

PR #235, opened 2018-07-30
Generate resolver on init

creativej:
## What's in this PR?
1. Update the `gen` command to generate skeleton resolvers. 
This should remove a lot of manual work of writing resolvers and making sure they have the correct interfaces. 
Resolver generation only happen if it's specified in the config. For example: 
```yml
resolver:
  filename: resolver.go
  type: Resolver
```
Also the generation is skipped if codegen finds an existing resolver in the specified location.

In the near future we will look at modifying existing resolver and update the interfaces as the schema file changes.

2. Updated the `init` command to also generate the skeleton code by default. 

```go
//go:generate gorunpkg github.com/vektah/gqlgen

package init

import (
	context "context"
)

type Resolver struct{}

func (r *Resolver) Mutation() MutationResolver {
	return &mutationResolver{r}
}
func (r *Resolver) Query() QueryResolver {
	return &queryResolver{r}
}

type mutationResolver struct{ *Resolver }

func (r *mutationResolver) CreateTodo(ctx context.Context, input NewTodo) (Todo, error) {
	panic("not implemented")
}

type queryResolver struct{ *Resolver }

func (r *queryResolver) Todos(ctx context.Context) ([]Todo, error) {
	panic("not implemented")
}

```


codegen/config.go
```diff
@@ -81,7 +83,7 @@ type TypeMapField struct {
 
 func (c *PackageConfig) normalize() error {
 	if c.Filename == "" {
-		return errors.New("Filename is required")
+		return nil
```
vektah: Can we keep this for model and exec? maybe by just not calling normalize if !Resolver.IsDefined?

codegen/codegen.go
```diff
@@ -57,23 +50,44 @@ func Generate(cfg Config) error {
...
+func generateResolver(cfg Config) error {
+	resolverBuild, err := cfg.resolver()
+	if err != nil {
+		return errors.Wrap(err, "resolver build failed")
+	}
+	filename := cfg.Resolver.Filename
+
+	if resolverBuild.ResolverFound {
+		log.Printf("Skipped resolver: %s.%s already exists\n", cfg.Resolver.ImportPath(), cfg.Resolver.Type)
+	} else {
```
vektah: would this read better if the happy path was left unindented?

codegen/config.go
creativej: Ah yes. my bad forgot it's shared! =x

PR #236, opened 2018-07-31
Generate server on running init

creativej:
Update the `init` command to also generate a runnable server to help getting started even easier.


codegen/templates/server.gotpl
```diff
@@ -0,0 +1,18 @@
+package main
+
+import (
+{{- range $import := .Imports }}
+	{{- $import.Write }}
+{{ end }}
+)
+
+func main() {
+	port := flag.Int("p", 8080, "port number")
```
vektah: the PORT env var is commonly used for this (and removes the flag dep)

codegen/build.go
```diff
@@ -99,6 +106,21 @@ func (cfg *Config) resolver() (*ResolverBuild, error) {
 	}, nil
 }
 
+func (cfg *Config) server(destDir string) *ServerBuild {
+	imports := Imports{
+		destDir: destDir,
+	}
+	imports.add(cfg.Exec.ImportPath())
```
vektah: you probably should include any imports that aren't covered by https://github.com/vektah/gqlgen/blob/master/codegen/import_build.go#L11-L27 here too (`http` and `githhub.com/vektah/gqlgen/handler`)

Without them, gqlgen can end up with duplicate package aliases in the generated code.

creativej: hrmm I didn't encounter that issue but sure. Updated.

ermik:
I am not getting a server on bare `init` run nor does it appear with `--server server.go`. Maybe I am doing it wrong?

ermik:
Aha! It works in a non-empty directory, but it only creates the yaml. Got it.

PR #239, opened 2018-08-01
Directive args

vektah:
Pass directive arguments in when calling directives.

This means the directives now look a lot like resolvers:
```go
type DirectiveRoot struct {
	HasRole func(ctx context.Context, next graphql.Resolver, role Role) (res interface{}, err error)
}
```


mathewbyrne:
💯 exactly what I was going for. Great to see this land.

PR #249, opened 2018-08-02
commit vendored gqlparser

vektah:
The parser dep is pretty closely tied to gqlgen, perhaps we should commit it? Thoughts?


vvakame:
I don't like this change...
https://github.com/alecthomas/gometalinter has `vendor/`, I can not build it successfully with a tool pruning a vendor like dep.

vektah:
It might be worth writing to see if there is a real problem here first. One thing I did notice is `dep ensure -update github.go/vektah/gqlgen` usually needs a parser update too

PR #250, opened 2018-08-02
Replace circleci with travis

vektah:
:roll_eyes:


dvic:
@vektah Just out of curiosity, why did you make the switch?

dvic:
Ahh I understand, maybe you can speed-up the builds by using non-`sudo` (docker container builds)? Or do you need `sudo`?

PR #260, opened 2018-08-06
Init Config Improvement

mathewbyrne:
Init will currently fail with an obscure error message if a config file is found.  This change fixes a couple of issues with the codepath, and outputs a better error message if a config file is located:

```
init failed: a configuration file already exists at /Users/mathew/Projects/go/src/github.com/99designs/example/gqlgen.yml
```

Also fixed a minor issue outputting the server filename.

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -105,28 +105,23 @@ func initConfig() *codegen.Config {
 		if err != nil {
 			fmt.Fprintln(os.Stderr, err.Error())
 			os.Exit(1)
-		} else if config != nil {
-			fmt.Fprintln(os.Stderr, "config file already exists")
-			os.Exit(0)
 		}
-	} else {
-		config, err = codegen.LoadConfigFromDefaultLocations()
-		if os.IsNotExist(errors.Cause(err)) {
-			if configFilename == "" {
-				configFilename = "gqlgen.yml"
-			}
-			config = codegen.DefaultConfig()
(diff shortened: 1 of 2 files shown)


cmd/init.go
```diff
@@ -94,39 +94,36 @@ func GenerateGraphServer(config *codegen.Config) {
...
-			os.Exit(1)
-		}
+	}
+
+	if config != nil {
+		fmt.Fprintf(os.Stderr, "init failed: a configuration file already exists at %s\n", config.FilePath)
+		os.Exit(1)
+	}
+
+	if !os.IsNotExist(errors.Cause(err)) {
```
creativej: What if the above code found a config file but there's parsing errors?

mathewbyrne: If there is a parse error then `config` will be `nil` and it will exit with `1` right?  I think that's the correct behaviour.  What do you think should happen?

creativej: Yeah cool yeah that's what I am thinking too so we will need to handle that edge case then because I don't think 
`!os.IsNotExist(errors.Cause(err))` will catch it.

mathewbyrne: Note the `!` — I know it's a bit hard to read, but it's checking for the opposite case; basically any error _except_ file not found.

creativej: hrmm okay I guess that makes sense.

PR #270, opened 2018-08-07
stop pickup "github.com/vektah/gqlgen/handler" from GOPATH

vvakame:
```
$ gqlgen init
Exec "go run ./server/server.go" to start GraphQL server

$ cat server/server.go
package main

import (
	"log"
	"net/http"
	"os"

	"github.com/vektah/gqlgen/handler"
	gqlgen_implements "github.com/vvakame/til/graphql/gqlgen-implements"
)

const defaultPort = "8080"

func main() {
	port := os.Getenv("PORT")
	if port == "" {
		port = defaultPort
	}

	http.Handle("/", handler.Playground("GraphQL playground", "/query"))
	http.Handle("/query", handler.GraphQL(gqlgen_implements.NewExecutableSchema(gqlgen_implements.Config{Resolvers: &gqlgen_implements.Resolver{}})))

	log.Printf("connect to http://localhost:%s/ for GraphQL playground", port)
	log.Fatal(http.ListenAndServe(":"+port, nil))
}
```

### DIFF
--- a/codegen/build.go
+++ b/codegen/build.go
@@ -86,6 +86,7 @@ func (cfg *Config) resolver() (*ResolverBuild, error) {
 	namedTypes := cfg.buildNamedTypes()
 	imports := buildImports(namedTypes, destDir)
 	imports.add(cfg.Exec.ImportPath())
+	imports.add("github.com/99designs/gqlgen/handler") // avoid import github.com/vektah/gqlgen/handler
 
 	cfg.bindTypes(imports, namedTypes, destDir, prog)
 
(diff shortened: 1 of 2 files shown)


vektah:
This should be added to the relevant build step, that list is shared by all the steps and handler should only be referenced by the generated resolvers.

So add it in here https://github.com/99designs/gqlgen/blob/master/codegen/build.go#L88

vvakame:
I pushed a commit. is this chages right?

PR #278, opened 2018-08-09
Fix Field Middlewares not Running for Subscriptions

### LINKED ISSUE #259: Subscriptions don't support directives
### Expected Behaviour
Directive should affect the subscriptions. At example below not authenticated user should not receive data about created messages.

### Actual Behavior
Query and Mutation support directives. And only authenticated users can read and create messages.
Subscriptions don't support directives. And all users can receive data about created messages.

```graphql
schema {
    query: Query
    mutation: Mutation
    subscription: Subscription
}
type Query {
    message(id: ID!): Message @isAuthenticated
}
type Mutation {
    messageCreate(subject: String!, text: String!): Message @isAuthenticated
}
type Subscription {
    messageCreated: Message! @isAuthenticated
}
type Message {
    messageID: ID!
    subject: String!
    text: String!
}
directive @isAuthenticated on FIELD_DEFINITION
```

mathewbyrne:
Fixes #259


mathewbyrne:
This is a bit undercooked as is, I'll get some test cases in when I get a chance.

mathewbyrne:
Going to close this since there has been a bit of churn on how resolvers are being run. This might need a bit of a rework.

PR #291, opened 2018-08-15
Fix for Trailing Slash in GOPATH

petersng:
If there is trailing slash in GOPATH, double slash causes the `EqualFold` matching to fail and `dir` variable to be incorrectly set.  If `/home/user/go/` is GOPATH, the string that is created is `/home/user/go//src/`.  This PR cleans up the original GOPATH before appending `/src/`.

Example:
`GOPATH=/home/user/go/` causes a "cannot import absolute path" error.
`GOPATH=/home/user/go` runs normally.


petersng:
Will look at these CI issues.

Issue #275, opened 2018-08-08
Response returns null on a non-nullable type


tkallioras:
I am a bit baffled by the way the framework deals with non-nullable types.

For example I have a mutation like that:

```
mutation SignUp($email: String!, $password: String!) {
  signUp(email: $email, password: $password) {
    success
  }
}
```
with the respective GraphQL schema:

`  signUp(email: String!, password: String!): Result!`

If I run this mutation through a client, in such a way that I get back an error (eg in this case I pass an empty email, which returns an error to the client):

```
{
	"data": {
		"signUp": null
	},
	"errors": [
		{
			"message": "Database error: email cannot be empty, message: email cannot be empty
			"path": [
				"signUp"
			]
		}
	]
}
```

_The issue is that the signUp field is not nullable, however the server responds with null._ In the beginning I believed it was a client issue, however after opening an issue in the [apollo-ios github repo](https://github.com/apollographql/apollo-ios/issues/342) it seems that there must be something off with the way the library handles nullability.

The issue is, that because the server nulls a non-nullable result, the client cannot handle correctly the error. Instead of getting back the correct error, it detects a null response (which should not be null) and reports the null value as the error.

vektah:
http://facebook.github.io/graphql/June2018/#sec-Errors-and-Non-Nullability 

Looks like nulls should bubble up, and in this case data should be null.

mikeifomin:
Faced with the same issue after update from 0.3 to 0.4.1 . All my schema now lack of a non nullable =(

vektah:
@mikeifomin that's a separate issue that should be fixed on master

PR #294, opened 2018-08-20
Simplfy concurrent resolver logic

vektah:
Before tackling #275 directly a little cleanup on concurrent resolvers is required. There currently isn't anywhere to add the required checks at the object level.

This pulls the concurrency logic out of individual graphql.Defer functions and moves it up into the object where we can put this bubbling logic. It also moves the recover function into the middleware stack, removing some duplicated code.


dvic:
@vektah @mathewbyrne shouldn't each item in an array also be resolved in parallel? (see line [here](https://github.com/99designs/gqlgen/blob/master/codegen/object.go#L203)) If not, how would one otherwise benefit from using dataloader implementations? Or am I missing something?

I noticed this when I upgraded to 0.4.x, which caused approximately a 10x increase in the number of SQL queries.

dvic:
@vektah No, the behavior is introduced with this commit, `0.4.4` works fine (produces the same amount of queries as previous versions). It's because of the sync resolving of array elements (done [here](https://github.com/99designs/gqlgen/blob/master/codegen/object.go#L218)). This commit changes the behavior so that resolving a field blocks on returning a `graphql.Marshaler`, while previously this was not the case.

dvic:
@vektah should I open a ticket for this?

vektah:
If you like, I'm hoping to look at it tomorrow.

Issue #295, opened 2018-08-20
DirectiveRoot struct fields' order often changes


heww:
### Expected Behaviour
generate DirectiveRoot struct fields in order

### Actual Behavior
DirectiveRoot struct fields' order often changes

### Minimal graphql.schema and models to reproduce

directive @hasPermission(obj: String!, act: String!) on FIELD_DEFINITION
directive @hasScope(object: String!, actions: [String!]) on FIELD_DEFINITION
directive @isAuthenticated on FIELD_DEFINITION

type Query {
  me: User! @isAuthenticated
}

type User {
  id: ID!
  username: String!
  email: String @isAuthenticated
  phone: String @isAuthenticated
  avatarUrl: String
}

PR #296, opened 2018-08-20
sort directives by name when gen

heww:
`ast.Schema` use map to store directives parsed from schema file, and map iteration order is random in Golang. The result is that `buildDirectives` will return random order directives when gen for the same schema file.

This pr sort directives by name, it will solve issue #295


vektah:
Thanks, looks good.

Issue #289, opened 2018-08-14
Question: How can we get parent object in directive function


vvakame:
### Expected Behaviour

Parent object can be retrieved from ResolverContext.

### Actual Behavior

It seems not to be able to get it.

### Minimal graphql.schema and models to reproduce

I made small demo project.
Is there a good way to do it?

https://github.com/vvakame/til/blob/ad9d192d297b4e0be358f3273442502497ff50fb/graphql/gqlgen-directive-parent/schema.graphql#L9
https://github.com/vvakame/til/blob/ad9d192d297b4e0be358f3273442502497ff50fb/graphql/gqlgen-directive-parent/server/server.go#L49-L56

JulienBreux:
Good question ;)

jekaspekas:
Hi!
The short answer is nothing.
```golang
// directive implement func
resp, err := next(ctx)
```
resp always contains the type of the field to which the directive is assigned and you receive a compilation error, or runtime error if you use a directive for fields with different types.
Performing a next(ctx) function is dangerous because if this directive assigned to mutation field, you run change this field anyway.
Also, you can't use correctly CurrentUser from context, because the context is recreated every request.

But if you use a websocket as a transport, you can see that wsConnection ("gqlgen/handler/websocket.go") struct is not change on every request. Exportable WSConnection with new CurrentUser field you can write to context. But this will require changing the package "99designs/gqlgen/handler" (you can copy this to project directory).
handler/websocket.go:
```golang
...
type WSConnection struct {
  ctx    context.Context
  conn   *websocket.Conn
  exec   graphql.ExecutableSchema
  active map[string]context.CancelFunc
  mu     sync.Mutex
  cfg    *Config
  // new field
  CurrentUser *gqlgen_directive_parent.User
}
type key string
const WSConnectionKey key = "WSConnectionKey"
...
func (c *WSConnection) subscribe(message *operationMessage) bool {
...
  reqCtx := c.cfg.newRequestContext(doc, reqParams.Query, vars)
  ctx := graphql.WithRequestContext(c.ctx, reqCtx)
  // write WSConnection into context
  ctx = context.WithValue(ctx, WSConnectionKey, c)
...
}
```
ID comparison must be performed at todoResolver func.
```golang
func (r *todoResolver) Text(ctx context.Context, obj *gqlgen_directive_parent.Todo) (string, error) {
  conn, ok := ctx.Value(handler.WSConnectionKey).(*handler.WSConnection)
  if !ok {
[...]

PR #301, opened 2018-08-21
add Result field to ResolverContext

vvakame:
refs #289
works fine https://github.com/vvakame/til/commit/8ddb178d489044de6c3daa137c9e38229289cfdf#diff-533b5224b1fd11ab427fe7114c0fdc13


vvakame:
this PR collide with #302 🤔 

This PR introduce `Result` field to `ResolverContext`. and I refactor `ResolverContext` that looks like immutable.

codegen/object.go
```diff
@@ -215,12 +215,13 @@ func (f *Field) doWriteJson(val string, remainingMods []string, astType *ast.Typ
 
 		return tpl(`{{.arr}} := graphql.Array{}
 			for {{.index}} := range {{.val}} {
-				{{.arr}} = append({{.arr}}, func() graphql.Marshaler {
-					rctx := graphql.GetResolverContext(ctx)
-					rctx.PushIndex({{.index}})
-					defer rctx.Pop()
+				{{.arr}} = append({{.arr}}, func(ctx context.Context) graphql.Marshaler {
+					rctx := &graphql.ResolverContext{
```
vektah: this is strange. array unmarshalling isn't a resolver, and it shouldn't need a new context?

codegen/templates/object.gotpl
```diff
@@ -30,6 +30,8 @@ func (ec *executionContext) _{{$object.GQLType}}(ctx context.Context, sel ast.Se
 		ctx = graphql.WithResolverContext(ctx, &graphql.ResolverContext{
 			Object: {{$object.GQLType|quote}},
 		})
+	{{else}}
+		graphql.GetResolverContext(ctx).Result = obj
```
vektah: This should probably happen using res around field.gotpl:62, here its going to be repeated for every child

vvakame: ah... make sense... 
It was a bit different from what I thought.
so I'll write what I want to do.

vvakame:
@vektah Result and ParentObject is not equals.
Resolver make `[]Todo`, this is Result.
I think that Directive want a `Todo`, this is ParentObject.
so, I want to implement both. what do you think about this opinion?

codegen/object.go
vvakame: yes, it is right.
this change required `default:` case under `case len(remainingMods) > 0 && remainingMods[0] == modList:`.
I'll move this code to right place and add some arguments to method signature.

vvakame:
I noticed that there are multiple ideas of ResolverContext...
1. Definition closely related to the current Resolver implementation.
2. Every time Path changes, a new child ResolverContext is created.
I think the 2 is more convenient.
For example, if we want to issue an error in Resolver for some Index in Resolver that returns Array, we need to be able to generate new child ResolverContext ourself.
Specifically, if an illegal id is passed to `nodes (ids: [ID!]!): [Node]!`.

Also, when users browse / create ResolverContext, it is not known where asynchronous processing will be occurred.
For this reason, I think that ResolverContext created once, it should be immutable.
So, I think it is complicated to use `PushIndex` and` Pop`, and I think it is dangerous.

vvakame:
It might be well refactored. 🤔

vektah:
I still think its a bit weird to be creating a new resolver context for each result, currently all of that state stays on the stack as method args. Perhaps this should work the same? see #314 for an alternative.

graphql/context.go
```diff
@@ -82,19 +82,18 @@ type ResolverContext struct {
 	Args map[string]interface{}
 	// The raw field
 	Field CollectedField
-
-	// indicies tracks the array indicies only. all field aliases come from the context stack
-	indicies []int
+	// The index of array in path.
+	Index *int
```
vektah: then again I think this is a bit nicer.

PR #302, opened 2018-08-21
fix index generation

vvakame:
https://play.golang.org/p/5R_inyGjapb


vektah:
The func should probably capture this as an argument:
```go
arr1 = append(arr1, func(idx1 int) graphql.Marshaler {
	// ...
	return arr2
}(idx1))
```

and a test would be nice, this is pretty easy to miss.

vvakame:
Where do you write the test now?

vektah:
If you can do it without changing the schema, use one of the examples.

Otherwise, dump it in https://github.com/99designs/gqlgen/blob/master/codegen/testserver/generated_test.go

vvakame:
this function is an immediate evaluate, it is not a problem even with the original code .... I have just noticed.

PR #314, opened 2018-08-24
Add obj to Directives

### LINKED ISSUE #289: Question: How can we get parent object in directive function

vektah:
Alternative solution to #301 that passes in obj into directive funcs, instead of creating new ResolverContexts around each element in array responses. Includes an updated todo example to show the usage.

Fixes #289

I have:


vektah:
Going to ship both of them, I think this is a better interface to use most of the time, but reaching up into an ancestor is sometimes useful.

Going to be making some changes to the json marshalling to fix parallelization so I want to get this stuff shipped.

vvakame:
wow, I just reading this PR. suddenly #301 was merged and I was surprise a bit 😉 
This PR seems to works fine for my issue. 👍

vvakame:
thanks!

Issue #313, opened 2018-08-23
Investigate performance regression due to simplification of concurrent resolver logic


dvic:
As discussed in #294, I noticed an increase in the number of queries produced by my dataloaders after #294.

lwc:
Hey @dvic this should be fixed on master now - can you confirm? thanks

dvic:
@lwc yes now it works again as before with the latest changes on master, thanks!

PR #316, opened 2018-08-24
use goroutine about processing each array elements

vvakame:
refs #313
#301 makes ResolverContext immutable.
We can use concurrent processing safety. 
cc/ @dvic 

I have:

performance regression test is so hard to writing... 🤔

### DIFF
--- a/codegen/object.go
+++ b/codegen/object.go
@@ -212,28 +212,37 @@ func (f *Field) doWriteJson(val string, remainingMods []string, astType *ast.Typ
 		}
 		var arr = "arr" + strconv.Itoa(depth)
 		var index = "idx" + strconv.Itoa(depth)
+		var wg = "wg" + strconv.Itoa(depth)
 		var usePtr bool
 		if len(remainingMods) == 1 && !isPtr {
 			usePtr = true
 		}
 
-		return tpl(`{{.arr}} := graphql.Array{}
+		return tpl(`
+			{{.arr}} := make(graphql.Array, len({{.val}}))
+			var {{.wg}} sync.WaitGroup
+			{{.wg}}.Add(len({{.val}}))
 			for {{.index}} := range {{.val}} {
-				{{- if not .isScalar }}
-					rctx := &graphql.ResolverContext{
-						Index: &{{.index}},
-						Result: {{ if .usePtr }}&{{end}}{{.val}}[{{.index}}],
-					}
-					ctx := graphql.WithResolverContext(ctx, rctx)
-				{{- end}}
-				{{.arr}} = append({{.arr}}, func() graphql.Marshaler {
-					{{ .next }} 
-				}())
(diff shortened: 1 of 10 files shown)


codegen/object.go
```diff
@@ -212,28 +212,37 @@ func (f *Field) doWriteJson(val string, remainingMods []string, astType *ast.Typ
...
-					rctx := &graphql.ResolverContext{
-						Index: &{{.index}},
-						Result: {{ if .usePtr }}&{{end}}{{.val}}[{{.index}}],
-					}
-					ctx := graphql.WithResolverContext(ctx, rctx)
-				{{- end}}
-				{{.arr}} = append({{.arr}}, func() graphql.Marshaler {
-					{{ .next }} 
-				}())
+				go func({{.index}} int) {
```
vektah: we should probably avoid creating a new gofunc if:
 - there is only one element in the array
 - the child field doesn't have a resolver (accessing properties is cheap and should happen sequentially)

codegen/object.go
```diff
@@ -212,28 +212,37 @@ func (f *Field) doWriteJson(val string, remainingMods []string, astType *ast.Typ
 		}
 		var arr = "arr" + strconv.Itoa(depth)
 		var index = "idx" + strconv.Itoa(depth)
+		var wg = "wg" + strconv.Itoa(depth)
```
vektah: this probably doesn't need to be nested, and can be shared by the whole field.

vektah:
Don't worry about testing performance, it shouldn't be too to test that `{ foo { bar { baz } } }` runs all `baz` concurrently, without any of them needing to complete.

vvakame:
updated!

codegen/object.go
```diff
@@ -212,37 +212,54 @@ func (f *Field) doWriteJson(val string, remainingMods []string, astType *ast.Typ
...
+				{{- end}}
 			}
-			{{.wg}}.Wait()
+			{{ if and .top (not .isScalar) }} wg.Wait() {{ end }}
 			return {{.arr}}`, map[string]interface{}{
 			"val":      val,
 			"arr":      arr,
 			"index":    index,
-			"wg":       wg,
+			"top":      depth == 1,
```
vektah: This probably can just go into WriteJson and be prefixed on? Doing it here in the recursive part is a bit hard to reason about.

vektah:
This really needs some tests :(

vvakame:
`integration/` is good test case. It caught my bug in this PR 😺

codegen/object.go
vvakame: It is little bit difficult.
We should exec `wg.Wait()` **before** return statement.
https://github.com/99designs/gqlgen/pull/316/files#diff-3995ea1c8b08158971f7c289f7f7d50aR256

PR #318, opened 2018-08-26
Add query cache

edsrzf:
This PR adds a query cache with a configurable maximum size. Past this size, queries are evicted from the cache on an LRU basis.

The default cache size is 1000, chosen fairly arbitrarily. If the size is configured with a non-positive value, then the cache is disabled.

Also ran `dep ensure` to add the new dependency to `Gopkg.lock`.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -26,7 +26,7 @@
   version = "v3.3.2"
 
 [[projects]]
-  digest = "1:78907d832e27dbfc6e3fdfc52bd2e5e2e05c1d0e3789d4825b824489fbeab233"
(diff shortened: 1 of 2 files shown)


vektah:
Neat.

Can you see any reason to change the implementation? It might be worth just shipping a good standard one. Given its an object tree and parsing isn't that expensive I can't see someone using memcache etc?

edsrzf:
Yeah, that's a fair point. My initial thinking was that people might want to tweak the caching/eviction policy, but it'd probably be fairly niche. Anyone that does need it can make their own handler, which isn't really that hard.

I'll have another go with an LRU cache and maybe just a configurable limit.

edsrzf:
Updated. I force-pushed since I didn't feel like a diff against the previous revision was useful.

edsrzf:
Updated

Issue #118, opened 2018-05-19
Add comments through generated code


iamclaytonray:
### Expected Behaviour
`gqlgen` strips away comments. Is this intentional? If not, is this a bug or a feature to be added in the future?

```go
// This is a User
type User struct {
  FirstName: string
  ID: string
}
```

### Actual Behavior
```go
type User struct {
  FirstName: string
  ID: string
}
```

### Minimal graphql.schema and models to reproduce
```gql
# This is a User
type User {
  firstName: String!
  id: ID!
}
```

vektah:
Not intentional, it just hasn't been built yet. Model generation is is pretty basic at the moment.

iamclaytonray:
Cool. I'll play around with this

andrioid:
@iamclaytonray Any progress?
@vektah Can you elaborate on what needs to be done for this to work? Maybe I can take a look.

vektah:
It should be pretty straightforward, all the descriptions have been parsed they just aren't being put in the generated models.

Enums do though, so heres an example https://github.com/99designs/gqlgen/blob/master/codegen/templates/models.gotpl#L31


The example provided isn't quite right any more, because descriptions are strings, not comments, so it should be:
```
"""This is a User"""
type User {
  firstName: String!
  id: ID!
}
```

andrioid:
Sorry to bother you again. But does this mean that if you're using your own models and models.go is not being generated, that the descriptions will not be populated?

vektah:
This issue is only for adding comments to the generated code, descriptions should already be exposed via introspection. If they arent showing up its probably because you're using comments instead of strings. 
```
# This is a comment, it wont do anything
type User {

"""This is a description, it will be exposed via introspection and maybe added to generated code"""
type User {
```

PR #320, opened 2018-08-26
Description added to generated Model code

andrioid:
I decided to take a stab at: https://github.com/99designs/gqlgen/issues/118

GraphQL descriptions are now written as comments on top of some of the Go types. I'm not sure if this covers all the basics. But it would be nice if someone could take a look and see if I'm on the right track.


andrioid:
Sorry I broke the build. It seems there are some newlines that don't match the expected code generation.

I have never used Go templates for anything before, so I'm a bit lost in how to fix that.

codegen/templates/models.gotpl
```diff
@@ -9,10 +9,13 @@ import (
 )
 
 {{ range $model := .Models }}
+	{{with .Description}} {{.|prefixLines "// "}} {{end}}
+
 	{{- if .IsInterface }}
 		type {{.GoType}} interface {}
 	{{- else }}
 		type {{.GoType}} struct {
+
```
vektah: Another extra newline

codegen/templates/models.gotpl
```diff
@@ -9,10 +9,13 @@ import (
 )
 
 {{ range $model := .Models }}
+	{{with .Description}} {{.|prefixLines "// "}} {{end}}
+
```
vektah: One of the newlines are here (use `{{- with .Description}}`) to remove it.

codegen/templates/models.gotpl
```diff
@@ -25,12 +28,13 @@ import (
 {{- end}}
 
 {{ range $enum := .Enums }}
+	{{with .Description}} {{.|prefixLines "// "}} {{end}}
 	type {{.GoType}} string
 	const (
-	{{ range $value := .Values -}}
+	{{ range $value := .Values}}
```
vektah: these -}} were removing unnecessary newlines

vektah:
There are also field level descriptions that could be added

andrioid:
I did briefly look into the FieldValue yesterday, but couldn't find Description to add.

I'll try to look into the newlines in the next few days.

vektah:
Awesome, thanks for the PR

PR #323, opened 2018-08-28
Add stub command

### LINKED ISSUE #216: gqlgen sync
It would be very useful to compare current schema to a set of RootResolvers (passed in as args? maybe in config?) and add any missing resolvers as stubs to the generated code.

Bonus points for dealing with conflicts.

This is a long standing idea thats been floating around since #9.

vektah:
WIP - very incomplete
===

This adds a new `gqlgen stubs` command that compares the generated interfaces and creates stubs for any missing ones, updating and signatures along the way.

Because this is mutating existing user-written source code I want to keep it in a separate command for now. As we gain some confidence in it I imagine it would happen automatically in `gqlgen generate` if configured.

Fixes #216 

I have:


mathewbyrne:
Closing for now, some of this might be useful later.

frederikhors:
@mathewbyrne why later and not now?

mathewbyrne:
> why later and not now?

I'm not sure what you mean. This PR itself will never merge, but some sections of it may be useful when we do address stub generation at some point.  We are currently addressing some more fundamental refactors to code generation and type mapping that are more of a priority at the moment.

PR #325, opened 2018-08-28
Only allow query operations on GET requests

### LINKED ISSUE #317: HTTP handler should not allow mutations on GET requests
The HTTP handler should not allow mutations on HTTP GET requests. Or at least make it configurable. This is asking for a CSRF vulnerability.

Alternatively, don't allow running queries on GET requests at all. GitHub's GraphQL API, for example, does the standard introspection query when you send it a GET request.

edsrzf:
This mitigates the risk of CSRF attacks.

Closes #317.


vektah:
Nice one!

PR #330, opened 2018-09-02
Use built-in less than operator instead of strings.Compare

edsrzf:
Just a small codegen simplification.


vektah:
:+1: good catch, no codegen chanages. Nice.

PR #331, opened 2018-09-03
Refactor arg codegen

edsrzf:
Now a function is generated for each field and directive that has arguments. This function can be used by both field methods as well as the `Complexity` method.

The `args.gotpl` template now generates the code for this function, so its purpose is a little different than it used to be.


vektah:
Looks like the complexity changes aren't included in this PR - It would help validate that the new code is fit for both purposes.

PR #338, opened 2018-09-11
Adds docs for how resolvers are bound

codyleyhan:
Adds documentation about all the different ways that resolvers can be bound 

I have:

### DIFF
--- a/docs/content/reference/resolvers.md
+++ b/docs/content/reference/resolvers.md
@@ -128,14 +128,14 @@ models:
 ```
 
 Here even though the graphQL type and Go struct have different field names, there is a Go struct tag field on ```longState```
-That matches and thus ```state``` will be bound to ```LongState```
+that matches and thus ```state``` will be bound to ```LongState```.
 
 
 The second way you can bind fields is by adding a line into the config file such as:
 ```go
 type Car struct {
     Make string
-    ShortState string 
+    ShortState string
     LongState string
     Model string
     Color string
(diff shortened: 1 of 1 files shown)


docs/content/reference/resolvers.md
```diff
@@ -0,0 +1,203 @@
...
+            state:
+                fieldName: LongState
+```
+
+## Binding to Anonymous or Embedded Structs
+All of the rules from above apply to a struct that has an embedded struct.
+
+```go
+type Truck {
+    car
```
vektah: should be `Car`?

docs/content/reference/resolvers.md
```diff
@@ -0,0 +1,203 @@
...
+
+    is4x4 bool
```
vektah: should be `Is4x4`?

docs/content/reference/resolvers.md
```diff
@@ -0,0 +1,203 @@
...
+```graphql
+type Truck {
+    make: String!
+    model: String!
+    state: String!
+    color: String!
+    odometerReading: Int!
+    is4x4: Bool!
+}
+```
```
vektah: May be worth mentioning embedding is also a good way to create thin presenters over database types without needing resolvers, eg

```
type Cat struct {
    db.Cat
}

func (c *Cat) ID() string {
     return fmt.Sprintf("%d:%d", c.Shard, c.Id)
}
```

vektah:
Thanks for the PR, looks good!

Issue #228, opened 2018-07-26
Plugin System


mathewbyrne:
As a part of furthering our support of directives, we would like to provide a system in the `gqlgen` binary to allow user and 3rd part code to interact with codegen.  Some of the use-cases we are considering include:

- Allowing a directive to generate a resolver for a field — e.g. `rest`/`grpc` calls.
- Allowing directive definitions and types to a schema.
- Altering a schema prior to codegen — e.g. enabling easy Relay connections.

The current plan is a plugin system which exposes a mechanism for allowing external plugins or code to register to be notified at various points during `gqlgen`'s execution, and modify the schema and codegen configuration.

These plugins are probably loaded and configured through the existing configuration file.  The plugins will probably be provided with the schema AST, and configuration that can be modified as required.

We're looking to start building this soon, but would welcome any additional use-cases for consideration.

ermik:
There should be a generic 'middleware' plugin for things like OpenTracing to patch in to request/response and resolution logic. E.g. sometimes it is important to know exactly how long unmarshall/marshall took.

vektah:
I think that already exists @ermik Did you see https://github.com/99designs/gqlgen/tree/master/opentracing and the example use in https://github.com/99designs/gqlgen/blob/master/example/dataloader/server/server.go?

ermik:
I saw it, hence the change in the wording of my comment. It's one middleware. If you fully support middleware pattern at a 'plugin' level I think there would be more pluggability (is that a real word?) without much additional development needed.

dannymidnight:
+1 for plugins. 

It would be great to have some contrib plugins for common directives e.g. form input validation  https://www.npmjs.com/package/graphql-constraint-directive

vetcher:
I see here 3 variants (or any combinations):
1. take approach from `protoc`: plugins is an executable, transport via stdin
1. use golang (plugin)[http://golang.org/pkg/plugin] package: define interface some interface, with will be used to call plugins from imported .so files
1. some template engine, e.g. default text/template: templates will override existing parts of code (functions/files?) or adds their own

I was thinking about same extention system for my codegen and may help with realization

mathewbyrne:
> take approach from protoc: plugins is an executable, transport via stdin

We did discuss this briefly, but we're not super keen on transport like that.  It adds a lot of complexity in coming up with a protocol, loses type safety etc.  It has the benefit of being easy to distribute.

> use golang (plugin)

[...]

PR #347, opened 2018-09-18
Custom Directive Implementations

mathewbyrne:
This adds a new configuration option for a user to bypass codegen for a directive and provide their own implementation like so:

```yml
directives:
  exampleDirective:
    implementation: "github.com/99designs/gqlgen.ExampleDirective"
```

This feature is primarily intended to support plugins (#228) where a plugin could provide an implementation during codegen and avoid it being added to the `DirectiveRoot`.

This is currently a pretty bare bones implementation, so I'd welcome any feedback.  One major question is how do we provide run-time dependencies to custom implementations?  Also is there a way we can provide a no-op for plugins that don't need a runtime implementation for a directive?

**TODO**
- [x] Validate implementation signature
- [x] Better test coverage


JulienBreux:
Nice work. Maybe need test?

mathewbyrne:
Yep, there's a couple of todos left on this at the moment; I thought I'd get it up for any feedback first.

codegen/templates/generated.gotpl
```diff
@@ -226,6 +226,20 @@ func (ec *executionContext) FieldMiddleware(ctx context.Context, obj interface{}
 		switch d.Name {
 		{{- range $directive := .Directives }}
 		case "{{$directive.Name}}":
+			{{- if $directive.Implementation }}
```
vektah:
```
{{- with $directive.Implementation}}
    // args
    {{ . }}
{{-end}}
```
?

also the indenting went a bit funky here, aim for readable templates, gofmt will clean the code up.

mathieupost:
Is this still possible? Cant seem to find documentation about it and if I try to use it like this 
```
directives:
  exampleDirective:
    implementation: "github.com/99designs/gqlgen.ExampleDirective"
```
I'm getting 
```
unable to parse config: yaml: unmarshal errors:
  line 41: field implementation not found in type config.DirectiveConfig
```

Issue #343, opened 2018-09-13
[help] How to read HTTP headers in subscription context?


JulienBreux:
Hello!

It's me again with a big thanks before.

My use case:

1. I've an header with `Authorization` key and `Bearer xxx.xxx.xxx` value
2. But my question is how to read in `ctx context.Context` of subscription to access to the value of header?

JulienBreux:
```go
func (sr *subscriptionResolver) Updated(ctx context.Context) (<-chan model.ObjectReturned, error) {
   // ctx? to read Authorization header value?
}
```

JulienBreux:
<img width="681" alt="screen shot 2018-09-14 at 12 47 29" src="https://user-images.githubusercontent.com/964330/45546143-b8927080-b81c-11e8-9365-a50e7841ccac.png">

With Apollo, for example. `Authorisation` is transmitted in the `connection_init` by `Payload`.
But in https://github.com/99designs/gqlgen/blob/master/handler/websocket.go#L82, the payload never handled at init.

```go
func (c *wsConnection) init() bool {
//...
	case connectionInitMsg:
		c.write(&operationMessage{Type: connectionAckMsg})
//...
}
```

How to read `Payload` at the initialisation of the WS ?

JulienBreux:
Up! Do you have an idea?

We can't use authenticated `subscriptions` with `gqlgen`?

JulienBreux:
Maybe the idea is to write the init Payload in the wsConnection? In `initPayload json.RawMessage`

DenisNeustroev:
Maybe pass your token in extensions field, it should work the same for http and ws

JulienBreux:
@DenisNeustroev how to?

vektah:
You can use http headers and some middleware that puts data into context, same as query.

Unfortunately websocket init payloads aren't currently supported (PRs welcome)

JulienBreux:
Thanks Adam, but this solution/recipe doesn't work.

The context between HTTP and WS seem different.
I can't retrieve my HTTP context in a subscription context :/

vektah:
Closing in favour of #355

Issue #242, opened 2018-08-01
[help] About Authorization HTTP header


JulienBreux:
Hello,

Thanks a lot for this amazing work!

I want to check JWT permissions from the Authorization HTTP header and inject identity in context to use in resolvers. 

Should I use `RequestMiddleware` (no access to HTTP headers?)? `ResolverMiddleware` (no access to HTTP headers)? `http.Handler` middleware (how to inject in graphql context)?

Thanks guys :)

JulienBreux:
Okey, I use a standard `http.Handler` as middleware :)

PR #348, opened 2018-09-19
Added parsing of the websocket init message payload

gissleh:
This PR adds parsing of the init payload sent when initializing the websocket transport, and adds functions to get this data from context. I also added support for it in the client package as a separate function since I didn't think it right to have it in options because it's transport specific.

I'm making this in response to the call for a PR in #343

I have:


handler/websocket.go
```diff
@@ -63,33 +63,42 @@ func connectWs(exec graphql.ExecutableSchema, w http.ResponseWriter, r *http.Req
...
+		return nil, false
 	}
 
+	initPayload = make(graphql.InitPayload)
+
 	switch message.Type {
 	case connectionInitMsg:
+		err := json.Unmarshal(message.Payload, &initPayload)
+		if err != nil {
+			// Treat an invalid payload as no payload
```
vektah: maybe for an empty payload this is what you want, but squashing all errors isn't good.

handler/websocket.go
```diff
@@ -63,33 +63,42 @@ func connectWs(exec graphql.ExecutableSchema, w http.ResponseWriter, r *http.Req
...
+	if !ok {
 		return
 	}
 
-	conn.run()
+	conn.run(initPayload)
 }
 
-func (c *wsConnection) init() bool {
+func (c *wsConnection) init() (initPayload graphql.InitPayload, ok bool) {
```
vektah: wsConnection is stateful per connection, the init payload could be added to there to avoid passing it around?

graphql/context.go
```diff
@@ -195,3 +195,55 @@ func (c *RequestContext) RegisterExtension(key string, value interface{}) error
...
 	return nil
 }
+
+const (
+	initpayload key = "initpayload_context"
+)
+
+// InitPayload is a structure that is parsed from the websocket init message payload. TO use
+// request headers for non-websocket, instead wrap the graphql handler in a middleware.
+type InitPayload map[string]interface{}
```
vektah: Does leaving this in the `handler` package cause an import loop? It doesn't feel right in here, surrounded by other graphql transport agnostic stuff.

handler/websocket.go
gissleh: I could change that to only run this when there is a payload, and pass the error up.

handler/websocket.go
gissleh: That's a good idea, actually. I'm not sure why I didn't think of that.

graphql/context.go
gissleh: I'll look into it.

gissleh:
> It looks like your expecting the user to wrap everything with their own (mutable?) context and call into this context to seed it?

As of right now that's the most efficient way of doing it to avoid multiple trips to the DB or session store. I'm also investigating adding a middleware option where the user can wrap the context so that every call to the `auth.FromContext` getter will behave as before, but I'm not sure how to contain that in websocket-specific code.

JulienBreux:
A big thanks, related to #242.

gissleh:
I didn't find a neat transport-specific solution to the problem of not having a common point of entry to wrap the context with appliaction-specific authorization, so for now the least inefficient solution would be to have every request be given a mutable session object that lazily handles the auth check / user load on the first call.

vektah:
I'm going to ship this as it's not a BC break and enables features that people want. 

I've created #355 to track clean up the desync between the two interfaces.

sijad:
for feature reference

this PR add ability to access to object that sent by [GQL_CONNECTION_INIT](https://github.com/apollographql/subscriptions-transport-ws/blob/faa219cff7b6f9873cae59b490da46684d7bea19/PROTOCOL.md#gql_connection_init)

client example:

```js
new WebSocketLink({
  uri,
  options: {
    reconnect: true,
    connectionParams: async () => {
      return {
        my_session_key: await getSession(),
      };
    },
  },
});
```

server example:

```go
import
    "log"

	"github.com/99designs/gqlgen/handler"
)

func (r *subscriptionResolver) MySubscription(ctx context.Context) (<-chan string, error) {
	// ...
	log.Println(handler.GetInitPayload(ctx).GetString("my_session_key"))
	// ...
}
```

PR #353, opened 2018-09-25
Parent middleware generated contexts

vektah:
Currently any context generated by resolver middleware is only passed to the resolver, not to the children. This means tracing middleware like OpenTracing and OpenCensus cant correctly set the parent span info. 

This is an alternative to #327 that uses the context that would be passed to the resolver, instead of returning one from the outer most middleware. 

Advantages:
 - Correct order when using multiple middleware
 - Not a BC break

I haven't addressed setting context from resolvers, but it's more obvious that you cant manipulate the context from inside a resolver (no next func). If we need it we can add a `ChildContext` prop to `ResolverContext` with a method that lets the user change it.

Thoughts @vvakame?

I have:


vvakame:
ah, looks good!
I'll try this change in my environment 👍

leebenson:
> I haven't addressed setting context from resolvers, but it's more obvious that you cant manipulate the context from inside a resolver (no next func). If we need it we can add a ChildContext prop to ResolverContext with a method that lets the user change it.

👍 We're running into exactly this issue with auth context that could do with being set inside a resolver.

Would you be open to a PR to add the `ChildContext` you mentioned?

Issue #356, opened 2018-09-26
gqlgen fail to find a field in a pointer embedded struct


MichaelMure:
Simple case to reproduce:

```go
type OpBase struct {
	Author Person `json:"author"`
}

type SetTitleOperation struct {
	*OpBase
	Title string `json:"title"`
	Was   string `json:"was"`
}
```

```graphql
type SetTitleOperation implements Operation & Authored {
  author: Person!

  title: String!
  was: String!
}
```

When `OpBase` is embedded as value, gqlgen find the `Author` field. When `OpBase` is embedded as a pointer, gqlgen fail to do that and ask me to implement a resolver, which would simply be:

```go
func (setTitleOperationResolver) Author(ctx context.Context, obj *operations.SetTitleOperation) (bug.Person, error) {
	return obj.Author, nil
}
```

## Expected behavior

gqlgen find the field and generate the proper resolver by itself.

MichaelMure:
If you need more details, the full code is here: https://github.com/MichaelMure/git-bug/tree/master/graphql

MichaelMure:
I have another breakage that is probably related:

```
panic: unexpected type operations.AddCommentOperation
goroutine 1029 [running]:
github.com/MichaelMure/git-bug/graphql/graph.(*executionContext)._Operation(0xc420144650, 0x99a660, 0xc4202e0030, 0xc420175340, 0x7, 0x7, 0xc420390f10, 0x0, 0xc4201d3f48)
	/home/travis/gopath/src/github.com/MichaelMure/git-bug/graphql/graph/gen_graph.go:5530 +0x43d
github.com/MichaelMure/git-bug/graphql/graph.(*executionContext)._OperationConnection_nodes.func2.1(0xc420144650, 0x99a660, 0xc4202e0030, 0xc420151980, 0xc420175340, 0x7, 0x7, 0xc420390e00, 0x12, 0x20, ...)
	/home/travis/gopath/src/github.com/MichaelMure/git-bug/graphql/graph/gen_graph.go:3322 +0x90
github.com/MichaelMure/git-bug/graphql/graph.(*executionContext)._OperationConnection_nodes.func2(0x11)
	/home/travis/gopath/src/github.com/MichaelMure/git-bug/graphql/graph/gen_graph.go:3323 +0xb2
created by github.com/MichaelMure/git-bug/graphql/graph.(*executionContext)._OperationConnection_nodes
[...]

PR #358, opened 2018-10-01
Fix Embedded Pointer

mathewbyrne:
Fixes the first part of #356 

`codegen.findField` was not accounting for anonymous pointer types and was generating a redundant resolver.  This change fixes this so that these pointers are unwrapped and the element type is considered recursively as a match for the field.

### DIFF
--- a/codegen/util.go
+++ b/codegen/util.go
@@ -132,17 +132,16 @@ func findField(typ *types.Struct, name, structTag string) (*types.Var, error) {
 		}
 
 		if field.Anonymous() {
-			if named, ok := field.Type().(*types.Struct); ok {
-				f, err := findField(named, name, structTag)
-				if err != nil && !strings.HasPrefix(err.Error(), "no field named") {
-					return nil, err
-				}
-				if f != nil && foundField == nil {
-					foundField = f
-				}
+
+			fieldType := field.Type()
+
+			if ptr, ok := fieldType.(*types.Pointer); ok {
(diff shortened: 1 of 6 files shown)


codegen/util.go
```diff
@@ -132,17 +132,16 @@ func findField(typ *types.Struct, name, structTag string) (*types.Var, error) {
...
-					return nil, err
-				}
-				if f != nil && foundField == nil {
-					foundField = f
-				}
+
+			fieldType := field.Type()
+
+			if ptr, ok := fieldType.(*types.Pointer); ok {
+				fieldType = ptr.Elem()
```
vektah: what happens when the embedded pointer is null? Worth adding a test?

mathewbyrne: Yep, good question, nothing will happen here. But I would expect a runtime panic. What do you think we should do?

vektah: can it return null and do the normal error bubbling dance?

mathewbyrne: Yeah ok, I'll come up with a test.  Might involve slightly more messing around here since we'd need to know if it was a pointer that we matched to in template.

mathewbyrne: Ok, after reviewing this a bit I _think_ this is going to be a bigger change.  Embeds can be pathological when they are more than a single embed deep with a mix of pointers thrown in.

I think for now we merge this and maybe log that as a separate issue.

PR #359, opened 2018-10-01
Fix Issue With Argument Pointer Type

### LINKED ISSUE #346: null parameters cause crashes
I'm dealing with a swath of issues around `gqlgen`'s implementation of defaults and nullable types. They seem bizarre, so I thought the issue was on my end, but i have been unable to find a sold root cause within my minimal code.

### Expected Behaviour
When a null is passed as a parameter, it should be replaced with its default parameter.

### Actual Behavior
Generated code attempts to parse `<nil>`, and then rejects the request.

```gql
type Query {
   history(start: Time = "anything", end: Time = "anything", limit: Int = 10)
}
```
image
image

mathewbyrne:
In #346 there was an issue revealed where a non-nullable argument with a default value was not passing an input `null` through to the resolver, and instead was giving an error.

The reason for this is that if an argument has a default and is a pointer (nullable), the pointer is stripped.  This cha This change removes this behaviour and keeps type a pointer.  If an argument is nullable, then null is always a valid value.

Fixes #346


mathewbyrne:
@vektah you think this is worth getting on `master` or BC enough for `next`?

PR #365, opened 2018-10-02
Don't let goimports guess import paths

### LINKED ISSUE #207: Ambiguous package names lead to incorrect import statements
This is basically just re-opening #60.  As mentioned there, as of commits abf85a1 / c0158f5 (the former doesn't build, so the last commit that works for me is a39c63a) that bug seems to have crept back into the code.

For example, say I have a package called `foo/query` and upon re-generating the code past that commit, I get the following diff:

```diff
--- a/foo/apigraphql/generated.go
+++ b/foo/apigraphql/generated.go
@@ -1,4 +1,4 @@
-// This file was generated by github.com/vektah/gqlgen, DO NOT EDIT
+// Code generated by github.com/vektah/gqlgen, DO NOT EDIT.
 
 package apigraphql
 
@@ -6,33 +6,154 @@ import (
        "bytes"
        context "context"
        apimodel "foo/apimodel"
-       query1 "foo/query"
+       query "foo/query"
        strconv "strconv"
 
        graphql "github.com/vektah/gqlgen/graphql"
        introspection "github.com/vektah/gqlgen/neelance/introspection"
-       query "github.com/vektah/gqlgen/neelance/query"
        schema "github.com/vektah/gqlgen/neelance/schema"
 )
 
```
as a result the code obviously doesn't compile anymore:
```
$ go test ./...
# foo/apigraphql
./generated.go:167:59: undefined: "foo/query".Operation
```

vektah:
gqlgen internally tracks all the imports used in a file and adds them to the import block, handling duplicate imports by appending names.

There are two sources of imports:
 - imports for user generated models, types in graphql, etc. These are all tracked as they are added, with the required deduping behaviours etc.
 - `ambientImports`: these are referenced in the templates and must keep the name they were given in the templates, eg `fmt.Println(` must get the import name `fmt`. Often these are in conditional parts of the templates and may not be used in a particular generated output.

Currently we run the generated code through `goimports`, this does a few things for us:
 - format the code: same as gofmt
 - group and sort imports: this makes our output a little more robust to changes
 - remove unused imports: required for ambient imports that don't actually get used.
 - add missing imports: **DANGER** This adds in any missing imports, these often end up being ambient imports that were forgotten about. This leads to subtle and hard to reproduce issues like #207 when a user imported package grabs the name used in by the import in the template.

This PR removes the "add missing imports" feature of goimports, while maintaining the other 3 by reimplementing the import removal logic and calling goimports with "Format only" to get grouping and sorting. 

Fixes #207

### DIFF
--- a/.gometalinter.json
+++ b/.gometalinter.json
@@ -9,5 +9,6 @@
       "PartitionStrategy": "packages"
     }
   },
+  "Skip": ["internal/imports/testdata"],
   "Disable": ["gas","golint","gocyclo","goconst", "gotype", "maligned", "gosec"]
 }
--- a/codegen/build.go
+++ b/codegen/build.go
@@ -112,6 +112,13 @@ func (cfg *Config) server(destDir string) *ServerBuild {
 	imports.add(cfg.Exec.ImportPath())
 	imports.add(cfg.Resolver.ImportPath())
 
+	// extra imports only used by the server template
+	imports.add("context")
+	imports.add("log")
+	imports.add("net/http")
+	imports.add("os")
+	imports.add("github.com/99designs/gqlgen/handler")
+
 	return &ServerBuild{
 		PackageName:         cfg.Resolver.Package,
 		Imports:             imports.finalize(),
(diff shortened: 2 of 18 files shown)


jufemaiz:
QQ on this: is there a way of implementing `local-prefixes` option for goimports?

PR #369, opened 2018-10-04
reverse errors and data order in response

vetcher:
Following the trend and recommendations in graphql: https://github.com/facebook/graphql/pull/384

> Note: When `errors` is present in the response, it may be helpful for it to 
appear first when serialized to make it more clear when errors are present
in a response during debugging.


vektah:
Happy to merge this, but the tests need to be updated.

PR #388, opened 2018-10-18
Resolver move into FieldMiddleware processing

vvakame:
I tried #353 today. It works fine!

## Now

I got below trace on DataDog. ✨ 
datadog-trace-view

## Problem

https://docs.datadoghq.com/tracing/visualization/#trace
the current problem is that parent span closes before child span closes.
Why it happens is that Resolver will work after Middleware's operation is finished.
Middleware will open & close Trace Span. but it is too early.

## Solution?

Currently, this PR make below trace. It can make graph pretty well.
datadog-trace-view-v2

but this change contains Breaking Change.
Resolver will be received arguments directly. It is not processed by middleware.
Middleware only can process return value after Resolver worked.

I do not have a good idea how to solve this issue... 😿 
Middleware now has a hook for start, but it seems like a problem that It does not have a finish hook. 🤔


vektah:
I think this was the way it was working:
graph 2


and this is what this pr does
graph 3

So this would break any directive or middleware that modifies the result. Plus its still missing the argument unpacking. (I don't know why that would be slow, but if it was it should show up)

I think what we really want is 
graph 4

Something that can time each part (unpacking, resolver execution, child execution, marshaling) with appropriate hooks to modify context. Maybe a super specific hook is more appropriate.
```go
type Tracer interface{
    StartFieldExecution(args, query) context.Context
    EndFieldExecution(stats from resolver) 
}
```
and we write the required code directly in the executors.

PR #392, opened 2018-10-23
Add Tracer layer

vvakame:
closes #388

@vektah thank you for your comment on https://github.com/99designs/gqlgen/pull/388#issuecomment-430966846

I have:
 - [ ] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content)) 🤔 

newtracer


codegen/testserver/generated.go
```diff
@@ -938,7 +939,11 @@ func (ec *executionContext) _Circle_radius(ctx context.Context, field graphql.Co
 	}
 	res := resTmp.(float64)
 	rctx.Result = res
-	return graphql.MarshalFloat(res)
+	resMarshaler := func() graphql.Marshaler {
+		return graphql.MarshalFloat(res)
+	}()
+	ec.Tracer.EndFieldExecution(ctx)
```
vektah: rather than needing to wrap all the marshallers in another func, could we `defer ec.Tracer.EndFieldExecution(ctx)`?

codegen/testserver/generated.go
```diff
@@ -3365,6 +3672,7 @@ func (ec *executionContext) ___Type_fields(ctx context.Context, field graphql.Co
 		Field:  field,
 	}
 	ctx = graphql.WithResolverContext(ctx, rctx)
+	ctx = ec.Tracer.StartFieldExecution(ctx)
```
vektah: This still wont capture the time spent unpacking args in `field.ArgumentMap`

vektah:
Not a blocker, but it would be good to find out what all these gaps are too:

image

graphql/tracer.go
```diff
@@ -0,0 +1,28 @@
+package graphql
+
+import "context"
+
+var _ Tracer = (*NopTracer)(nil)
+
+type Tracer interface {
```
vektah: Maybe something like:
```go
		StartOperationExecution(ctx context.Context) context.Context
	EndOperationExecution(ctx context.Context)
	StartFieldArgParsing(ctx context.Context)
	StartFieldExecution(ctx context.Context) context.Context // is this still the right place? should it be on the first one
        StartFieldChildExecution(ctx context.Context) // or StartFieldMarshalling... not sure
	EndFieldExecution(ctx context.Context
```

codegen/testserver/generated.go
vvakame: I want to access ResolverContext in StartFieldExecution method.
e.g. https://github.com/mercari/mtc2018-web/blob/feat-refactor-tracer/server/gqlapi/gqlopencensus/tracer.go#L46

vvakame:
> Not a blocker, but it would be good to find out what all these gaps are too:

My example resolver is worked on fully on-memory only.
Operation is done in 0.4ms(400μs).
I guess the other process used the container's CPU for a moment.

BTW, I want to find the bottleneck on my resolvers.
It's main reason is maybe DB or Storage or Network access.
unpacking argument and result value marsharing is CPU bound process.
I hope it is as correct as possible, but I do not think we need to be able to do something complicated.

I guess App Developer is not want to tracking gqlgen's performance on Tracing.

vvakame:
note: https://gitter.im/gqlgen/Lobby?at=5bceb8dfc7bf7c36625604b1

vvakame:
why appveyor reference old version's test case...? 🤔

vektah:
Looks good. Shouldn't be a BC break right, The default preserves existing behaviour?

vvakame:
sure, I think so.

codegen/testserver/generated_test.go
```diff
@@ -35,9 +35,60 @@ func TestForcedResolverFieldIsPointer(t *testing.T) {
...
+type testTracer struct {
+	id     int
+	append func(string)
+}
+
+func (tt *testTracer) StartOperationExecution(ctx context.Context) context.Context {
+	line := fmt.Sprintf("op:start:%d", tt.id)
+
+	tracerLogs, _ := ctx.Value("tracer").([]string)
+	ctx = context.WithValue(ctx, "tracer", append(tracerLogs, line))
```
vektah: I think the tests are flaking out because appending to a slice isn't thread safe.

You now have multiple concurrent goroutines all working on the same underlying slice. You probably want something like:
`append(append([]string, tracerLogs...), line)`

vvakame:
AppVeyor
```
  - go generate ./...
  - go test -timeout 20m ./...
```

Circle CI
```
go generate $(go list ./... | grep -v codegen/tests)
go test -race $(go list ./... | grep -v codegen/tests)
```

Why is there a difference? 🤔

vektah:
Nice find, Fixed in https://github.com/99designs/gqlgen/pull/396

PR #403, opened 2018-10-29
copy complexity to RequestContext

vvakame:
I want to get access to operation complexity on Tracing.

I have:

### DIFF
--- a/graphql/context.go
+++ b/graphql/context.go
@@ -17,6 +17,10 @@ type RequestContext struct {
 	RawQuery  string
 	Variables map[string]interface{}
 	Doc       *ast.QueryDocument
+
+	ComplexityLimit     int
+	OperationComplexity int
+
 	// ErrorPresenter will be used to generate the error
 	// message from errors given to Error().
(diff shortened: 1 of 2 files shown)


vvakame:
I want to see the complexity of each path, but it looks difficult 🤔

graphql/context.go
```diff
@@ -17,6 +17,10 @@ type RequestContext struct {
 	RawQuery  string
 	Variables map[string]interface{}
 	Doc       *ast.QueryDocument
+
+	ComplexityLimit     int
```
vektah: why is ComplexityLimit here, its static?

vvakame: Yes it's static value determined at compile time.

BTW, I want to modify it on run time...

vektah: > BTW, I want to modify it on run time...

:thinking: 

 when? ideally a request over the limit doesn't run anything, especially not in the resolver stack.

vvakame: For example, If we want to change ComplexityLimit between 1st party client and 3rd party client.

Issue #25, opened 2018-02-23
Default Args not implemented.


Teddy-Schmitz:
Hello,

Thanks again for this library.  It seems that default arguments in the graph schema isn't implemented?

`blockchain(limit: Int = 10): BlockchainConnection!`

Generates code that has limit as `*int` but I would expect it to be an `int` that sets it to 10 or the passed in value.

Teddy-Schmitz:
O and I guess for reference, what I am currently doing is converting a toy project I wrote a few weeks ago using neelance/graph-go to this project before we evaluate switching over our main production application as we like the features this supports.

Thanks again for sharing thisl ibrary!

vektah:
Is the toy project public? If so, would you mind If i linked to it as an example?

Teddy-Schmitz:
thanks for getting this in so quickly, its not public right now but if that changes I'll let you know.

PR #406, opened 2018-11-01
Use value type if default value

millerlogic:
Hi,

If a default value is present, I'd like it to use a value type:

```graphql
type Query {
    items(limit: Int = 10): ItemList!
}
```
to generate:

```go
func (r *queryResolver) Items(ctx context.Context, limit int) (ItemList, error) {
// ...
```

I was hoping doing ```limit: Int! = 10``` (required type with default value) would suffice, and it sort of did, but the playground web UI returned errors when I didn't explicitly provide a limit value.

I believe I've used other go code generators that switch to value types when default values are present.

Also it looks like the original request for default values implied that the type would switch to a value type: https://github.com/99designs/gqlgen/issues/25 "Generates code that has limit as ```*int``` but I would expect it to be an ```int```"

This is more of a feature request that I implemented. I am new to this project and I'm not sure if this change is correct, though it appears to be doing what I expect. Feel free to reject/close it.
This would be a breaking change, so I understand any hesitation, however maybe now is the time to do such changes.

Thanks,
-Chris

### DIFF
--- a/codegen/input_build.go
+++ b/codegen/input_build.go
@@ -63,6 +63,9 @@ func (cfg *Config) buildInput(types NamedTypes, typ *ast.Definition) (*Object, e
 		if field.DefaultValue != nil {
 			var err error
 			newField.Default, err = field.DefaultValue.Value(nil)
+			if newField.Type.IsPtr() {
+				newField.Type.StripPtr()
+			}
 			if err != nil {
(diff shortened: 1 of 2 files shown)


vektah:
It might be worth bumping the playground version too, just in case this has been fixed.

stereosteve:
playground issues:

* https://github.com/prisma/graphql-playground/issues/930
* https://github.com/prisma/graphql-playground/issues/871

PR #410, opened 2018-11-02
Move tracing to contrib

vektah:
We now have 3 competing implementations of tracing. I've moved them out of the core and into [gqlgen-contrib](https://github.com/99designs/gqlgen-contrib).

We still need docs for all the tracers, Maybe a new subheading on the doc site for contrib?


vvakame:
> Maybe a new subheading on the doc site for contrib?

👍 

I think Go Developer needs https://godoc.org/github.com/99designs/gqlgen-contrib . 👀

Issue #415, opened 2018-11-06
Deprecate gqlgen Binary


mathewbyrne:
Currently we ship a binary with gqlgen, but increasingly we're running into several issues with it:

1. Bootstrapping is hard — we want to update the getting started docs to remove the `go get` usage, but then how do you run gqlgen in your project?

2. Versioning differences — the getting started docs switch to using dep by their end, but this will cause issues if `master` is ahead of the latest release.

3. Shipping Plugins — this becomes a lot harder since we need some way to load code dynamically from a fixed binary.  We'd rather avoid trying to solve this by letting users roll their own binary anyway.

We're attempting to address the getting started issues at the moment with documentation, but we'd also like to make this code change at the same time.  So starting from the next release out recommendation will be that you integrate gqlgen into your project by calling a new method in the `cmd` package.  The documentation will be updated to reflect this.

This approach isn't set in stone yet, and I'm open to any other suggestions people might have.  We've already seen a few projects do something similar, calling into the `codegen` package to have a custom binary.

## Migration away

Follow these steps to migrate away from the gqlgen binary:

1. In your project, create a `scripts/gqlgen.go` file with the following contents:

  ```go
  // +build ignore

  package main

  import "github.com/99designs/gqlgen/cmd"

  func main() {
  	cmd.Execute()
  }
  ```

2. If you are using [`dep`](https://github.com/golang/dep) to manage dependencies for your project, ensure that you have them all now:

  ```bash
  dep ensure
  ```

3. Update any generate commands in your project to use this new script.  Ensure that the path to the script is correct:

   ```go
   //go:generate go run scripts/gqlgen.go -v
   ```

4. _(optional)_ Remove the previous binary from your bin path so that you aren't accidentally running a previous version of gqlgen.

MichaelMure:
Can't agree enough. I hacked my way into this approach by copy/pasting internal code from gqlgen (see https://github.com/MichaelMure/git-bug/blob/master/graphql/gen_graphql.go), it works really well, no risk of version mismatch and go dep will vendor anything that might be required.

That said, why the `cmd`package ? Intuitively, this package should be for a command line tool, not for package that you import and use. Why not simply the `codegen` package or even the root `gqlgen` package ?

vektah:
We want to make sure there are no breaking changes here, so once someone copies the stub in they should only need to change it again if they want to add a plugin.

[...]

Issue #408, opened 2018-11-02
Improve Getting Started Documentation


mathewbyrne:
There are a lot of users getting stuck during the [getting started documentation](https://gqlgen.com/getting-started/) or shortly thereafter.  Some of the issues include:

1. User installs both gqlgen globally (and probably gets `master`), then installs gqlgen from dep.  So generated code is no sometimes not compatible depending on the state of master.

2. Users trying to execute `gqlgen` outside of their `GOPATH`.  We should at least make this clear that it's not currently possible and why.

3. User cannot find `gqlgen` binary — some people are struggling binding the binary, either they don't have it in their `PATH`, or have another configuration error.

We should address these issues and ensure that the described path does not easily run into any of them.

hiucimon:
Can someone PLEASE tell me how to be "in the GO PATH"? I have tired everything I can think of and get the same message, and have searched until I am blue in the face and can not see a clear answer anywhere. I have tried setting GOPATH to various places, and tried CD'ing to where it seems to be and it never changes the results. gqlgen gets installing into ~/go/bin/ on the Ubuntu instance I have tried it on (also tried the same things on Mac).

hiucimon:
OK, I figured it out. For anyone who needs to know, you have to be in the source directory of this code when you execute it.

lowe1111:
@hiucimon what do u mean with this code? I cant get it running

PR #416, opened 2018-11-06
Improve Getting Started Documentation — No Binary Approach

mathewbyrne:
Addresses #408 — this is a set of improvements to the getting started documentation.  The main change is deprecating using a global binary, and moving to always using `dep` to ensure users have a consistent version.

#415 discusses the whys for this approach.


docs/content/getting-started.md
```diff
@@ -6,32 +6,54 @@ menu: main
...
+Assuming you already have a working [Go environment](https://golang.org/doc/install), create a directory for the project in your `$GOPATH`:
+
+```sh
+$ mkdir -p $GOPATH/src/github.com/[username]/gqlgen-todos
+```
+
+Add the following file to your project under `scripts/gqlgen.go`:
+
+```go
+// +build ignore
```
vektah: there needs to be an empty line between this comment and package main for the build tag to work

docs/content/getting-started.md
```diff
@@ -60,17 +82,17 @@ type Mutation {
...
+This has created an empty skeleton with all files you need:
 
- - gqlgen.yml - The gqlgen config file, knobs for controlling the generated code.
- - generated.go - The graphql execution runtime, the bulk of the generated code
- - models_gen.go - Generated models required to build the graph. Often you will override these with models you write yourself. Still very useful for input types.
- - resolver.go - This is where your application code lives. generated.go will call into this to get the data the user has requested. 
+ - `gqlgen.yml` — The gqlgen config file, knobs for controlling the generated code.
+ - `generated.go` — The GraphQL execution runtime, the bulk of the generated code.
+ - `models_gen.go` — Generated models required to build the graph. Often you will override these with your own models. Still very useful for input types.
+ - `resolver.go` — This is where your application code lives. `generated.go` will call into this to get the data the user has requested.
```
vektah: did this also create a server entry point?

docs/content/getting-started.md
```diff
@@ -242,34 +264,14 @@ query findTodos {
...
-
-```bash
-dep init
-dep ensure
-```
-
 At the top of our resolvers.go a go generate command was added that looks like this:
 ```go
-//go:generate gorunpkg github.com/99designs/gqlgen
+//go:generate go run ./scripts/gqlgen.go
```
vektah: did you update the template for resolvers.go?

mathewbyrne: Ah interesting.  Yeah what should we do there?  We can't assume users will have a script to run at `scripts/gqlgen.go`.

cmd/root.go
```diff
@@ -9,6 +9,9 @@ import (
 	"github.com/99designs/gqlgen/graphql"
 	"github.com/99designs/gqlgen/internal/gopath"
 	"github.com/urfave/cli"
+
+	// Required since otherwise dep will prune away these unused packages before codegen has a chance to run
+	_ "github.com/99designs/gqlgen/handler"
```
mathewbyrne: This seems to be enough to get `init` to run cleanly.  Attempting to run `server/server.go` afterwards however still requires some additional deps (websockets I think).  So I added a `dep ensure` after `gqlgen init` in the docs.

emil-nasso:
Is it possible to add an explanation to what `./...` means? 

This is not unlikely to be the first run in with `go generate` for many users. The go documentation is not very good or clear on this.

To my understanding it runs go generate in all subdirectories so `./...` means "the subdirectories (...) of the current directory (.)".

I think it would help to alleviate some confusion to just have a short explaination like:

> You can now run any generation commands for all subdirectories in your project with:

mathewbyrne:
@emil-nasso yeah I think that's a fair comment.  I'll add a note.

docs/content/getting-started.md
```diff
@@ -88,22 +117,27 @@ type Todo struct {
 }
 ```
 
-And then tell gqlgen to use this new struct by adding this to the gqlgen.yml:
+Next tell gqlgen to use this new struct by adding it to `gqlgen.yml`:
+
 ```yaml
 models:
   Todo:
     model: github.com/vektah/gqlgen-tutorials/gettingstarted.Todo
```
samjbobb: Earlier the doc says:

> `mkdir -p $GOPATH/src/github.com/[username]/gqlgen-todos`

So I think this doesn't match. Probably should be:

`model: github.com/[username]/gqlgen-todos.Todo`

docs/content/getting-started.md
```diff
@@ -152,17 +186,16 @@ better than generating.
 ### Write the resolvers
 
 This is a work in progress, we have a way to generate resolver stubs, but it only cant currently update existing code. We can force it to run again by deleting `resolvers.go` and re-running gqlgen:
+
 ```bash
-rm resolvers.go
-gqlgen
+$ rm resolvers.go
```
samjbobb: For me, this is `resolver.go` no `resolvers.go`

docs/content/getting-started.md
```diff
@@ -152,17 +186,16 @@ better than generating.
...
+$ go run scripts/gqlgen.go
 ```
 
-Now we just need to fill in the `not implemented` parts
-
+Now we just need to fill in the `not implemented` parts.  Update `graph/graph.go`
```
samjbobb: I think (for me) this should be "Update `resolver.go`"

docs/content/getting-started.md
```diff
@@ -242,34 +275,14 @@ query findTodos {
...
-
-*Note*: If you dont have dep installed yet, you can get it [here](https://github.com/golang/dep)
-
-First uninstall the global version we grabbed earlier. This is a good way to prevent version mismatch footguns.
-
-```bash
-rm ~/go/bin/gqlgen
-rm -rf ~/go/src/github.com/99designs/gqlgen
-``` 
+At the top of our `resolvers.go` add the following line:
```
samjbobb: `resolvers.go` -> `resolver.go` (for me at least, I think)

samjbobb:
Hey thanks for this project and this PR. This updated documentation really helped me get started.

I created a new project and set my `Gopkg.toml` to:

```
[[constraint]]
  name = "github.com/99designs/gqlgen"
  branch = "improved-getting-started"
```

And then ran through the Getting Started guide. 

Overall, was successful and the new docs helped. I noticed a couple of discrepancies, so I added some commends to the PR. Hope that's helpful!

Thanks again.

PR #419, opened 2018-11-09
fix unexpected ctx variable capture on Tracing

vvakame:
I tried Tracing with reqCtx.GetErrors, but I can't get errors expectedly.
BAD: https://play.golang.org/p/AswrHXxcgT7
GOOD: https://play.golang.org/p/Ftk_SLFVVgw
https://golang.org/ref/spec#Defer_statements

I have:


vvakame:
did it!

codegen/testserver/generated_test.go
```diff
@@ -276,67 +462,109 @@ func TestGeneratedServer(t *testing.T) {
...
+			"StartOperationExecution",
+			"StartFieldExecution",
+			"StartFieldResolverExecution",
+			"StartFieldChildExecution",
+		}
 
-			var msg struct {
-				resp struct {
-					Updated string
+		assertStep := func(target string) func(ctx context.Context) {
```
vektah: Are the arrays required in the configurable tracer? this would a bit clearer with something like:
```go
configurableTracer := &configurableTracer{
    StartOperationParsingFuncs: func(ctx context.Context) context.Context {
        return context.WithValue("StartOperationParsing", true)
    },
    StartOperationValidation: func(ctx context.Context) context.Context {
        assert.True(t, ctx.Value("StartOperationParsing")
        return context.WithValue("StartOperationValidation", true)
    },
    // etc
}
```

no need for anonymous helper funcs or appending to arrays?

codegen/testserver/generated_test.go
```diff
@@ -276,67 +444,84 @@ func TestGeneratedServer(t *testing.T) {
...
+			return func(ctx context.Context) {
+				for _, step := range steps {
+					assert.NotEmpty(t, ctx.Value(step))
+					if step == target {
```
vektah: I still think its better to write this out the long way in the test, so it is easy to understand what you're testing. It took me a few reads to get it.

Issue #375, opened 2018-10-10
[Proposal] Always use pointers in resolver return types


vektah:
Currently the return value for a resolver is based on the schema Non-Nullability. 

eg
for `query { user: User!}` we generate `func (r *Query) User(ctx context.Context) (User, error)`
for `query { user: User}` we generate `func (r *Query) User(ctx context.Context) (*User, error)`

Initially, I thought this would be useful as the compiler can validate the non-nullability, however it:
 - Forces you to return an empty objects with errors, eg `return User{}, fmt.Errorf("not found")`
 - Isn't very idiomatic. Most go apis will return (*Thing, error), and return nil when error is set
 - Fragments function signatures; if you have a database function looking up a user it probably already returns a (*User, error), being able to return it directly would be nice.

**Proposal**

Regardless of nullability always use a pointer. Gqlgen already knows how to check for nulls in fields (used when binding directly to model fields that have pointers).

This would need to go behind a config flag for now. We can opt in new users in init, and output a deprecation message for some release in the future to drop. Or maybe we just keep it. Its probably a pretty small change either way.

What are peoples thoughts on this?

related https://github.com/prisma/prisma/issues/3288

edsrzf:
I think defaulting to a pointer makes sense. The current behavior has definitely been an annoyance at times.

What about types that are normally passed by value (except when nullable), like `int`? Maybe exclusions for certain built-in types? (Possibly even standard library types that are normally passed by value like `time.Time`.) It might also be nice to default to pointers, but allow overriding to values via configuration/directive.

mathewbyrne:
Having worked with gqlgen with a small to medium sized schema for about 6 months now, I can definitely say that I would prefer this change, largely for the benefits listed above.

azavorotnii:
@vektah real drawback of pointers - they can be nil for non-null schema. This makes unnecessary null-check and makes it less clear for developer if there can return `nil` value or can not. I think what there is right now is correct and clear for developer what expected as a result of resolver method.

azavorotnii:
@mathewbyrne I understand that this is only about return values. But assuming schema:
```
type Query {
  user (id: ID!): User!
}
```
We will get resolver with nullable return value which can be defined like this by developer:
```
type Resolver {
  users map[string]*User
}

func (r *Resolver) User(ctx context.Context, id string) (*User, error) {
  return r.users[id], nil
}
```
[...]

PR #422, opened 2018-11-11
accept an optional ctx parameter on model methods

gracenoah:
This PR allows you to take a context on model methods. This makes it much easier to transition from graph-gophers/graphql-go to this library. By configuring all of your graph-gophers resolvers as "models" in gqlgen, you can reuse existing code and migrate gradually instead of having to rewrite all resolvers in order to migrate.

Testing and docs coming up after I get some initial feedback on the concept and implementation.

Edit: I ran into one more thing that prevents me from being able to transition gradually. Models can't be returned from resolvers as pointers unless they are optional. All of my graph-gophers resolvers are pointers. I'm considering a temporary hack to make all models passed as pointers or converting all graph-gophers resolvers to non-pointers (there was no real reason for them to be pointers). I'd love some tips if anyone has done this before. 

I have:

### DIFF
--- a/codegen/object.go
+++ b/codegen/object.go
@@ -33,16 +33,17 @@ type Object struct {
 
 type Field struct {
 	*Type
-	Description    string          // Description of a field
-	GQLName        string          // The name of the field in graphql
-	GoFieldType    GoFieldType     // The field type in go, if any
-	GoReceiverName string          // The name of method & var receiver in go, if any
-	GoFieldName    string          // The name of the method or var in go, if any
-	Args           []FieldArgument // A list of arguments to be passed to this field
-	ForceResolver  bool            // Should be emit Resolver method
-	NoErr          bool            // If this is bound to a go method, does that method have an error as the second argument
-	Object         *Object         // A link back to the parent object
-	Default        interface{}     // The default value
+	Description      string          // Description of a field
+	GQLName          string          // The name of the field in graphql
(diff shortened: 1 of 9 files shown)


vektah:
We are thinking about making all non primative return types pointers, which may help the second issue. See https://github.com/99designs/gqlgen/issues/375

gracenoah:
Ah, thanks for the tip about not running in parallel. That might make it infeasible to reuse the resolvers without rewriting them. I'll try to address that in this PR as well. Parallelizing if there's a context makes sense to me.

gracenoah:
I don't see any existing tests for model methods. Any suggestions on where I should add the tests?

vektah:
You can create a model in [codegen/testserver](https://github.com/99designs/gqlgen/tree/master/codegen/testserver) and wire it into that config, then add tests to generated_test.go (or create a new file like modelctx_test.go) in that directory.

gracenoah:
Tests added. Please review.

vektah:
You need to run a `go generate ./...` and commit the changes. Seeing how the generated code has changed makes reviewing easier (even if it artificially inflates line counts)

gracenoah:
Sorry about the testing screwups. It's green now.

vektah:
Awesome, thanks for the PR.

Looks like we could use a guide for migrating from gophers, If anyone has a large codebase they have successfully migrated it would be great to get their notes (maybe in the recipes section?).

PR #425, opened 2018-11-12
Switch to hosting docs on render.com

vektah:
[Render.com](https://render.com) has offered to host our static site for free, and have a pretty simple setup for rebuilding on merge to master. I've switched the DNS records and updated the docs.


anurag:
Brilliant!

PR #434, opened 2018-11-15
Update context.go

urakozz:
Right now code generated with latest master fails since there are usages of Trace but there is no any single write to this variable


nmijailovic:
Yes, that fixed the issue with the generated code not having the Tracer: graph/generated.go:204:10: ec.Tracer undefined (type *executionContext has no field or method Tracer)

PR #435, opened 2018-11-16
Update README.md comparison with graph-gophers

matiasanaya:
I've fixed issues with the comparison chart in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,10 +15,10 @@ Looking for typescript? Check out [graphqlgen](https://github.com/prisma/graphql
 | Docs | [docs](https://gqlgen.com) & [examples](https://github.com/99designs/gqlgen/tree/master/example) | [examples](https://github.com/graph-gophers/graphql-go/tree/master/example/starwars) | [examples](https://github.com/graphql-go/graphql/tree/master/examples) | [examples](https://github.com/samsarahq/thunder/tree/master/example)|
 | Query | :+1: | :+1: | :+1: | :+1: |
 | Mutation | :+1: | :construction: [pr](https://github.com/graph-gophers/graphql-go/pull/182) | :+1: | :+1: |
-| Subscription | :+1: | :construction: [pr](https://github.com/graph-gophers/graphql-go/pull/132) | :+1: | :+1: |
+| Subscription | :+1: | :construction: [pr](https://github.com/graph-gophers/graphql-go/pull/182) | :+1: | :+1: |
 | Type Safety | :+1: | :+1: | :no_entry: | :+1: | 
 | Type Binding | :+1: | :construction: [pr](https://github.com/graph-gophers/graphql-go/pull/194) | :no_entry: | :+1: |
-| Embedding | :+1: | :construction: [pr](https://github.com/graphql-go/graphql/pull/274) | :no_entry: | :no_entry: |
+| Embedding | :+1: | :no_entry: | :construction: [pr](https://github.com/graphql-go/graphql/pull/371) | :no_entry: |
 | Interfaces | :+1: | :+1: | :+1: | :no_entry: [is](https://github.com/samsarahq/thunder/issues/78) |
 | Generated Enums | :+1: | :no_entry: | :no_entry: | :no_entry: |
 | Generated Inputs | :+1: | :no_entry: | :no_entry: | :no_entry: |


README.md
```diff
@@ -14,11 +14,11 @@ Looking for typescript? Check out [graphqlgen](https://github.com/prisma/graphql
 | Boilerplate | less | more | more | some |
 | Docs | [docs](https://gqlgen.com) & [examples](https://github.com/99designs/gqlgen/tree/master/example) | [examples](https://github.com/graph-gophers/graphql-go/tree/master/example/starwars) | [examples](https://github.com/graphql-go/graphql/tree/master/examples) | [examples](https://github.com/samsarahq/thunder/tree/master/example)|
 | Query | :+1: | :+1: | :+1: | :+1: |
-| Mutation | :+1: | :construction: [pr](https://github.com/graph-gophers/graphql-go/pull/182) | :+1: | :+1: |
-| Subscription | :+1: | :construction: [pr](https://github.com/graph-gophers/graphql-go/pull/132) | :+1: | :+1: |
+| Mutation | :+1: | :+1: | :+1: | :+1: |
```
vektah: This one hasn't merged yet, but it looks active.

README.md
```diff
@@ -14,11 +14,11 @@ Looking for typescript? Check out [graphqlgen](https://github.com/prisma/graphql
...
+| Subscription | :+1: | :+1: | :+1: | :+1: |
 | Type Safety | :+1: | :+1: | :no_entry: | :+1: | 
 | Type Binding | :+1: | :construction: [pr](https://github.com/graph-gophers/graphql-go/pull/194) | :no_entry: | :+1: |
-| Embedding | :+1: | :construction: [pr](https://github.com/graphql-go/graphql/pull/274) | :no_entry: | :no_entry: |
+| Embedding | :+1: | :no_entry: | :no_entry: | :no_entry: |
```
vektah: This was re-opened in another PR, but its in the wrong column (not gophers)

vektah:
Thanks for the update

README.md
matiasanaya: @vektah not sure why that PR is being singled out. `graph-gophers/graphql-go` supports mutations and what I understand of that PR is that it is an enhancement to the way mutation resolvers are implemented.

README.md
matiasanaya: should I move it to the next column?

README.md
vektah: Because it prevents a pretty common use case

matiasanaya: i.e. "GraphQL allows you to have a mutation and a query with the same name, but graph-gophers/graphql-go does not."?

vektah: Yeah in graphql they (query, mutation, subscription) are all separate types, but there it has been merged them into one and just hopes they don't collide. 

I think it's pretty important to draw attention to when deciding which library to use, because you won't hit it until you are well into building the server and swapping at that point becomes pretty costly.

matiasanaya: That makes sense, thanks for clarifying. I'll remove this change from the PR.

README.md
```diff
@@ -15,10 +15,10 @@ Looking for typescript? Check out [graphqlgen](https://github.com/prisma/graphql
 | Docs | [docs](https://gqlgen.com) & [examples](https://github.com/99designs/gqlgen/tree/master/example) | [examples](https://github.com/graph-gophers/graphql-go/tree/master/example/starwars) | [examples](https://github.com/graphql-go/graphql/tree/master/examples) | [examples](https://github.com/samsarahq/thunder/tree/master/example)|
 | Query | :+1: | :+1: | :+1: | :+1: |
 | Mutation | :+1: | :construction: [pr](https://github.com/graph-gophers/graphql-go/pull/182) | :+1: | :+1: |
-| Subscription | :+1: | :construction: [pr](https://github.com/graph-gophers/graphql-go/pull/132) | :+1: | :+1: |
+| Subscription | :+1: | :construction: [pr](https://github.com/graph-gophers/graphql-go/pull/182) | :+1: | :+1: |
```
matiasanaya: Since subscriptions are supported but not namespaced I've referenced here the same PR as for mutations.

README.md
```diff
@@ -15,10 +15,10 @@ Looking for typescript? Check out [graphqlgen](https://github.com/prisma/graphql
...
 | Type Safety | :+1: | :+1: | :no_entry: | :+1: | 
 | Type Binding | :+1: | :construction: [pr](https://github.com/graph-gophers/graphql-go/pull/194) | :no_entry: | :+1: |
-| Embedding | :+1: | :construction: [pr](https://github.com/graphql-go/graphql/pull/274) | :no_entry: | :no_entry: |
+| Embedding | :+1: | :no_entry: | :construction: [pr](https://github.com/graphql-go/graphql/pull/371) | :no_entry: |
```
matiasanaya: Moved the PR to the correct column

PR #439, opened 2018-11-20
Fix type binding validation for slices of pointers like []*foo

snormore:
I was running into this during `gqlgen -v`:
```
field has wrong type: []github.com/snormore/foobar/pkg/app.Foo is not compatible with []*github.com/snormore/foobar/pkg/app.Foo
```

I have:


vektah:
I'm planning a refactor for this, but have been very time poor recently.

Thanks for the PR, looks good.

PR #440, opened 2018-11-20
Updated README: added another alternative for typescript

dotansimha:
Hi @99designs!

In this PR I updated the README file, pointing to another alternative for generating TypeScript from GraphQL.
GraphQL Code Generator generates output from a GraphQL schema and GraphQL documents (query/mutation/subscription/fragment) for both client and server. And also supports generating TypeScript resolvers similar to this package.


vektah:
That library looks great, but I don't think its gqlgens job to index other projects. I'm going to drop the link to graphqlgen instead and keep this library focused on whats happening in golang.

PR #441, opened 2018-11-23
Replaced "encoding/json" by "github.com/json-iterator/go"

Daavidaviid:
## Replace encoding/json by a faster library

Replaced the default Marshal/Unmarshal operations from the standard library to the similar operations from a faster library: [github.com/json-iterator/go](https://github.com/json-iterator/go).

As you can see, according to THEIR benchmark, it might help improve the performance of gqlgen as it relies heavily on json marshalling/unmarshalling.

Benchmark results

I haven't tried doing a solid benchmark using gqlgen as I haven't had the occasion yet. But if you feel like doing it, go for it.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -8,8 +8,11 @@ import (
 	"io/ioutil"
 	"net/http"
 
+	jsoniter "github.com/json-iterator/go"
 	"github.com/mitchellh/mapstructure"
 )
+
+var jsonIterator = jsoniter.ConfigCompatibleWithStandardLibrary
 
 // Client for graphql requests
 type Client struct {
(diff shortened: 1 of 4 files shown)


vektah:
Thanka for the pr, However as you can see in the changes, we don't really use the core json marshalled much. Insted of introducing another library I would rather move that logic into gqlgen. Everything else is already statically generated. Wdyr? Maybe benchmarks would show if there is a real benefit?

tugberkugurlu:
> Insted of introducing another library I would rather move that logic into gqlgen.

👋 do you have any examples on how we can replace the JSON marshaling with a custom marshaling code, e.g. using with json-iterator/go for example?

PR #458, opened 2018-11-30
[WIP]add types int32/int64/float32/uint* to UnmarshalID and add test

andrey1s:
- use  int32 or int64 or float32 or uint as id in types
 - add test to UnmarshalID


vektah:
Could use a description. What are you solving here? The json decoder only has a few types it emits

frederikhors:
@andrey1s why did you closed this one?

PR #460, opened 2018-12-01
add execute argument and input directive

andrey1s:
added execution of system directives for ARGUMENT_DEFINITION and INPUT_FIELD_DEFINITION

generating code for arguments for example when adding a Directive instead of 
```golang
	if tmp, ok := rawArgs["Id"]; ok {
		var err error
		arg0, err = graphql.UnmarshalID(tmp)
		if err != nil {
			return nil, err
		}
	}
```
will be 
```golang
	if tmp, ok := rawArgs["Id"]; ok {

		argm0, err := chainFieldMiddleware([]graphql.FieldMiddleware{
			func(ctx context.Context, n graphql.Resolver) (res interface{}, err error) {
				min := "0"
				return e.directives.AssertRange(ctx, tmp, n, &min, nil, nil)
			},
		}...)(ctx, func(ctx2 context.Context) (args0 interface{}, err error) {
			args0, err = graphql.UnmarshalID(tmp)
			if err != nil {
				return nil, err
			}
			return
		})
		if err != nil {
			return nil, err
		}
		if data, ok := argm0.(string); ok {
			arg0 = data
		} else {
			return nil, errors.New("expect string")
		}

	}
```

if the argument is type input, the code will be added in addition to the previous code and the previous code will be added (for example)
```golang
	if tmp, ok := rawArgs["input"]; ok {

		var err error
		arg0, err = UnmarshalDataInput(tmp)
		if err != nil {
			return nil, err
		}

		arg0, err := e.DataInputMiddleware(ctx, arg0)
		if err != nil {
			return nil, err
		}
	}
```
and added input middleware(for example) or empty middleware
```golang
func (e *executableSchema) DataInputMiddleware(ctx context.Context, obj *models.DataInput) (*models.DataInput, error) {

	if obj.InternalData != nil {
		var err error
		obj.InternalData, err = e.CoordinatesInputMiddleware(ctx, obj.InternalData)
		if err != nil {
			return nil, err
		}
	}

	return obj, nil
}
```


mathewbyrne:
Thanks for the PR, I'll try to review it as I get time.  It's a pretty big change though, a bit more description would be appreciated as to how you've implemented this and how it affects the current directives implementation.

andrey1s:
update description
the current implementation (master) only supports directives for FIELD_DEFINITION and Executable directives, and does not support arguments. I left them unchanged, I hope we'll share them via the Location Directive in the future

vektah:
This is a large PR to drop without any discussion beforehand :( We have been thinking to cover this use case (validation) via a codegen plugin rather than getting the user to wire them up for each graph.

I'm getting lots of compile errors if I add a directive to some args in starwars:
```
directive @length(min: Int!, max: Int, message: String!) on ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION

# The query type, represents all of the entry points into our object graph
type Query {
    hero(episode: Episode = NEWHOPE @length(min: 2, max: 5)): Character
    reviews(episode: Episode! @length(min: 2, max: 5), since: Time): [Review!]!
# ...
type Starship {
    # The ID of the starship
    id: ID!
    # The name of the starship
    name: String! @length(min: 2, max: 5)
```
```
go generate ./example/starwars/
warning: running gqlgen from this binary is deprecated and may be removed in a future release. See https://github.com/99designs/gqlgen/issues/415
/home/adam/go/src/github.com/99designs/gqlgen/example/starwars/generated.go:308:19: invalid operation: (&args0) (value of type *interface{}) has no field or method UnmarshalGQL
/home/adam/go/src/github.com/99designs/gqlgen/example/starwars/generated.go:305:54: cannot convert nil (untyped nil value) to string
/home/adam/go/src/github.com/99designs/gqlgen/example/starwars/generated.go:268:54: cannot convert nil (untyped nil value) to string
```

andrey1s:
> cannot convert nil (untyped nil value) to string

invalid schema
`directive @length(min: Int!, max: Int, message: String!)`
message MUST NOT be null
`@length(min: 2, max: 5, message:"not valid")`
and if not set `ARGUMENT_DEFINITION` schema invalid, but there is no error

and for example if use `@hasRole` from gqlgen/example/todo/schema.graphql without role MUST be error when generate go files, but its working as if you were the owner

> invalid operation: (&args0) (value of type *interface{}) has no field or method UnmarshalGQL

updated generate args

PR #491, opened 2019-01-08
Add typhon support

jamesallison:

I have:


handler/graphql.go
```diff
@@ -12,6 +12,8 @@ import (
 	"github.com/99designs/gqlgen/graphql"
 	"github.com/gorilla/websocket"
 	"github.com/hashicorp/golang-lru"
+	"github.com/monzo/terrors"
+	"github.com/monzo/typhon"
```
vektah: Rather than support all transports forever we should work how to make handler open for extension.

The way this is setup currently, everyone using gqlgen would need to fetch monzo/typoon.

PR #492, opened 2019-01-09
Unified merge pass

vektah:
This refactor removes all the bootstrapping logic and moves it into the root `gqlgen` package, removing the main entirely, see https://github.com/99designs/gqlgen/issues/415 for more info.

This public interface now looks like this:
```go
func Generate(cfg *config.Config) error {}
func GenerateResolver(schema *unified.Schema) error  {}
func GenerateServer(filename string, cfg *config.Config) error {}
```
Much cleaner interface to expose, and pretty similar to what @MichaelMure proposed in https://github.com/99designs/gqlgen/issues/415#issuecomment-449756685.

I've also extracted the builder into its own object, to try to clean up the boundary between the methods the templates have access to and the merge process. 

Only generated code changes were ordering.


MichaelMure:
Awesome !

> Inside the unified package (Naming is hard, give me a cache to invalidate!)

A good tip for naming packages in go is to not consider what code is inside, but how it will be used. That is, the name of the package should be a good namespace for the types/functions exposed. Here, maybe it should simply be `gqlgen.Generate(cfg *config.Config)` ?

vektah:
> A good tip for naming packages in go is to not consider what code is inside, but how it will be used. That is, the name of the package should be a good namespace for the types/functions exposed. Here, maybe it should simply be gqlgen.Generate(cfg *config.Config) ?

We did recently (0.7?) deprecate using `main.go`, maybe we can reclaim the root package?

PR #493, opened 2019-01-09
Return string

jamesallison:

I have:


cmd/init.go
```diff
@@ -7,7 +7,7 @@ import (
 	"os"
 	"strings"
 
-	"github.com/99designs/gqlgen/codegen"
+	"github.com/monzo/gqlgen/codegen"
```
vektah: Most vendoring tools will let you rewrite the repo without having to change the imports, for dep it you would add this to your `Gopkg.toml` of your app using gqlgen:

```toml
[[constraint]]
  branch = "master"
  name = "github.com/99designs/gqlgen"
  source = "https://github.com/monzo/gqlgen"
```

handler/graphql.go
```diff
@@ -279,6 +281,102 @@ func GraphQL(exec graphql.ExecutableSchema, options ...Option) http.HandlerFunc
 	return handler.ServeHTTP
 }
 
+func TyphonHandlerFromGraphQL(exec graphql.ExecutableSchema, options ...Option) typhon.Service {
```
vektah: Rather than support all transports forever we should work how to make handler open for extension.

The way this is setup currently, everyone using gqlgen would need to fetch `monzo/typoon`.

PR #497, opened 2019-01-14
Small fixes

azavorotnii:
No particular issue, just small things that I found in code.

### DIFF
--- a/codegen/testserver/generated_test.go
+++ b/codegen/testserver/generated_test.go
@@ -125,6 +125,7 @@ func TestGeneratedServer(t *testing.T) {
 
 	t.Run("subscriptions", func(t *testing.T) {
 		t.Run("wont leak goroutines", func(t *testing.T) {
+			runtime.GC() // ensure no go-routines left from preceding tests
 			initialGoroutineCount := runtime.NumGoroutine()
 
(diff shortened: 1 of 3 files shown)


azavorotnii:
Strange failure:
```
    --- FAIL: TestGeneratedServer/subscriptions (2.01s)
        --- FAIL: TestGeneratedServer/subscriptions/wont_leak_goroutines (2.01s)
	Error Trace:	generated_test.go:151
	Error:      	Not equal: 
	            	expected: 9
	            	actual  : 8
	Test:       	TestGeneratedServer/subscriptions/wont_leak_goroutines
```
So at the end one less goroutine active than was at the beginning of test. Should there be settlement loop at the beginning of test as well? It just looks one extra goroutine was active at the beginning of test.

vektah:
Thanks for the PR. 

Please make sure any future work is based on the `next` branch, there are some large refactors that would make rebasing between the two pretty hard.

vektah:
You can track the 0.8 progress on the [board](https://github.com/99designs/gqlgen/projects/1). The refactor card is much much larger than the features we are adding, so probably the next week or two.

Non breaking bug fixes could probably go straight to master for a quick 0.7.2, but right now, with such a large refactor it would basically need to be written twice.

PR #498, opened 2019-01-15
add `content-type: text/html` header to playground handler

vilterp:
This ensures that the browser doesn't think it should download the page
instead of rendering it, if the handler goes through a gzipping
middleware such as https://github.com/phyber/negroni-gzip.

I have:

(Sorry, not quite sure where relevant docs or tests are)

### DIFF
--- a/handler/playground.go
+++ b/handler/playground.go
@@ -42,6 +42,7 @@ var page = template.Must(template.New("graphiql").Parse(`<!DOCTYPE html>
 
 func Playground(title string, endpoint string) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Add("Content-Type", "text/html")
 		err := page.Execute(w, map[string]string{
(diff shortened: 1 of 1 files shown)


azavorotnii:
Please rebase on top of "next" branch.

vektah:
I think your gzip middleware is misbehaving.

`net/http` will sniff the content type by calling [DetectContentType](https://golang.org/src/net/http/sniff.go?s=646:688#L11), if your middleware isn't doing that lots of things would be broken. Maybe internally negroni does something, but this sniffing is part of the normal http contract for net/http.

vilterp:
Hm, it seems to call that: https://github.com/phyber/negroni-gzip/blob/master/gzip/gzip.go#L71 But it doesn't come out as `text/html`. Not sure where it's getting messed up. Will come up with a minimal repro when I have a minute.

vilterp:
Thanks!

PR #502, opened 2019-01-21
Model plugin

vektah:
This changes model generation to be nearly entirely independent of codegen, it only shares the config and some utils for binding to the go program.

Was going to do this much later, but pulling model generation out of codegen entirely lets us add a new constraint to the binding phase: All types have backing go implementations. This should allow us to delete a whole step (Named Type collection), removing a whole lot of indirection. Will probably roll this change into allowing multiple types as its going to be a pretty large change to the type system.

Plugin interfaces still very experimental.


gedw99:
@vektah Wow this is a huge change but going to make things much simpler !

Issue #506, opened 2019-01-23
Input types: cant use methods for setters


danilobuerger:
### Expected Behaviour

As per https://gqlgen.com/reference/resolvers/ for types to bind to a method, I want to have the same ability for inputs.

### Actual Behavior

gqlgen errors out.

### Minimal graphql.schema and models to reproduce

It should be possible to define 

```
type SomeInput struct {
	ID string `json:"id"`
}

func (i *SomeInput) SomeField(v string) {
}
```

to be able to set `SomeField: String!` through a method.

On a side note: This is one potential solution to #505 because we then know if the field has been specified or not.

danilobuerger:
@mathewbyrne @vektah I would like to create a PR for this but I am stuck.

In `codegen/util.go` I have extended `bindObject` to also bind to setter methods and set the related fields on `Field` like `bindMethod` does for getters.

Now I wonder how to actually generate this. In `codegen/type.go` the unmarshaling is generated but this is bound to `Type` not `Field` so I don't have access to the previous set information on the setter. How should I proceed here?
* Extend `Type`?
* Overwrite `Unmarshal` in `Field`?
* Something else?

vektah:
Make sure you are starting from the `next` branch, there have been a lot of changes https://github.com/99designs/gqlgen/compare/next

The next thing on my list is to see if TypeReference (formerly Type) can be folded back directly into field, might be better to wait?

danilobuerger:
Sure, can you please ping me when its done so I can start working on this?

danilobuerger:
@vektah any news? I would really like to start with this...

mathewbyrne:
The `next` branch is still in flux, and will be merging into master with the 0.8 release. No specific timeframe but we're thinking the next week or two.

vektah:
@danilobuerger the large refactor ( #534 ) has landed, make sure you base off next.

steebchen:
The stale bot is very dangerous here, as it seems the issue is resolved while in reality it's still open...

jszwedko:
Agreed, I would very much like to see this issue solved as a solution for #505. Is it possible to reopen this @vektah ?

moxar:
Is there any news regarding this feature ? It labeld for v0.10, but it seems that it is not implemented in v0.10 nor v0.11 @vektah

PR #509, opened 2019-01-24
Do not try to bind method on input object

danilobuerger:
Related to #506, it shouldn't be possible to bind a getter method to an input object.


vektah:
We aren't accepting codegen changes against master, all feature work should be on `next`.

PR #514, opened 2019-01-26
Add support for go modules

### LINKED ISSUE #226: Go Modules
This morning I tried to convert our project to use go modules instead of dep. 
I found that gqlgen makes this currently impossible, hence I take this opportunity to document my findings here.

The first minor annoyance is the tags in gqlgen are not recognized as version tags for go modules. This can be easily fixed by adding new version tags or renaming the existing ons ("0.3.0" -> "v0.3.0").

Using the gqlgen code generator outside the GOPATH in a module is currently not working. This is because gqlgen uses go/build to retrieve package information. go/build is currently not go modules aware.
There is an issue https://github.com/golang/go/issues/26504 and a CL https://go-review.googlesource.com/c/go/+/125296 trying to fix this. It is also mentioned that ideally users should use https://godoc.org/golang.org/x/tools/go/packages instead, which however  is not finalized and depends on go1.11.

So where does this leave us. I would recommend to switch the version tags to the go module format rather sooner than later. If something depends on the current format, tagging with both is an option.

For the second problem I think we should just wait for the go/build fix and see if it works for us. Switching to go/packages seems to restricting in my point of view.

vektah:
Recent refactors have made this pretty easy. 

Replaces all references to `go/build` and `golang.org/x/tools/go/loader` with `golang.org/x/tools/go/packages` to get module support.

Looks like this will speed up generation by ~2x.

Closes #226


Genert:
Amazing!

PR #522, opened 2019-02-02
Shows: adjustments to submodule use

### LINKED ISSUE #521: model from submodule is not resolved during codegen
### Expected Behaviour
`go generate` via script would run after conversion to go modules
### Actual Behavior
I'm getting `modelgen: unable to find type` with type referencing `gqlgen.yaml` directive to look for one of the types in a submodule.

Here's the project structure:
```
gqlgen.yaml
scripts/ — (gqlgen script is here)
src/ — (root module, the project)
    gqlgen/  — (module)
        models.go — (model)
        graphql.go — (exec)
    pkg/
        transport/ — (submodule)
            graphql.go — (resolver with go generate)
        endpoint/ — (submodule)
            sometype.go — (referenced type is here)
        ...  
```
so `gqlgen.yaml` is something along the lines of:
```yaml
schema: src/gqlgen/schema.graphql
exec:
  filename: src/gqlgen/graphql.go
model:
  filename: src/gqlgen/models.go
struct_tag: json
models:
  SomeType:
    model: domain.com/projectname/pkg/endpoint.SomeType
resolver:
  filename: src/pkg/transport/graphql.go
  type: Resolver

```
while `go.mod` in `domain.com/project/gqlgen` module will have to say:
```go
require (
	domain.com/projectname/pkg/endpoint v0.0.0
	// and dependencies of the endpoint module
)

replace (
	domain.com/projectname/pkg/endpoint => ../pkg/endpoint
	// and dependencies of the endpoint module
)
```

The local (there is no `https://` repo) `pkg` modules are specified explicitly via `require` & `replace` directives. `src` directory is not part of module names, but the project builds fine with old gqlgen code (`go build` finds everything just fine). `go generate` shows network `go get`'s, so I'm not sure if this is a gqlgen or go problem.

ermik:
This addresses failing execution due to `go mod` resolving old-style `/path/based` submodules to non-mod packages off `github.com`. Explicit linking (in some cases to the parent) and existence of `go.mod` files in subdirectories that can be seen in `import` statements tells the compiler to "color within the lines" i.e. stay in the fetched code. This is achieved with `require` and [`replace` directives](https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive) across module paths.

I have:
 - N/A Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - N/A Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

This PR is first and foremost a working example, and may be closed rather than merged to allow maintainers to investigate the module system further as well as add tests that would **eliminate all `$GOPATH`-based `gqlgen`-related code`** as one of the measures that would enforce the encapsulated nature of module use.

Closes #521


ermik:
Re: CI:
```sh
Step 4/6 : COPY go.* /projects/gqlgen/
 ---> 97bbf03d95d2
Step 5/6 : RUN go mod download
 ---> Running in f63dea3ffd16
```
the download fails because only first-level `go.` files are added to the container. In our project we forfeited the precision of `dep` when used in Docker builds and instead copy everything (all buildable code) at once. That hooks all submodule directories and their `go.` module definitions.

cmd/go.mod
```diff
@@ -0,0 +1,26 @@
...
+	github.com/99designs/gqlgen/graphql/introspection => ../graphql/introspection
+	github.com/99designs/gqlgen/handler => ../handler
+	github.com/99designs/gqlgen/internal/code => ../internal/code
+	github.com/99designs/gqlgen/internal/imports => ../internal/imports
+	github.com/99designs/gqlgen/plugin => ../plugin
+	github.com/99designs/gqlgen/plugin/modelgen => ../plugin/modelgen
+)
+
+require (
+	github.com/99designs/gqlgen v0.4.5-0.20190203203210-e4679b668de0
```
vektah: This is whack

vektah:
I'm going to close this, I think you're environment is broken or has constraints that havnet been considered by the go modules team.

From https://github.com/golang/go/wiki/Modules#modules
> A module is a **collection of related Go packages** that are versioned together as a single unit.
>
> Modules record precise dependency requirements and create reproducible builds.
> 
> Most often, a version control repository contains exactly one module defined in the repository root. (Multiple modules are supported in a single repository, but typically that would result in more work on an on-going basis than a single module per repository).

This is inline with my understanding of modules, you should only need one in the root of the repo.

Here is my test process, using docker for a clean working environment:
```bash
docker run -it golang:1.11.5 bash
mkdir /testproject
cd /testproject
go mod init example.com/project
cat > gqlgen.go <<EOL
// +build ignore

package main

import "github.com/99designs/gqlgen/cmd"

func main() {
    cmd.Execute()
}
EOL

# grab the next version for modules support
go get github.com/99designs/gqlgen@next

# everything generates ok
go run gqlgen.go init

# everything runs ok
go run server/server.go

# only one gqlgen has been installed
find / -iname 'introspection'
# /go/pkg/mod/github.com/99designs/gqlgen@v0.4.5-0.20190127090136-055fb4bc9a6a/codegen/testserver/introspection
# /go/pkg/mod/github.com/99designs/gqlgen@v0.4.5-0.20190127090136-055fb4bc9a6a/graphql/introspection
```

PR #524, opened 2019-02-03
Include OperationDefinition in RequestContext

enjoylife:
A motivating example is where you have a query doc with multiple definitions and you wish to know which one is executed.

E.g. Given the following query doc, which is common when using graphiql or another explorer:
```
query apiOp {
  # ...
}

query apiOpTwo {
 # ...
}
```
And the request is executed with:

```
{"query":"...", operationName":"apiOpTwo"}
```
You want to log the operation name in logs tracing, etc.  But without the specific operation definition parsed from the request you don't know which one. 

By including this PR, the following becomes possible
```
func OperationNameFromCtx(ctx context.Context) string {
	requestContext := graphql.GetRequestContext(ctx)
	// we use the potentially added Op to find the name of the query being executed
	if requestContext.Op != nil && requestContext.Op.Name != "" {
		return requestContext.Op.Name // => apiOpTwo
	}
	return "nameless-operation"
}
```

### DIFF
--- a/graphql/context.go
+++ b/graphql/context.go
@@ -17,6 +17,7 @@ type RequestContext struct {
 	RawQuery  string
 	Variables map[string]interface{}
 	Doc       *ast.QueryDocument
+	Op        *ast.OperationDefinition
 
 	ComplexityLimit      int
 	OperationComplexity  int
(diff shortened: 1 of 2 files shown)


enjoylife:
Oh, I see. Yeah, for batching this would fall flat in its goal. Now that you mention it, I'm curious on how the canonical name is generated by Apollo, I'll look into it. 

Also I have to admit, I have been prioritizing the dev experience lately as the generic `Graphql Query` span names I see in dev are personally annoying.

As  for the security/attack surface, one of the big reasons we see this package as being great, is in the 1st class api for preventing complexity explosion via `ComplexityLimit` 👍.
This has been great for a number of reasons, foremost the following:
1. It's freed us up to to rely on operation names, and other brittle markers but instead, codify hard limits. 
2. And it's helped to highlight whats going to break/cap out in prod, via swapped out interfaces during automated stress tests. 

You can probably tell but are focus leans more towards dev friendliness as our usage of the lib is mostly for internal tools atm.

In any case, it seems like there's some interesting higher level thoughts around this. e.g. the batching, so I'm totally cool with this being closed and relooked at in the future when batching or other related tasks are prioritized.

PR #529, opened 2019-02-04
Add websocket keepalive support

sinamt:
This adds websocket keepalive support for subscriptions.

By default keepalive is not enabled. A new `handler.Option` function called `handler.WebsocketKeepAliveDuration(duration time.Duration) Option` is added to allow configuration of the keepalive delay duration. The keepalive only fires when no other write has been issued for the delay duration.

This code has been extracted from the previously closed https://github.com/99designs/gqlgen/pull/125.

I have:
 - ~~[ ] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))~~ N/A


handler/graphql.go
```diff
@@ -241,6 +243,14 @@ func CacheSize(size int) Option {
 
 const DefaultCacheSize = 1000
 
+// WebsocketKeepAliveDuration allows you to reconfigure the keepAlive behavior.
+// By default, keep-alive is disabled.
```
vektah: should we enable this by default and provide a way to opt out?

sinamt:
I've changed this to a constant tick rate for the keepalive, ignoring any activity written to the socket. It turns out that the [client we use](https://github.com/apollographql/subscriptions-transport-ws/) expects a constant tick rate for the keepalive. The upside is it simplifies the code a little.

> Does it work OK in a browser?

Yeah, it does now with the constant tick rate change.

handler/graphql.go
sinamt: 👍 I've enabled it by default in https://github.com/99designs/gqlgen/pull/529/commits/b2f6b845c7ce9614bbde43206b0bbbb9edc4c6f3 with a ping interval of 25s (matching the expected default interval of 30s in our client https://github.com/apollographql/subscriptions-transport-ws/) if this seems reasonable?

Issue #519, opened 2019-01-30
Missing content type with invalid json as query


vektah:
### Expected Behaviour
curl --header "Content-Type: application/json" --request POST --data 'fff' http://localhost:8080/graphql

should return Content-Type application/json

### Actual Behavior

Isn't setting the content type.

Sonna:
Is it not being set within the GraphQL handler `ServeHTTP`?
https://github.com/99designs/gqlgen/blob/39216361225c6dc824331ad1a218c9c931cc0985/handler/graphql.go#L323

Or would you want the generated server template prefix with a response writer to fix the headers?

https://github.com/99designs/gqlgen/blob/39216361225c6dc824331ad1a218c9c931cc0985/codegen/templates/server.gotpl#L22

```gotpl
        http.With(func(i http.Handler) http.Handler {
		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			w.Header().Set("Content-Type", "application/json")
		})
	}).Handle("/query", handler.GraphQL({{ lookupImport .ExecPackageName }}.NewExecutableSchema({{ lookupImport .ExecPackageName}}.Config{Resolvers: &{{ lookupImport .ResolverPackageName}}.Resolver{}})))
```

PR #532, opened 2019-02-05
Fix set header to JSON earlier in GraphQL response

Sonna:
Update the GraphQL handler to set the Response Header to JSON earlier for
error messages to be returned as JSON and not text/html.

Fixes https://github.com/99designs/gqlgen/issues/519

## Notes:
- Add checks for JSON Content-Type checks in decode bad queries tests


Sonna:
Why do the CircleCI checks fail?

PR #534, opened 2019-02-06
Multiple bind types

vektah:
:hot_pepper: :hot_pepper: :hot_pepper: :hot_pepper: :hot_pepper: 

This PR allows `models` in the config to optionally be an array:
```
  ID:
    model:
      - github.com/99designs/gqlgen/graphql.IntID
```
When mapping onto structs any of these types can be used. When used as an argument the first one is used.

This should make working with other generated code like protobuf much nicer by allowing automatic marshaling of `timestamp.Time` or `time.Time` into graphql Time.

This was made possible by some really heavy refactors of codegen:
 - all templates are now colocated next to the code that builds the data
 - server, stub and model generation are now plugins
 - no more recursive templates
 - all marshalling now happens in small functions, instead of in place at each site
 - the NamedTypes pass has been removed, TypeReferences are built by calling the Binder instead.


codegen/templates/templates.go
```diff
@@ -22,27 +22,105 @@ import (
...
+
+func Render(cfg Options) error {
 	if CurrentImports != nil {
 		panic(fmt.Errorf("recursive or concurrent call to RenderToFile detected"))
 	}
-	CurrentImports = &Imports{destDir: filepath.Dir(destFile)}
+	CurrentImports = &Imports{destDir: filepath.Dir(cfg.Filename)}
+
+	// load path relative to calling source file
+	_, callerFile, _, _ := runtime.Caller(1)
```
mathewbyrne: Neat trick 😄 I was super confused about how templates were being referenced for a minute there. May eventually require docs since it will probably be something plugins want to commonly do (or a separate interface)

plugin/resolvergen/resolver.gotpl
```diff
@@ -0,0 +1,40 @@
+// THIS CODE IS A STARTING POINT ONLY. IT WILL NOT BE UPDATED WITH SCHEMA CHANGES.
```
mathewbyrne: Yep, maybe a decent middle ground for now rather than remove this entirely.

mathewbyrne:
Ok I've spent about 30 mins reading so far, tough PR to give a review on given it's scope and I'd like to get a better understanding of the overall changes.  Couple of quick thoughts:

1. I really like the layout of generated code now, very intuitive to read and to find where it was generated from.

2. I'm not seeing many new test cases around the type mapping — is that something we should work on after this merges?

codegen/args.go
```diff
@@ -0,0 +1,114 @@
...
+	Value         interface{} // value set in Data
+}
+
+func (f *FieldArgument) Stream() bool {
+	return f.Object != nil && f.Object.Stream
+}
+
+func (b *builder) buildArg(obj *Object, arg *ast.ArgumentDefinition) (*FieldArgument, error) {
+	def := b.Schema.Types[arg.Type.Name()]
+	if !def.IsInputType() {
```
mathewbyrne: Are these sorts of checks more in the realm of parser validation? Should gqlgen always be operating under the assumption that the schema is valid and well-formed?

vektah: Good point, Will update the parser.

mathewbyrne: Cool, yeah I noticed a few of these around. I think it's a reasonable assumption, and our fixes for codegen panics recently have been to catch issues in parser validation.

vektah:
> I'm not seeing many new test cases around the type mapping — is that something we should work on after this merges?

Yeah, Just the one in todo with ID. Will flesh it out more later along with the docs. Dont want to keep this PR sitting around too long.

PR #542, opened 2019-02-09
Add more validation checks on .yml config file

Elgarni:
This PR adds two validation checks after parsing gqlgen.yml:

- Check if the config file has two package names defined in the same directory
- Check if the same filename has been used twice. For instance using **gql/gen.go** for both _exec_ and _model_ fields.

Although this may seem intuitive, but it can get tricky in a few cases.
For instance, the following gqlgen.yml config file:
```yml
schema:
- schema.graphql
exec:
  filename: generated/exec.go
  package: graphql
model:
  filename: generated/models.go
resolver:
  filename: generated/resolver.go
  type: Resolver
```
When generating the code, the following error message shows up:
```bash
generating resolver failed: resolver build failed: required package was not loaded: github.com/elgarni/project/generated.User
exit status 2
```
The message could be more descriptive. Moreover, the code was generated for exec.go and models.go, which will not compile anyways since they have different package names.

After this PR, the error message becomes:
```bash
invalid config format: filenames generated/exec.go and generated/models.go are in the same directory but have different package definitions
exit status 1
```
And no code was generated.

The objective of this PR is to stop execution early on, if there are errors in the config file, and not generating any code unless everything is defined correctly in gqlgen.yml

### DIFF
--- a/codegen/config.go
+++ b/codegen/config.go
@@ -216,10 +216,10 @@ func (cfg *Config) Check() error {
 		if inSameDir {
 			for _, previous := range prevPkgList {
 				if current.Package != previous.Package {
-					eitherPackageEmpty := previous.Package != "" || current.Package != "";
+					eitherPackageEmpty := previous.Package != "" || current.Package != ""
 					if eitherPackageEmpty {
 						if current.Package == filepath.Base(current.Dir()) && previous.Package == "" {
-							break;
+							break
 						}
 						if previous.Package == filepath.Base(previous.Dir()) && current.Package == "" {
 							break
--- a/codegen/config.go
+++ b/codegen/config.go
@@ -191,6 +191,47 @@ func (cfg *Config) Check() error {
 	if err := cfg.Resolver.Check(); err != nil {
 		return errors.Wrap(err, "config.resolver")
 	}
+
+	// check packages names against conflict, if present in the same dir
+	// and check filenames for uniqueness
+	packageConfigList := []PackageConfig{
+		cfg.Model,
+		cfg.Exec,
+		cfg.Resolver,
+	}
+	filesMap := make(map[string]bool)
+	pkgConfigsByDir := make(map[string][]PackageConfig)
+	for i, current := range packageConfigList {
+		if i == 0 {
+			filesMap[current.Filename] = true
+			pkgConfigsByDir[current.Dir()] = []PackageConfig{current}
+			continue
+		}
+		_, fileFound := filesMap[current.Filename]
+		if fileFound {
+			return fmt.Errorf("filename %s defined more than once", current.Filename)
+		}
+		filesMap[current.Filename] = true
+		prevPkgList, inSameDir := pkgConfigsByDir[current.Dir()]
+		if inSameDir {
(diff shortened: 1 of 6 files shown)


codegen/config.go
```diff
@@ -191,6 +191,47 @@ func (cfg *Config) Check() error {
...
+	// and check filenames for uniqueness
+	packageConfigList := []PackageConfig{
+		cfg.Model,
+		cfg.Exec,
+		cfg.Resolver,
+	}
+	filesMap := make(map[string]bool)
+	pkgConfigsByDir := make(map[string][]PackageConfig)
+	for i, current := range packageConfigList {
+		if i == 0 {
```
vektah: this branch looks unnecessary

codegen/config.go
```diff
@@ -191,6 +191,47 @@ func (cfg *Config) Check() error {
...
+			return fmt.Errorf("filename %s defined more than once", current.Filename)
+		}
+		filesMap[current.Filename] = true
+		prevPkgList, inSameDir := pkgConfigsByDir[current.Dir()]
+		if inSameDir {
+			for _, previous := range prevPkgList {
+				if current.Package != previous.Package {
+					eitherPackageEmpty := previous.Package != "" || current.Package != ""
+					if eitherPackageEmpty {
+						if current.Package == filepath.Base(current.Dir()) && previous.Package == "" {
```
vektah: This is reproducing logic from https://github.com/99designs/gqlgen/blob/next/codegen/config/config.go#L138
which is called by c.Exec.Check() earlier in this function

Maybe its best to let normalise run first?

codegen/config.go
```diff
@@ -191,6 +191,47 @@ func (cfg *Config) Check() error {
...
+			pkgConfigsByDir[current.Dir()] = []PackageConfig{current}
+			continue
+		}
+		_, fileFound := filesMap[current.Filename]
+		if fileFound {
+			return fmt.Errorf("filename %s defined more than once", current.Filename)
+		}
+		filesMap[current.Filename] = true
+		prevPkgList, inSameDir := pkgConfigsByDir[current.Dir()]
+		if inSameDir {
```
vektah: I think there is only one valid value for a dir? we can drop the slice.

Taking into account the comment below, I think this can all collapse down into:
```go
pkgNameForDir := map[string]string{}
for i, current := range packageConfigList {
	// snip

	prevName, inSameDir := pkgConfigsByDir[current.Dir()]
	if inSameDir && prevName != current.Package {
	    return fmt.Errorf("filenames %s and %s are in the same directory but have different package definitions", current.Filename, previous.Filename)
	}

	pkgConfigsByDir[current.Dir()] = current.Package
}
```

codegen/config.go
Elgarni: True. Will be removed

codegen/config.go
Elgarni: Exactly. I think config.Check() should be called after it's been normalized.
I'll see how that will work.

codegen/config.go
Elgarni: This look much better. Assuming the config is normalized, this should be doable.

PR #544, opened 2019-02-09
Fix directives on fields with custom scalars

enjoylife:
The bug occurs when using a custom defined scalar, an input type, and  directive together.
E.g.
```gql
#schema.gql
scalar UUID
directive @direc(s: String!) on FIELD_DEFINITION | ARGUMENT_DEFINITION | INPUT_FIELD_DEFINITION
input CustomInput {
    stuff:[UUID]  @direc(s:"stuff")
}
```
```go
// custom models
import "github.com/gofrs/uuid"
func MarshalUUID(id uuid.UUID) graphql.Marshaler{ /* ... */ }
func UnmarshalUUID(input interface{}) (uuid.UUID, error) { /* ... */ }
```

would generate:
```go
// ...
// shouldn't be whole path of github.com/gofrs/...
if data, ok := tmp.([]*github.com/gofrs/uuid.UUID) ; ok {
    it.Stuff = data
} else {
    return it, fmt.Errorf(`unexpected type %T from directive, should be []*github.com/gofrs/uuid.UUID`, tmp)
}
```
now: 

```go
if data, ok := tmp.([]*uuid.UUID); ok {
    it.Stuff = data
} else {
    return it, fmt.Errorf(`unexpected type %T from directive, should be []*github.com/gofrs/uuid.UUID`, tmp)
}

```

### DIFF
--- a/codegen/input.gotpl
+++ b/codegen/input.gotpl
@@ -35,7 +35,7 @@
 					if err != nil {
 						return it, err
 					}
-					if data, ok := tmp.({{ $field.TypeReference.GO }}) ; ok {
+					if data, ok := tmp.({{ $field.TypeReference.GO | ref }}) ; ok {
 						it.{{$field.GoFieldName}} = data
 					} else {
 						return it, fmt.Errorf(`unexpected type %T from directive, should be {{ $field.TypeReference.GO }}`, tmp)
--- a/codegen/testserver/generated.go
+++ b/codegen/testserver/generated.go
@@ -759,10 +759,13 @@ input OuterInput {
     inner: InnerInput!
 }
 
+scalar ThirdParty
+
 input InputDirectives {
     text: String! @length(min: 0, max: 7, message: "not valid")
     inner: InnerDirectives!
     innerNullable: InnerDirectives
+    thirdParty: ThirdParty @length(min: 0, max: 7, message: "not valid")
 }
 
 input InnerDirectives {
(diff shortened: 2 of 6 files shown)


vektah:
Needs a test

enjoylife:
Added to the integration schema such that it will catch said bug during go generate.

Prior to diff running go generate with the added schema:

```
❯ go generate ./...
gofmt failed on generated.go: /Users/matt/Projects/gqlgen/codegen/testserver/generated.go:3856:37: expected ')', found '/' (and 5 more errors)
Generated github.com/99designs/gqlgen/codegen/testserver in 1.23s
Generated github.com/99designs/gqlgen/example/chat in 1.04s
Generated github.com/99designs/gqlgen/example/config in 1.04s
Generated github.com/99designs/gqlgen/example/dataloader in 1.33s
Generated github.com/99designs/gqlgen/example/scalars in 1.18s
Generated github.com/99designs/gqlgen/example/selection in 1.03s
Generated github.com/99designs/gqlgen/example/starwars in 1.15s
Generated github.com/99designs/gqlgen/example/todo in 1.17s
Generated github.com/99designs/gqlgen/example/type-system-extension in 1.08s
Generated github.com/99designs/gqlgen/integration in 1.24s
```

After diff it builds fine

enjoylife:
Not sure if the failure is from me force pushing or something else as it just works local.

vektah:
> After diff it builds fine. @vektah I'm hoping that would cover the "test" ask? If not, I would need some guidance on what would be a good example test to follow/improve.

Yeah, that looks good. As long as it cant happen again without something breaking. Most of the templates are covered like this.

> Not sure if the failure is from me force pushing or something else as it just works local.

Its failing because you've changed the testserver schema but havent run `go generate ./...` and committed the changes.

vektah:
I see, looks like theres some instability in the generated code. It looks unrelated to this change so Ill fix it on next.

Issue #533, opened 2019-02-06
Improve Field Collection API


mathewbyrne:
We get a number of support requests around users wanting to know which fields were selected in a query.  This is possible via the `CollectFieldsCtx` helper method, but it's a little unintuitive, not well documented, and could be simplified a bit.

[Field collection](https://facebook.github.io/graphql/June2018/#sec-Field-Collection) is the process of resolving a selection set down to the fields that are relevant for the resolved object.  In generated code, field collection is run after the resolver, because we need to know what the GraphQL Type of the resolved value will be so that we can ensure we only relevant collect fields.

The common use-case that is encountered is a resolver that fetches from a data source, and the user wants to prevent over-fetching by limiting their query to only the set of requested fields.  This is _possible_ with `CollectFieldsCtx`, but requires a value for `satisfies []string`, which is not documented.

The other use-case that should be supported is to get the collected fields that satisfy the resolved value.  This is also possible with `CollectFieldsCtx`, however the resolver needs to know all of the possible implementors for a type, which is a problem codegen should be able to solve.

I'm proposing to add two methods to solve these use-cases:

```
func CollectFields(ctx context.Context) []CollectedField
```
This would return a set of all fields, ignoring fragment Type Conditions.  This would be the common use-case for simple sets of fields a user wants to query.

```
func CollectFieldsFor(ctx context.Context, obj interface{}) []CollectedField
```

This would return a set of fields, with respect to the resolved type.  Internally gqlgen would resolve the set of Implementors for a given type and pass it through to `graphql.CollectFields` as `satisfies []string`.

I'm not sure the best way to do the resolving of the type yet, perhaps a method `GetImplementors(obj interface{})` could be generated to support this?  Open to suggestions.

Along with these changes would come some documentation explaining field collection and how users can access this in resolvers.

vektah:
I think the first use case (overfetching) shouldnt expose collected fields at all, its just a `[]string` of the graphql field names. CollectedField exposes child selection sets too which is rarely what you want.

> This would return a set of fields, with respect to the resolved type.

what does this actually look like?
```golang
row := db.Query("SELECT id, type FROM common")

var fields []string
switch row.Type {
   case "left":
      var leftEvent Left
      db.Query("SELECT ${graphql.CollectedFieldsFor(ctx, leftEvent)} FROM left_events", &leftEvent)\
      return leftEvent
   case "joined":
      var joinedEvent Joined
[...]

Issue #280, opened 2018-08-09
Improve interface to collect query fields from resolver 


oskanberg:
Here's an example schema:

```graphql
interface Thing {
    name: String!
    permissions: [String]
}

type BigThing implements Thing {
    name: String!
    permissions: [String]
}

type SmallThing implements Thing {
    name: String!
    permissions: [String]
}

type Query {
    getThings(...) [Thing]
}
```

And queries:

```graphql
getThings(...) {
    name
}

or

getThings(...) {
    name
    permissions
}

or

getThings(...) {
    name
    ... on BigThing {
        permissions
    }
}
```

In the query resolver for getThings(...), I want to be able to change behaviour based on whether `permissions` was requested. In order to understand whether or not a given query has included the `permission` property, I need to collect the fields from all fragments and check them:

```golang
fields := graphql.CollectFieldsCtx(ctx, ["BigThing", "SmallThing"])
for _, field := range fields {
   if field.Name == "permissions" {
       // it was included
   }
}
```

Where `fields` may well include `"permissions"` several times, if there are conditional fragments in the query. If I need separate logic for each inherited type, I would need to do make the `CollectFieldsCtx` call multiple times.

@vektah suggested that this interface could be improved, in order to more easily collect the requested fields, regardless of type.

I'm fairly new to this, so I'm not sure how such a simplified interface would look. I would tentatively suggest that something like this might be helpful:

```golang
// returns a map of [type]bool to indicate whether the given property was requested on that type
func gqlgen.ResolverQueryHasProperty(ctx context.Context, property string) map[string]bool
```

Alternatively a new resolver interface method could be introduced that also accepts a method to check if a given property was accepted.

```golang
type RequestChecker func(subtype string, property string) bool

func QueryWithCheck_getThings(ctx context.Context, params ..., check RequestChecker) ([]*Thing, error) {
    requested := check("BigThing", "permissions")
    if requested {
        // permissions was requested on BigThings
    }
}
```

thechriswalker:
I use this sort of functionality a lot, mostly to check the fields requested in order to make minimal database queries. 

Although it seems complex, a couple of simple utility functions cover the required functionality (including arbitrarily nested fragments/inline fragments) and I feel muddying the resolver API would be counter-productive. 
Maybe the functions themselves could be added to the library as available utilities, rather than imposed in the resolver API.

```go
// get the fields requested in a graphql query
// Useful for constructing SQL queries from the GraphQL query
[...]

PR #551, opened 2019-02-18
Improved Collect Fields API and Documentation

mathewbyrne:
This adds a new API helper method `CollectAllFields(context.Context) []string` as discussed in #280 and #533.  The goal is to add a simple API for the common case of wanting to know which fields were requested, and add some documentation around how this feature works.

### DIFF
--- a/graphql/exec.go
+++ b/graphql/exec.go
@@ -16,6 +16,9 @@ type ExecutableSchema interface {
 	Subscription(ctx context.Context, op *ast.OperationDefinition) func() *Response
 }
 
+// CollectFields returns the set of fields from an ast.SelectionSet where all collected fields satisfy at least one of the GraphQL types
+// passed through satisfies. Providing an empty or nil slice for satisfies will return collect all fields regardless of fragment
+// type conditions.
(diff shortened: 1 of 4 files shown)


graphql/exec.go
```diff
@@ -38,7 +41,7 @@ func collectFields(reqCtx *RequestContext, selSet ast.SelectionSet, satisfies []
 			if !shouldIncludeNode(sel.Directives, reqCtx.Variables) {
 				continue
 			}
-			if !instanceOf(sel.TypeCondition, satisfies) {
+			if len(satisfies) > 0 && !instanceOf(sel.TypeCondition, satisfies) {
```
vektah: should this be an explicit nil check? is there a difference between satisfies = nil and satisfies = []? maybe [] is not useful so the distinction is not important?

mathewbyrne: Yeah perhaps.  Currently on `next` an empty slice will collect object fields, and skip any Fragments.  Is that useful?  Maybe, I guess it does potentially break BC if anyone is using it for this use-case.

Still, a `nil` slice vs and empty slice having different behaviours seems a little fragile to me?

vektah: :+1:

PR #553, opened 2019-02-18
Update Getting Started for 0.8 and Go Modules

mathewbyrne:
This updates the Getting Started documentation to work with Go Modules and the 0.8 release in general.

We decided to keep the `dep` documentation around for now in case people aren't ready to upgrade to Go Modules.  But the addition of Go Modules simplifies the setup and execution a lot, however it now makes less sense to deprecate the binary, so it is reintroduced here.

### DIFF
--- a/api/generate.go
+++ b/api/generate.go
@@ -1,4 +1,4 @@
-package gqlgen
+package api
 
 import (
 	"syscall"
--- a/ambient.go
+++ b/ambient.go
@@ -1,4 +1,4 @@
-package gqlgen
+package main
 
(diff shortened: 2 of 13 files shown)


api/generate.go
```diff
@@ -1,4 +1,4 @@
-package gqlgen
+package api
```
mathewbyrne: I started by moving this into `codegen`, but it caused an import loop between `plugin` and `codegen`.  I've moved it here for now, do you have another suggestion?

PR #555, opened 2019-02-18
Handling queries with partial path errors

jszwedko:
Opening this as a PR since I felt like providing a diff adding a test was the simplest way to convey my question.

In the case where only part of the query path fails, we currently only receive the error without also receiving the data from the paths that successfully resolved. E.g. for a query like:

```
{
  user {
    name
  }
  organization {
    name
  }
}
```

Where resolving `user` fails, but resolving `organization` is successful, we currently get a response like:

```
{
  "errors": [
    {
      "message": "Some error",
      "path": [
        "user"
      ]
    }
  ],
  "data": null
}
```

Where `organization` is not returned.

My expected behavior would have been:

```
{
  "errors": [
    {
      "message": "Some error",
      "path": [
        "user"
      ]
    }
  ],
  "data": {
    "organization": {
      "name": "Bob"
    }
  }
}
```

I've added an integration test here that currently fails, but would pass with my desired behavior.

If this current behavior is unintentional, I can take a shot at completing this PR to return the expected result (errors only for the paths that fail to resolve and the data for the rest).

### DIFF
--- a/integration/integration-test.js
+++ b/integration/integration-test.js
@@ -70,4 +70,13 @@ describe('Errors', () => {
 
         expect(res.errors[0].message).toEqual('Normal error');
     });
+
+    it('should return errors only for failed resolvers', async () => {
+        let res = await client.query({
+            query: gql`{ viewer { user { name } } error(type: NORMAL) }`,
+        });
+
+        expect(res.errors[0].message).toEqual('Normal error');
+        expect(res.data.viewer.user.name).toEqual('Bob');
+    });
 });


mathewbyrne:
I'd need to see the schema you have, but my guess is that you're fields are non-nullable, and we are therefore bubbling the error up to the parent.  From the spec:

> If the field which experienced an error was declared as Non-Null, the null result will bubble up to the next nullable field.

So I'm pretty sure this is intended behaviour.  If you want to support partial successes, you will need to ensure that the fields are nullable in your schema.

jszwedko:
Ah, yes, apologies, I should have included the schema:

An excerpt:
```
type User {
  id: ID!
  email: String!
}

type Query {
  me: User!
  users: [User]
}
```

In our case, an error in the `me` query will result in no `users` being returned even if that resolution was successful. The query looks like:
```
query {
  me {
    email
   }
  users {
    email
  }
```

I do see that the spec says that if the field is `null` it will bubble up to the next nullable field.

My mistake was misreading them as separate "operations" where it is actually one operation.

Thanks for the conversation @mathewbyrne !

PR #565, opened 2019-03-01
Fix cli config getters

### LINKED ISSUE #560: gqlgen@next disrespects config file
### Expected Behaviour
Running `go run gqlgen/cmd.go -c gqlgen/gqlgen.yml` on gqlgen@latest works fine.
When using with gomodules, installing `next` with `go get github.com/99designs/gqlgen@next` and running `go run gqlgen/cmd.go -c gqlgen/gqlgen.yml` it should work too.

### Actual Behavior
`$ go run gqlgen/cmd.go -c gqlgen/gqlgen.yml` 
`unable to open schema: open schema.graphql: no such file or directory`

-

Extra note: I noticed When I put the gqlgen.yml in the root dir, it works, even though explicitly setting the config file to `gqlgen/gqlgen.yml`.

project structure:
```
main.go
prisma/
  schema.graphql
gqlgen/
  gqlgen.yml
```

If a separate repro repo is needed, let me know and I'll provide one

steebchen:
Fix #560 by using the correct cli flag `"config"`

I have:

Maybe some tests would be good but it may be easier to just create a `const config = "config"` but I wasn't sure on how to integrate it with urfave/cli because that uses `"config, c"` as the flag name.

### DIFF
--- a/cmd/gen.go
+++ b/cmd/gen.go
@@ -20,7 +20,7 @@ var genCmd = cli.Command{
 	Action: func(ctx *cli.Context) {
 		var cfg *config.Config
 		var err error
-		if configFilename := ctx.String("gen"); configFilename != "" {
+		if configFilename := ctx.String("config"); configFilename != "" {
 			cfg, err = config.LoadConfig(configFilename)
 			if err != nil {
 				fmt.Fprintln(os.Stderr, err.Error())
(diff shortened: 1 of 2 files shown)


vektah:
Awesome, thanks!

PR #569, opened 2019-03-04
Introduction Improvements

mathewbyrne:
This aims to improve our introduction and README sections with the following changes:

1. `README.md` and `docs/content/introduction.md` now have the same content.
2. Moved Feature Comparison into the main side out of `README.md`
3. Added and linked to some Contribution Guidelines

~~Currently README and introduction are maintained separately.  I couldn't figure out a way to have them read the same source; GitHub MD has no include functionality, and hugo cannot read files outside of it's root directory (`/docs`).~~

~~They also use different link syntax (hugo uses `<ref >` shorttags) making them a little incompatible.~~

~~Maybe we could lint for this?  It's always going to be a bit fuzzy though.  Otherwise some compile step that writes both files from a single source?~~

Of course, git can track links. With this we are able to link `README.md` into the hugo content dir and munge it's contents there.


docs/content/introduction.md
```diff
@@ -1,31 +1,35 @@
...
+ - **Type safe** — You should never see `map[string]interface{}` here.
+ - **Codegen** — Let us generate the boring bits, so you can build your app quickly. 
 
+[Feature Comparison]({{< ref "feature-comparison.md" >}})
 
-## Getting started
+## Getting Started
 
-First take a look at the [Getting Started]({{< ref "getting-started.md" >}}) tutorial.
+First work your way through the [Getting Started]({{< ref "getting-started.md" >}}) tutorial.
```
mathewbyrne: **TODO** there definitely another paragraph here that mentions something about looking at Resolvers, Type Mapping etc, in roughly the order we think it's important to further learn gqlgen concepts.

PR #579, opened 2019-03-05
Take care about commonInitialisms in ToGo

vvakame:
in v0.8.0, `RelatedURLs` is convert to `RelatedUrls`, and `ImageIDs` be `ImageIds`.

I have:

### DIFF
--- a/codegen/templates/templates.go
+++ b/codegen/templates/templates.go
@@ -129,7 +129,6 @@ func Funcs() template.FuncMap {
 		"lcFirst":       lcFirst,
 		"quote":         strconv.Quote,
 		"rawQuote":      rawQuote,
-		"toCamel":       ToCamel,
 		"dump":          Dump,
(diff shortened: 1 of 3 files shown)


codegen/templates/templates.go
```diff
@@ -232,29 +232,52 @@ func ToCamel(s string) string {
 	if s == "_" {
 		return "_"
 	}
-	buffer := make([]rune, 0, len(s))
+	buf := bytes.NewBuffer(make([]byte, 0, len(s)))
```
vvakame: https://play.golang.org/p/QFS0VH1V90s

vektah: Not needed, safe because all replacements are < 128

https://github.com/golang/lint/blob/06c8688daad7faa9da5a0c2f163a3d14aac986ca/lint.go#L734-L736

PR #582, opened 2019-03-05
Load the playground sources from HTTPS by default

demdxx:
For some browsers on non-secure domains resources from CDN doesn't loads, so I made all cdn.jsdelivr.net resources of the playground by HTTPS by default

Here is no need to add any tests or documentation updates


vektah:
> For some browsers on non-secure domains resources from CDN doesn't loads

which ones?

Does always using https affect cors boundries in a negative way?

demdxx:
It doesn't work in Safari Version 12.0 (14606.1.36.1.9)
Access-Control-Allow-Origin: *

screenshot 2019-03-06 at 1 42 37 pm

PR #584, opened 2019-03-06
Fix for @deprecated directive not appearing

### LINKED ISSUE #580: Deprecated directive on fields not working
### Using v0.8.0

Adding `@deprecated` on a field results with 

```sh
$ go run github.com/99designs/gqlgen -v
merging failed: unable to build object definition: directive deprecated not found
exit status 3
```

### Note

This previously worked on v0.7.1 without issue.

Deprecating Enum still works

```graphql
type Todo {
  id: ID!
  text: String!
  done: Boolean! @deprecated(reason: "Because I can")
}

type Query {
  todos: [Todo!]!
}
```

mathewbyrne:
Fixes #580

During `buildDirectives` in codegen, the builtin directives `skip`, `include`, and `deprecated` are skipped over.  This causes them to not be registered with the builder later when building objects.  It seems like `deprecated` is definitely needed, since it's a schema directive.  `skip` and `include` I'm not so sure about.  @vektah what's the reason for skipping these in build?

### DIFF
--- a/codegen/data.go
+++ b/codegen/data.go
@@ -62,9 +62,16 @@ func BuildData(cfg *config.Config) (*Data, error) {
 		return nil, err
 	}
 
+	dataDirectives := make(map[string]*Directive)
+	for name, d := range b.Directives {
+		if !d.Builtin {
+			dataDirectives[name] = d
+		}
+	}
+
 	s := Data{
 		Config:     cfg,
-		Directives: b.Directives,
+		Directives: dataDirectives,
 		Schema:     b.Schema,
 		SchemaStr:  b.SchemaStr,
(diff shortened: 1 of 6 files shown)


vektah:
That line comes from when they were first added, I'm guessing what that skip actually does has changed a little. 

https://github.com/99designs/gqlgen/commit/09242061c7ce1d6e304f06344914c7bc6788c8b7#diff-5e244cf316790fef0b9112abc3070944R5

mathewbyrne:
Ok, so instead of skipping these directives then, we set a flag and only include directives with the flag in the `DirectivesRoot`.

PR #604, opened 2019-03-11
Fix directives on args with custom type

cevou:

When generating code with a directive for an argument which is a custom type or input the full identifier of the type is returned, which causes errors.

Example:
```
if data, ok := tmp.(github.com/user/repository/model.MyInput) ; ok {
  arg0 = data
} else {
  return nil, fmt.Errorf(`unexpected type %T from directive, should be github.com/user/repository/model.MyInput`, tmp)
}
```

After applying the change, the code is fixed:
```
if data, ok := tmp.(model.MyInput); ok {
  arg0 = data
} else {
  return nil, fmt.Errorf(`unexpected type %T from directive, should be github.com/user/repository/model.MyInput`, tmp)
}
```

### DIFF
--- a/codegen/args.gotpl
+++ b/codegen/args.gotpl
@@ -24,7 +24,7 @@ func (ec *executionContext) {{ $name }}(ctx context.Context, rawArgs map[string]
 				if err != nil {
 					return nil, err
 				}
-				if data, ok := tmp.({{ $arg.TypeReference.GO }}) ; ok {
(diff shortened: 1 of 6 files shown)


vektah:
Needs a test to prevent future regressions :pray:

Should be pretty easy to add something into the codegen/testserver that exercises this edge case

cevou:
Thanks for the hint where to add the tests. Just added them.

vektah:
Thanks for the PR

PR #610, opened 2019-03-12
Allow configuring the complexity limit dynamically per request

tgwizard:
This can be useful if you have different types of clients (internal, 3rd party), or if you want to allowed authorized users to query for more data. `ComplexityLimitFunc` only takes a context now, under the assumption that all data relevant for computing the limit can be fetched from there. Similar to how other middleware works.

I have:


vektah:
Thanks, that seems convenient.

PR #614, opened 2019-03-12
Adding entry for workshop

ivancorrales:
First of all thank you for so great library and I want to recognize that I've tried some Golang libraries and this is the one that I more like!

I did a workshop in my company about GraphQL. And I prepared a workshop with challenges. All the workshop is based 99designs/gqlgen  and that's the reason because I thought maybe you could be interested in publishing the link as part of the README

https://graphql-go.wesovilabs.com/


vektah:
Thanks this looks good! great to have more learning resources.

Issue #603, opened 2019-03-10
Enum argument for directive generating wrong code


cevou:
### Expected Behaviour

I want to specify a directive for an argument which has an enum as an argument.
I expect the generated code to compile afterwards.

### Actual Behavior

The following error is returned when starting the server:
```
cannot use []interface {} literal (type []interface {}) as type []model.Role in argument to ec.executableSchema.directives.MyDirective
```

I assume the `templates.Dump` function return a faulty string.

### Minimal graphql.schema and models to reproduce

```graphql
enum Role {
  ADMIN
  USER
}

directive @myDirective(role: [Role!]!) on ARGUMENT_DEFINITION

schema {
  query: Query
}

type Query {
  test(someId: ID! @hasRole(role: [ADMIN])): [Test]!
}

type Test {
  id: ID!
}
```

PR #615, opened 2019-03-12
Handle enum slices as directive parameter

cevou:
When an enum is defined as an parameter of an directive, the code that is generated contains `[]interface{}` for example `[]interface{}{"ADMIN", "USER"}` instead of the actual type of the enum. This is incompatible with the generated directive function signature which has the explicit type of the enum slice. (see #603)

This change checks after the argument is dumped if the arg is a slice and enum. If this is the case it will replace `[]interface{}` by the actual type.

To me this seems to be a little hacky, but I couldn't find a better place to implement it. I cannot be done in `dump` as there is no reference to the actual type. But there might be a better way to solve this?

I have:

Fixes: #603

### DIFF
--- a/codegen/directive.go
+++ b/codegen/directive.go
@@ -130,6 +130,11 @@ func (d *Directive) ResolveArgs(obj string, next string) string {
 			} else {
 				dArg = templates.Dump(arg.Default)
 			}
+			// for enum slices replace interface{} with actual type to match directive function signature
+			if arg.TypeReference.IsSlice() && arg.TypeReference.Definition.Kind == ast.Enum {
+				t := templates.CurrentImports.LookupType(arg.TypeReference.GO)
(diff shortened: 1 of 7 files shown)


vektah:
Thanks for reporting this issue, but I'm going to close this PR. It looks like there are a heap of bugs in this area and its going to take a bit of time to straighten it out, and I don't want this getting in the way of straightening out any other 0.8 issues.

Its not limited to just enums, eg:
```
directive @complexDirectiveArgs(in: [RecursiveInputSlice!]!) on ARGUMENT_DEFINITION
```
generates
```
return ec.directives.ComplexDirectiveArgs(ctx, tmp, n, []interface{}{map[string]interface{}{"self": map[string]interface{}{"self": nil}}, nil})
```
but that function is expecting a []RecursiveInputType

Issue #616, opened 2019-03-12
Be able to determine if resolver is a function


codyleyhan:
Currently there is no way to determine in a resolver middleware if the field that is being resolved is a function or if its result is just a value on an already resolved instance.  This would be good information to have in order to reduce operations like only tracing spans or logs to when the resolver is a function.

codyleyhan:
After some looking through the code, it seems like all that is needed is to add a field to the resolver context maybe something like `IsMethod` and then adjust the template so that this field is populated during codegen.  If this approach sounds right I can get a PR open in the next few days @mathewbyrne

mathewbyrne:
Yeah I think that makes sense.  `IsMethod` is already available at codegen time, we just don't expose it through the `ResolverContext`.

PR #618, opened 2019-03-13
Adds way to determine if a resolver is a function call or value

codyleyhan:
Addresses #616  by exposing `IsMethod` from codegen to the `ResolverContext` when a resolver is called.

I have:


vektah:
Looks good, thanks for the PR

PR #633, opened 2019-03-18
Fix Having Pointers to Union Types

### LINKED ISSUE #594: [0.8.1] Nullable union fields with resolvers generate methods expecting to return a pointer to an interface
### Expected Behaviour

A schema of:

```graphql
type Foo {
  bar: String
}

union Bar = Foo

type Query {
  bar: Bar
}
```

With:

```yaml
models:
  Baz:
    fields:
      foo:
        resolver: true
```

Will generate a resolver with signature:

```golang
func (r *queryResolver) Bar(ctx context.Context) (Bar, error) {
        panic("not implemented")
}
```

### Actual Behavior

A resolver with signature:

```golang
func (r *queryResolver) Bar(ctx context.Context) (*Bar, error) {
        panic("not implemented")
}
```

is generated where `*Bar` is a pointer to the Go interface `Bar`.

I've worked around this for now by making the type not nullable in the GraphQL schema:

```graphql
type Query {
  bar: Bar!
}
```

But this means the schema is not accurately expressing the fact that `null` can also be returned.

This seems similar to https://github.com/99designs/gqlgen/issues/484

schema.graphql
```graphql
type Foo {
  bar: String
}

union Bar = Foo

type Query {
  bar: Bar
}
```

gengql.yml
```yml
# .gqlgen.yml example
#
# Refer to https://gqlgen.com/config/
# for detailed .gqlgen.yml documentation.

schema:
- schema.graphql
exec:
  filename: generated.go
model:
  filename: models_gen.go
resolver:
  filename: resolver.go
  type: Resolver

models:
  Baz:
    fields:
      foo:
        resolver: true
```

mathewbyrne:
Fixes #594

Adds unions to the list of things that aren't turned into pointers based on the schema.  I also rewrote `CopyModifiersFromAst` to make it a bit simpler to use.

### DIFF
--- a/codegen/config/binder.go
+++ b/codegen/config/binder.go
@@ -399,7 +399,7 @@ func (b *Binder) TypeReference(schemaType *ast.Type, bindTarget types.Type) (ret
 			ref.GO = obj.Type()
 		}
 
-		ref.GO = b.CopyModifiersFromAst(schemaType, def.Kind != ast.Interface, ref.GO)
+		ref.GO = b.CopyModifiersFromAst(schemaType, ref.GO)
 
 		if bindTarget != nil {
 			if err = code.CompatibleTypes(ref.GO, bindTarget); err != nil {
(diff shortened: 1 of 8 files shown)


codegen/config/binder.go
```diff
@@ -414,12 +414,15 @@ func (b *Binder) TypeReference(schemaType *ast.Type, bindTarget types.Type) (ret
...
-func (b *Binder) CopyModifiersFromAst(t *ast.Type, usePtr bool, base types.Type) types.Type {
+func (b *Binder) CopyModifiersFromAst(t *ast.Type, base types.Type) types.Type {
 	if t.Elem != nil {
-		return types.NewSlice(b.CopyModifiersFromAst(t.Elem, usePtr, base))
+		return types.NewSlice(b.CopyModifiersFromAst(t.Elem, base))
 	}
 
-	if !t.NonNull && usePtr {
+	def := b.schema.Types[t.Name()]
+	usePtr := def.Kind != ast.Interface && def.Kind != ast.Union
```
mathewbyrne: @vektah this potentially changes the logic a little bit here, since `usePtr` is recalculated at each recursion, but I think it's maybe more technically correct?  Nothing unexpected seemed to shift after codegen.

vektah: :+1:  I think its effectively the same thing.

PR #634, opened 2019-03-18
Use graphql.String for types wrapping a basic string

### LINKED ISSUE #595: Enums can't be imported
### Expected Behaviour
I want to import existing enums:

```yaml

models:
  Scope:
    model: github.com/steebchen/api/prisma.Scope
```

### Actual Behavior
Some weird errors appear at runtime, after generating the files:
```
$ go run .
# github.com/steebchen/api/gqlgen
gqlgen/gqlgen.go:2393:2: not enough arguments to return
gqlgen/gqlgen.go:2393:11: ec.unmarshalInputScope undefined (type *executionContext has no field or method unmarshalInputScope)
gqlgen/gqlgen.go:2397:11: ec._Scope undefined (type *executionContext has no field or method _Scope)
```

```go
# prisma/prisma.go
type Scope string

const (
	ScopeCompany  Scope = "Company"
	ScopeBranch   Scope = "Branch"
	ScopeEmployee Scope = "Employee"
	ScopeCustomer Scope = "Customer"
)
```

gqlgen@v0.8.1

vektah:
Adds a special case that allows `type Foo string` to fallback to calling whatever marshaler is defined for String.

Not the cleanest approach, but it closes out the regression until we have a better solution.

Fixes #595 

I have:

### DIFF
--- a/codegen/config/binder.go
+++ b/codegen/config/binder.go
@@ -158,9 +158,11 @@ func (b *Binder) PointerTo(ref *TypeReference) *TypeReference {
 	newRef := &TypeReference{
 		GO:          types.NewPointer(ref.GO),
 		GQL:         ref.GQL,
+		CastType:    ref.CastType,
 		Definition:  ref.Definition,
 		Unmarshaler: ref.Unmarshaler,
 		Marshaler:   ref.Marshaler,
+		IsMarshaler: ref.IsMarshaler,
 	}
 
 	b.References = append(b.References, newRef)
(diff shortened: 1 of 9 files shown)


codegen/config/binder.go
```diff
@@ -172,28 +174,34 @@ type TypeReference struct {
 	Definition  *ast.Definition
 	GQL         *ast.Type
 	GO          types.Type
+	BaseType    types.Type  // Before calling marshalling functions cast from/to this base type
 	Marshaler   *types.Func // When using external marshalling functions this will point to the Marshal function
 	Unmarshaler *types.Func // When using external marshalling functions this will point to the Unmarshal function
+	IsMarshaler bool        // Does the type implement graphql.Marshaler and graphql.Unmarshaler
```
mathewbyrne: Could you maybe wrap `IsMarshaler` and `BaseType` into a struct or something to indicate they're only set together?  The names seems a bit generic and since `TypeReference` is used everywhere I can imagine being confused by this field in the future.

vektah: They are mutually exclusive options, builtins never implement Marshal.

Open to naming suggestions, naming is hard. Its the underlying type but only when the type cant be marshalled and is currently only set to string.

PR #652, opened 2019-03-29
add extra builtin types only when not defined

andrey1s:
In 0.8.1 you could override the builtin types (eg `Time`) in config, this beaviour was lost in https://github.com/99designs/gqlgen/pull/635.


vektah:
Why?

andrey1s:
you cannot set the type in the config, for example
```
models:
    Time:
        model:
           - githab.com/.../Time
           - githab.com/.../OtherTime
```

PR #654, opened 2019-03-31
doc: fix mistake on introspection doc page

sharkyze:
This PR fixes a small mistake in the documentation page of `introspection`

I have:


vektah:
Woudlnt it make more sense to disable introspection if you're not an admin? The default for this prop is false, so your change wont do anything?

sharkyze:
> Woudlnt it make more sense to disable introspection if you're not an admin? The default for this prop is false, so your change wont do anything?

Yes I agree, sorry I made the change too quickly. It should be better now.

Issue #473, opened 2018-12-12
Invalid code generation given underscore-only fields


wabain:
### Expected Behaviour

A bit similar to #376 - I added a placeholder field named `_` to my schema and found that gqlgen generated invalid code. I'd expect gqlgen to generate something legal, if maybe ugly. (As far as I can tell, this [is allowed](https://facebook.github.io/graphql/June2018/#sec-Names) by GraphQL.)

### Actual Behavior

The resolver and complexity interface definitions generate invalid Go code:

```go
type ComplexityRoot struct {
	Query struct {
		 func(childComplexity int) int
	}
}

type QueryResolver interface {
	_(ctx context.Context) (*bool, error)
}
```

### Minimal graphql.schema and models to reproduce

```graphql
type Query {
    _: Boolean
}
```

Issue #376, opened 2018-10-11
exposing multiple fields with differing case / underscores causes invalid code generation


vektah:
Given the schema:
```
type Foo {
   sizeMapping: [Sizemapping!]!
   size_mapping: [Sizemapping!]! @deprecated
}
```

The generated complexity struct ends up duplicating a method, which is a compile error:
```
struct {
        SizeMapping func(childComplexity int) int
        SizeMapping func(childComplexity int) int
}
```

Worth checking the other field binding logic isn't doing anything too crazy here.

PR #668, opened 2019-04-10
Fix: complexity case selection

mbranch:
While working on limits to complexity, we found an issue with how code generation builds the Complexity func. Basically, the Complexity func was receiving `field` not as `GoFieldName`, but as the GraphQL name.

It's possible this isn't the correct solution (and `field` param should be a Go name), but based on what's happening in [complexity_test.go](https://github.com/99designs/gqlgen/blob/5ff60925ce417b4689cd745361bb800608192dfe/complexity/complexity_test.go#L210-L220), I think this is correct. I'll try to add tests this week.

I have:

### DIFF
--- a/codegen/generated!.gotpl
+++ b/codegen/generated!.gotpl
@@ -46,7 +46,8 @@ type ComplexityRoot struct {
 {{ range $object := .Objects }}
 	{{ if not $object.IsReserved -}}
 		{{ $object.Name|go }} struct {
-		{{ range $field := $object.UniqueFields -}}
+		{{ range $_, $fields := $object.UniqueFields }}
+			{{- $field := index $fields 0 -}}
 			{{ if not $field.IsReserved -}}
 				{{ $field.GoFieldName }} {{ $field.ComplexitySignature }}
 			{{ end }}
@@ -84,20 +85,25 @@ func (e *executableSchema) Complexity(typeName, field string, childComplexity in
 	switch typeName + "." + field {
 	{{ range $object := .Objects }}
 		{{ if not $object.IsReserved }}
-			{{ range $field := $object.UniqueFields }}
-				{{ if not $field.IsReserved }}
-					case "{{$object.Name}}.{{$field.GoFieldName}}":
-						if e.complexity.{{$object.Name|go}}.{{$field.GoFieldName}} == nil {
-							break
-						}
-						{{ if $field.Args }}
-							args, err := ec.{{ $field.ArgsFunc }}(context.TODO(),rawArgs)
-							if err != nil {
-								return 0, false
+			{{ range $_, $fields := $object.UniqueFields }}
+				{{- $len := len $fields }}
+				{{- range $i, $field := $fields }}
+					{{- $last := eq (add $i 1) $len }}
+					{{- if not $field.IsReserved }}
+						{{- if eq $i 0 }}case {{ end }}"{{$object.Name}}.{{$field.Name}}"{{ if not $last }},{{ else }}:
+							if e.complexity.{{$object.Name|go}}.{{$field.GoFieldName}} == nil {
+								break
 							}
+							{{ if $field.Args }}
+								args, err := ec.{{ $field.ArgsFunc }}(context.TODO(),rawArgs)
+								if err != nil {
+									return 0, false
+								}
+							{{ end }}
+							return e.complexity.{{$object.Name|go}}.{{$field.GoFieldName}}(childComplexity{{if $field.Args}}, {{$field.ComplexityArgs}} {{ end }}), true
 						{{ end }}
-						return e.complexity.{{$object.Name|go}}.{{$field.GoFieldName}}(childComplexity{{if $field.Args}}, {{$field.ComplexityArgs}} {{end}}), true
-				{{ end }}
+					{{- end }}
+				{{- end }}
 			{{ end }}
 		{{ end }}
 	{{ end }}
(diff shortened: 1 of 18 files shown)


vektah:
Theres a bit of legacy on this one, and probably a missing test.

We switched to using go field names because _ is a valid graphql identifier - https://github.com/99designs/gqlgen/issues/473

https://github.com/99designs/gqlgen/issues/376 and https://github.com/99designs/gqlgen/pull/622 followed on from there to fix a few of the issues with that approach.

mbranch:
> Theres a bit of legacy on this one, and probably a missing test.
> 
> We switched to using go field names because _ is a valid graphql identifier - #473
> 
> #376 and #622 followed on from there to fix a few of the issues with the current approach.

Okay. I'm not sure if you mean this is the wrong solution or not, but let me know and I'm happy to spend time revising.

vektah:
Yeah its not quite right. The complexity funcs are set and deduped based on the go field name, not the graphql one to avoid the listed bugs.

Heres a failing that hits on the issue (drop it into `testserver/complexity_test.go`):


```go
func TestComplexityFuncs(t *testing.T) {
	resolvers := &Stub{}
	cfg := Config{Resolvers: resolvers}
	cfg.Complexity.OverlappingFields.Foo = func(childComplexity int) int { return 1000 }
	cfg.Complexity.OverlappingFields.NewFoo = func(childComplexity int) int { return 5 }

	srv := httptest.NewServer(handler.GraphQL(NewExecutableSchema(cfg), handler.ComplexityLimit(10)))
	c := client.New(srv.URL)

	resolvers.QueryResolver.Overlapping = func(ctx context.Context) (fields *OverlappingFields, e error) {
		return &OverlappingFields{
			Foo:    2,
			NewFoo: 3,
		}, nil
	}

	t.Run("with high complexity limit will not run", func(t *testing.T) {
		ran := false
		resolvers.OverlappingFieldsResolver.OldFoo = func(ctx context.Context, obj *OverlappingFields) (i int, e error) {
			ran = true
			return obj.Foo, nil
		}

		var resp struct {
			Overlapping interface{}
		}
		err := c.Post(`query { overlapping { oneFoo, twoFoo, oldFoo, newFoo, new_foo } }`, &resp)

		require.EqualError(t, err, `http 422: {"errors":[{"message":"operation has complexity 1009, which exceeds the limit of 10"}],"data":null}`)
		require.False(t, ran)
	})

	t.Run("with low complexity will run", func(t *testing.T) {
		ran := false
		resolvers.QueryResolver.Overlapping = func(ctx context.Context) (fields *OverlappingFields, e error) {
			ran = true
			return &OverlappingFields{
				Foo:    2,
				NewFoo: 3,
			}, nil
		}

		var resp struct {
			Overlapping interface{}
		}
		c.MustPost(`query { overlapping { newFoo } }`, &resp)

		require.True(t, ran)
	})

	t.Run("with multiple low complexity will not run", func(t *testing.T) {
		ran := false
		resolvers.QueryResolver.Overlapping = func(ctx context.Context) (fields *OverlappingFields, e error) {
			ran = true
			return &OverlappingFields{
				Foo:    2,
				NewFoo: 3,
			}, nil
		}

		var resp interface{}
		err := c.Post(`query { 
			a: overlapping { newFoo },
			b: overlapping { newFoo },
			c: overlapping { newFoo },
		}`, &resp)

		require.EqualError(t, err, `http 422: {"errors":[{"message":"operation has complexity 1009, which exceeds the limit of 10"}],"data":null}`)
		require.False(t, ran)
	})
}
```

Note that the schema for this has two graphql fields that with underscores removed and case normalised to fit go standards would collide:
```
type OverlappingFields {
  # ... snip ...
  newFoo: Int!
  new_foo: Int!
}

```

mbranch:
@vektah This is revised to handle overlapping fields.

DBL-Lee:
What do you think about excluding __typename in complexity calculation by default?
I am encountering this problem because client generated code queries typename by default and any multiplicative complexity functions will cause this to blow up. And custom complexity functions cannot be defined on this field.

mbranch:
> What do you think about excluding __typename in complexity calculation by default?
> I am encountering this problem because client generated code queries typename by default and any multiplicative complexity functions will cause this to blow up. And custom complexity functions cannot be defined on this field.

I'm not a maintainer of this library, but I think `__typename` having a base-cost of `1` is reasonable. Perhaps your complexity calculation should be multiplying larger numbers at the query/mutation level so that adding a handful extra fields doesn't exceed your complexity limit.

codegen/generated!.gotpl
```diff
@@ -84,9 +85,14 @@ func (e *executableSchema) Complexity(typeName, field string, childComplexity in
 	switch typeName + "." + field {
 	{{ range $object := .Objects }}
 		{{ if not $object.IsReserved }}
-			{{ range $field := $object.UniqueFields }}
+			{{ range $_, $fields := $object.UniqueFields }}
+			{{ $len := len $fields }}
+			{{ range $i, $field := $fields }}
+				{{ $ii := add $i 1 }}
```
vektah: You know go can have multiple cases on the same line to avoid this fallthrough? eg

```go
	case "OverlappingFields.oneFoo", "OverlappingFields.twoFoo":
		... code ...
```

Should clean up both the generated code and the templates.

mbranch: Okay. I preferred one-field-one-line, but am happy to revise.

mbranch: @vektah Updated to avoid the `fallthrough`.

Issue #672, opened 2019-04-13
Add Template To templates.Options


marwan-at-work:
The plugin system currently walks the filepath of the directory to where the plugin was originally written, but if you want to publish a plugin as a binary that others download, that binary might not have the associated template files. 

We should allow the templates package to accept a template string to bypass looking up the file location.

mathewbyrne:
I think this change makes sense.  I don't think you could currently publish a plugin as a binary currently though, unless you published a full custom gqlgen binary.

vektah:
Yeah, I think I've given up on the idea that gqlgen could be shipped as a binary, its too closely tied to the local go version. Plugin loading also becomes much tricker, and there isnt much benefit - everyone is going to have a go toolchain installed where they are generating code.

marwan-at-work:
Yes, the go toolchain is still required. But more in the sense that someone could potentially download the plugin binary from Github Releases and not from `go get` and therefore the `.gotpl` files would be missing.

mathewbyrne:
Sure, but what is a plugin binary?  How does gqlgen use this binary as a plugin?  Currently this is not possible, and as @vektah mentioned probably isn't the direction gqlgen is headed.

That said I still think it's a good change, just because it gives plugins a bit more flexibility in how they provide templates.

marwan-at-work:
@mathewbyrne  my current use case is that I have a `protoc` plugin that generates a GraphQL server from a Twirp `service.proto` file. So `gqlgen` does not use the binary as a plugin. But the plugin itself does what `gqlgen generate` does and adds my custom plugins within it.

vektah:
Cool, I think that makes sense. :+1:

PR #673, opened 2019-04-13
codegen/templates: allow templates to be passed in options instead of…

marwan-at-work:
Fixes https://github.com/99designs/gqlgen/issues/672

### DIFF
--- a/codegen/templates/templates.go
+++ b/codegen/templates/templates.go
@@ -19,19 +19,36 @@ import (
 	"github.com/pkg/errors"
 )
 
+// CurrentImports keeps track of all the import declarations that are needed during the execution of a plugin.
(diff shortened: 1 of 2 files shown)


codegen/templates/templates.go
```diff
@@ -24,6 +24,7 @@ var CurrentImports *Imports
 
 type Options struct {
 	PackageName     string
+	Template        string
```
vektah: This could use a comment describing the pretty large change in behaviour.

marwan-at-work: @vektah I noticed a lot of the code is uncommented so I took the liberty to add documentation to what I know about the plugin system so far. Feel free to suggest edits/additions/removals.

vektah:
Sweet!

Issue #674, opened 2019-04-14
Run GraphQL Playground while offline


jonatasbaldin:
Hey there 👋 

Today, `gqlgen` has no offline support for the GraphQL Playground handler, since it uses CND for its files.

I was thinking if we could download the necessary files (css, js etc) and insert it in the template.

I made a *non-working* example of what I want to achieve, but we need a `/static` endpoint to work, which I don't think its the best way. 

Another idea is to read the files and append it to the template, on the string level.

What do you think?

Thanks!

mathewbyrne:
I think my preference would be to _not_ include these files in the repo; the current method is easier to update and means we aren't committing third party javascript into the repo.  If you want to run offline, I would recommend downloading the [desktop version](https://github.com/prisma/graphql-playground) of the Playground.

jonatasbaldin:
Yeah, I agree! Closed the PR and closing this too :)

TalShperling:
There is another solution I came up with.

If you are working on other network/locally you can clone this project:
https://github.com/TalShperling/graphql-playground-offline

and run the CDN locally for your/your environment's needs.

PR #675, opened 2019-04-14
[WIP/Not Working] Add GraphQL Playground static files

jonatasbaldin:
All information at #674.


vektah:
I dont love large binary blobs git like this:
 - They remain in git history and all versions need to be fetched to clone gqlgen.
 - I cant establish trust. Is this even the same binary, you could have tampered with it (not saying you have, just that it leaves the door open for someone to do in the future)?

The second point could be worked around by adding a check in CI that fetches a fresh copy and makes sure its the same the pinned version.

This could also be done as a standalone repo, outside of gqlgen, eg github.com/jonatasbaldin/playground-go. It doesnt even really have any ties to gqlgen itself.

jonatasbaldin:
Yeah, I agree. I had this idea, but wasn't in love with it.

Let's close this. I may try to create my own handler in another package.

Thanks!

PR #680, opened 2019-04-17
Better CollectFields performance

asp24:
CollectFields was increased thanks to following improvements:
 - [x] Pre-allocate mem for method result slice
 - [x] graphql.CollectFields now accept *RequestContext as first arg
 - [x] Early return in shouldIncludeNode if directives empty

Before
`BenchmarkSimpleQueryNoArgs-4   	  100000	     20736 ns/op	    6628 B/op	     120 allocs/op`
After
`BenchmarkSimpleQueryNoArgs-4   	  100000	     19513 ns/op	    6692 B/op	     120 allocs/op`

In practice, difference is much bigger on complex models

### DIFF
--- a/codegen/object.gotpl
+++ b/codegen/object.gotpl
@@ -4,7 +4,7 @@ var {{ $object.Name|lcFirst}}Implementors = {{$object.Implementors}}
 
 {{- if .Stream }}
 func (ec *executionContext) _{{$object.Name}}(ctx context.Context, sel ast.SelectionSet) func() graphql.Marshaler {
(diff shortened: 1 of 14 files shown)


vektah:
Came into this pretty cynical, but it looks pretty straight forward. Nice!

My last round of perf digging - https://github.com/99designs/gqlgen/pull/465.

I think the next big wins are in eliminating as many goroutines as possible during execution, probably by shipping a dataloader solution integrated with gqlgen that doesnt require goroutines and forcing users to be more explicit around which resolvers need to be concurrent. Going to require a bit of planning.

PR #686, opened 2019-04-22
Adds default custom scalar of interface{}

qhenkart:
Given that the Map default custom scalar is automatically provided of map[string]interface{}, it seemed strange to me that interface{} itself has no scalar. These seems like a common (albeit not optimal) enough design pattern in Go that it should be included. At first I thought it might push people not to strictly type their schema, but since Map is supported, I don't see any reason why this shouldn't be.

I didn't add and link an issue because of the simplicity of this PR

I have:
 - [] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
I can add tests if they are deemed necessary, but the other custom scalars don't have any, and this particular scalar is much simpler

### DIFF
--- a/codegen/config/config.go
+++ b/codegen/config/config.go
@@ -363,8 +363,9 @@ func (c *Config) InjectBuiltins(s *ast.Schema) {
 
 	// These are additional types that are injected if defined in the schema as scalars.
 	extraBuiltins := TypeMap{
-		"Time": {Model: StringList{"github.com/99designs/gqlgen/graphql.Time"}},
-		"Map":  {Model: StringList{"github.com/99designs/gqlgen/graphql.Map"}},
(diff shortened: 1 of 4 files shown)


vektah:
I'm on the fence about this, Map and Time are universal, Interface is leaking goisms.

OTOH I can see it might be useful.

Any thoughts @mathewbyrne?

mathewbyrne:
I don't love the name `Interface` in GraphQL since it's a Go-ism, we're thinking maybe `Any` instead?

qhenkart:
@vektah @mathewbyrne I originally went with Untyped (that's what I ended up calling my custom scalar anyway). I wasn't sure what direction you guys would would want to take so I went with Interface for the PR. I will switch it to Any, although I do feel like Untyped is a bit more clear than Any, but it's up to you guys

vektah:
Awesome, thanks for the PR.

PR #687, opened 2019-04-22
Fix: omit deprecated fields when includeDeprecated=false

stereosteve:
Introspection with `includeDeprecated=false` would still return deprecated fields.  In practice this is a small deal because the introspection query typically has `includeDeprecated=true`.

I have:


vektah:
Nice, thanks!

PR #689, opened 2019-04-24
Enforce content type for POST requests

tgwizard:
Currently, the gqlgen HTTP handler doesn't look at the request `Content-Type` header, but just assumes that the request is JSON.

This opens up for a CSRF vulnerability. Servers that accept only `application/json` usually don't need to care about CSRF if they've disabled CORS, since HTML forms on other pages cannot submit data with `application/json` content type. However, currently, a HTML form on another site can submit a `text/plain` request but with a valid JSON body, and the gqlgen server will happily accept it. If the service does auth through cookies, this can be dangerous.

This PR validates the request `Content-Type` header, similarly to how `express-graphql` does it: https://github.com/graphql/express-graphql/blob/1eeaa141b5419834c11872741d9c5d12cfa227c3/src/parseBody.js#L51-L59. `express-graphql` supports `application/x-www-form-urlencoded` and `application/graphql` too though, which we could if we wanted.

This PR shouldn't break well-behaving clients, but it definitely could. We could add a config option `AllowedContentTypes`, where the server could specify which content types it allows. The default (nil/empty slice) could be to support all content types, maintaining backwards compatibility but continuing the slightly weird behaviour, or the default could be to allow "all supported content types", which could mean the `application/x-www-form-urlencoded` and `application/graphql` in the future too.

What do you think? 

I have:


vektah:
> supports application/x-www-form-urlencoded and application/graphql too though, which we could if we wanted.

Hmm, Maybe I've actually broken this in the rebase. https://github.com/99designs/gqlgen/pull/655 adds support for multipart forms. Can I get a sanity check before I cut a release?

> This PR shouldn't break well-behaving clients, but it definitely could. 

Lets make it very clear in the release notes. Between this and the complexity changes we probably should flag this as 0.9, they both have a decent chance to ruin someones day.

tgwizard:
Oh, sorry, I didn't realize this was merged. The changes look good to me after the rebase too.

Issue #693, opened 2019-04-27
panic: interface conversion: types.Type is *types.Basic, not *types.Named


adigunhammedolalekan:
### versions
 - `gqlgen version`? `v0.8.3`
 - `go version`? - `go version go1.12.1 linux/amd64`
 - dep or go modules? - Go modules

#### panic output
```

panic: interface conversion: types.Type is *types.Basic, not *types.Named

goroutine 1 [running]:
github.com/99designs/gqlgen/codegen.(*builder).bindField(0xc0010e18e8, 0xc000915880, 0xc00013d320, 0x0, 0x0)
        /home/adigun/GoApps/pkg/mod/github.com/99designs/gqlgen@v0.8.3/codegen/field.go:111 +0x103d
github.com/99designs/gqlgen/codegen.(*builder).buildField(0xc0010e18e8, 0xc000915880, 0xc00026e4d0, 0x0, 0xe9ca00, 0x0)
        /home/adigun/GoApps/pkg/mod/github.com/99designs/gqlgen@v0.8.3/codegen/field.go:64 +0x31d
github.com/99designs/gqlgen/codegen.(*builder).buildObject(0xc0010e18e8, 0xc0000b4e40, 0x9db536, 0x6, 0x0)
        /home/adigun/GoApps/pkg/mod/github.com/99designs/gqlgen@v0.8.3/codegen/object.go:73 +0x4ec
github.com/99designs/gqlgen/codegen.BuildData(0xc0002120c0, 0xab5c40, 0xe9ba48, 0x0)
        /home/adigun/GoApps/pkg/mod/github.com/99designs/gqlgen@v0.8.3/codegen/data.go:84 +0x461
github.com/99designs/gqlgen/api.Generate(0xc0002120c0, 0x0, 0x0, 0x0, 0x0, 0xa1c290)
        /home/adigun/GoApps/pkg/mod/github.com/99designs/gqlgen@v0.8.3/api/generate.go:37 +0x253
github.com/99designs/gqlgen/cmd.glob..func1(0xc00021a000)
        /home/adigun/GoApps/pkg/mod/github.com/99designs/gqlgen@v0.8.3/cmd/gen.go:39 +0xa2
github.com/urfave/cli.HandleAction(0x91f560, 0xa1b560, 0xc00021a000, 0xc0000724e0, 0x0)
        /home/adigun/GoApps/pkg/mod/github.com/urfave/cli@v1.20.0/app.go:492 +0x7c
[...]

Issue #586, opened 2019-03-06
Defining a schema with recursive types is not possible


janflyborg:
### Expected Behaviour

If a define a schema like this (which is valid GraphQL) and run gqlgen:

```
schema {
    query: Query
}

type Query {
     a: A
}

type A {
    b: B!
}

type B {
    a: A!
}
```

I would expect gqlgen to generate types in models_gen.go that contained pointers to the types A and B (in order to not define a recursive type in Go). 

Instead I get the following output in the 0.7.2 release:

```
/home/janf/go/src/api/models_gen.go:5:6: illegal cycle in declaration of A
/home/janf/go/src/api/models_gen.go:5:6: 	A refers to
/home/janf/go/src/api/models_gen.go:9:6: 	B refers to
/home/janf/go/src/api/models_gen.go:5:6: 	A
validation failed: couldn't load packages due to errors: api

```
### Actual Behavior

The generated model looks like this:

```
// Code generated by github.com/99designs/gqlgen, DO NOT EDIT.

package api

type A struct {
        B B `json:"b"`
}

type B struct {
        A A `json:"a"`
}
```
Which compiles with an error:
`./models_gen.go:9:6: invalid recursive type B`

A temporary solution would be to change the schema to contain a pointer like this.
```
type A {
    b: B # b is now nullable, which breaks the recursive type definition
}
```
But that would change the semantics and since this is a valid way to define schemas in GraphQL it would be a good thing if use cases like this was supported. Maybe with a schema directive?

AFAIK there is no problems with recursions for queries like these since the consumer who is writing the query decides at what depth the recursion should stop (when he selects the attributes to query for).
Since the query is finite, so is the result.

### Minimal graphql.schema and models to reproduce
Please see above

atomicleads:
possible dup of #585

PR #701, opened 2019-05-08
Use pointers when embedding structs in generated structs

### LINKED ISSUE #585: not possible to use schema with cyclic graph?
### Expected Behaviour

Schema should be able to have cycles.

### Actual Behavior

Schema with cycles generates structs with dependencies on other structs that lead to an error.

I've taken the following example from https://swapi.graph.cool/ to illustrate a cycle.

To be able to perform this query:
```
query {
    allPersons {
      name
      birthYear
      species {
        people {
          name
          birthYear
        }
      }
    }
}
```

You would need to have this part of the schema:
```graphql
interface Node {
  id: ID!
}

type Person implements Node {
  name: String!
  birthYear: String
  species: [Species!]
}

type Species implements Node {
  name: String!
  people: [Person!]
}
```

But that leads to a problem with gengql generating cycling dependency between structs.

vektah:
modelgen currently tries to turn
```graphql
type LoopA {
    b: LoopB!
}

 type LoopB {
    a: LoopA!
}
```

into 
```go
type LoopA struct {
	B LoopB
}

 type LoopB struct {
	A LoopA
}
```
Which is invalid. 

This PR adds pointers whenever embedding a struct in a struct to avoid this.

Fixes #585 #586 #693 

I have:

### DIFF
--- a/plugin/modelgen/out/generated.go
+++ b/plugin/modelgen/out/generated.go
@@ -16,34 +16,36 @@ type MissingUnion interface {
 	IsMissingUnion()
 }
 
-type ExistingType struct {
-	Name     *string           `json:"name"`
-	Enum     *ExistingEnum     `json:"enum"`
-	Int      ExistingInterface `json:"int"`
-	Existing *MissingType      `json:"existing"`
-}
-
-func (ExistingType) IsMissingUnion()      {}
-func (ExistingType) IsMissingInterface()  {}
-func (ExistingType) IsExistingInterface() {}
-func (ExistingType) IsExistingUnion()     {}
-
 type MissingInput struct {
 	Name *string      `json:"name"`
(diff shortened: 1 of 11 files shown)


plugin/modelgen/out/generated.go
```diff
@@ -37,7 +37,7 @@ type MissingType struct {
 	Name     *string          `json:"name"`
 	Enum     *MissingEnum     `json:"enum"`
 	Int      MissingInterface `json:"int"`
-	Existing *ExistingType    `json:"existing"`
+	Existing **ExistingType   `json:"existing"`
```
lwc: is this double pointer intentional?

vektah: Good catch!

PR #702, opened 2019-05-08
Drop automatic conversion of IsZero to null

### LINKED ISSUE #684: IsZero method usage is problematic
### What happened?

This PR introduced special behavior when a model has an `IsZero` method: https://github.com/99designs/gqlgen/pull/540 (first released in gqlgen 0.8.0)

When `v.IsZero()` returns `true`, the value is treated as `null`: https://github.com/99designs/gqlgen/blob/3e39b57a98fac8d4db1a85e0b998bd83e29fb9c5/codegen/type.gotpl#L59-L68

This happens even when the field is not nullable, which causes an error.

Our concrete use case is that we use this package to model numbers where precision is important, such as monetary values: https://github.com/shopspring/decimal

`decimal.Decimal` has an `IsZero` method and is, by convention, passed by value. Its zero value definitely should not be treated as null.

### What did you expect?

I expected zero to mean zero, as it did in previous gqlgen versions. Not null.

### versions
 - `gqlgen version`? 0.8.3
 - `go version`? go version go1.11.4 darwin/amd64
 - dep or go modules? dep v0.5.0

vektah:
Moves the conversion of IsZero timestamps into the Time scalar, other custom scalars can return `graphql.Null` to get this behaviour.

If you don't want time to behave this way you can bring your own custom Time implementation, its pretty short.

Fixes #684

I have:
 - [x] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md)) - Existing time tests are decent

### DIFF
--- a/codegen/config/binder.go
+++ b/codegen/config/binder.go
@@ -238,25 +238,6 @@ func (t *TypeReference) IsScalar() bool {
 	return t.Definition.Kind == ast.Scalar
 }
 
-func (t *TypeReference) HasIsZero() bool {
-	it := t.GO
-	if ptr, isPtr := it.(*types.Pointer); isPtr {
-		it = ptr.Elem()
-	}
-	namedType, ok := it.(*types.Named)
-	if !ok {
-		return false
-	}
-
-	for i := 0; i < namedType.NumMethods(); i++ {
-		switch namedType.Method(i).Name() {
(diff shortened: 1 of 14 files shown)


lwc:
LGTM 👌

Issue #715, opened 2019-05-17
`[Post!]!` becomes `[]*Post` not `[]Post` in 0.9.0


ZackKorman:
### What happened?
Generate is turning the following schema:

```
type Post {
        id: Int!
}

type Article {
	posts: [Post!]!
}
```

into:

```
type Post struct {
        ID         int        `json:"id"`
}

type Article {
	 Posts         []*Post        `json:"posts"`
}
```

### What did you expect?

I expected the following generated types, which has been the case with previous versions: 

```
type Post struct {
        ID         int        `json:"id"`
}

type Article {
	 Posts         []Post        `json:"posts"`
}
```

That is, I expected `[Post!]!` to become `[]Post` not `[]*Post`. I cannot tell whether this is intentional, based on #710 and #711, or not. 

### Minimal graphql.schema and models to reproduce

See above.

### versions
 - `gqlgen version`? 0.9.0
 - `go version`? 1.12.0
 - dep or go modules? go  modules

vektah:
This was intentional, see https://github.com/99designs/gqlgen/pull/710

It makes gqlgen more consistent with other codegen tools like prisma and protobuffer and generally reduces boilerplate.

See also https://github.com/99designs/gqlgen/issues/375#issuecomment-480779700

ZackKorman:
Ahh, I didn't see #375. Alright, thanks.

gissleh:
I've discussed this in the Gitter and plan to make a PR about making this specific instance of pointer-slices being used instead of struct-value-slices a configuration choice.

mtibben:
> a configuration choice

Seems this would add considerable complexity, without much upside. What's the rationale?

gissleh:
I've used gqlgen since 0.5, and while the refactor for 0.8 was mostly done by search-replace and about 1 block of boilerplate per model, refactoring only the resolvers for this one would mean a lot of loops to convert between. I can understand the rationale for pointers to single structs due to compilation errors, but struct-of-slice-pointers seems to be a choice mentioned in a single comment and then accepted.

I'll post the PR and let you review it. If the added complexity isn't worth it, feel free to reject it.

Daavidaviid:
Yeah me too, seems kind of complicated to solve this.

Tomarron:
this break the spec of graphql, which define the syntax of "!", what the plan  or what this tool will be.

mathewbyrne:
[...]

PR #719, opened 2019-05-19
Made choice of whether to use slice-of-struct-pointers vs slice-of-st…

gissleh:
Related to: #715

Version 0.9 broke backwards compatibility with 0.8 in a way that required major refactoring by changing resolver arguments, generated struct fields and resolver return values that are slices of structs to slices of struct pointers. This PR adds a configuration option to use the old 0.8 behavior.

I have tested whether the configuration has its intended effects. If there is a more elegant way to do this or a more proper way to test it, please let me know.

I understand that this adds considerable complexity in that a bug might occur for one choice of this configuration option but not the other, so I completely understand if you reject this because you do not want to support it.

I have:


gissleh:
I took the time to update my resolvers to use the new format by copying into pointer arrays. I don't like the added bloat, but it wasn't as great a refactor workload as I expected.

I won't really need this anymore then, but I'm willing to put in the work if you're up for merging this after a few changes.

mathewbyrne:
Thanks for your work, but we don't currently believe that making this configurable is something we want to maintain.  This change has been to bring slice returns from resolvers in line with regular structs, which were changed to always be pointers in 0.8.

Good to hear that the refactor wasn't too troublesome.  We hear you that copying into pointer arrays isn't ideal, but also remember that users have been doing the opposite for a long time now, as many services/storage layers return slices of arrays.

gissleh:
Fair enough. I'll close this PR then.

Issue #721, opened 2019-05-22
Panic when model is a type definition of another package


marwan-at-work:
### What happened?
if `gqlgen.yml` points to a custom type, and that custom type is defined as follows: 

```
type Custom somepkg.Custom
```

Then the `gqlgen` command panics with the following error: 

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xb0 pc=0x12f29f8]

goroutine 1 [running]:
github.com/99designs/gqlgen/codegen/config.(*Binder).ObjectPosition(0xc0011b1980, 0x16b5920, 0xc0053c57c0, 0x0, 0x0, 0x0, 0x0, 0x0)
        /Users/me/gqlgen/codegen/config/binder.go:63 +0x1f8
github.com/99designs/gqlgen/codegen.(*builder).bindField(0xc00895f8c8, 0xc001b38a80, 0xc00cc493b0, 0x0, 0x0)
        /Users/me/gqlgen/codegen/field.go:116 +0x3b8
github.com/99designs/gqlgen/codegen.(*builder).buildField(0xc00895f8c8, 0xc001b38a80, 0xc0001efea0, 0x0, 0x1a86220, 0x0)
        /Users/me/gqlgen/codegen/field.go:64 +0x31d
github.com/99designs/gqlgen/codegen.(*builder).buildObject(0xc00895f8c8, 0xc0000cad80, 0x15c8f85, 0x6, 0x0)
        /Users/me/gqlgen/codegen/object.go:73 +0x4d4
github.com/99designs/gqlgen/codegen.BuildData(0xc0000ca240, 0x16a17a0, 0x1a854e0, 0x0)
        /Users/me/gqlgen/codegen/data.go:84 +0x461
github.com/99designs/gqlgen/api.Generate(0xc0000ca240, 0x0, 0x0, 0x0, 0x0, 0x16094e8)
        /Users/me/gqlgen/api/generate.go:37 +0x243
github.com/99designs/gqlgen/cmd.glob..func1(0xc0001e0000)
        /Users/me/gqlgen/cmd/gen.go:39 +0xa2
github.com/urfave/cli.HandleAction(0x150d860, 0x16086a0, 0xc0001e0000, 0xc000032300, 0x0)
        /Users/me/pkg/mod/github.com/urfave/cli@v1.20.0/app.go:492 +0x7c
[...]

PR #722, opened 2019-05-22
resolve all pkg dependencies

marwan-at-work:
Fixes https://github.com/99designs/gqlgen/issues/721


mathewbyrne:
Thanks for the fix. Would you mind please adding a test case that demonstrates the issue?

marwan-at-work:
@mathewbyrne no problem. I'm not too familiar with the codebase. I'd be happy to add a test case if I knew where/how I can define Go code with type declarations and somehow target this function in my int tests only. Thanks :)

PR #728, opened 2019-05-26
resolvergen: use the resolver type as base name for dependent types

fgallina:
The template was outputing invalid code since the resolver type was
not used in places like the embedding at {query,mutation}Resolver.

This change also ensures that objects like {query,mutation}Resolver
also use the user provided type name as suffix.

Here's the resulting diff on the code generation with `type:
GeneratedResolver` in the resolver config:

```
diff -u resolver.go resolvernew.go
--- resolver.go 2019-05-26 20:04:15.361969755 -0300
+++ resolvernew.go      2019-05-26 20:04:54.170737786 -0300
@@ -7,20 +7,20 @@
 type GeneratedResolver struct{}

 func (r *GeneratedResolver) Mutation() MutationResolver {
-       return &mutationResolver{r}
+       return &mutationGeneratedResolver{r}
 }
 func (r *GeneratedResolver) Query() QueryResolver {
-       return &queryResolver{r}
+       return &queryGeneratedResolver{r}
 }

-type mutationResolver struct{ *Resolver }
+type mutationGeneratedResolver struct{ *GeneratedResolver }

-func (r *mutationResolver) CreateTodo(ctx context.Context, input NewTodo) (*Todo, error) {
+func (r *mutationGeneratedResolver) CreateTodo(ctx context.Context, input NewTodo) (*Todo, error) {
        panic("not implemented")
 }

-type queryResolver struct{ *Resolver }
+type queryGeneratedResolver struct{ *GeneratedResolver }

-func (r *queryResolver) Todos(ctx context.Context) ([]*Todo, error) {
+func (r *queryGeneratedResolver) Todos(ctx context.Context) ([]*Todo, error) {
        panic("not implemented")
 }
```

I haven't yet found the place where this is tested nor the sensible way to add test for this use case. Will update as soon as I find it.

I have:

### DIFF
--- a/plugin/resolvergen/resolver.go
+++ b/plugin/resolvergen/resolver.go
@@ -19,6 +19,7 @@ type Plugin struct{}
 var _ plugin.CodeGenerator = &Plugin{}
 
 func (m *Plugin) Name() string {
+	// TODO: typo, should be resolvergen
 	return "resovlergen"
 }
 func (m *Plugin) GenerateCode(data *codegen.Data) error {
--- a/plugin/resolvergen/resolver_test.go
+++ b/plugin/resolvergen/resolver_test.go
@@ -1,0 +1,163 @@
+package resolvergen
+
+import (
+	"fmt"
+	"go/types"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"unicode"
+
+	"github.com/99designs/gqlgen/codegen"
+	"github.com/99designs/gqlgen/codegen/config"
+	"github.com/vektah/gqlparser/ast"
+)
+
+func TestPlugin_Name(t *testing.T) {
+	t.Run("test plugin name", func(t *testing.T) {
+		m := &Plugin{}
+		if got, want := m.Name(), "resovlergen"; got != want {
+			t.Errorf("Plugin.Name() = %v, want %v", got, want)
+		}
+	})
+}
+
+// Types for testing code generation, both Mutation and
+// MutationResolver must implement types.Type.
+type Mutation struct{}
+
+func (m *Mutation) Underlying() types.Type {
+	return m
+}
+
+func (m *Mutation) String() string {
+	return "Mutation"
+}
+
+type MutationResolver struct{}
+
+func (m *MutationResolver) Underlying() types.Type {
+	return m
+}
+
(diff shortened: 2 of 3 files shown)


mathewbyrne:
Can you please add a test case demonstrating the issue?  It looks like none of the generated code has changed from to your update.

fgallina:
@mathewbyrne I think that may be related to the fact that the `resolver.go` generation is skipped if the file already exists. Also there must be a custom resolver type in the `gqlgen.yml` to exercise these changes. Excerpt from my original config that drove me to propose this change:

```
...
resolver:
  filename: graphql/resolver.go
  package: graphql
  type: GeneratedResolver
```

plugin/resolvergen/resolver_test.go
```diff
@@ -0,0 +1,163 @@
...
+		return &codegen.Data{
+			Config: &config.Config{
+				Resolver: cfg,
+			},
+			Objects: codegen.Objects{obj},
+		}
+	}
+
+	t.Run("renders expected contents", func(t *testing.T) {
+		m := &Plugin{}
```
seriousben: Maybe an easier and simpler way to write this test would be to use golden files (https://medium.com/@jarifibrahim/golden-files-why-you-should-use-them-47087ec994bf)

plugin/resolvergen/resolver.go
```diff
@@ -19,6 +19,7 @@ type Plugin struct{}
 var _ plugin.CodeGenerator = &Plugin{}
 
 func (m *Plugin) Name() string {
+	// TODO: typo, should be resolvergen
```
vektah: :laughing:

plugin/resolvergen/resolver_test.go
vektah: We already do this using code review of the committed generated files. No need for update, just commit the generated code and push.

All this test should need to do is call the generator with some config, [eg](https://github.com/99designs/gqlgen/blob/master/plugin/modelgen/models_test.go). It may not look like its asserting anything, but really its output is tested in three ways:
 - firstly we run goimports when saving go files, so running generate should error out in the test if it generates invalid code.
 - if that fails the linter should see invalid code, and fail the branch.
 - code review catches changes. diffs are readable in github.

Issue #587, opened 2019-03-06
allow for custom implementations of the @deprecated directive


gracenoah:
### Expected Behaviour

I expect to be able to write a directive middleware for @deprecated that logs a warning every time a deprecated field is accessed.

### Actual Behavior

There's nothing in `DirectiveRoot` for the @deprecated directive.

### Minimal graphql.schema and models to reproduce
```
{
  query: String! @deprecated(reason: "no reason")
}
```

vektah:
related to #584

mathewbyrne:
Yeah, this is the use-case I was discussing with @vektah yesterday.  It's a case of not marking deprecated as a builtin directive, which will output it in the `DirectiveRoot`.  I wonder if this is something we could expose through config?

swkumar:
Can builtin directives now be accessed through `DirectiveRoot`?

thumphries:
Apologies for commenting on a closed issue; just leaving a note for those finding this issue in the future...

To get `Deprecated` into your `DirectiveRoot` for customisation I had to first set `skip_runtime` to false, like so:

```yaml
directives:
  deprecated:
    skip_runtime: false
```

(as of gqlgen v0.11.3)

grzn:
with v0.17.55, it doesn't work.

it adds generated code do call the directive: 
```
		directive1 := func(ctx context.Context) (interface{}, error) {
			reason, err := ec.unmarshalOString2ᚖstring(ctx, "Use tagsV2 instead")
			if err != nil {
				var zeroVal []string
				return zeroVal, err
			}
			if ec.directives.Deprecated == nil {
				var zeroVal []string
				return zeroVal, errors.New("directive deprecated is not implemented")
			}
			return ec.directives.Deprecated(ctx, obj, directive0, reason)
		}
```

but it doesn't add `Deprecated` to the DirectiveRoot, so the code doesn't compile.

Issue #726, opened 2019-05-24
New Proposal: Mark field as resolver with graphql directives


sunfmin:
Right now you can either use custom defined models compare with schema.graphql definition to generate resolvers. 

Or you can use gqlgen.yml to config resolvers like

```yaml
models:
  Topic:
    fields:
      replies:
        resolver: true
```

I suggest another way to mark a field as resolver by using https://graphql.github.io/graphql-spec/draft/#sec-Type-System.Directives

For example:

```graphql
directive @resolver on FIELD_DEFINITION

type Topic {
    id: ID!
    title: String!
    replies: [Reply!] @resolver
}
```

This way, Field definition and whether it is a resolver are in the same place. So you won't forget to remove one or another when changing

vektah:
Yeah there are a bunch of these, `@model`, `@resolver`, `@name` could be interesting

lwc:
Done!

PR #732, opened 2019-05-30
Add a plugin for configuring gqlgen via directives

lwc:
Uses the plugin system / `MutateConfig` interface to allow for gqlgen configuration directly in the schema files, via two directives:

```graphql
directive @goModel(model: String, models: [String!]) on OBJECT 
    | INPUT_OBJECT 
    | SCALAR 
    | ENUM 
    | INTERFACE 
    | UNION

directive @goField(forceResolver: Boolean, name: String) on INPUT_FIELD_DEFINITION 
    | FIELD_DEFINITION
```
I think this captures all the features available in gqlgen.yml.

### Benefits
- Config lives alongside the schema
- Allows config to be split across multiple schema files

### Questions
- Naming is hard - originally I'd called these `bind` and `field` but was concerned they were too vague. Thoughts?

Supersedes: https://github.com/99designs/gqlgen/pull/727 
Fixes: https://github.com/99designs/gqlgen/issues/587, https://github.com/99designs/gqlgen/issues/726

I have:

I'll update docs if we decide to go ahead with this approach


sunfmin:
make these directive builtin? Or better not?

Issue #401, opened 2018-10-28
Add support for Automatic Persisted Queries


icco:
This is more of a feature request than a bug, and I'd be willing to do the work if it's something people are interested in.

Apollo added this neat feature which allows for dynamic query caching.

 - https://blog.apollographql.com/improve-graphql-performance-with-automatic-persisted-queries-c31d27b8e6ea
 - https://github.com/apollographql/apollo-link-persisted-queries

This would require an addition to the parsing logic so that a request without a query would still work. For example:

```
{
  operationName: 'MyQuery',
  variables: null,
  extensions: {
    persistedQuery: {
      version: 1,
      sha256Hash: hashOfQuery
    }
  }
}
```

currently fails at https://github.com/99designs/gqlgen/blob/1b17b5a2489094911e701a299aedee3d1d1a2319/handler/graphql.go#L287,

Instead this would change to not send a 422, but rather a 200 with `    { message: 'PersistedQueryNotFound' }
`

vektah:
It's something we've talked a bit about internally and a few others have expressed interest. It needs a little thought and care around the interfaces for saving and fetching queries, as these will need to actually be persisted.

icco:
I was imagining a first pass would just be to augment the LRU being used to also store a hash. Not great for storing lots of queries, but would probably work well without drastically changing the performance profile of the app.

vektah:
My current understanding of APQ is roughly like this:
 - In CI you enumerate all queries, and register the query with a central database, which returns the hash
 - Store the hash against the query locally somewhere, committed in your client repo
 - at query send the hash instead of the query
 - all running servers need to be able to lookup the registered query from a central data source

Without some kind of persistent storage, shared by all the servers in the pool, this wont work.

icco:
That's how relay works, APQ though is different.

![apq](https://cdn-images-1.medium.com/max/1600/1*7yEdVjVXcXcEeMEJF6CQGw.png)

Basically Apollo first hashes the query, and sends the operation name and a hash. If the server knows that hash + operation name combo, it returns the results, otherwise it returns a unique error that says either:

 1) I don't support persisted queries
 2) I haven't stored that hash yet

if #1, the client stops sending hashes. if #2, the client sends the full query, with the hash it expects it to be stored at. There isn't any shared storage between the client and the server.

vektah:
Nice pic :+1:

[...]

PR #734, opened 2019-06-01
Automatic Persisted Queries

DBL-Lee:
This PR implements automatic persisted queries as described [here](https://blog.apollographql.com/improve-graphql-performance-with-automatic-persisted-queries-c31d27b8e6ea)

It follows apollo server implementation [here](https://github.com/apollographql/apollo-server/blob/master/packages/apollo-server-core/src/requestPipeline.ts) to make sure it is compatible with apollo clients.

This have been running in our production environment for a few weeks and works well with Apollo client with APQ enabled.

This implements #401 

I have:

### DIFF
--- a/docs/content/reference/apq.md
+++ b/docs/content/reference/apq.md
@@ -1,0 +1,77 @@
+---
+title: "Automatic persisted queries"
+description:   
+linkTitle: "APQ"
+menu: { main: { parent: 'reference' } }
+---
+
+When you work with GraphQL by default your queries are transferred with every request. That can waste significant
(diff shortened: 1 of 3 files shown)


radev:
See https://github.com/DBL-Lee/gqlgen/pull/1 for proposed implementation

RxAssim:
this looks solid 👍

vektah:
Thanks for the PR. I gave it a test run on one of our projects, docs are good, everything seems to work.

The handler package is becoming a bit big, we will have to do something to make it a bit more modular in the future.

PR #742, opened 2019-06-12
Add middleware to intercept sendError

tgwizard:
The reason for this PR is so that the server can easily log errors whenever clients send invalid queries (e.g. not valid GraphQL query, or doesn't conform to the schema). Currently this is short-circuited in the handler, with no easy way to access those errors.

Services can of course send in a wrapper of the response writer to the gqlgen handler, and in that wrapper intercept response writes, and log them if the status code is a 4xx. That's more general, but it seems more complicated that necessary (especially if you want to handle the Hijacker interface for websocket connections).

Therefore, I think this PR is a good start. I'm not at all attached to the names I've come up with, feedback much appreciated.

I have:

### DIFF
--- a/docs/content/reference/errors.md
+++ b/docs/content/reference/errors.md
@@ -106,3 +106,22 @@ server := handler.GraphQL(MakeExecutableSchema(resolvers),
 }
 ```
 
+
+### The send error middleware
+
+To get full control over all errors sent to the client, including when the client-sent GraphQL query is invalid, you can
+pass in a send error handler.
+
+Example:
+
+```go
+server := handler.GraphQL(MakeExecutableSchema(resolvers),
+	handler.SendErrorMiddleware(
+		func(ctx context.Context, errors []*gqlerror.Error) []*gqlerror.Error {
+			fmt.Printf("%v\n", errors)
+			return errors
+		}
+	),
+)
+```
+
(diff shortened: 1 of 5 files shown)


tgwizard:
To see how a wrapper of a response writer could accomplish logging of 4xx responses, see https://github.com/Shopify/goose/pull/37. Perhaps that is the cleaner approach.

marwan-at-work:
@tgwizard out of curiosity, can't you just use a `ResolverMiddleware` like this: 

```golang
func middleware(ctx, next graphql.Resolver) (interface{}, error) {
  resp, err := next(ctx)
  if err != nil {
    log.Println(err)
  }
  return resp, err
}
```

tgwizard:
@marwan-at-work unfortunately not. These errors that I want to intercept and log - invalid graphql query, either syntactically or semantically for the server's schema - are emitted before we have a request context and before the `RequestMiddleware` or `ResolverMiddleware` are called. See https://github.com/99designs/gqlgen/blob/694f90aa089c34a80bf3007a6d298a96ba7f132c/handler/graphql.go#L422-L441.

Still, I think this PR's `SendErrorMiddleware` is a bit weird, and I think something like https://github.com/Shopify/goose/pull/37 is more appropriate. Perhaps there's a way the gqlgen http processing can have a nice generic error middleware, but I think this PR isn't quite it.

marwan-at-work:
@tgwizard that makes sense, thank you for the helpful answer :v:

PR #744, opened 2019-06-14
 add Execute QUERY/MUTATION/SUBSCRIPTION Directives

andrey1s:
add run Executable directives by location QUERY, MUTATION and SUBSCRIPTION


codegen/generated!.gotpl
```diff
@@ -196,6 +208,126 @@ type executionContext struct {
 	*executableSchema
 }
 
+{{ if and .QueryDirectives .QueryRoot }}
```
vektah: Probably need to dry this out. I suggest creating a new template called `directives.gotpl` and feed it data from a new function in directives.go. Try to paramatize out the parts that would be hard to do in the template.

andrey1s: moved the template to a separate file

vektah:
:tada: Thanks!

Issue #747, opened 2019-06-15
The resolvergen and servergen plugins do not fully support a configured resolver type.


seriousben:
### What happened?

Generated code does not compile.

### What did you expect?

Generated code to compile.

### Minimal graphql.schema and models to reproduce

gqlgen.yml:

```
schema:
- schema.graphql
exec:
  filename: generated.go
model:
  filename: models_gen.go
resolver:
  filename: resolver.go
  type: ResolverRoot
```

PR #748, opened 2019-06-15
Fix: support configured resolver type in server and resolver plugins

seriousben:
Fixes: https://github.com/99designs/gqlgen/issues/747

Generated code now compiles when specifying a custom Resolver type.


seriousben:
Kind of a duplicate of https://github.com/99designs/gqlgen/pull/728

vektah:
Going to close in favor of #728 which getting tests sorted out. Thanks for the PR though.

PR #750, opened 2019-06-18
[websocket] Add a config to reject initial connection

eddeee888:
### Problem

Currently, we are allowing clients to always connect to the server unless the message is invalid. Current implementation [here](https://github.com/99designs/gqlgen/blob/694f90aa089c34a80bf3007a6d298a96ba7f132c/handler/websocket.go#L88-L97)

This is not safe as a malicious user can connect to the WebSocket and perform queries/mutations. An example of such message:

Screen Shot 2019-06-18 at 11 05 15 am

We would have to check in every query/mutation/subscription to make sure that a user can perform such action (related PR: https://github.com/99designs/gqlgen/pull/348). 

### Proposal

This PR adds a new function to handler config called `websocketOnInitFunc`. This can be used by the server to check the init payload to decide whether to establish the connection or reject it. If this function is not given, the connection is always established.

### Example workflow

This is an example workflow, different apps may have different approaches:

1. The client sends a WebSocket upgrade request with init payload to the server and a token in cookie
2. The server receives the request, picks out the cookie and put it in context
3. When the init function is called, it will use the init payload and compare with the cookie in context
4a. If they are the same, create the connection
4b. If they are not the same, close the connection and send back a `invalid init payload` message

Accepted connection:
Screen Shot 2019-06-18 at 10 53 02 am

Rejected connection:
Screen Shot 2019-06-18 at 10 53 18 am

I have:
 - [x] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content)) - None

### DIFF
--- a/handler/graphql.go
+++ b/handler/graphql.go
@@ -30,7 +30,7 @@ type params struct {
 	Variables     map[string]interface{} `json:"variables"`
 }
 
-type websocketInitFunc func(ctx context.Context, initPayload InitPayload) bool
+type websocketInitFunc func(ctx context.Context, initPayload InitPayload) error
 
 type Config struct {
 	cacheSize                       int
@@ -255,7 +255,7 @@ func (tw *tracerWrapper) EndOperationExecution(ctx context.Context) {
 
 // WebsocketInitFunc is called when the server receives connection init message from the client.
 // This can be used to check initial payload to see whether to accept the websocket connection.
-func WebsocketInitFunc(websocketInitFunc func(ctx context.Context, initPayload InitPayload) bool) Option {
+func WebsocketInitFunc(websocketInitFunc func(ctx context.Context, initPayload InitPayload) error) Option {
 	return func(cfg *Config) {
 		cfg.websocketInitFunc = websocketInitFunc
 	}
--- a/handler/graphql.go
+++ b/handler/graphql.go
@@ -40,6 +40,7 @@ type Config struct {
 	tracer                          graphql.Tracer
 	complexityLimit                 int
 	complexityLimitFunc             graphql.ComplexityLimitFunc
+	websocketOnInitFunc             func(ctx context.Context, initPayload InitPayload) bool
 	disableIntrospection            bool
 	connectionKeepAlivePingInterval time.Duration
 	uploadMaxMemory                 int64
@@ -248,6 +249,14 @@ func (tw *tracerWrapper) EndFieldExecution(ctx context.Context) {
 func (tw *tracerWrapper) EndOperationExecution(ctx context.Context) {
 	tw.tracer2.EndOperationExecution(ctx)
 	tw.tracer1.EndOperationExecution(ctx)
+}
+
+// WebsocketOnInitFunc is called when the server receives connection init message from the client.
+// This can be used to check initial payload to see whether to accept the websocket connection.
+func WebsocketOnInitFunc(websocketOnInitFunc func(ctx context.Context, initPayload InitPayload) bool) Option {
+	return func(cfg *Config) {
+		cfg.websocketOnInitFunc = websocketOnInitFunc
+	}
 }
 
 // CacheSize sets the maximum size of the query cache.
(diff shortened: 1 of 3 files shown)


handler/websocket.go
```diff
@@ -12,7 +12,7 @@ import (
 
 	"github.com/99designs/gqlgen/graphql"
 	"github.com/gorilla/websocket"
-	"github.com/hashicorp/golang-lru"
+	lru "github.com/hashicorp/golang-lru"
```
Sonna: 🤔

handler/graphql.go
```diff
@@ -250,6 +251,14 @@ func (tw *tracerWrapper) EndOperationExecution(ctx context.Context) {
 	tw.tracer1.EndOperationExecution(ctx)
 }
 
+// WebsocketOnInitFunc is called when the server receives connection init message from the client.
+// This can be used to check initial payload to see whether to accept the websocket connection.
+func WebsocketOnInitFunc(websocketOnInitFunc func(ctx context.Context, initPayload InitPayload) bool) Option {
```
Sonna: We will not support custom errors, and only return true or false on initialise failure?
```suggestion
func WebsocketOnInitFunc(websocketOnInitFunc func(ctx context.Context, initPayload InitPayload) (bool, error)) Option {
```

Sonna: _hmm ..._ I guess it depends on `func (c *wsConnection) init() bool` and whether or not you send the custom error message via `c.sendConnectionError`; like with [`c.sendConnectionError("invalid init payload")`](https://github.com/99designs/gqlgen/pull/750/files#diff-857d352c30d3f33e4fed201e953d953cR100)

handler/graphql.go
```diff
@@ -40,6 +40,7 @@ type Config struct {
 	tracer                          graphql.Tracer
 	complexityLimit                 int
 	complexityLimitFunc             graphql.ComplexityLimitFunc
+	websocketOnInitFunc             func(ctx context.Context, initPayload InitPayload) bool
```
mathewbyrne: Maybe type this func, and consider slight name change:
```suggestion
	websocketInitFunc             WebsocketInitFunc
	
	// ...
	
	type WebsocketInitFunc func(ctx context.Context, initPayload InitPayload) bool
```

handler/websocket.go
```diff
@@ -94,6 +94,15 @@ func (c *wsConnection) init() bool {
 			}
 		}
 
+		if c.cfg.websocketOnInitFunc != nil {
+			isValidPayload := c.cfg.websocketOnInitFunc(c.ctx, c.initPayload)
+			if !isValidPayload {
```
mathewbyrne: If you don't need error handling, it's a bit more idiomatic to write:
```suggestion
			if ok := c.cfg.websocketOnInitFunc(c.ctx, c.initPayload); !ok {
```

handler/websocket.go
eddeee888: My IDE linter does this automatically 😬

handler/graphql.go
eddeee888: I didn't want to give this too many options. I don't think the server needs to tell the client why it fails either. If there's a need for it we can create another PR to follow up

vektah: Maybe just `func(ctx context.Context, initPayload InitPayload) error`?

handler/graphql.go
```diff
@@ -30,6 +30,8 @@ type params struct {
 	Variables     map[string]interface{} `json:"variables"`
 }
 
+type websocketInitFunc func(ctx context.Context, initPayload InitPayload) error
```
eddeee888: Updated the interface to return `error` instead of `bool`

handler/websocket.go
```diff
@@ -94,6 +94,14 @@ func (c *wsConnection) init() bool {
 			}
 		}
 
+		if c.cfg.websocketInitFunc != nil {
+			if err := c.cfg.websocketInitFunc(c.ctx, c.initPayload); err != nil {
+				c.sendConnectionError(err.Error())
```
eddeee888: The server now has the option to return an error. The error body is then sent to the client

PR #751, opened 2019-06-20
Allow disabling generation

tt:
I'm using this package for a server implementation and would like to also use it to generate typed models for a client library.

I can ignore the generated.go and either remove it or avoid checking it in but with this small change, I can have gqlgen output only the file I care about.

I realize this is irrelevant "for building GraphQL servers" so I won't mind if you decide to close the pull request.

I have:
 - [ ] ~Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))~

(I'm not sure this option should be documented so let me know if you think so.)

### DIFF
--- a/codegen/config/config.go
+++ b/codegen/config/config.go
@@ -182,7 +182,7 @@ func (c *PackageConfig) Check() error {
 	if strings.ContainsAny(c.Package, "./\\") {
 		return fmt.Errorf("package should be the output package name only, do not include the output filename")
 	}
-	if c.Filename != "" && !strings.HasSuffix(c.Filename, ".go") {
+	if c.Filename != "" && c.Filename != os.DevNull && !strings.HasSuffix(c.Filename, ".go") {
 		return fmt.Errorf("filename should be path to a go source file")
 	}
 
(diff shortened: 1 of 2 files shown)


vektah:
You should already be able to do this by calling the modelgen plugin directly:
```

func main() {
	cfg, err := config.LoadConfigFromDefaultLocations()
	if os.IsNotExist(errors.Cause(err)) {
		cfg = config.DefaultConfig()
	} else if err != nil {
		fmt.Fprintln(os.Stderr, err.Error())
		os.Exit(2)
	}
	
	err = modelgen.New().(*modelgen.Plugin).MutateConfig(cfg)
	if err != nil {
		panic(err)
	}
}
```

tt:
The intent was to be able to invoke `go run github.com/99designs/gqlgen` and not have to add a second step to delete the generated.go file.

If we have to maintain a tool to generate these models, we're probably better off deleting the file.

vektah:
Its pretty low maintenance for you to plonk down a 13 line go file next to wherever your running go generate, and update the stanza to be

//go:generate go run generatemodels.go

tt:
It's more maintenance than `rm generated.go` (especially as we need to generate models in separate repositories for at least two different APIs) assuming the command line interface is as or more stable than the programmatic one.

I'm also not entirely sure how I would structure the repository to prevent gqlgen from becoming a dependency of the client package. (It's of course great for hermetic builds if it is but not for consumers importing this package only to get the already generated client.)

Thank you for considering it and for the proposal.

Issue #757, opened 2019-06-25
Dataloader code exemple may be wrong


franxois:
### What happened?

It seems there is a mistake in the dataloader code exemple

### What did you expect?

I expect to see IDs given as input used in sql code generation

PR #758, opened 2019-06-25
Update dataloaders.md

franxois:
I may have seen a mistake in documentation with dataloader. I update code to make SQL request use requested IDs.

I use a map to store result by IDs, then I recreate output array to return values

fix https://github.com/99designs/gqlgen/issues/757


vektah:
:+1: this probably serves a better base to start working from.

Issue #760, opened 2019-06-26
EPIC: Directive support


vektah:
Here to track the status of directives

:warning: these features are subject to change - if you are using any of these features make sure you have tests on your directives that actually run queries against a graph. :warning: 

Execution directives
====
location | implemented | obj is
-----|----- | ---
QUERY | :heavy_check_mark: https://github.com/99designs/gqlgen/pull/744 | *ast.OperationDefinition 
MUTATION | :heavy_check_mark: https://github.com/99designs/gqlgen/pull/744 | *ast.OperationDefinition
SUBSCRIPTION | :heavy_check_mark: https://github.com/99designs/gqlgen/pull/744 | *ast.OperationDefinition
FIELD | :heavy_check_mark: https://github.com/99designs/gqlgen/pull/756 | parent type
FRAGMENT_DEFINITION | :no_entry: 
FRAGMENT_SPREAD | :no_entry: 
INLINE_FRAGMENT | :no_entry: 

Schema directives
===
location | implemented | obj is 
---|--- | --
SCHEMA | :no_entry:  
SCALAR | :no_entry: 
OBJECT | :no_entry: 
FIELD_DEFINITION | :heavy_check_mark: | parent type 
ARGUMENT_DEFINITION | :heavy_check_mark: https://github.com/99designs/gqlgen/pull/460 | argument bag
INTERFACE | :no_entry: 
UNION | :no_entry: 
ENUM | :no_entry: 
ENUM_VALUE | :no_entry: 
INPUT_OBJECT | :no_entry: 
INPUT_FIELD_DEFINITION | :heavy_check_mark:  | parent input

dbadura:
Hello,
what happen with `ARGUMENT_DEFINITION` directive?

dtrckd:
Hi,
I've commented on a potential problem with OBJECT directive, [here](https://github.com/99designs/gqlgen/issues/1120#issuecomment-650748143).

logrusorgru:
Hi, there.

There are a few questions about directives. Commonly most about INPUT_FIELD_DEFINITION. It is marked with a check mark and its input object is parent object.

1. The gqlgen generator doesn't allow directives with INPUT_FIELD_DEFINITION on input fields. But allows them on the _input fields_ with FIELD_DEFINITION.
2. Object received in directive handler is map[string]interface{}, not Go object. Is it by design? Why not Go object?
3. The following directive produce invalid server that panics just before the directive called
```graphqls
[...]

PR #776, opened 2019-07-05
updates directive root structure

qhenkart:
The documentation was incorrect in the way directives are generated, missing the body interface{}. I originally thought this was a bug until I checked what the object was and figured that should be in the documentation too

I have:
 - [N/A ] Added tests covering the bug / feature (see [testing] (https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [x ] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

### DIFF
--- a/docs/content/reference/directives.md
+++ b/docs/content/reference/directives.md
@@ -22,12 +22,16 @@ enum Role {
 }
 ```
 
-When we next run go generate, gqlgen will add this directive to the DirectiveRoot 
+When we next run go generate, gqlgen will add this directive to the DirectiveRoot
 ```go
 type DirectiveRoot struct {
-	HasRole func(ctx context.Context, next graphql.Resolver, role Role) (res interface{}, err error)
+	HasRole func(ctx context.Context, obj interface{}, next graphql.Resolver, role Role) (res interface{}, err error)
 }
 ```
+*ctx*: passed from the resolver
+*obj*: the payload aggregated so far
+*next*: signals that the directive was successful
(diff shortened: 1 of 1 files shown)


docs/content/reference/directives.md
```diff
@@ -22,12 +22,16 @@ enum Role {
...
-When we next run go generate, gqlgen will add this directive to the DirectiveRoot 
+When we next run go generate, gqlgen will add this directive to the DirectiveRoot
 ```go
 type DirectiveRoot struct {
-	HasRole func(ctx context.Context, next graphql.Resolver, role Role) (res interface{}, err error)
+	HasRole func(ctx context.Context, obj interface{}, next graphql.Resolver, role Role) (res interface{}, err error)
 }
 ```
+*ctx*: passed from the resolver
+*obj*: the payload aggregated so far
```
vektah: This can actually vary a bit, depending on where the directive was - https://github.com/99designs/gqlgen/issues/760

docs/content/reference/directives.md
```diff
@@ -22,12 +22,16 @@ enum Role {
...
+*next*: signals that the directive was successful
```
vektah: `next` is the middleware pattern, it's not signaling success it's fetching the value by going through the rest of the stack.

vektah:
Shipped with some tweaks:
https://github.com/99designs/gqlgen/commit/139e4e8d444b5c6fb27406ad610ca4fce897710a

PR #780, opened 2019-07-07
fixed generating a description to golang comments for enum type

zdebra:
Hello, I've created a bug fix for plugin modelgen. I and my team have recently started using the tool in production and we really like it. 

I wanted to dig into the code by submitting an easy fix of something that I found out. When models are generated from GraphQL schema, a description is used as golang comments. It works fine for all GraphQL types but enum. It worked in a way that the description was formatted into a single word CamelCased. I've added test cases and fixed it by removing calling `go` in a template pipe. 

Those tests from me could be more sophisticated as I am checking only for the number of words in a comment.

Let me know if this is welcomed. I would like to contribute more.

### DIFF
--- a/plugin/modelgen/models.gotpl
+++ b/plugin/modelgen/models.gotpl
@@ -36,7 +36,7 @@
 {{- end}}
 
 {{ range $enum := .Enums }}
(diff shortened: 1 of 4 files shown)


vektah:
:joy: thanks.

PR #788, opened 2019-07-09
Add Windows compatibility to ImportPathForDir

### LINKED ISSUE #786: Init not works on Windows (eternal loop)
### What happened?
`go run scripts/gqlgen.go init `
lasts forever

### What did you expect?
Behavior as described here: https://gqlgen.com/getting-started-dep/

Schema and models are not used, just gqlgen.go script as in instructions

### versions
gqlgen version = 0.9.1
go version = go1.12.4 windows/amd64
GOPATH = C:\Users\Username\Develop\go

Actually, the problem is in ImportPathForDir function in internal/code/imports.go path.
There is an infinite loop at 64 line: firstly it checks for the go.mod file, then it tries to find absolute package path and just doing comparison to "/".
I have not tried it in Unix-like systems, but in Windows 10 with dep (and without go mod) installation the loop will never stop because after several cycles variable `modDir` is became equal to "C:\\".
Reference to the problem: https://github.com/99designs/gqlgen/blob/d2c5bf2ae2d8d7da647aded0b3287be4ad2547a9/internal/code/imports.go#L64

dissipative:
Fixes #786

### DIFF
--- a/internal/code/imports.go
+++ b/internal/code/imports.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


vektah:
Sorry, we generally try to avoid `if windows` checks (except in tests).

I've fixed this in https://github.com/99designs/gqlgen/pull/822, instead of looking for `/` it stops as soon as the parent dir has the same path as the current dir.

Issue #790, opened 2019-07-10
dataloader alternative


bugzpodder:
I am trying an alternative to dataloader for dealing with N+1 queries.  I want to see if this proposal is reasonable.

The underlying idea is pretty simple:
If you are resolving a field for a struct, eagerly resolve it for all other structs.
for example, if you have a User {} and you want to resolve its Todos, just resolve any other User's Todos at the same time.

See https://github.com/99designs/gqlgen/pull/808 for a working example.

Consider the models:
```
package api

type User struct {
  Todos []*Todo
}
```

What happens is that you'll need a UserLoader { users []api.User }, and you can expose a method LoadTodos().  This will get all todo ids from each user struct, and instantiate a TodoLoader that will load all the Todos for you.

Any time you need to resolve the TODOs field for a user, just go back to the UserLoader and call LoadTodos.

```
package loader

type UserLoader {
  users []api.User
}

func UserLoaderByIds(ids []int) *UserLoader {...}
func (s *UserLoader) LoadTodos() {
  uids := []int{}
  for _, user := s.users {
     uids = append(uids, user.id)
   }
  todoIds = db.LoadTodosByUserIds(uids)
  todoLoader = TodoLoaderByIds(todoIds)
  // Put todos in users
}
```

See https://github.com/99designs/gqlgen/pull/808 for a working example.

bugzpodder:
I have a working prototype, I can send a PR with an example app if this sounds interesting.  Works well with 1k items with nested resolvers.

I don't like dataloader that much because of having to deal with context and blocking for 1ms every time, but that approach may work better if your service handle multiple requests simutaneously (so it will batch data from these requests at the same time, while my approach only batches per request).

molon:
```
{
  users(first:2) {
    nickname
    orders(first: 2) {
      nodes {
        id
        seller {
          nickname
        }
      }
    }
  }
}
```

If user list has two elements, then there are two orders every user. 
The order_ids: "a","b","c","d".
Then if all orders have same `seller_id:xxx`
How do you ensure that execute sql `select nickname from user where id = 'xxx'` once?

bugzpodder:
In my implementation,  select nickname from user where id = 'xxx' will execute twice (even if first = 100 for both instances). 

So here is what happens in the request:
1. graphql sees users(first:2) query.  It constructs a UserLoader and it will contain two rows of users.  This will execute select nickname from user where id = 'xxx' once here.
[...]

PR #808, opened 2019-07-27
Use a batch loader example

bugzpodder:
This is a demonstration of a batch loader without using data loader.
This should be a new example, but just piggybacking on existing data loader approach for now that shows you what I've changed.  Just requesting feedback for now and if everything looks good I can update it to a new example.

https://github.com/99designs/gqlgen/issues/790


vektah:
I don't think its good to compare this to a dataloader, this is just doing a batched request (with some unnecessary indirection).

Dataloaders collect the batch from disconnected parts of the tree, eg a request like
```
{
   messages {
     sender { name }
     receiver { name, inviter { name } }
     message
   }
}
```

with a dataloder it will collect the requests for both the sender and receiver, making a single batch request to the underlying service, and then cache it. If the inviter lookup is in the cache no backend request needs to be made. 
For the case where the inviter = sender the above query should only make one batch request.

with the example code you have given it will always make 3 requests.

Issue #745, opened 2019-06-14
Keep alive message does not follow the Ack message


tretne:
### What happened?

According to the [GraphQL over WebSocket Protocol](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md#session-init-phase) the `GQL_CONNECTION_ACK` message should be followed by a `GQL_CONNECTION_KEEP_ALIVE` message in the session init phase, if keep alive is used. However, the current implementation of `gqlgen` does not send the first `GQL_CONNECTION_KEEP_ALIVE` message until 25 seconds have passed, using the default value for `DefaultConnectionKeepAlivePingInterval`.  
This often causes issues with our client that uses [`subscriptions-transport-ws`@0.9.16](https://github.com/apollographql/subscriptions-transport-ws) that uses an interval with a default timeout of 30 seconds to check the connection using the keep alive messages, since the first `GQL_CONNECTION_KEEP_ALIVE` after a reconnection does not reach the client before this connection check triggers and determines that it shall close the websocket since it no keep alive message has been received.

What really happens is that our client enters a continuous reconnection loop, but there is more to that problem than just this issue. This issue is just a frequent trigger of that problem after the backend has restarted.

### What did you expect?

That `GQL_CONNECTION_ACK` should be followed by `GQL_CONNECTION_KEEP_ALIVE` and then that `GQL_CONNECTION_KEEP_ALIVE` should be sent according to `DefaultConnectionKeepAlivePingInterval`

### versions

- `gqlgen version` 0.9.0
- `go version` 1.12.6
- go modules

sneko:
Hi @tretne ,

It seems I'm also having this exact issue https://github.com/apollographql/subscriptions-transport-ws/issues/598

Did you set a workaround up? Like increasing the timeout on the frontend side?

Thank you,

sneko:
I made my own workaround. I think that's the simpliest we can do waiting for an update of gqlgen.

On your frontend, just add this parameter when instantiating your WebSocket client:
```
    connectionCallback: () => {
      // TODO: remove this workaround
      // WORKAROUND: prevent infinite reconnection (https://github.com/99designs/gqlgen/issues/745)
      (wsClient as any).wasKeepAliveReceived = true;
    },
```

Like that it simulates the missing `GQL_CONNECTION_KEEP_ALIVE` after reconnection on gqlgen server.

@janflyborg @mitsuyoshi4 @hallgren I hope it could help you

[...]

PR #809, opened 2019-07-29
Allow send first keep alive message instantly

pmachan:
#745 Allow send keep alive message instantly (after connection_ack). 

[Linked with subscription-transport-ws](https://github.com/apollographql/subscriptions-transport-ws/issues/598)

### DIFF
--- a/handler/graphql.go
+++ b/handler/graphql.go
@@ -55,21 +55,22 @@ type PersistedQueryCache interface {
 type websocketInitFunc func(ctx context.Context, initPayload InitPayload) error
 
 type Config struct {
-	cacheSize                       int
-	upgrader                        websocket.Upgrader
-	recover                         graphql.RecoverFunc
-	errorPresenter                  graphql.ErrorPresenterFunc
-	resolverHook                    graphql.FieldMiddleware
-	requestHook                     graphql.RequestMiddleware
-	tracer                          graphql.Tracer
-	complexityLimit                 int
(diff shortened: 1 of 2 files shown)


handler/graphql.go
```diff
@@ -319,6 +320,16 @@ func WebsocketKeepAliveDuration(duration time.Duration) Option {
 	}
 }
 
+// WebsocketKeepAliveFirstInstantly allows you to reconfigure the keepalive behavior.
+// By default, keepalive is sent when first keep alive ping interval is over.
+// Set handler.connectionKeepAliveFirstInstantly = true to send first keepalive
+// instantly. This behaviour is required by a few frontend libraries.
+func WebsocketKeepAliveFirstInstantly(enabled bool) Option {
```
vektah: Lets make this the default, looks like [the spec is pretty clear](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md#gql_connection_init)
> Client sends this message after plain websocket connection to start the communication with the server
> 
> The server will response only with GQL_CONNECTION_ACK + GQL_CONNECTION_KEEP_ALIVE (if used) or GQL_CONNECTION_ERROR to this message.
> 
> payload: Object : optional parameters that the client specifies in connectionParams

vektah:
I've updated this to read closer to the spec and removed the config. Sorry, I would have waited but I want to cut a release today.

PR #816, opened 2019-08-05
Update cors.md to allow CORS for websockets

nii236:
The cors.md file did not cover websockets in particular. I have updated the example in the cors.md file to show how this is done.

### DIFF
--- a/docs/content/recipes/cors.md
+++ b/docs/content/recipes/cors.md
@@ -2,7 +2,7 @@
 title: "Setting CORS headers using rs/cors for gqlgen"
 description: Use the best of breed rs/cors library to set CORS headers when working with gqlgen
 linkTitle: CORS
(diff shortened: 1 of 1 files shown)


docs/content/recipes/cors.md
```diff
@@ -34,9 +34,18 @@ func main() {
 		Debug:            true,
 	}).Handler)
 
+	upgrader := websocket.Upgrader{
+		CheckOrigin: func(r *http.Request) bool {
+			// Check against your desired domains here
+			return true
```
vektah: probably should read something like:
```go
   return r.Host == "example.org"
```

defaulting to true is a huge security hole, as it allows any domain to open websockets to your graph.

nii236:
No worries @vektah, I've changed it.

PR #818, opened 2019-08-07
add configure cache for the query

andrey1s:
- add cache interface 
- configure Lou cache by default
- add option set custom cache

### DIFF
--- a/handler/cache.go
+++ b/handler/cache.go
@@ -1,0 +1,9 @@
+package handler
+
+// Cache for the graphqlHandler
+type Cache interface {
+	// Get looks up a key's value from the cache.
+	Get(key interface{}) (value interface{}, ok bool)
+	// Add adds a value to the cache.  Returns true if an eviction occurred.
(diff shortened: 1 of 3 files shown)


handler/cache.go
```diff
@@ -0,0 +1,9 @@
+package handler
+
+// Cache for the graphqlHandler
+type Cache interface {
+	// Get looks up a key's value from the cache.
+	Get(key interface{}) (value interface{}, ok bool)
+	// Add adds a value to the cache.  Returns true if an eviction occurred.
+	Add(key, value interface{}) (evicted bool)
```
vektah: `evicted` is not used

andrey1s: its interface  for instead https://github.com/99designs/gqlgen/blob/master/handler/graphql.go#L384

vektah: It forces other implementations to return a value we don't use

vektah:
Going to roll this into https://github.com/99designs/gqlgen/pull/885

PR #823, opened 2019-08-09
Generated structs implements graphql interfaces

maqdev:
I have:
 - Updated generated interfaces to have Get methods for fields defined in an interface
 - Updated generated models to actually implement interfaces they inherit from
 - Added test to cover that generated models always match reference, it's easier to support and see diff in case of issues


maqdev:
I realise this have little if any real use... closing.

PR #826, opened 2019-08-12
Provide option to customize request parameter parsing

enjoylife:
**TL;DR** Allow users to override the default query parsing with their own logic

**Problem:**
The current `ServeHTTP` assumes many details when populating and parsing  the struct holding`query`, `operationName`, etc. This "out of the box" functionality  is great for getting up and running quickly  but it is limiting for many use cases:

* Want to use `PUT` or `Content-Type: multipart/alternative` for your requests? Unfortunately we currently can't customize the http method and content-types used during requests for they are hard coded in switch statements. 
* Do you want to handle large file uploads by streaming files in with`MultipartReader`? Again, unfortunately the current implementation simply reads all data into temp files via `ParseMultipartForm`. 

**Solution:**
To enable the use cases I listed above, this pull request does the following:

* creates a new option, `ParserMiddleware`  allowing users to define custom ways of parsing out the query from a request. This option takes in a new type `ParserHandlerFunc` which is somewhat obtuse, but exact in what is required to enable this kind of option.
* In order to enable said option, I had to make public the type holding onto the query params which were ultimately parsed, see `ParsedParams`.  

I have:
- [ ] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))  _If this has a chance to be merged I'll fill out the documentation on the types and provide some simple examples for the usage of the option_

**Origin**

This all started when I was running into an error when making `POST` requests with `"multipart/form-data"`. I have a custom http middleware which does some internal file streaming house keeping and starts a `MultipartReader`. 

When the request made its way into graphql related code to be parsed out, the `processMultipart` call  would always cause and error due to the earlier handling of files.  I couldn't avoid this error which originates from the `http.request` package
```
       // this prevents double parsing of form data after request body has been
       // handed off to a MultipartReader instead of ParseMultipartFrom.
	if r.MultipartForm == multipartByReader {
		return errors.New("http: multipart handled by MultipartReader")
	}
```
Initially to fix and avoid this condition, I just completely gutted the internals of `ServeHTTP` to use `MultipartReader` within my fork, but I figured making this configurable might help others to avoid similar situations. 

With that being said I admit this is a rather intrusive option to provide, in how it's exposing what was once internal. Therefore I'm totally open to change the naming, options, etc. if that helps gets this merged.

### DIFF
--- a/graphql/params.go
+++ b/graphql/params.go
@@ -1,0 +1,22 @@
+package graphql
+
+import "net/http"
+
+type ParsedParams struct {
+	Query         string                 `json:"query"`
+	OperationName string                 `json:"operationName"`
+	Variables     map[string]interface{} `json:"variables"`
+	Extensions    *Extensions            `json:"extensions"`
+}
+
+type Extensions struct {
+	PersistedQuery *PersistedQuery `json:"persistedQuery"`
+}
+
+type PersistedQuery struct {
+	Sha256  string `json:"sha256Hash"`
+	Version int64  `json:"version"`
+}
+
+type ParserFunc func(http.ResponseWriter, *http.Request, ParsedParams)
+type ParserHandlerFunc func(ParserFunc) http.HandlerFunc
--- a/handler/graphql.go
+++ b/handler/graphql.go
@@ -26,22 +26,6 @@ import (
 	"github.com/vektah/gqlparser/validator"
 )
 
-type params struct {
-	Query         string                 `json:"query"`
-	OperationName string                 `json:"operationName"`
-	Variables     map[string]interface{} `json:"variables"`
-	Extensions    *extensions            `json:"extensions"`
-}
-
-type extensions struct {
-	PersistedQuery *persistedQuery `json:"persistedQuery"`
-}
-
-type persistedQuery struct {
-	Sha256  string `json:"sha256Hash"`
-	Version int64  `json:"version"`
-}
-
 const (
 	errPersistedQueryNotSupported = "PersistedQueryNotSupported"
 	errPersistedQueryNotFound     = "PersistedQueryNotFound"
@@ -62,6 +46,7 @@ type Config struct {
 	resolverHook                    graphql.FieldMiddleware
 	requestHook                     graphql.RequestMiddleware
 	tracer                          graphql.Tracer
+	parserHandler                   graphql.ParserHandlerFunc
 	complexityLimit                 int
 	complexityLimitFunc             graphql.ComplexityLimitFunc
 	websocketInitFunc               websocketInitFunc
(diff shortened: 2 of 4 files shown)


graphql/params.go
```diff
@@ -0,0 +1,22 @@
+package graphql
+
+import "net/http"
+
+type ParsedParams struct {
```
enjoylife: was once internal... needs docs

graphql/params.go
```diff
@@ -0,0 +1,22 @@
...
+type Extensions struct {
+	PersistedQuery *PersistedQuery `json:"persistedQuery"`
+}
+
+type PersistedQuery struct {
+	Sha256  string `json:"sha256Hash"`
+	Version int64  `json:"version"`
+}
+
+type ParserFunc func(http.ResponseWriter, *http.Request, ParsedParams)
```
enjoylife: new types

handler/graphql.go
```diff
@@ -390,187 +388,199 @@ func computeQueryHash(query string) string {
...
-	if len(listErr) != 0 {
-		sendError(w, http.StatusUnprocessableEntity, listErr...)
-		return
-	}
+}
 
-	if gh.cache != nil && !cacheHit {
-		gh.cache.Add(reqParams.Query, doc)
-	}
+func newDefaultParser(cfg *Config) graphql.ParserHandlerFunc {
```
enjoylife: maintains the exact logic which was originally found in `ServeHTTP`

handler/graphql_test.go
```diff
@@ -182,6 +182,53 @@ func TestHandlerHead(t *testing.T) {
 	assert.Equal(t, http.StatusMethodNotAllowed, resp.Code)
 }
 
+func TestCustomParser(t *testing.T) {
+	h := GraphQL(&executableSchemaStub{}, ParserMiddleware(func(parserFunc graphql.ParserFunc) http.HandlerFunc {
```
enjoylife: new tests and an example of the option being used

handler/graphql.go
```diff
@@ -390,187 +388,199 @@ func computeQueryHash(query string) string {
...
 
-		if extensions := r.URL.Query().Get("extensions"); extensions != "" {
-			if err := jsonDecode(strings.NewReader(extensions), &reqParams.Extensions); err != nil {
-				sendErrorf(w, http.StatusBadRequest, "extensions could not be decoded")
+		ctx := r.Context()
+
+		var queryHash string
+		apqRegister := false
+		apq := reqParams.Extensions != nil && reqParams.Extensions.PersistedQuery != nil
+		if apq {
```
vektah: Will be easier to see what's changed in here once this flattens back out to the same level, this could be a middleware stack.

imagine
```go
handler.Use(func(next graphql.HandlerFunc)) {
   return graphql.HandlerFunc(func(w, r, params)) {
        /// apq stuff that modifies params
        next(w, r, modifiedParams)
   }
}
```
not inline of course, APQ would just be a function that could be used.

This is really the interface we need to get right for any of this to go ahead IMO.

handler/graphql.go
```diff
@@ -390,187 +388,199 @@ func computeQueryHash(query string) string {
...
 
-	ctx, op, vars, listErr := gh.validateOperation(ctx, &validateOperationArgs{
-		Doc:           doc,
-		OperationName: reqParams.OperationName,
-		CacheHit:      cacheHit,
-		R:             r,
-		Variables:     reqParams.Variables,
+		switch op.Operation {
+		case ast.Query:
+			b, err := json.Marshal(gh.exec.Query(ctx, op))
```
vektah: this json.Marshal might need to be pluggable too, maybe part of the same transport hook?

What would this look like if we wanted to use GRPC as a transport instead of json+POST?

vektah:
Would love to get your thoughts on https://github.com/99designs/gqlgen/pull/885.

Issue #799, opened 2019-07-22
Using directives on a field returning an interface does not allow `nil`


jszwedko:
### What happened?

Using a directive on a field that returns a GraphQL/Go `interface` type prohibits you from returning `nil`, even if the field is nullable,  as the generated type assertion fails.

### What did you expect?

Returning `nil` for an interface field that is nullable is possible when using a directive.

### Minimal graphql.schema and models to reproduce

schema.graphql

```graphql
directive @user on FIELD_DEFINITION

interface Node {
  id: ID!
}

type Query {
  node(id: ID!): Node @user
}
```

gqlgen.yml

```yml
# .gqlgen.yml example
#
# Refer to https://gqlgen.com/config/
# for detailed .gqlgen.yml documentation.

schema:
- schema.graphql
exec:
  filename: generated.go
model:
  filename: models_gen.go
resolver:
  filename: resolver.go
  type: Resolver
autobind: []
```

resolver.go

```go
package gqlgen_test

import (
        "context"
)

type Resolver struct{}

func (r *Resolver) Query() QueryResolver {
        return &queryResolver{r}
}

type queryResolver struct{ *Resolver }

func (r *queryResolver) Node(ctx context.Context, s string) (Node, error) {
        return nil, nil
}

type Node interface {
        IsNode()
}
```

The issue appears to be with this part of the generated code:

```go
  resTmp, err := ec.ResolverMiddleware(ctx, func(rctx context.Context) (interface{}, error) {
    directive0 := func(rctx context.Context) (interface{}, error) {
      ctx = rctx // use context from middleware stack in children
      return ec.resolvers.Query().Node(rctx, args["id"].(string))
    }
    directive1 := func(ctx context.Context) (interface{}, error) {
      return ec.directives.User(ctx, nil, directive0)
    }
    tmp, err := directive1(rctx)
    if err != nil {
      return nil, err
    }
    if data, ok := tmp.(Node); ok {
      return data, nil
    }
    return nil, fmt.Errorf(`unexpected type %T from directive, should be gqlgen-test.Node`, tmp)
  })
```

I believe it should check for `nil` before the type assertion check as type assertions will return false if the stored value is `nil`.

### versions

```
module gqlgen-test

go 1.12

require (
        github.com/99designs/gqlgen v0.9.1
        github.com/vektah/gqlparser v1.1.2
)
```

jszwedko:
[...]

PR #838, opened 2019-08-17
fix directives return nil handling

vvakame:
related #799 and #819 🤔 ??

`.TypeReference.IsNilable` provides Go side type information.
directives.gotpl, args.gotpl and input.gotpl use it.
so, I think usage is split to 2 ways.
1. The value bind to Go field (or variable). ( args.gotpl and input.gotpl )
2. The value map to graphql.Marshaler. ( directives.gotpl )

I think `#2` pattern should use `.TypeReference.GQL.NonNull` instead of `.TypeReference.IsNilable`.

My change returns nil, and if it is invalid. caught by here → https://github.com/99designs/gqlgen/blob/5c644a6fbef1a9bc1c50ef6975686711ec31ff28/codegen/field.gotpl#L79-L86

I have:


codegen/field.gotpl
```diff
@@ -104,14 +104,12 @@
 		if err != nil {
 			return nil, err
 		}
+		if tmp == nil {
```
vektah: :thinking: 
Right, because directives always use `interface{}` they can always be nil, and we don't care what the datatype is. Unlike everywhere else where we cant compare a string to nil without getting a compile-time error.

:+1:

Issue #518, opened 2019-01-29
Optimized batch loading without blocking


atombender:
I'm trying to understand how `gqlgen` can optimize batch loading lazily/concurrently, given that all resolvers are synchronous — i.e. they return the value itself, instead of a thunk (future).

I was reading through the code example for data loading to try to understand how `dataloaden` works. In the its code, I noticed that `Load(key)` simply calls `LoadThunk(key)()`. Then I noticed this line in the readme: "This method will block for a short amount of time, waiting for any other similar requests to come in". Is this was gqlgen assumes?

If you look at the [graphql-go project](https://www.github.com/graphql-go/graphql), all resolvers can return either values or thunks. This means the entire GraphQL tree can be resolved to thunks, which can then queue up all the keys they need so that when the thunks are called, all the objects can be fetched at once.

Can someone fill me in on how gqlgen is supposed to handle this? A resolver like this:

```
func (r *rootQueryResolver) Things(ctx context.Context, uids []*string, limit *int) ([]*models.Thing, error) {
	// ...
}
```

...cannot defer any work at all; it has to return. The only way to batch this is to execute all the resolvers concurrently in goroutines, then have some kind of queue that keeps track of when they've all resolved their keys — or just wait for an arbitrary interval. That seems wrong to me, and all the attendant synchronization sounds like it would be super inefficient.

vektah:
> The only way to batch this is to execute all the resolvers concurrently in goroutines ...  or just wait for an arbitrary interval.

This is what gqlgen does, we eliminate a bunch of places where its unnessicary (binding to fields directly, or methods that dont take a context) and goroutines are cheap enough - [benchmarks](https://github.com/appleboy/golang-graphql-benchmark). There are still a few optimization targets here.

There are optimizations we can make on both sides:
 - a. pooling goroutines to avoid growstack (see https://github.com/99designs/gqlgen/pull/465)
 - b. notifying dataloaden when all goroutines have been dispatched and the scheduler has had a chance to run. This is enough to avoid the sleep.

Overarching goal for gqlgen is for clean, readable, safe code. I dont think trading clean stacktraces and return values for extra milliseconds (way less if b is actioned)

atombender:
Thanks. Since you're the one who's done benchmarks, I'm not going to argue until I could possibly offer contrary evidence, but I followed that benchmark link, and it's not actually doing any concurrent batch loading. It's just measuring a single resolver that returns a static value.

[...]

PR #841, opened 2019-08-19
Pool goroutines

edsrzf:
This is a draft proposal/approach for pooling goroutines. It's not ready to merge, but available for early feedback on the general approach before I go through the effort of trying to add tests, docs, and add further configurability.

There are two problems I'd like to solve:

* Limiting concurrency available for the execution of a single operation.
* Allow a better mechanism for to trigger dataloaders that doesn't rely on timeouts (somewhat related to #518, definitely related to https://github.com/vektah/dataloaden/issues/14). Not solved by pooling goroutines, but I have thoughts around how we might build on top of the pool. If we get to the point that we're happy with this, I'll draft up a proposal issue.

There may also be performance benefits to pooling goroutines, but I haven't measured yet. Judging from #465 it looks like there's potential.

The solution is relatively simple. It adds a new `Pool` type which can be configured with a goroutine limit. The limit defaults to 0, in which case there is no limit, but goroutines are still pooled and reused. `RequestContext` now has a `Pool` field that's used for kicking off goroutines, so goroutines are pooled per-request.

I have:

### DIFF
--- a/codegen/object.gotpl
+++ b/codegen/object.gotpl
@@ -76,7 +76,7 @@ func (ec *executionContext) _{{$object.Name}}(ctx context.Context, sel ast.Selec
 			panic("unknown field " + strconv.Quote(field.Name))
 		}
 	}
-	out.Dispatch()
+	out.Dispatch(ec.Pool)
 	if invalids > 0 { return graphql.Null }
 	return out
 }
--- a/codegen/testserver/generated.go
+++ b/codegen/testserver/generated.go
@@ -8198,7 +8198,7 @@ func (ec *executionContext) _A(ctx context.Context, sel ast.SelectionSet, obj *A
 			panic("unknown field " + strconv.Quote(field.Name))
 		}
 	}
-	out.Dispatch()
+	out.Dispatch(ec.Pool)
 	if invalids > 0 {
 		return graphql.Null
 	}
@@ -8225,7 +8225,7 @@ func (ec *executionContext) _AIt(ctx context.Context, sel ast.SelectionSet, obj 
 			panic("unknown field " + strconv.Quote(field.Name))
 		}
 	}
-	out.Dispatch()
+	out.Dispatch(ec.Pool)
 	if invalids > 0 {
 		return graphql.Null
 	}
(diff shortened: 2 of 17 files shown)


vektah:
I had a similar experiment a while ago to benchmark the difference, one pool for everything though. Most of the performance savings were from reusing stacks, and avoiding calls to morestack.

I think go already does some pooling of goroutines, but they don't hold onto the grown stacks. I get the feeling as this is a separate pool for each request it probably won't have much of a performance impact on but it's worth measuring.

Limiting concurrency can be done very cheaply using atomic. I guess this hinges on what you want to do with dataloading, which probably needs to be moved into gqlgen. 

I've wondered for a long time if dataloaders are actually doing too much, caching can happen more naturally in the graph based on Id+type, perhaps collecting batches becomes much simpler addressing it on it's own, with some interface changes gqlgen probably doesn't need goroutines per item.

edsrzf:
While limiting concurrency would be kind of nice, I'm mostly after a solution to the dataloader issues.

I'm curious whether you've put much thought into this:

> with some interface changes gqlgen probably doesn't need goroutines per item

I'm keen to find a solution that most people are happy with and willing to do some work toward it.

vektah:
> I'm curious whether you've put much thought into this:
>
> > with some interface changes gqlgen probably doesn't need goroutines per item

A little, not enough. My rough thought process goes like this:
 - we could simply make resolvers work on slices, but that interface would be terrible most of the time where single loading is fine.
 - could use an `@batch` annotation to make some resolvers as a batch, but still means repeating the "fetch N users by n ids" call all the places that could return users
 - for the relay node and apollo federation interfaces what we really want is a way to "fetch N entities by N ids + types"

I'm not really sure how the relationship between the graph and the fetching interface works though. 
eg given the ability to lookup users 

```graphql
type ChatMessage {
   author: User
}
```

What should chat message implement? can we use directives?
 - could be thunks?
 - could be returning a ref to id + type. is there some crossover with federated external references?
 - could be goroutines + pooling, but getting smarter about where concurrency actually happens and remove most of them

Definitely more questions than answers.

bugzpodder:
> we could simply make resolvers work on slices, but that interface would be terrible most of the time where single loading is fine.

My approach was to just separate out the resolver into two resolveLoader and resolveField steps.  resolveLoader loads the data and assign it to the field in the api struct.  resolveField just returns whatever that's in the field.  It works well now that resolvers in gqlgen returns a slice of pointers.

edsrzf:
I'm still keen to see some of the underlying issues here addressed, but it seems like this isn't the solution.

Issue #740, opened 2019-06-12
Support for Apollo Federation


BitPhinix:
Are there any plans to add this functionality to GQLGen?

vektah:
Docs for implementing
https://www.apollographql.com/docs/apollo-server/federation/federation-spec/

vektah:
@marwan-at-work has started looking at this

marwan-at-work:
WIP branch https://github.com/marwan-at-work/gqlgen/tree/fed-wip

BitPhinix:
@marwan-at-work Let me know if I can do anything to help you, like testing etc.

StevenACoffman:
Ref: #5

StevenACoffman:
BTW, If you are looking for a golang graphql which composes the overall schema and executes federated queries, consider [nautilus gateway](https://github.com/nautilus/gateway/) as a starting point. It does not currently implement the Apollo Spec, but does provide federated graphql gateway functionality.

benjaminjkraft:
We (Khan Academy) are also interested in this -- @marwan-at-work if there's anything we can do to help (either writing code or testing/etc.) let me know (here or benkraft at khanacademy dot org)!  I have some time allocated this month to work on federation backend support and we're very keen to start using it as soon as possible!

BitPhinix:
Is there any progress here? I'm in dire need of this feature and I'd be happy to help in any way that's required / useful.

marwan-at-work:
@BitPhinix I've gotten a little busy at work but I'm back working on it. The `@key` is working as intended for MVP in my branch above. The `@requires` will change the signatures and that's what im working on now

bugzpodder:
If you are not using any of the new directives (ie, just want apollo-gateway to connect to your graphql service) possibly because you are only running one graphql service currently or just testing things out, you get a little bit done by just copying the graphql schema in https://www.apollographql.com/docs/apollo-server/federation/federation-spec/ and put it into your own schema (you'll need to manually return a string block that consistutes your schema as a whole).  And then the gateway is able to connect to the gql server just fine and act as a pass through.

marwan-at-work:
as a matter of fact, if you're not using any of the new directives, all you need to include in your schema is the `_sdl` that returns the string representation of your graphql.schema file, and you're federation-ready. 

But if you start to need the directives, you can always implement them yourself as part of your business logic. 

This issue is more about creating an abstraction layer that separates the Federation Spec details from the business logic.

[...]

PR #851, opened 2019-08-30
Apollo Federation MVP

marwan-at-work:
This PR is mostly to get some eyes on the current architecture and what can be improved/re-written. 

Once that looks good, I'll move on to adding unit/e2e tests 

Fixes https://github.com/99designs/gqlgen/issues/740


codegen/directives.gotpl
```diff
@@ -1,6 +1,9 @@
 {{ define "implDirectives" }}{{ $in := .DirectiveObjName }}
 	{{- range $i, $directive := .ImplDirectives -}}
 		directive{{add $i 1}} := func(ctx context.Context) (interface{}, error) {
+			if ec.directives.{{$directive.Name|ucFirst}} == nil {
```
JulienBreux: `{{ $directive.Name|ucFirst }}`

api/generate.go
```diff
@@ -17,16 +18,32 @@ func Generate(cfg *config.Config, option ...Option) error {
 	_ = syscall.Unlink(cfg.Exec.Filename)
 	_ = syscall.Unlink(cfg.Model.Filename)
 
+	if err := cfg.Check(); err != nil {
+		return err
```
JulienBreux: `errors.Wrap(err, "generating core failed")`

go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/99designs/gqlgen
 
+go 1.12
```
JulienBreux: Sure?

plugin/federation/federation.go
```diff
@@ -0,0 +1,335 @@
...
+	Entities []*Entity
+}
+
+// New returns a federation plugin that injects
+// federated directives and types into the schema
+func New() plugin.Plugin {
+	return &federation{}
+}
+
+func (f *federation) Name() string {
```
JulienBreux: `// Name returns the plugin name`

plugin/federation/federation.go
```diff
@@ -0,0 +1,335 @@
...
+	return "federation"
+}
+
```
JulienBreux: `// MutateConfig mutates the configuration`

plugin/federation/federation.go
```diff
@@ -0,0 +1,335 @@
...
+			cfg.Models[e.Name].Fields[r.Name] = config.TypeMapField{Resolver: true}
+		}
+	}
+	builtins := config.TypeMap{
+		"_Service": {
+			Model: config.StringList{
+				"github.com/99designs/gqlgen/graphql/introspection.Service",
+			},
+		},
+		"_Any": {Model: config.StringList{"github.com/99designs/gqlgen/graphql.Map"}},
```
JulienBreux: Formatting as below?

plugin/federation/federation.go
```diff
@@ -0,0 +1,335 @@
...
+	f.setEntities(cfg)
+	s := "type Entity {\n"
+	for _, e := range f.Entities {
+		s += fmt.Sprintf("\t%s(%s: %s): %s!\n", e.ResolverName, e.FieldName, e.FieldTypeGQL, e.Name)
+	}
+	s += "}"
+	cfg.AdditionalSources = append(cfg.AdditionalSources, &ast.Source{Name: "entity.graphql", Input: s, BuiltIn: true})
+}
+
+func (f *federation) MutateSchema(s *ast.Schema) error {
```
JulienBreux: `// ...`

plugin/federation/federation.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+	cfg.Directives["external"] = config.DirectiveConfig{SkipRuntime: true}
+	cfg.Directives["requires"] = config.DirectiveConfig{SkipRuntime: true}
+	cfg.Directives["provides"] = config.DirectiveConfig{SkipRuntime: true}
+	cfg.Directives["key"] = config.DirectiveConfig{SkipRuntime: true}
+	cfg.Directives["extends"] = config.DirectiveConfig{SkipRuntime: true}
+
+	return nil
+}
+
```
JulienBreux: `// ...`

plugin/federation/federation.go
```diff
@@ -0,0 +1,335 @@
...
+// RequireField is similar to an entity but it is a field not
+// an object
+type RequireField struct {
+	Name          string                // The same name as the type declaration
+	NameGo        string                // The Go struct field name
+	TypeReference *config.TypeReference // The Go representation of that field type
+}
+
+func (f *federation) GenerateCode(data *codegen.Data) error {
```
JulienBreux: `// GenerateCode ...`

go.mod
hboylan: 1.13 [just released](https://golang.org/doc/go1.13)

plugin/federation/federation.go
```diff
@@ -0,0 +1,335 @@
...
+	union := &ast.Definition{
+		Name:        "_Entity",
+		Kind:        ast.Union,
+		Description: "A union unifies all @entity types (TODO: interfaces)",
+		Types:       []string{},
+	}
+	for _, ent := range f.Entities {
+		union.Types = append(union.Types, ent.Name)
+		s.AddPossibleType("_Entity", ent.Def)
+		// s.AddImplements(ent.Name, union) // Do we need this?
```
vektah: I think this is what's causing the `if ent` in model gen.

https://github.com/99designs/gqlgen/pull/851/files#diff-777f78d80bba845e7f5493310d21fce6R119

plugin/federation/federation.go
```diff
@@ -0,0 +1,335 @@
...
+directive @key(fields: _FieldSet!) on OBJECT | INTERFACE
+directive @extends on OBJECT
+`,
+		BuiltIn: builtin,
+	}
+}
+
+// Entity represents a federated type
+// that was declared in the GQL schema.
+type Entity struct {
```
vektah: I think I would prefer if this referenced the underlying types where possible, rather than copying them just out, just in case another plugin is modifying them.
 - We already have `Def`, so `Name` is redundant
 - FieldName and FieldTypeGQL come from `FieldDef`, so let's add that instead?

If we apply the same logic to `RequireField` I think it can be replaced entirely by a `codegen.Field`

codegen/directives.gotpl
JulienBreux: As you want, it's just a question of uniformity. But you're right.

codegen/field.go
```diff
@@ -95,6 +95,18 @@ func (b *builder) bindField(obj *Object, f *Field) error {
 		f.GoReceiverName = "ec"
 		f.GoFieldName = "introspectType"
 		return nil
+	case f.Name == "_entities":
```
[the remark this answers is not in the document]
vvakame: It would be nice to have a mechanism that can be extended with plugins.

plugin/federation/federation.go
```diff
@@ -0,0 +1,335 @@
...
+				model.Fields = map[string]config.TypeMapField{}
+				cfg.Models[e.Name] = model
+			}
+			cfg.Models[e.Name].Fields[r.Name] = config.TypeMapField{Resolver: true}
+		}
+	}
+	builtins := config.TypeMap{
+		"_Service": {
+			Model: config.StringList{
+				"github.com/99designs/gqlgen/graphql/introspection.Service",
```
vvakame: Can we move `github.com/99designs/gqlgen/graphql/introspection.Service` to `github.com/99designs/gqlgen/plugin/federation/federation.Service` ?

benjaminjkraft:
> I was hoping for more feedback from people who want to use this.

There are some other users over in #740, but for us (Khan Academy) I can say we're already using this and are very happy to see it merged!

cliedeman:
Interested in `SchemaMutator` to implement a relay plugin

nurdism:
couple issues I've found while messing with this

there should be a way to define 
```yml
service:
  filename: "..."
  package: "..."
```
in the `gqlgen.yml` (there is a PR for this on your fork)


When generating service.go, you use `_Entity` interface (which goes in the generated models package), in my setup models are generated on a `models` package (from `gqlgen.yml`) , the _Entity interface should be declared in the service.go package (or use a different name). Underscore values are not exported and will cause errors when the packages differ. 


When using custom scalars as the key in the @key directive, it should import the package and use the type when casting it
IE
```
id, ok := rep["id"].(my.pkg/custom/scalar.type)
```
to 
```
import "my.pkg/custom/scalar"
...
id, ok := rep["id"].(scalar.type)
```
In my case this is causing validation errors. 


should be a way for users to define the federation spec (for external validations or the schema files) and fill in the blanks (of it is not provided)


lastly, more of a nit pick, but the errors for failed type casting on __resolve_entities func in service.go should be more declarative (ie "Failed casting id to scalar.type", rather than it saying "opsies")

Other than that, working pretty well for me.  Still doing some testing. I'll report back if I find anything else. Keep up the good work 👍

sothychan:
This is great.  I'm going to try using this now and will provide feedback if any.

sothychan:
@marwan-at-work   @vektah   I have a PR that pulls in the latest gqlgen into @marwan-at-work 's federation branch and fixed the merge conflicts.  Here it is: https://github.com/marwan-at-work/gqlgen/pull/3

marwan-at-work:
@sothychan that's great thank you 🙏 -- i've been incredibly busy the last couple of months (switched companies), but will get back to this ASAP.

codegen/field.go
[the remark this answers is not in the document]
marwan-at-work: @vektah maybe if the user has an `_entities` query *without* federation...we can possibly check `cfg.Federated && f.Name == _entities` ... that said, it would be great if the plugin can inject this part, but it looks like there's no way right now...can you confirm? Thanks!

sothychan:
@marwan-at-work I'm running into a weird issue saying "_Entity not found in typemap" when I use the autobind feature that's in v0.10.  I'm going to look into why and submit a PR.

codegen/field.go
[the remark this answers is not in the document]
vektah: Welcome back!

I wonder if we could add something to `TypeMapField` in config that replaces all of this logic?
```go
type TypeMapField struct {
	Resolver  bool   `yaml:"resolver"`
	FieldName string `yaml:"fieldName"`

	// new
	GeneratedMethod string `yaml:"-"`
}
```

which would be set [here](https://github.com/99designs/gqlgen/pull/851/files#diff-6edbafe93c80509bb043690f2a456213R104) to reference `__resolve_entities`

vvakame:
hi, I tried to implement https://github.com/apollographql/federation-demo by gqlgen with this PR.
It's works fine! 👏 

I can make public my repository but federation-demo repository's license is unclear.
I opened https://github.com/apollographql/federation-demo/issues/12 . please wait.

josemarluedke:
@marwan-at-work @vektah 

Any new development in this area? We are starting to look for a service-oriented graphql, and the federation would be perfect for that. Any changes to get this moving?

StevenACoffman:
@nurdism We implemented [that here](https://github.com/99designs/gqlgen/commit/dd02f6da7598df6a76d3ee1f7f375d25349d4ed2)

marwan-at-work:
Now that I'm fully onboarded with my new job...I'm very happy to come back to this, I've cleared my Sunday for this PR 👍 

Hopefully I can wrap it up within that day but I did notice there are some merge conflicts also when I generate tests a lot of them do fail...although federation itself works okay 😄

plugin/federation/federation.go
marwan-at-work: @vektah does that mean I won't need to check for the `key` directive if I add this line?

codegen/field.go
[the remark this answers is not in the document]
marwan-at-work: @vektah that's an interesting idea...I'll give it a shot 👍

plugin/federation/federation.go
marwan-at-work: @vektah I did it for `Def` and `FieldDef`. However, `RequireField` would make `setEntities` create a dummy `codegen.Field` so I can have a Name to match against in `GenerateCode`, then `GenerateCode` will have to replace its codegen.Field with the dummy one which fields a bit ugly. Would you like me to switch that up as well?

codegen/field.go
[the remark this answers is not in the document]
marwan-at-work: Actually, TypeMapField is in `MutateConfig` not `MutateSchema` where you linked  above, can you expand more on how that would work? Thanks!

marwan-at-work:
@vektah all of the comments are addressed except: 

1. field.go (would love to know more about how we can fix that)
2. making the build/tests pass as I'm not familiar with the codebase enough to know why they're failing :)

go.mod
marwan-at-work: AFAIK, go 1.12 will make gqlgen compatible with any users not on 1.13

plugin/federation/federation.go
marwan-at-work: Nevermind, I uncommented the line and it added _Entity to `implementors`. This seems to only be used in fragment-related things?

marwan-at-work:
@vektah I managed to get the builds to pass but federation tests are still super preliminary. 

This leaves the field.go which I'd need some guidance as to how to inject into it from the plugin or I can a TODO comment for a follow up PR if that's something that requires more figuring out. 

Thanks

jepp2078:
Quick question @marwan-at-work. Won't the name _Entity in the generated model give problems for people that use packages for their models, since _Entity isn't exported?

codegen/field.go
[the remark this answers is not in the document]
vektah: Lets leave it like this for now and get this on master. Might do some cleanup once it lands to make this more pluggable.

vektah:
I've merged the gqlparser pr and tagged a new release. Lets update the this and get it :ship:'d.

> This leaves the field.go which I'd need some guidance as to how to inject into it from the plugin or I can a TODO comment for a follow up PR if that's something that requires more figuring out.

Lets ship this as is and iterate on master.

marwan-at-work:
@vektah sounds great, I updated the go.mod file :v: 

@jepp2078 that's a valid point, although I'm not sure in which use case you'd need to programmatically import the _Entity interface. That said, I imagine it would be possible to let the user customize those names just like any other model.

vektah:
I think the mod file still needs a version bump to 1.2.1
https://github.com/99designs/gqlgen/pull/851/commits/d2a13d33cdcb27e8e141c54e1ab1fa0aebad2d2b#diff-37aff102a57d3d7b797f152915a6dc16L25

marwan-at-work:
@vektah done

jepp2078:
> @jepp2078 that's a valid point, although I'm not sure in which use case you'd need to programmatically import the _Entity interface. That said, I imagine it would be possible to let the user customize those names just like any other model.

@marwan-at-work I'm pretty sure it would break as soon as you set the package for exec and model to two different values, since generated.go needs to use _Entity.

marwan-at-work:
@jepp2078 now that this is merged, it would be incredibly helpful to open an issue with reproduction steps of how it would break. I can dig into it afterwards, thanks!

nitin302:
> hi, I tried to implement https://github.com/apollographql/federation-demo by gqlgen with this PR.
> It's works fine! 👏
> 
> I can make public my repository but federation-demo repository's license is unclear.
> I opened [apollographql/federation-demo#12](https://github.com/apollographql/federation-demo/issues/12) . please wait.

@vvakame @marwan-at-work @vektah can you point to this project where there is an working example?

vvakame:
I cant make public it. I cant get response from Apollo org members yet...
We need write full-scratched example

marwan-at-work:
I haven't updated this repo in a while, but I also Go'fied the federation-demo repo when I was implementing the spec in gqlgen: https://github.com/marwan-at-work/federation-demo

nitin302:
> I cant make public it. I cant get response from Apollo org members yet...
> We need write full-scratched example

@vvakame license is now added.

vvakame:
wow, thanks!
I made public about https://github.com/vvakame/gqlgen-apollo-federation-demo

PR #853, opened 2019-09-08
Added graphiql handler

andrioid:
I've been using a [rather awesome version of GraphiQL](https://github.com/OneGraph/graphiql-with-extensions) on Gatsby recently and found myself missing it on my gqlgen project.

This PR adds an optional handler called `GraphiQL` that can be used exactly like the `Playground` handler. Assets and code is fetched through the unpkg.com CDN.

I will update documentation and examples if that's what we want to do. There is nothing wrong with Playground. I just prefer this one.

I have not:
 - Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 -  Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

### DIFF
--- a/handler/graphiql.go
+++ b/handler/graphiql.go
@@ -1,0 +1,85 @@
+package handler
+
+import (
+	"html/template"
+	"net/http"
+)
+
+var graphiqlPage = template.Must(template.New("graphiql").Parse(`<!DOCTYPE html>
+<html>
+<head>
+	<meta charset="utf-8"/>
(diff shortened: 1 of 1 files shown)


vektah:
I think this can easily live in its own repo - would be happy to link to it in the readme but dont want to add handlers for every graphql ui to gqlgen.

PR #854, opened 2019-09-08
Fix for nested fields backed by map or interface

wabain:
In order to support distinguishing between null and missing values in GraphQL, gqlgen allows [binding GraphQL types](https://github.com/99designs/gqlgen/blob/55b2144289debeb5ca104a4d01d36f96be5ed84c/docs/content/reference/changesets.md) to `map[string]interface{}`. However, currently those types can't be embedded in generated models because model generation doesn't know about the special cases in type binding that make it work. Trying with master or v0.9.3 gives the error `modelgen: package cannot be nil`.

In order to fix this, this PR makes two changes:

* Check for a type name of `interface{}` and `map[string]interface{}` in `Binder.FindType`. I added a variant function that takes responsibility for splitting the package and type names, but the old function is still there for backwards compatibility.

* Extend the nilable type logic used in codegen for the executor to model field generation, so that an optional GraphQL value is bound to a field of type `map[string]interface{}` instead of `*map[string]interface{}`. In order to maintain backwards compatibility, this change only applies to the special cases where the map or interface type is specified directly in the gqlgen config, not to other Go types that have an underlying nilable type.

I have:
 - [x] **N/A** Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

### DIFF
--- a/codegen/config/binder.go
+++ b/codegen/config/binder.go
@@ -75,7 +75,22 @@ func (b *Binder) ObjectPosition(typ types.Object) token.Position {
 	return pkg.Fset.Position(typ.Pos())
 }
 
+func (b *Binder) FindTypeFromName(name string) (types.Type, error) {
+	pkgName, typeName := code.PkgAndType(name)
+	return b.FindType(pkgName, typeName)
+}
+
 func (b *Binder) FindType(pkgName string, typeName string) (types.Type, error) {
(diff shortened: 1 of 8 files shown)


vektah:
This looks awesome, thanks for the PR.

PR #859, opened 2019-09-11
Hotfix/857

jamesabbottsmith:
Added err check


vektah:
Please a look at https://github.com/99designs/gqlgen/pull/861 and see if it covers your use case

jamesabbottsmith:
I haven't had time to test it yet, but #861 looks like it does the trick. Thanks!

PR #885, opened 2019-10-01
Refactor handler package

vektah:
The handler package has grown organically for a long time, its time to clean up the interfaces and get it ready for 1.0.

### New handler extension API

The core of this changes the handler package to be a set of composable extensions. The extensions implement a set of optional interfaces:
- **OperationParameterMutator** runs before creating a OperationContext (formerly RequestContext). allows manipulating the raw query before parsing.
- **OperationContextMutator** runs after creating the OperationContext, but before executing the root resolver.
- **OperationInterceptor** runs for each incoming query after parsing and validation, for basic requests the writer will be invoked once, for subscriptions it will be invoked multiple times.
- **ResponseInterceptor** runs around each graphql operation response. This can be called many times for a single operation the case of subscriptions.
- **FieldInterceptor** runs around each field

Anatomy of a request@2x

Users of an extension should not need to know which extension points are being used by a given extension, they are added to the server simply by calling `Use(extension)`.

There are a few convenience methods for defining middleware inline, instead of creating an extension
 - AroundFields
 - AroundOperations
 - AroundResponses

Middleware:
 - [x] APQ
 - [x] Query Complexity
 - [x] Error Presenters and Recover Func
 - [x] Introspection Query support (I've flipped this to opt-in)
 - [x] Query AST cache
 - [x] Tracing API

```go
srv.Use(extension.Introspection{})
srv.Use(extension.AutomaticPersistedQuery{
	Cache: lru.New(100),
})
srv.Use(apollotracing.Tracer{})
```

### Transports

In addition to extensions, transports can also be registered against the server. Only one transport can serve a given request.

Transports:
 - [x] GET
 - [x] JSON
 - [x] Multipart form
 - [x] Websockets

new usage looks like this
```go
srv := New(es)

})
srv.AddTransport(transport.Options{})
srv.AddTransport(transport.MultipartForm{})
````

### More contesistent naming

As part of cleaning up the names the RequestContext has been renamed to OperationContext, as there can be multiple created during the lifecycle of a request. A new ResponseContext has also been created and error handling has been moved here. This allows each response in a subscription to have its own errors. I'm not sure what bugs this might have been causing before...

### Removal of tracing

Many of the old interfaces collapse down into just a few extension points:
Anatomy of a request@2x (1)

The tracing interface has also been removed, tracing stats are now measured in core (eg time to parse query) and made available on the operation/response contexts. Much of the old interface was designed so that users of a tracer dont need to know which extension points it was listening to, the new handler extensions have the same goal. 

### Backward compatibility

I've added a backwards compatibility layer that keeps most of the original interface in place. There are a few places where BC is known to be broken:
 - ResponseMiddleware: The signature used to be `func(ctx context.Context, next func(ctx context.Context) []byte) []byte` and is now `func(ctx context.Context) *Response`. We could maintain BC by marshalling to json before and after, but the change is pretty easy to make and is likely to cause less issues.
 - The Tracer interface has been removed, any tracers will need to be reimplemented against the new extension interface.

### CODE REVIEW

File: graphql/handler/server.go [hunk 1]
```diff
@@ -0,0 +1,184 @@
...
+}
+
+func (s *Server) parseOperation(ctx context.Context, rc *graphql.RequestContext) (*ast.QueryDocument, *gqlerror.Error) {
+	ctx = rc.Tracer.StartOperationValidation(ctx)
+	defer func() { rc.Tracer.EndOperationValidation(ctx) }()
+
+	return parser.ParseQuery(&ast.Source{Input: rc.RawQuery})
+}
+
+func (gh *Server) validateOperation(ctx context.Context, rc *graphql.RequestContext) (context.Context, *ast.OperationDefinition, gqlerror.List) {
```
vvakame: can remove `ctx`? same as parseOperation.

graphql/handler/tracer.go
```diff
@@ -0,0 +1,26 @@
...
+		return func(ctx context.Context, writer Writer) {
+			rc := graphql.GetRequestContext(ctx)
+			rc.AddTracer(tracer)
+			rc.AddRequestMiddleware(func(ctx context.Context, next func(ctx context.Context) []byte) []byte {
+				ctx = tracer.StartOperationExecution(ctx)
+				resp := next(ctx)
+				tracer.EndOperationExecution(ctx)
+
+				return resp
+			})
```
vvakame: How about add `StartOperation` and `EndOperation` ?
I don't understanding correctly about Subscription...

vektah: covered by OperationInterceptor

vvakame:
hi, I tried this branch in our internal tool.
I think It's definitely improved than v0.10! 👍 

so, I write some suggestions.

* If we have`srv.UseRecommended()` that is good for newbies...? 🤔 
* use `extension.ComplexityLimit` is little bit difficult
  * I want `extension.ComplexityLimitSize(10000)`
  * ...and `extension.ComplexityLimit` (skip complexity checking when user is admin)
* `HandlerExtension interface{}` should be something like `HandlerExtension interface{ IsSuperCoolExtension() }` 🤔 
  * prevent invalid use & runtime error to compile time error
* `extension.AutomaticPersistedQuery{Cache: lru.New(100)}` should be `extension.NewAutomaticPersistedQuery(...)`
  * nil reference error occurred when `Cache` is nil. We wanna get an error on early stage (e.g. server startup timing)
* `OperationContextMutator` and other extension requires return `*gqlerror.Error` value, but application returns normal error...
  * I use `srv.SetErrorPresenter(graphQLErrorPresenter())` and `return graphQLErrorPresenter()(ctx, err)`. little bit weird 🤔 
* so, OperationContext doesn't have `ComplexityLimit` and `OperationComplexity`
  * I want to write it to logs
  * I have no good idea ... 😿 
  * current implementations can't communicate data with each extensions. it required? (e.g. ComplexityLimit extension <-> Logger extension

and, I found bugs.

* `transport.POST{}` implementations called w.WriteHeader multiply.
  * https://golang.org/src/net/http/server.go#L1114
  * https://github.com/golang/appengine/blob/16bce7d3dc4e458f2f6f56a1349cbbfcdc8a8fdf/internal/api.go#L376
* `AutomaticPersistedQuery` doesn't match any `Use` func supported types.
  * maybe `OperationParameterMutator` ?
* APQ will be failed because `w.WriteHeader(http.StatusUnprocessableEntity)` is called always when returns `PersistedQueryNotFound`.
  * client receive 422, but 200 is expected.
* generated code doesn't have `// Code generated by github.com/99designs/gqlgen, DO NOT EDIT.` header
  * linter makes noises 🙉

vektah:
Thanks for the thorough review

> If we havesrv.UseRecommended() that is good for newbies...? thinking

There is `handler.NewDefaultServer` with the reccomended defaults.

> use extension.ComplexityLimit is little bit difficult
> I want extension.ComplexityLimitSize(10000)

Ok I've added in the fixed limit func again

> ...and extension.ComplexityLimit (skip complexity checking when user is admin)

This should already be possible

> HandlerExtension interface{} should be something like HandlerExtension interface{ IsSuperCoolExtension() } thinking
> prevent invalid use & runtime error to compile time error

Sure :+1:

> extension.AutomaticPersistedQuery{Cache: lru.New(100)} should be extension.NewAutomaticPersistedQuery(...)
> nil reference error occurred when Cache is nil. We wanna get an error on early stage (e.g. server startup timing)

I've added a Validate method instead, wrapping everything in new makes forward compatibility hard. We should be able to add new optional params easily to any of these.

> OperationContextMutator and other extension requires return *gqlerror.Error value, but application returns normal error...

I dont have a great solution to this, its there so that extensions can manipulate the raw graphql response to conform to their relevant specs. These are often not application level issues, they are dealing with things like APQ. Whats are you using OperationContextMutator for that has application errors?


>> so, OperationContext doesn't have ComplexityLimit and OperationComplexity. I want to write it to logs

Cool, I've taken a stab at allowing extensions to communicate. What do you think?


> transport.POST{} implementations called w.WriteHeader multiply. https://golang.org/src/net/http/server.go#L1114
https://github.com/golang/appengine/blob/16bce7d3dc4e458f2f6f56a1349cbbfcdc8a8fdf/internal/api.go#L376

Fixed :+1:

> AutomaticPersistedQuery doesn't match any Use func supported types.
maybe OperationParameterMutator ?
APQ will be failed because w.WriteHeader(http.StatusUnprocessableEntity) is called always when returns PersistedQueryNotFound.
client receive 422, but 200 is expected.

Fixed :+1:

> generated code doesn't have // Code generated by github.com/99designs/gqlgen, DO NOT EDIT. header linter makes noises hear_no_evil

Oops! Yeah I turn off those headers while working on codegen so my editor actually complains about errors. Forgot to turn it off :sweat_smile:

vektah:
I've run some benchmarks of this vs master and its slightly about 9% slower:
before:
```
595623             21525 ns/op            6443 B/op        119 allocs/op
```
after:
```
546052             26687 ns/op            7556 B/op        137 allocs/op
```

Most of this seems to be due to treating all requests like subscriptions, theres a little less overhead but it drys out the implmentations. I think this is acceptable, as long as we spend sime time tuning the bigger offenders after this lands:
image

vvakame:
> There is handler.NewDefaultServer with the reccomended defaults.

👍 

> I've added a Validate method instead

cool! 👍 
I'm loving `Validate(schema ExecutableSchema) error`

> Whats are you using OperationContextMutator for that has application errors?

user login checker. out (internal) application doesn't response any contents when user is not loggged-in to our company's SSO.
I want to access something likes `srv.GetErrorPresenter` from ctx. But I have workaround now, It's not serious matter.

<details>

```go
func userLoginChecker() interface {
	graphql.HandlerExtension
	graphql.OperationContextMutator
} {
	errPresenter := graphQLErrorPresenter() // `srv.SetErrorPresenter(graphQLErrorPresenter())`

	return &operationContextMutatorImpl{
		name: "UserLoginChecker",
		f: func(ctx context.Context, rc *graphql.OperationContext) *gqlerror.Error {
			user := domains.CurrentUser(ctx)
			if user == nil {
				loginURL, err := aeuser.LoginURL(ctx, "/")
				if err != nil {
					return gqlerror.Errorf("%s", err.Error())
				}
				return errPresenter(ctx, domains.ErrLoginRequired.Detailf("loginURL: %s", loginURL))
			}
			return nil
		},
	}
}
```

</details>

> I've taken a stab at allowing extensions to communicate. What do you think?

`rc.Stats.GetExtension` is looks good for me 👍 
so, If I say something greedy, I wanna access it from APQ too. I'm logging PersistedQueryHash with 

```
	var persistedQueryHash string
	{
		b := sha256.Sum256([]byte(rc.RawQuery))
		persistedQueryHash = hex.EncodeToString(b[:])
	}
```

but `rc.Stats.GetExtension("APQ").(*FooBar)` is more easy.

---

suggestions

* `OperationParameterMutator` (and other HE) should be embed `HandlerExtension` interface
  * `var _ graphql.OperationContextMutator = (*foo)(nil)` is not enough. need `var _ graphql.HandlerExtension = (*foo)(nil)`, but it's not convenience.
* Does HandlerExtension have naming convention?
  * something like `CamelCase` or `camelCase` or...? basically, `CamelCase` is popular in code now.
* `graphql.GetOperationContext` should not be return nil in all situations
  * problem is ... when `executor#CreateOperationContext` returns `return nil, err`, `executor#DispatchError` called. but `apollotracing#InterceptResponse` doesn't check `rc` is nil. nil reference occurred.
  * all of hook point (HandlerExtension) can access to OperationContext via ctx. It's easy to understand & coding I think.

vvakame:
👀

vektah:
@vvakame I've injected operation context earlier now, so it should be availble during all of the handler hooks, and it now panics when trying to fetch it.

This makes it possible for APQ to push stats in (it runs very early, on the raw request params before decoding)/

vvakame:
cool! 👍 
I made #947 . if its merged, our internal tools will be work with this PR branch.

vektah:
Ok, time to ship this then. Lets start moving towards 0.11 (or maybe 1.0, if we get the plugin interfaces sorted too...)

Issue #380, opened 2018-10-13
Conditionally removing fields from introspection based on auth


vektah:
Enable consuming apps with different application roles to generate clients with only the methods they have access to.

Note there is terminology overlap here between the end user role and the application role. Each end user would probably still see the same view of the graph.

saward:
I'm interested in hiding certain fields/queries/mutations from the schema introspection depending on whether the user is authenticated or not, and varying based on roles.  Am I right in thinking that this issue relates to doing something like that?

mathewbyrne:
@saward correct.

sdalezman:
@vektah @mathewbyrne was looking through the introspection code - would it make sense to add a directive that allows disabling of fields and inputs in the schema similar to how deprecated is currently handled? 

If so, think this would be a quick win we'd be able to help out with. Let me know and someone on our team can allocate some time in the next few days to solve this

mathewbyrne:
I think a specific directive is probably not the right way to implement this; `@deprecated` is special in that it is a [built-in directive from the spec](https://facebook.github.io/graphql/June2018/#sec--deprecated) and field deprecation is also [a feature from the spec](https://facebook.github.io/graphql/June2018/#sec-Field-Deprecation).

So I think this feature needs to be thought out a little more.  Currently the only input into introspection is the parsed schema, so I imagine there might be some sort of additional input here that introspection exports so allow granular control of which types are exposed (maybe?).  If you guys are keen to address this I'd love to see a proposal.

sdalezman:
That's a good point regarding directives - reason I reached for adding a built in directive is that it would be embedded in the workflow of a developer. 

I think the other option would be to introduce another key in gqlgen so that we can specify the fields/input to be excluded from introspection. I think this would be a happy medium and fit into the overall workflow of gqlgen. I don't think this should be something that's introduced in the options to `GraphQL`. To me the `GraphQL` instantiation is something that should change very infrequently and isn't part of my day to day workflow as a developer.

The nice part of introducing a new key to gqglen is it's already part of my existing workflow (whenever I define queries I always think about 1. updating schema 2. any updates that need to happen to gqlgen.yml).

I can understand not wanting to "pollute" gqlgen, but maybe it becomes a new key under `schema` called `disableIntrospection` and it would include a list of fields and inputs to be disabled? That would fit very cleanly into existing workflows I think. 

If that works for you we can map out the code updates itself to facilitate that update.

mathewbyrne:
[...]

PR #888, opened 2019-10-05
Conditionally removing fields from introspection

andcan:
First work on https://github.com/99designs/gqlgen/issues/380.

 - Added new interface for plugins: `SchemaMutator`: I need to be able to mutate schema before code generation.
	Plugins implementing this interface get called between `ConfigMutator.MutateConfig` and `CodeGenerator.GenerateCode`.
	I had to modify `codegen.BuildData` signature and add a couple rows into `api.Generate` to make this work
 - Added `introspection` plugin.
	Injects `@introspection` directive on `__Type.inputFields` and `__Type.fields`.
	Generates `@introspection` directive implementation.
- Added `introspection` example.

This PR is to receive comments/suggestions on how to implement this.
If/when looks good I'll add tests and docs.


plugin/introspection/plugin.go
```diff
@@ -0,0 +1,96 @@
...
+	return &Plugin{cfg: cfg}
+}
+
+func (p *Plugin) Name() string {
+	return "introspection"
+}
+
+func (p *Plugin) MutateConfig(cfg *config.Config) error {
+	for _, d := range p.cfg.Directives {
+		cfg.Directives[d] = config.DirectiveConfig{SkipRuntime: true}
```
vvakame: I'm already using `@hasRole` directive, and it has runtime implementation.
I think `MutateConfig` is not necessary. 
user defines `@hide` and `@requireAuth` in schema. `skip_runtime` settings should write gqlgen.yml explicitly.

plugin/introspection/plugin.go
```diff
@@ -0,0 +1,96 @@
...
+	introspection := &ast.DirectiveDefinition{
+		Description: "",
+		Name:        "introspection",
+		Arguments:   nil,
+		Locations: []ast.DirectiveLocation{
+			ast.LocationInputFieldDefinition,
+			ast.LocationFieldDefinition,
+		},
+		Position: nil,
+	}
```
vvakame: `@introspection` does not have an any (common) specs. I want to hide it in IntrospectionQuery.
I recommend add something like `SkipRuntime: true` or `Builtin: true` .
but wanna generate directives in code. 🤔 

@vektah do you have any opinion in this topic?

plugin/introspection/plugin.go
```diff
@@ -0,0 +1,96 @@
+package introspection
```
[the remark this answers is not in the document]
andcan: Yes, you're right. I renamed it to introauth, could't find a better naeìme.

plugin/introspection/plugin.go
andcan: Yes you're right, this is not necessary.

codegen/data.go
```diff
@@ -35,17 +35,14 @@ type builder struct {
 	Directives map[string]*Directive
 }
 
-func BuildData(cfg *config.Config) (*Data, error) {
+func BuildData(cfg *config.Config, schema *ast.Schema, schemaStr map[string]string) (*Data, error) {
```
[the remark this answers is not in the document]
vektah: I think I like this implementation better https://github.com/99designs/gqlgen/pull/851/files#diff-86f7257ff3b7e8564cb7fae47bd53676R37-R46

Do we actually need to modify schema or can it be purely additive?

plugin/introspection/plugin.go
vektah: Yep, it already exists https://github.com/99designs/gqlgen/blob/master/codegen/directive.go#L24

vektah:
The [handler refactor](https://github.com/99designs/gqlgen/pull/885) pr has some overlap with this, between the two we should be able to pull introspection out entirely into a [handler extension](https://github.com/99designs/gqlgen/pull/885/files#diff-a2eafa7f07f95598428b68504f03e5ebR1-R18).

Might be good to sort out what that looks like before we cut a release, so we get the interfaces for generated handler plugins right.

vvakame:
@andcan #885 is merged. could you merge master/HEAD into this PR?
I really need this changes 😉 please let me know I can help you anything.

andcan:
Sorry lost that, I will give a look to the handler refactor but I'm a bit out of time right now it might take some time. I'll let you know if I need help.

vvakame:
@andcan 👀

vektah:
Closing this until its rebased and ready for review again, feel free to reopen.

Issue #876, opened 2019-09-24
Question/Suggestion: Hook plugin into modelgen


nurdism:
I was looking for a way to add a tag to the model gen with a directive, for things like ORMs and Validation of stuct fields, is there a way to hook into that? Or is this not currently possible?

PR #897, opened 2019-10-11
Add possibility to hook into modelgen plugin

mskrip:
I realize there already is a branch [modelgen-structtags](https://github.com/99designs/gqlgen/tree/modelgen-structtags) for the mentioned issue and for what this change tries to accomplish. I'm opening a PR nonetheless at least as a suggestion since the feature has more general implications (possibility to mutate the whole generated models before rendering). Feel free to close this if the mentioned branch is close to solving the issue.

---

This change introduces option to implement custom hook for model
generation plugin without the need to completly copy the whole `modelgen` plugin.

One very possible case is as described in #876 and with this change the solution for
that can be:

```golang
func mutateHook(b *ModelBuild) *ModelBuild {
	for _, model := range b.Models {
		for _, field := range model.Fields {
			field.Tag += ` orm_binding:"` + model.Name + `.`  +  field.Name + `"`
		}
	}

	return b
}

...

func main() {
    p := modelgen.Plugin {
        MutateHook: mutateHook,
    }

    ...
}

```

I have:

### DIFF
--- a/plugin/modelgen/models_test.go
+++ b/plugin/modelgen/models_test.go
@@ -1,6 +1,7 @@
 package modelgen
 
 import (
+	"go/ast"
 	"go/parser"
 	"go/token"
 	"io/ioutil"
@@ -14,7 +15,9 @@ import (
 func TestModelGeneration(t *testing.T) {
 	cfg, err := config.LoadConfig("testdata/gqlgen.yml")
 	require.NoError(t, err)
-	p := Plugin{}
+	p := Plugin{
+		MutateHook: mutateHook,
+	}
 	require.NoError(t, p.MutateConfig(cfg))
 
 	require.True(t, cfg.Models.UserDefined("MissingTypeNotNull"))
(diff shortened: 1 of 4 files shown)


JulienBreux:
Good job!

lulucas:
Very useful feature for bson, mapstructure, camel to under score...

plugin/modelgen/models_test.go
```diff
@@ -42,4 +45,30 @@ func TestModelGeneration(t *testing.T) {
 			require.True(t, len(words) > 1, "expected description %q to have more than one word", text)
 		}
 	})
+
+	t.Run("tags are applied", func(t *testing.T) {
+		node, err := parser.ParseFile(token.NewFileSet(), "./out/generated.go", nil, 0)
+		require.NoError(t, err)
+		for _, obj := range node.Scope.Objects {
```
vektah: holy moly batman.

would this be good enough (and a lot more understandable) if we just searched for the expected string in the file?

mskrip: Yea, I guess that's enough and a lot simpler. Changed it.

vektah:
This needs some docs before we can :shipit:

vektah:
Those docs look great, thanks!

Issue #896, opened 2019-10-11
Nondeterministic output when query and subscription return (non)nullable list elements


zannen:
With the schema below, output is nondeterministic.

This behaviour occurs when a query returns a list of nullable things and a subscription returns a list of non-nullable things (or vice versa), the generated code for the _subscription_ (it seems to always be the subscription) flips between using the normal function (`marshalOThing2ᚖgqlgentestᚐThing`) and the function with the null check (`marshalNThing2ᚖgqlgentestᚐThing`).

This behaviour does **not** occur:
* with _two queries_, one returning `[Thing]` and one returning `[Thing!]`
* with _two subscriptions_, one returning `[Thing]` and one returning `[Thing!]`
* with one query and one subscription, if _`Thing` is replaced by `String`_

### What did you expect?
Deterministic output.

### Minimal graphql.schema to reproduce
```gql
schema {
  query: Query
  subscription: Subscription
}

type Thing {id: Int}

type Query {
  things1: [Thing] # Note the "!" or lack thereof.
}

type Subscription {
  things2: [Thing!] # Note the "!" or lack thereof.
}
```

### gqlgen.yml
```yaml
---
schema: graphql.schema
exec:
  filename: generated.go
model:
  filename: models.go
```

Here's the script I've been using to test. Run it and wait for a while to see if files with different checksums get created.

```bash
rm -f *generated.go
while test "$(find -name '*-generated.go' | wc -l)" -lt 2 ; do
  date
  gqlgen -c gqlgen.yml
  cp -a generated.go "$(md5sum generated.go | awk '{print $1}')-generated.go"
  sleep 1
done
```

### versions
 - `gqlgen version` v0.10.1
 - `go version` go1.11.13 linux/amd64
 - dep or go modules? go modules: `go.mod`
    ```
    module gqlgentest
    
    require (
    	github.com/99designs/gqlgen v0.10.1
    	github.com/vektah/gqlparser v1.1.2
    )
    ```

alexsn:
I have the exact some issue.

zannen:
The template creating the code line that flips is [codegen/type.gotpl L93](https://github.com/99designs/gqlgen/blob/master/codegen/type.gotpl#L93).

zannen:
PR https://github.com/99designs/gqlgen/pull/900 should fix this.

colinschoen:
Pretty sure I am running into this too. Thanks for the PR @zannen.

zannen:
Thanks @vektah.

PR #900, opened 2019-10-13
Update UniquenessKey for when Element is/isn't nullable (#896)

zannen:
With a schema:
```gql
type Query {
  things1: [Thing] # Note the lack of "!"
}

type Subscription {
  things2: [Thing!] # Note the "!"
}
```

the `UniquenessKey` for the two lists is the same, which causes non-deterministic output.

I have:
 - [x] Added tests covering the bug (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md)) - **Note** I realise I have not added tests. I took a look, but as the testing doc says, it's not easy. And this Issue causes flaky output, so I'm not sure what kind of test to write. **Assistance requested**.
    * *Edit 1*: In place of a unit test, I have added a check (https://github.com/zannen/gqlgen/commit/67f8031c02aef393b134f52ce0600d75b8aa20af) for obviously different TypeReferences being overwritten in the `map`.
    * *Edit 2*: I have added an example (https://github.com/zannen/gqlgen/commit/1a4230c8655f7f8ba66ffec86bce19b5a834826e) to check for regression. I think this is sufficient to check the box.
 - [ ] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content)) - **Note** I read all the docs, didn't see anywhere relevant to note this change.


zannen:
I have no idea why the `plugin/resolvergen` race test is failing on CircleCI.

I ran the test over 2000 times locally and it succeeded every time:

```bash
while true ; do
  date
  go test -count=1 -race github.com/99designs/gqlgen/plugin/resolvergen
done
```

example/issue_896/graphql.schema
```diff
@@ -0,0 +1,14 @@
+schema {
```
vektah: Examples are there for people to understand how gqlgen works.

This should be in `codegen/testserver/issue896.graphql` and `extend schema {`. There should be plenty of examples in that directory.

zannen: Thanks for the pointer, I have moved the schema file and added `extend` as you suggest.

vektah:
Nice one!

Issue #798, opened 2019-07-19
ID! validation is returning an error of type string.


thedodd:
### What happened?
When submitting a request where a literal integer is supplied by the client instead of a string literal for the `ID!` type, the framework's error handling is quite different. The error is presented to the error handler configured by the user, instead of just being flagged as a malformed request and returned as a 400 to the caller automatically.

There appears to be a difference between the type validation of the high-level request object vs the validation of field parameters. This inconsistency complicates error handling, such that it is unclear where in the stack the error is coming from. Perhaps it would be good to use a more identifiable error type so that we can handle it appropriately in the error handler logic.

The question needs to be asked: "is this some unhandled error coming from bad business logic in my code (an ISE), or is this an error which can be presented as-is to the caller?"

### What did you expect?
I expected consistency. This is a type validation error. The gqlgen stack should, arguably, be handling this error during the main validation phase and return an error before it ever leaves the gqlgen stack.

### Minimal graphql.schema and models to reproduce
```graphql
type Query {
    getProject(id: ID!): String!
}
```

Then just query that field with an int instead of a string.

### versions
 - `gqlgen version`? `v0.9.1`
 - `go version`? `1.12.6`
 - dep or go modules? `go modules`

thedodd:
@vektah &| @neelance just wanted to see if either of you have any thoughts on the best way to deal with this. Thanks for all of the work that you've put into this project!

thedodd:
Just adding another heartbeat here (to keep the stale bot reaper from killing this issue).

cfilby:
@thedodd I had a PR that was related to this, #902 - does that resolve your issue?

thedodd:
@cfilby that probably will address the issue. Have those changes been released/tagged? I'll test and report back.

cfilby:
@thedodd it's merged into master but hasn't yet been included in a tagged release. You could test it directly by using the latest master commit instead of a versioned release?

vvakame:
I close this issue. if its still haven't solved, please mention to me.

PR #902, opened 2019-10-16
Add support for int64 IDs

cfilby:
Based on the [GraphQL Spec](https://graphql.github.io/graphql-spec/June2018/#sec-ID), Input ID types should be coerced into strings if provided as an integer during input. In a local GraphQL API we started getting "int64 is not a string" errors when trying to pass an integer in as a resolver ID parameter while using the default graphql.ID type. We were able to recreate this behavior using the gqlgen starwars example project as well. This error does _not_ happen while using graphql.IntID type for marshalling.

This PR adds an int64 case to the MarshalID method, which appears to resolve the issue in my quick smoke test.

Let me know if you want me to update the formatting for the test, since it looks like it diverges from the repo style.

Related Issues:
* #798

I have:
 - [*] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [*] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

### DIFF
--- a/graphql/id.go
+++ b/graphql/id.go
@@ -20,6 +20,8 @@ func UnmarshalID(v interface{}) (string, error) {
 		return string(v), nil
 	case int:
 		return strconv.Itoa(v), nil
+	case int64:
+		return strconv.Itoa((int(v))), nil
 	case float64:
 		return fmt.Sprintf("%f", v), nil
 	case bool:
--- a/graphql/id.go
+++ b/graphql/id.go
@@ -21,7 +21,7 @@ func UnmarshalID(v interface{}) (string, error) {
 	case int:
 		return strconv.Itoa(v), nil
 	case int64:
-		return strconv.Itoa(int(v)), nil
+		return strconv.FormatInt(v, 10), nil
 	case float64:
 		return fmt.Sprintf("%f", v), nil
 	case bool:
--- a/graphql/id.go
+++ b/graphql/id.go
@@ -21,7 +21,7 @@ func UnmarshalID(v interface{}) (string, error) {
 	case int:
 		return strconv.Itoa(v), nil
 	case int64:
-		return strconv.Itoa((int(v))), nil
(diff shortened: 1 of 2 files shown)


graphql/id_test.go
```diff
@@ -0,0 +1,36 @@
...
+		Input       interface{}
+		Expected    string
+		ShouldError bool
+	}{
+		{
+			Name:        "int64",
+			Input:       int64(12),
+			Expected:    "12",
+			ShouldError: false,
+		},
```
vvakame: https://golang.org/pkg/math/#pkg-constants
how about `math.MaxInt64`?

graphql/id.go
```diff
@@ -20,6 +20,8 @@ func UnmarshalID(v interface{}) (string, error) {
 		return string(v), nil
 	case int:
 		return strconv.Itoa(v), nil
+	case int64:
+		return strconv.Itoa(int(v)), nil
```
vvakame: how is https://golang.org/pkg/strconv/#FormatInt ?

cfilby: @vvakame I opted to use strconv.Itoa to be consistent with the UnmarshalIntID function. It looks like Itoa is just an alias to FormatInt(int64(x), 64):
https://golang.org/pkg/strconv/#Itoa

If you want I can update both to use strconv.FormatInt explicitly though.

graphql/id_test.go
cfilby: Added math.MaxInt64 and math.MinInt64 test cases

graphql/id.go
cfilby: Done

frederikhors:
@cfilby is this working using `go run github.com/99designs/gqlgen -v` today?

I still have this message using `int64`: `ID is incompatible with int64`.

PR #919, opened 2019-11-03
Fixes #843: Bind to embedded struct method or field

### LINKED ISSUE #843: Embedded struct methods are not considered
### What happened?
Embedded struct methods are not considered by `gqlgen` when looking at a model. If I'm reading [the docs](https://gqlgen.com/reference/resolvers/#binding-to-anonymous-or-embedded-structs) correctly, this should be allowed.

### What did you expect?
Embedded struct methods be considered by `gqlgen` when looking at a model.

### Minimal `graphql.schema` and models to reproduce
```graphql
type Query {
  todos: [Todo!]!
}

type Todo {
  id: ID!
  name: String!
}
```

```go
// Todo model
type Todo struct {
	*ider
	*namer
}

// // ID resolver
// func (t *Todo) ID(ctx context.Context) (string, error) {
// 	return t.ider.ID(ctx)
// }
//
// // Name resolver
// func (t *Todo) Name(ctx context.Context) (string, error) {
// 	return t.namer.Name(ctx)
// }

type ider struct{}

func (*ider) ID(ctx context.Context) (string, error) {
	return "1", nil
}

type namer struct{}

func (*namer) Name(ctx context.Context) (string, error) {
	return "name", nil
}
```

I've commented out the two methods that make `gqlgen` happy, which are redundant the way I understand go.

You can also clone [this repo](https://github.com/matiasanaya/gqlgen-test/commit/d5a21764b49b265c15ea9fe6c35ffd6969590015) and run:

``` cd embedding && go run server/server.go```

### versions
 - `gqlgen version` v0.9.1 
 - `go version` go1.12.5 darwin/amd64
 - go modules

### Thoughts

Is there a better way of doing what I'm trying to do?

Ideally I'd be able to compose a `struct` out of other structs or interfaces that I use internally and `gqlgen` wouldn't force me to define on the composed object all the methods.

But would it be better if `gqlgen` generated an `interface` instead of a `struct` with fields for every type? That way I don't have to be running `go generate ./...` each time my `struct` changes and would only depend on the `schema.graphql` and the `gqlgen.yml` files.

We are using this in production and I'm wondering if we are using it in the wrong way because I can't seem to understand the use case for this kind of `interface`:

```go
type TodoResolver interface {
	ID(ctx context.Context, obj *Todo) (string, error)
	Name(ctx context.Context, obj *Todo) (string, error)
}
```
I'd much prefer this to be generated as:
```go
type TodoResolver interface {
	ID(ctx context.Context) (string, error)
	Name(ctx context.Context) (string, error)
}
```

Thanks!

matiasanaya:
This is an attempt to write test cases for the buggy `codegen.builder#findBindTarget` method as initially raised in #843 .

Some of the issues:
* It errors (and fails silently) if it finds an embedded exported interface within a struct.
* It won't look into methods from un-exported embedded structs.
* It can't bind to an interface.

I'm not entirely sure what's the expected behaviour in some of this cases since [this docs](https://gqlgen.com/reference/resolvers/), code comments [here](https://github.com/99designs/gqlgen/blob/4c35356cbe7bf886fd8c59c9f754d2d98f6987b8/codegen/field.go#L189) and examples (I'm certain I've seen an example on how to bind to an embedded interface) seem to contradict at times each other.

A way to fix some of these issues would be to use `types.LookupFieldOrMethod`. If there's some alignment on what the spec should be, I'm happy to update this PR with code that solves the issues.

The best way to run these tests is:
```
go generate ./codegen/testserver/embedded/... && go test ./codegen/testserver/embedded/... -count=1
```

I have:

### DIFF
--- a/codegen/field.go
+++ b/codegen/field.go
@@ -184,75 +184,111 @@ func (b *builder) bindField(obj *Object, f *Field) (errret error) {
 	}
 }
 
-// findField attempts to match the name to a struct field with the following
-// priorites:
-// 1. Any method with a matching name
-// 2. Any Fields with a struct tag (see config.StructTag)
-// 3. Any fields with a matching name
-// 4. Same logic again for embedded fields
+// findBindTarget attempts to match the name to a struct field or method
+// with the following priorites:
+// 1. Any Fields with a struct tag (see config.StructTag). Errors if more than one match is found
+// 2. Any method or field with a matching name. Errors if more than one match is found
+// 3. Same logic again for embedded fields
 func (b *builder) findBindTarget(named *types.Named, name string) (types.Object, error) {
+	strukt, isStruct := named.Underlying().(*types.Struct)
+	if isStruct {
+		// NOTE: a struct tag will override both methods and fields
+		// Bind to struct tag
+		found, err := b.findBindStructTagTarget(strukt, name)
+		if found != nil || err != nil {
+			return found, err
+		}
+	}
+
+	// Search for a method to bind to
+	var foundMethod types.Object
 	for i := 0; i < named.NumMethods(); i++ {
 		method := named.Method(i)
-		if !method.Exported() {
+		if !method.Exported() || !strings.EqualFold(method.Name(), name) {
 			continue
 		}
 
-		if !strings.EqualFold(method.Name(), name) {
-			continue
+		if foundMethod != nil {
+			return nil, errors.Errorf("found more than one matching method to bind for %s", name)
 		}
 
-		return method, nil
+		foundMethod = method
 	}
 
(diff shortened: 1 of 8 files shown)


vektah:
Looks good, PR description and title could use an update.

matiasanaya:
@vektah thanks for the review. I've updated the PR title, fixed up the commits and cleaned up the unused code previously committed.

This PR fixes the following to issues:
* `findBindTarget` fails to bind to embedded structs if it finds an embedded exported interface within a struct.
* `findBindTarget` fails to bind to exported fields / methods of un-exported embedded structs.

I'm happy for this to be merged as is.
I'd love to do a followup PR to support embedded interfaces.

Issue #918, opened 2019-11-02
generation performance


vikstrous:
### What happened?

We've been using gqlgen for a long time and it has been taking almost 1.5 mins to generate from scratch now and 30 seconds to re-generate with no changes. We have 5 graphql servers that share a lot of resolvers. In total this is still multiple minutes in the best case. This is the slowest part of our `go generate` process.

The "from scratch" number if after running `go clean -r ./...`.

The machine I'm running on is linux with an SSD and 16 GB ram and 4 cores.

In the from scratch case, it used 600MB memory and in the no changes case, 200MB.

On a less powerful machine, I've given up entirely on running gqlgen.

### What did you expect?
Generation time should be measured in seconds, not minutes.

### Minimal graphql.schema and models to reproduce
Our project is very large and closed source. I can't share too much. I'll share some stats for one of the servers:

* ~100 models listed in gqlgen.yaml
* 1 autobind directory
* 106 types
* 35 enums
* 2 scalars
* 6200 lines of schema
* I'm guessing at least 1000 resolvers based on the number of resolvers per line I'm seeing

### versions
 - gqlgen 0.10.1
 - go 1.13.3
 - go modules

I'm looking for a few things:

* Ideas for how to reduce generation time without code changes on either side
* Ideas for how to reduce generation time with code changes either in gqlgen or our codebase
* Ideas for how to profile things to get more detailed info about what the slow part is. For now I'll try to fork gqlgen and hack in something like https://golang.org/pkg/runtime/pprof/#hdr-Profiling_a_Go_program to profile it.

Let me know if you guys have a chat I can join so we can discuss ideas.

Edit: here's the output from `time -v` for the best case scenario with no changes:
```
        User time (seconds): 76.74
        System time (seconds): 14.24
        Percent of CPU this job got: 275%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:33.06
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 198192
        Average resident set size (kbytes): 0
        Major (requiring I/O) page faults: 1
        Minor (reclaiming a frame) page faults: 869804
        Voluntary context switches: 198697
        Involuntary context switches: 52535
        Swaps: 0
        File system inputs: 0
        File system outputs: 47184
        Socket messages sent: 0
        Socket messages received: 0
        Signals delivered: 0
[...]

PR #926, opened 2019-11-10
various performance optimizations

vikstrous:
This is not a PR that's meant to be merged. It's just a POC of some performance optimizations discussed in https://github.com/99designs/gqlgen/issues/918

### DIFF
--- a/internal/code/imports.go
+++ b/internal/code/imports.go
@@ -14,8 +14,6 @@ import (
 	"golang.org/x/tools/go/packages"
 )
 
-var nameForPackageCache = sync.Map{}
-
 var gopaths []string
 
 func init() {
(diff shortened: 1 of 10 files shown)


vektah:
Nice exploration, good to see someone dig in here.

Sadly, I dont think its going to be this straightforward. The reason for multiple `packages.Load` calls is plugins change the code on disk for a package (eg modelgen).

If we want to go to a single pass we need a way for plugins to invalidate and refetch some packages, this didnt exist in `packages` when the code was written but maybe it does now.

One of the ideas we threw around at the time was using one of the long running language servers, which should handle reloading packages better - eg gopls or guru. But we would need to be careful that `go generate` can still run cleanly inside a docker container where inotify events can be quite when using docker for mac and volume mounts.

internal/code/imports.go
```diff
@@ -92,24 +90,39 @@ func ImportPathForDir(dir string) (res string) {
 
 var modregex = regexp.MustCompile("module (.*)\n")
 
+type NameForPackage struct {
+	cache    *sync.Map
```
vektah: Why did this need to be moved out of a private global?

vikstrous: Technically it didn't need to be moved, but it was weird to me to use global state when we were already creating a struct to hold the packages field.

In addition to the fact that global state is a dangerous anti-pattern, I suspect that this would also have caused issues if I wanted to use the same binary to generate multiple graphql servers.

vektah:
This has been split out into 
https://github.com/99designs/gqlgen/pull/940
https://github.com/99designs/gqlgen/pull/941
https://github.com/99designs/gqlgen/pull/942
https://github.com/99designs/gqlgen/pull/943
https://github.com/99designs/gqlgen/pull/944
https://github.com/99designs/gqlgen/pull/945

PR #927, opened 2019-11-10
Bind to embedded interface

matiasanaya:
This is a follow up PR to #919 that allows `gqlgen` to bind to embedded interfaces. There's a test case that covers this use case.

I have:


vektah:
Awesome work, tested this with some of our code and everything looks good. Theres some nice cleanups in here too!

PR #948, opened 2019-11-29
fix apq status

vvakame:
I don't have good solutions... 🤔 

so, APQ works in CreateOperationContext method.
and handler exec `w.WriteHeader(http.StatusUnprocessableEntity)` when its returns error.
but It's unexpected behavior in client.

I have:


vvakame:
It doesn't seems right to always return `200`... 🤔

vvakame:
We needs some status code modifying process from middleware...
I'll try to implement it.

vektah:
Sorry its taken so long to get back to this, went on holiday :tropical_drink:.

Maybe its best to keep it simple? We dont need to reimplement error presenter, just a way for handlers during early request processing to set the status code right?

what about 
```go
interface HttpStatusCoder {
   error
   HttpStatusCode() int
}
```
?

PR #965, opened 2019-12-24
add StatusCodeFunc for HTTP Transport implementations

vvakame:
revise #948
add method that modify HTTP status code

This PR is works, but maybe not the best.

I want to add `StatusCode(ctx context.Context, resp *graphql.Response) int` but graphql.Transport doesn't limited to HTTP handling.

so, how about below API design?

```
type TransportConfigurator interface {
  AttachTransports (ctx context, transports []graphql.Transport)
}
```
and
```
func (a AutomaticPersistedQuery) AttachTransports (ctx context, transports []graphql.Transport) {
  for _, transport := range transports {
    httpTransport, ok := transport.(interface { HookStatusCode(f func(ctx context.Context, resp *graphql.Response) int) })
    if !ok { continue }
    httpTransport.HookStatusCode(func(ctx context.Context, resp *graphql.Response) int {
      if len(resp.Errors) == 1 && resp.Errors[0].Message == "PersistedQueryNotFound" {
        return http.StatusOK
      }
      return 0
    })
  }
}
```

I have:


vektah:
Alternative https://github.com/99designs/gqlgen/pull/978, what do you think?

Even if we dont take it we should keep the integration tests.

PR #967, opened 2019-12-29
docs: typo in code on the getting start page

m-bal:
The package names in the files weren't the same as the name of the go module. The Go Module at creation was named github.com/[username]/gqlgen-todos but the todo.go and resolver.go files named the package gqlgen_todos.

### DIFF
--- a/docs/content/getting-started.md
+++ b/docs/content/getting-started.md
@@ -81,7 +81,7 @@ This has created an empty skeleton with all files you need:
 The generated model for Todo isn't right, it has a user embeded in it but we only want to fetch it if the user actually requested it. So instead lets make a new model in `todo.go`:
 
 ```go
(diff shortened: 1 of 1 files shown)


docs/content/getting-started.md
```diff
@@ -81,7 +81,7 @@ This has created an empty skeleton with all files you need:
 The generated model for Todo isn't right, it has a user embeded in it but we only want to fetch it if the user actually requested it. So instead lets make a new model in `todo.go`:
 
 ```go
-package gqlgen_todos
+package gqlgen-todos
```
vektah: package names in go cant have -, they must be valid identifiers. 

import paths can, but the package name cant. Maybe this is a little confusing and we should avoid it all together in the getting started docs.

PR #969, opened 2020-01-02
Optimizes CollectAllFields function 

qhenkart:

I have:
 - [x] Added tests covering the bug / feature (see [testing] (https://github.com/99designs/gqlgen/blob/master/TESTING.md))

I was working on optimizing my gqlgen implementation, and when looking into using the CollectAllFields method. I noticed there was an unnecessary embedded loop. This PR simply uses a map for constant time lookup of unique entries instead of relooping through the results list for every index.

### DIFF
--- a/docs/content/reference/field-collection.md
+++ b/docs/content/reference/field-collection.md
@@ -54,4 +54,4 @@ The type `Circle` would satisfy `Circle`, `Shape`, and `Shapes` — these values
 
 > Note
 >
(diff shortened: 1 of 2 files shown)


qhenkart:
it seems circle has some intermittent issues running tests. But the failure in the circleci:cover is unrelated to this PR

vektah:
Would be good to benchmark this:
 - both of those lists should be pretty small, n < 100 max
 - maps are heap allocated, gc pressure causes real performance issues
 - this function is called everywhere

vektah:
Feel free to reopen with a benchmark

PR #978, opened 2020-01-07
Allow customizing http and websocket status codes for errors

vektah:
This PR sets a general rule that:
 - parse and validation errors are 422s
 - resolver errors are 200s
 - anything else is a 200.

same applies to websockets, except they use GQL_ERROR and GQL_DATA message types instead of status codes.

But we need a way to distinguish between the types of errors. Apollo sets a "code" extension, lets copy them and match their codes. I've included the two core ones in the `errcode` package, you can return anything and they will be treated as 200s by default. To register a custom 422:
```go
func init()  {
   errcode.Register("MY_CODE", errcode.KindProtocol) // make this code a 422 
}
// when you return an error:
errcode.Set(err, "MY_CODE")
```

I've also added integration tests for websockets and fixes a bunch of websocket related issues, mostly around error handling.

closes https://github.com/99designs/gqlgen/pull/948
closes https://github.com/99designs/gqlgen/pull/965

I have:


graphql/handler/transport/websocket.go
```diff
@@ -126,6 +127,7 @@ func (c *wsConnection) init() bool {
 }
 
 func (c *wsConnection) write(msg *operationMessage) {
+	//fmt.Println("write", msg.ID, msg.Type, string(msg.Payload))
```
vvakame: 👀

graphql/handler/transport/websocket.go
```diff
@@ -284,6 +297,8 @@ func (c *wsConnection) readOp() *operationMessage {
 		return nil
 	}
 
+	//fmt.Println("read", message.ID, message.Type, string(message.Payload))
```
vvakame: 👀

graphql/handler/transport/websocket.go
```diff
@@ -292,4 +307,5 @@ func (c *wsConnection) close(closeCode int, message string) {
 	_ = c.conn.WriteMessage(websocket.CloseMessage, websocket.FormatCloseMessage(closeCode, message))
 	c.mu.Unlock()
 	_ = c.conn.Close()
+	//fmt.Println("CLOSE", closeCode, message)
```
vvakame: 👀

graphql/handler/transport/websocket_test.go
```diff
@@ -164,7 +164,7 @@ func TestWebsocketWithKeepAlive(t *testing.T) {
 	msg := readOp(c)
 	assert.Equal(t, connectionKeepAliveMsg, msg.Type)
 
-	// server message
+	// server messageapollotracing_test
```
vvakame: is this change right?

integration/readme.md
```diff
@@ -0,0 +1,15 @@
...
+
+First start the go server
+```bash
+go run integration/server/server.go
+```
+
+And in another terminal:
+```bash
+cd integration
+npm install
```
vvakame:
```suggestion
npm ci
```
https://docs.npmjs.com/cli/ci
npm ci is better than install.
it doesn't update package-lock.json unexpectedly.

graphql/errcode/codes.go
```diff
@@ -0,0 +1,49 @@
...
+	codeType[code] = kind
+}
+
+// Set the error code on a given graphql error extension
+func Set(err *gqlerror.Error, value string) {
+	if err.Extensions == nil {
+		err.Extensions = map[string]interface{}{}
+	}
+
+	err.Extensions["code"] = value
```
vvakame: I think `"code"` is little bit commonly used in user namespace.
shoud we have an API to change property name?

vektah: It matches the purpose and name of the apollo [code](https://www.apollographql.com/docs/apollo-server/data/errors/#codes) so I dont think we should have any issues here.

graphql/handler/transport/websocket.go
vektah: I was going to leave them in, I end up adding them every time I need to debug websockets.

I didnt want to go down the path of adding low overhead configurable logging. what do you think?

PR #982, opened 2020-01-07
fix: explicitly exclude trailing comma from link

DavidJFelix:
- this looks dumb, but when the page is rendered, the link resolves with the comma, despite the comma being excluded in github rendering.


DavidJFelix:
image

Inspect on current page

PR #983, opened 2020-01-09
single packages.Load for NameForPackage

vikstrous:
I simplified this PR to make the minimum necessary changes to get the job done. This PR supersedes #944 
Unfortunately, I had to use global state because the existing code was using global state. I'm willing to refactor this in a follow up, ending up with something similar to #944 if that makes more sense.

I didn't add any new tests because exists tests already cover this change.

Non-scientific test comparing master vs this PR on our proprietary codebase:

Before:
```
40.92user 23.75system 0:27.99elapsed 231%CPU (0avgtext+0avgdata 197584maxresident)k
0inputs+0outputs (1major+430432minor)pagefaults 0swaps
```
After:
```
21.46user 8.37system 0:13.77elapsed 216%CPU (0avgtext+0avgdata 201152maxresident)k
0inputs+0outputs (1major+160637minor)pagefaults 0swaps
```
This is done on a pretty weak machine with somewhat slow IO. I used the second results for each after running the tests twice.

I have:
 - n/a Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - n/a Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

### DIFF
--- a/codegen/data.go
+++ b/codegen/data.go
@@ -6,9 +6,11 @@ import (
 	"sort"
 
 	"github.com/99designs/gqlgen/codegen/config"
+	"github.com/99designs/gqlgen/internal/code"
 	"github.com/pkg/errors"
 	"github.com/vektah/gqlparser/ast"
 	"github.com/vektah/gqlparser/formatter"
+	"golang.org/x/tools/go/packages"
 )
 
 // Data is a unified model of the code to be generated. Plugins may modify this structure to do things like implement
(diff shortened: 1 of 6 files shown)


vektah:
Thanks, this looks a bit cleaner. Shame about the global but we can iterate. I think the other PR had issues with multiple levels of caching.

PR #985, opened 2020-01-10
Correctly generate a federated schema when no entity has a `@key`.

csilvers:
Normally, when a service is taking part in graphql federation, it will
have at least one type defined with a `@key` field, so that other
services can link to (that is, have an edge pointing to) the type that
this service provides.  The previous federation code assumed that was
the case.

But it's not required: a service could not define `@key` on any of its
types.  It might seem that would mean the service is unreachable,
since there is no possibility of edges into the service, but there are
two edges that can exist even without a `@key`: top level Query edges
and top level Mutation edges.  That is, if a service only provides a
top-level query or top-level mutation, it might not need to define a
`@key`.

This commit updates the federation code to support that use case.

I have:

### DIFF
--- a/plugin/federation/federation.go
+++ b/plugin/federation/federation.go
@@ -56,9 +56,11 @@ func (f *federation) MutateConfig(cfg *config.Config) error {
 		"_Any": {
 			Model: config.StringList{"github.com/99designs/gqlgen/graphql.Map"},
 		},
-		"Entity": {
+	}
+	if len(entityFields) > 0 {
+		builtins["Entity"] = config.TypeMapEntry{
 			Fields: entityFields,
-		},
+		}
 	}
 	for typeName, entry := range builtins {
 		if cfg.Models.Exists(typeName) {
@@ -79,7 +81,14 @@ func (f *federation) MutateConfig(cfg *config.Config) error {
 // the fields that had the @key directive
 func (f *federation) InjectSources(cfg *config.Config) {
 	cfg.AdditionalSources = append(cfg.AdditionalSources, f.getSource(false))
+
 	f.setEntities(cfg)
+	if len(f.Entities) == 0 {
+		// It's unusual for a service not to have any entities, but
+		// possible if it only exports top-level queries and mutations.
+		return
+	}
+
 	s := "type Entity {\n"
 	for _, e := range f.Entities {
 		s += fmt.Sprintf("\t%s(%s: %s): %s!\n", e.ResolverName, e.Field.Name, e.Field.Type.String(), e.Def.Name)
@@ -88,9 +97,9 @@ func (f *federation) InjectSources(cfg *config.Config) {
 	cfg.AdditionalSources = append(cfg.AdditionalSources, &ast.Source{Name: "entity.graphql", Input: s, BuiltIn: true})
 }
 
-// MutateSchema creates types and query declarations
-// that are required by the federation spec.
-func (f *federation) MutateSchema(s *ast.Schema) error {
+// addEntityToSchema adds the _Entity Union and _entities query to schema.
+// This is part of MutateSchema.
+func (f *federation) addEntityToSchema(s *ast.Schema) {
 	// --- Set _Entity Union ---
 	union := &ast.Definition{
 		Name:        "_Entity",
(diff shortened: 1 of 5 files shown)


csilvers:
I don't know what the protocol is for marking people who I think would be good to review this, so I'll just do it as a comment: @marwan-at-work I think you're the best one to review this!

plugin/federation/federation.go
```diff
@@ -56,9 +56,11 @@ func (f *federation) MutateConfig(cfg *config.Config) error {
 		"_Any": {
 			Model: config.StringList{"github.com/99designs/gqlgen/graphql.Map"},
 		},
-		"Entity": {
+	}
+	if len(entityFields) != 0 {
```
ravisastryk: Can you consider changing it as follows?
`
if len(entityFields) > 0
`

csilvers: Sure!

csilvers:
(I maybe did the wrong thing by amending rather than pushing a new commit, but this is ready for another look!)

csilvers:
Anything else I need to do to help get this merged?  I don't understand the 'cover' failure -- it's in an area unrelated to my change, and the diff seems non-contentful to me.  Hopefully that's not the thing holding things up, since I don't know what to do about it...

marwan-at-work:
I believe @vektah could help get this merged :) 

As for `cover`, I think you have a failing test. Does `go test ./...`  succeed locally?

csilvers:
I do have a failing test locally:
```
--- FAIL: TestNameForDir (0.24s)
    imports_test.go:45: 
                Error Trace:    imports_test.go:45
                Error:          Not equal: 
                                expected: "tmp"
                                actual  : "linewrap"
                                
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -tmp
                                +linewrap
                Test:           TestNameForDir
```
This differs from the failing test in circleci:
```
--- FAIL: TestPrune (0.04s)
    require.go:157: 
        	Error Trace:	prune_test.go:13
        	Error:      	Not equal: 
        	            	expected: "package testdata\n\nimport _ \"underscore\"\nimport a \"fmt\"\nimport \"time\"\n\ntype foo struct {\n\tTime time.Time `json:\"text\"`\n}\n\nfunc fn() {\n\ta.Println(\"hello\")\n}\n\ntype Message struct {\n\tID        string    `json:\"id\"`\n\tText      string    `json:\"text\"`\n\tCreatedBy string    `json:\"createdBy\"`\n\tCreatedAt time.Time `json:\"createdAt\"`\n}\n"
        	            	actual  : "package testdata\n\nimport (\n\ta \"fmt\"\n\t\"time\"\n\t_ \"underscore\"\n)\n\ntype foo struct {\n\tTime time.Time `json:\"text\"`\n}\n\nfunc fn() {\n\ta.Println(\"hello\")\n}\n\ntype Message struct {\n\tID        string    `json:\"id\"`\n\tText      string    `json:\"text\"`\n\tCreatedBy string    `json:\"createdBy\"`\n\tCreatedAt time.Time `json:\"createdAt\"`\n}\n"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -2,5 +2,7 @@
        	            	 
        	            	-import _ "underscore"
        	            	-import a "fmt"
        	            	-import "time"
        	            	+import (
        	            	+	a "fmt"
        	            	+	"time"
        	            	+	_ "underscore"
        	            	+)
        	            	 
        	Test:       	TestPrune
FAIL
```

Neither of them should be affected by any code I changed.

csilvers:
I merged and the tests are still failing.  I really believe the test failures are not due to my code, but some sort of flakiness.  This time it's this:
```
/usr/local/go/pkg/tool/linux_amd64/link: signal: killed
FAIL	github.com/99designs/gqlgen/client [build failed]
```
I don't know what is causing the linker to be killed -- OOM? -- but it looks like that's what's causing the test failure.

I don't know if it's possible to just force-rerun the tests, or if it even matters -- I guess we can merge even with the test failure, if we're confident this PR isn't responsible.

vvakame:
I re-run the tests manually, its passed.
https://stackoverflow.com/questions/52129612/golang-circleci-2-0-test-failing-with-signal-killed
This is a problem I haven't seen much before...
so, test is passed. and LGTM.
thanks to your contributions. and thanks to your review @marwan-at-work !

PR #993, opened 2020-01-20
Resolver regenerator

vektah:
This PR adds a new way to generate the resolver scaffolding that can actually run on generate:
 - theres a central `resolver.go` that has the root type in it, this is only generated once and is meant to be configured by the user to add dependencies etc.
 - each input graphql file gets a `{filename}_resolvers.go`
 - every time generate is run the code from the old resolvers are copied through to the new bodies
 - any code left over will be pushed to the bottom of the file, so if you are renaming or deleting things we wont delete any user code.

A tiny example of it in action (+ intellij file watcher triggering go generate on graphql file changes)
example

Reimplementation of https://github.com/99designs/gqlgen/pull/323


neelofer-appdirect:
Thank you. When will this be released for use?

shaofu88:
Is it possible to separate resolvers into different files by type (instead of file)?

PR #995, opened 2020-01-20
Remove empty lines on DirectiveRoot generation

alexsn:
Signed-off-by: Alex Snast <alexsn@fb.com>

DirectiveRoot methods are spaced with an empty line which is unnecessary.


vvakame:
thanks!

PR #998, opened 2020-01-29
Add package config for federation file

josemarluedke:
When generating a project with the federation, the federation service file is always on the root of the project. To allow using federation in another project layout, we must have a way to customize where the federation file should be located. 

This PR adds the option to configure the federation package, ideally, it should be part of the Exec config, but that is no clear path on adding that there.

I have:
 - [ ] Added tests covering the bug / feature
-> I could not figure out how to test this. 
 - [ ] Updated any relevant documentation
-> There is no documentation for federation

### DIFF
--- a/codegen/config/config.go
+++ b/codegen/config/config.go
@@ -24,6 +24,7 @@ type Config struct {
 	Exec                     PackageConfig              `yaml:"exec"`
 	Model                    PackageConfig              `yaml:"model,omitempty"`
 	Resolver                 PackageConfig              `yaml:"resolver,omitempty"`
+	Federation               PackageConfig              `yaml:"federation,omitempty"`
 	AutoBind                 []string                   `yaml:"autobind"`
 	Models                   TypeMap                    `yaml:"models,omitempty"`
 	StructTag                string                     `yaml:"struct_tag,omitempty"`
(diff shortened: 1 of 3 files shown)


vektah:
> This PR adds the option to configure the federation package, ideally, it should be part of the Exec config, but that is no clear path on adding that there.

Care to elaborate on that? Is it because the generated apollo codegen code must be in the same package?

josemarluedke:
Yeah, it's because the code it generates assumes it's running next to the generated.go (Exec).

vektah:
I think this is the right approach regardless, as we may change it so this can generate in its own package later. It feels cleaner giving federation its own config rather than lumping it in with the server runtime. 

I've just landed a couple of big prs on master cleaning up config loading a bit, so this is going to need a rebase. Pay extra attention to the validation now happening in check()
https://github.com/99designs/gqlgen/blob/8ed2ec599b8faed3751177fd4335b1b3c3a79922/codegen/config/config.go#L323-L332

it handles when two different sections try to declare the different package names for the same directory (eg exec and federation).
We should also add a check for when federation isn't in the same directory as exec at all, and let the user know this configuration isn't (currently) supported.

vektah:
Sorry, I've added the required validation and tests to https://github.com/99designs/gqlgen/pull/1012

PR #999, opened 2020-01-30
skip searching directives when entity key is found

ravisastryk:
skip searching directives when entity key is found


vektah:
What are you trying to fix?

ravisastryk:
> What are you trying to fix?
The existing code checks if an entity has been found. However, once it is found there is no need to continue further. This fix is to just skip once  the directive is found. Notice `break` statement when looking up in Directives. Does this make sense?
https://github.com/99designs/gqlgen/pull/999/files#diff-777f78d80bba845e7f5493310d21fce6R109

plugin/modelgen/models.go
```diff
@@ -96,16 +96,17 @@ func (m *Plugin) MutateConfig(cfg *config.Config) error {
...
 
-	var hasEntity bool
+	hasEntity := false
 	for _, schemaType := range schema.Types {
 		if cfg.Models.UserDefined(schemaType.Name) {
 			continue
 		}
-		var ent bool
+		ent := false
 		for _, dir := range schemaType.Directives {
```
vektah: Using `ForName` is probably clearer
```go
if schemaType.Directives.ForName("key") != nil {
    hasEntity = true
    ent = true
}
```

PR #1004, opened 2020-02-04
Conditionally removing fields from introspection

### LINKED ISSUE #380: Conditionally removing fields from introspection based on auth

andcan:
Fixes #380 
## How to use
Use introspection plugin with `AllowFieldFunc` and `AllowInputValueFunc`.
## How it works
`Introspection` extension now implements `FieldInterceptor`.
When `InterceptField` is called checks if the result returned by `next()` is 
`[]introspection.Field` or `[]introspection.InputValue`.
If any of the two conditions is true proceeds to filter field using the correct `AllowFunc` specified when creating the extension.

I have:

### DIFF
--- a/graphql/handler/extension/introspection.go
+++ b/graphql/handler/extension/introspection.go
@@ -39,11 +39,11 @@ func (c Introspection) InterceptField(ctx context.Context, next graphql.Resolver
 	res, err = next(ctx)
 
 	fc := graphql.GetFieldContext(ctx)
-	t := fc.Parent.Result.(*introspection.Type)
 	if fields, ok := res.([]introspection.Field); ok {
 		if c.AllowFieldFunc == nil {
 			return
 		}
+		t := fc.Parent.Result.(*introspection.Type)
 		var newFields []introspection.Field
 		for _, field := range fields {
 			allow, err := c.AllowFieldFunc(ctx, t, &field)
@@ -59,6 +59,7 @@ func (c Introspection) InterceptField(ctx context.Context, next graphql.Resolver
 		if c.AllowInputValueFunc == nil {
 			return
 		}
+		t := fc.Parent.Result.(*introspection.Type)
 		var newFields []introspection.InputValue
 		for _, field := range fields {
 			allow, err := c.AllowInputValueFunc(ctx, t, &field)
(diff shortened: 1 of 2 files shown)


andcan:
Still work in progress breaks if FieldContext.Parent.Result is not an *introspection.Type

graphql/handler/extension/introspection.go
```diff
@@ -27,3 +34,43 @@ func (c Introspection) MutateOperationContext(ctx context.Context, rc *graphql.O
...
+			if allow {
+				newFields = append(newFields, field)
+			}
+		}
+		res = newFields
+	} else if fields, ok := res.([]introspection.InputValue); ok {
+		if c.AllowInputValueFunc == nil {
+			return
+		}
+		t := fc.Parent.Result.(*introspection.Type)
```
vvakame: `intropection.Type`, `intropection.Directive` and `introspection.Field` will be coming.
https://spec.graphql.org/June2018/#sec-Schema-Introspection `__InputValue`

andcan: I was thinking to change it to something like
```
type Introspection struct {
	Auth IntrospectionAuth
}

type IntrospectionAuth interface {
	AllowEnumValues(ctx context.Context, enumValues []introspection.EnumValue) ([]introspection.EnumValue, error)

	AllowFields(ctx context.Context, fields []introspection.Field) ([]introspection.Field, error)

	AllowInputValues(ctx context.Context, inputValues []introspection.InputValue) ([]introspection.InputValue, error)
}
```
and the allow func would be something like:
```
AllowFieldsFunc: func(ctx context.Context, fields []introspection.Field) ([]introspection.Field, error) {
						fc := graphql.GetFieldContext(ctx)
						newFields := make([]introspection.Field, 0, len(fields))
						if t, ok := fc.Parent.Result.(*introspection.Type); ok && *t.Name() == "TestType1" {
							for _, field := range fields {
								if field.Name == "TestField1" {
									continue
								}
								newFields = append(newFields, field)
							}
							return newFields, nil
						}
						return fields, nil
					},
```
This way the allow func is more verbose, but is called just one time and there is no need to have one function for each of intropection.Type, intropection.Directive and introspection.Field.
What do you think?

PR #1006, opened 2020-02-04
skip searching directives when entity is found

ravisastryk:
* skip searching directives when entity key is found.
* The existing code checks if an entity has been found. However, once it is found there is no need to continue further. This fix is to just skip once the directive is found.


ravisastryk:
I had merge conflicts and needed to close the previous ones. Please review and merge @vektah

vektah:
Nice one

PR #1010, opened 2020-02-05
Make sure there's a Query node before trying to add a field to it.

csilvers:
Federation adds some queries to the schema.  There already existed
code to insert a Query node if none existed previously.  But that code
was only put on addEntityToSchema(), and not the other place we update
the query, addServiceToSchema().

Almost always the old code was good enough, since we call
addEntityToSchema() before addServiceToSchema().  But there's on
exception: when the schema has no `@key`.  In that case we only call
addServiceToSchema(), so we need to do the query-existence check there
too.

I have:

### DIFF
--- a/plugin/federation/federation.go
+++ b/plugin/federation/federation.go
@@ -97,6 +97,17 @@ func (f *federation) InjectSources(cfg *config.Config) {
 	cfg.AdditionalSources = append(cfg.AdditionalSources, &ast.Source{Name: "entity.graphql", Input: s, BuiltIn: true})
 }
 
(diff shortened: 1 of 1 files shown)


csilvers:
@marwan-at-work -- this is appropriate for you to review I think.

This fixes a bug from my last PR (that allowed code to work without a `@key` field).  It fixes the even rarer case that there's no `@key` field *and* there's no `Query`.

csilvers:
(Also, I can't figure out a good way to test MutateSchema -- if you know of one, let me know!)

vektah:
Sweet, good to see this getting some polish before 0.11 :+1:

PR #1011, opened 2020-02-05
Compound key support in federation

lizfaubell:
Compound keys aren't currently supported in the federation plugin, but are allowed per the Apollo spec. This patch adds support by modifying the federation plugin to handle a list of key fields on an entity rather than a single top-level key field. It will now look for "findEntityNameByKeyField1AndKeyField2..." in the resolver, rather than the original "FindEntityNameByKeyField". The change does not support more complicated FieldSets in the key, such as nested selections.

We are using this internally with Go services that use compound keys in federation.

References:
- Apollo federation spec: https://www.apollographql.com/docs/apollo-server/federation/federation-spec/

I have:
 - [ ] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content)) - I didn't find anything docs specific to federation.

### DIFF
--- a/plugin/federation/federation.go
+++ b/plugin/federation/federation.go
@@ -91,7 +91,11 @@ func (f *federation) InjectSources(cfg *config.Config) {
 
 	s := "type Entity {\n"
 	for _, e := range f.Entities {
-		s += fmt.Sprintf("\t%s(%s: %s): %s!\n", e.ResolverName, e.Field.Name, e.Field.Type.String(), e.Def.Name)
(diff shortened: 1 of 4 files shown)


vektah:
Nice, another incremental. Any thoughts @marwan-at-work ?

marwan-at-work:
Looks awesome! I'll clone the branch and give it a try as soon as I can. 

Feel free to merge in the mean time if it looks good to everyone 👍

PR #1012, opened 2020-02-06
Allow configuring the federation output file location

vektah:
I'm going through writing some docs on federation and this is the first bit I hit. How do I turn it on, and where does it output to.

:eyes: @marwan-at-work  @csilvers @lizfaubell This is removing the `federated` property and replacing it with new top level package config in the form:
```yaml
federation:
  filename: graph/generated/federation.go
```

Theres an error message that will say pretty much exactly that that. Hope this isn't too much of a break, but I want to get it in before we release.

replaces https://github.com/99designs/gqlgen/pull/998

I have:


dkapadia:
Looks good to me (as someone who had a similar PR up on marwan's branch) - I like that we'll just have a single configuration point to enable federation as well as configure the location

PR #1015, opened 2020-02-06
Federation cleanup

vektah:
Still a work in progress.

A bit of a cleanup getting ready for release:
 - Remove the LoadSchema call happening in the federation plugin by splitting inject schema in to two phases, one before load schema for adding directives etc, and one after which has all of the user schema loaded but may be still missing some from other plugins.
 - Stop running the ast printer over the sources. This breaks line counts and file attribution, and exposes bugs to people not using federation eg https://github.com/vektah/gqlparser/pull/115
 - Moves the _Entity implementation into `fedruntime` package, so its no longer private. This means models dont need to be in the same package as federation.

I think this is pretty close, but I'm working on an example/docs for federation to test against but am hitting a few issues.


vektah:
I slept on this, I dont think I like it. I can see cases where mutating the schema is going to be required - like automatic relay connection generation.

Will break this into smaller parts.

PR #1018, opened 2020-02-08
Federation docs and examples

vektah:
Adds an example implementing the [federation introduction](https://www.apollographql.com/docs/apollo-server/federation/introduction/) server, a short recipe on how it was created and an end to end test to make sure federation doesn't get completely broken.

@marwan-at-work 

I have:


vektah:
I don't want to split the examples out into their own repo because its really hard to keep external resources in sync. Look how well it turned out for https://github.com/vektah/gqlgen-tutorials :(

In this case though, the go mod changes aren't coming from the example, but from a bit of debugging middleware I needed to debug an issue.  
image

**side note:** Federation gateway is pretty dumb hey?

That said, these new deps are pretty tiny and shouldn't cause any issues. Its a shame go mod tidy doesn't prune test/unused deps itself. At least go doesn't build these packages, and the downloads are fast (via proxy)

PR #1019, opened 2020-02-08
Remove source reprinting

vektah:
Removes the `MutateSchema` plugin hook, the only schema changes we need right now are additive. This means:
 - The embedded source in generated.go keeps correct line/file/position info for error messages.
 - Federation code can be simplified a little
 - We can now use `extend type Foo` to extend remote federated types, rather than `@extends`
 - sdl method now echos back the real user schema, rather than a synthesised one

We will likely need to bring MutateSchema back at some point to handle relay Connections, but lets cross that bridge when we get to it.

@marwan-at-work 

Re-up of https://github.com/99designs/gqlgen/pull/1015


codegen/config/config.go
```diff
@@ -138,6 +138,19 @@ func LoadConfig(filename string) (*Config, error) {
 		}
 	}
 
+	for _, filename := range config.SchemaFilename {
+		filename = filepath.ToSlash(filename)
+		var err error
+		var schemaRaw []byte
+		schemaRaw, err = ioutil.ReadFile(filename)
+		if err != nil {
+			fmt.Fprintln(os.Stderr, "unable to open schema: "+err.Error())
```
marwan-at-work: why not return an error like the rest of the error blocks within the function?

vektah: Just copy paste, thanks :)

PR #1021, opened 2020-02-09
WIP: Update Example Tracing to exclude trivial resolvers

cliedeman:

I have:

### DIFF
--- a/graphql/handler/apollotracing/tracer.go
+++ b/graphql/handler/apollotracing/tracer.go
@@ -2,6 +2,7 @@ package apollotracing
 
 import (
 	"context"
+	"github.com/vektah/gqlparser/ast"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


graphql/handler/apollotracing/tracer.go
```diff
@@ -60,28 +61,43 @@ func (a Tracer) InterceptField(ctx context.Context, next graphql.Resolver) (res
...
+				StartOffset: start.Sub(rc.Stats.OperationStart),
+				Duration:    end.Sub(start),
+			})
+		}()
+	}
 
 	return next(ctx)
 }
 
+func isScalar(definition ast.FieldDefinition) bool {
```
cliedeman: Is there any easy to check if a field is a scalar?

vektah: Not easily, you would need to store the `schema` passed into validate and lookup the named type from `graphql.GetFieldContext(ctx).Field.Definition.Type.Name()` in schema.Types.

But, you probably don't want to do that, scalars have no bearing on how hard something is to calculate. Take for example a `totalBalance` on an `Account` type that needs to go sum all the rows in the account.

OTOH there are Objects which are really cheap to calculate, like the User which can be pulled from context and costs nothing.

vektah:
Keep in mind that this isnt just some random example tracing, this is the [apollo tracing spec](https://github.com/apollographql/apollo-tracing), if we start dropping things from it we should be certain apollo will handle it.

PR #1027, opened 2020-02-13
propagate resolver errors to response error in ResponseMiddleware

sonatard:
We want to get errors in ResponseMiddleware with Response.Errors

- issue

```go
func MyResponseMiddleware(ctx context.Context, next graphql.ResponseHandler) *graphql.Response {
	resp := next(ctx) // resp.Errors is empty when resolver return err
	
	return resp
}
```

I have:


vektah:
Nice!

PR #1038, opened 2020-02-17
check slice length

vvakame:
I tried New resolver layout, but I got below error.
```
panic: runtime error: index out of range [0] with length 0
```
maybe my usage is wrong. I wanna get some error message.

I have:


vektah:
This could use a test, did you find out what the cause was?

vvakame:
🤔 this problem occured when *.go output directory is not exists. maybe...?

PR #1039, opened 2020-02-17
Fix link to examples directory in Federation docs

VitaliiLakusta:
Fix link to examples directory in doc https://gqlgen.com/recipes/federation/


vektah:
Thanks

Issue #1040, opened 2020-02-17
0.11.0 Cannot use type *Server as type http.HandlerFunc


frederikhors:
### What happened?

Upgrading from `v0.10.2` to `v0.11.0` and using https://github.com/go-chi/chi I can't use this code anymore:

```go
r := chi.NewRouter()
gqlgenServer := handler.GraphQL(graphql.NewExecutableSchema(graphql.Config{}}),config.GraphqlHandlerOptions)
r.Post("/query", gqlgenServer)
```

because of:

`Cannot use 'gqlgenServer' (type *Server) as type http.HandlerFunc`

### What did you expect?

I expected to be able to use it like this again. Or how?

### versions
 - `gqlgen version`? 0.11.0
 - `go version`? 1.13
 - dep or go modules? mod

vektah:
*Server is a http.Handler, not a http.HandlerFunc.

Use 
```
r.Handle("/query", gqlgenServer)
```
or
```
r.Post("/query", gqlgenServer.ServeHTTP)
```

PR #1050, opened 2020-02-19
Executor

technoweenie:
This is a first attempt at extracting an executor from `graphql/handler`, as discussed in #1040. 

I started by creating a new `graphql/executor.Executor` type, with all the necessary bits from `graphql/handler.Server`:

* extension
* errorPresenter
* recoverFunc
* queryCache

I also added `(*Executor) Exec(context.Context, *graphql.RawParams) *graphql.Response`, which is ripped from `graphql/handler/transport.Post`, minus all the HTTP stuff. I then removed the `graphql/handler.executor` private type in favor of the new `graphql/executor.Executor` type. All tests pass.

I temporarily had a `graphql/executor.ExtensionList` type to implement `Use()`, `AroundFields()`, etc. But it didn't seem necessary when I replaced the internal `executor` type. 

NOTE: This does remove `OperationFunc`, `ResponseFunc`, and `FieldFunc` from the `graphql/handler` package. They didn't seem necessary since they were only used inside the `Around*` funcs on `graphql/handler.Server`.


graphql/executor/extensions.go
```diff
@@ -0,0 +1,147 @@
...
+func (e *Executor) AroundOperations(f graphql.OperationMiddleware) {
+	e.Use(OperationFunc(f))
+}
+
+// AroundResponses is a convenience method for creating an extension that only implements response middleware
+func (e *Executor) AroundResponses(f graphql.ResponseMiddleware) {
+	e.Use(ResponseFunc(f))
+}
+
+func (e *Executor) setExtensions() {
```
vektah: `setExtensions` is called multiple times, once for each `Use()`. It used to throw the whole executor away and recreate it, but now its reusing it so we need to be careful to zero out all the props.

`operationParameterMutators` and `operationContextMutators` aren't being reset, so if you were to use an extension with it, they would get registered multiple times - once for each other extension thats added after it - and called multiple times at runtime.

vektah:
This looks great. I like that server is now back to one responsibility (+ wrapping, maybe we can embed exec instead?)

> NOTE: This does remove OperationFunc, ResponseFunc, and FieldFunc from the graphql/handler package. They didn't seem necessary since they were only used inside the Around* funcs on graphql/handler.Server.

We should probably mark those as private and update the names to reflect they are just for `Around...`

graphql/executor/executor.go
```diff
@@ -11,115 +11,58 @@ import (
...
+	now := graphql.Now()
+	params.ReadTime = graphql.TraceTiming{
+		Start: now,
+		End:   now,
+	}
 
-			return resp
-		}
-	})
+	rc, err := e.CreateOperationContext(ctx, params)
```
vektah: This already happens in `DispatchOperation`, In fact, If `GetStartTime()` fell back to Now() then Exec could be removed again in favour of calling `DispatchOperation` directly?

I don't like widening this interface for a fairly minor convenience method, and one that wont support subscriptions.

graphql/executor/extensions.go
technoweenie: Ah, the difference is that the old server recreates the private executor (in the [Server constructor](https://github.com/99designs/gqlgen/blob/b165568cce7d5027ba1ee2da68dad3012b12d189/graphql/handler/server.go#L37) and [in `Use()`](https://github.com/99designs/gqlgen/blob/b165568cce7d5027ba1ee2da68dad3012b12d189/graphql/handler/server.go#L78-L90)), ensuring those values are reset properly.

I'll see if I can replicate this behavior in tests and then fix it in this PR.

Thanks for your patience. Judging by your comments, it looks like this PR will take a bit of time to get right :)

technoweenie:
Can someone with push access rerun the cover test? Looks like an intermittent network issue.

graphql/executor/executor.go
technoweenie: I ended up just removing the convenience method since it was only used in tests.

technoweenie:
Last comment before I get more feedback :) I brought back the old around func types (https://github.com/99designs/gqlgen/pull/1050/commits/a84d657791b8f11a31b5af768d429b662faca312) in case any users are using them directly. That's technically breaking the Go package, and I don't feel comfortable as a new contributor making that kind of change.

vektah:
> That's technically breaking the Go package, and I don't feel comfortable as a new contributor making that kind of change.

To be clear the `Around*` methods themselves are 100% required, but the types they use shouldnt need to be exported. They were only just added in 0.11 and arent documented, I would hope people arent using them yet. Even if they are, its a simple break and we aren't 1.0 yet.

ddevault:
Would you consider adding a page under `docs/reference/` about this?

ruskovweb:
In addition to this it would be very useful if you export the `parseQuery` function. I want to parse and validate the query in advance and then pass it to the executor. In my case I have a list of predefined queries (stored in the service) which can be executed by name. I am parsing and validating those queries during app initialization. Each query can be executed by its name, but now this parsing and validating functionality happens on every request, because I have to pass a string. I want to parse and validate the queries only once and then just to execute them, without the overhead for parsing and validating.

StevenACoffman:
@ruskovweb This was merged three years ago, so please make a new issue (or more preferably create a PR) if you would like any attention.

Issue #954, opened 2019-12-13
Can't get all request fields from query


RichardLindhout:
### What happened?
I'm using https://relay.dev/ as client library where you can put data inside fragments to structure your query more logically.
This is query Relay executes.
```graphql
query FlowTreeQuery(
  $flowId: ID!
) {
  flowBlocks(flowId: $flowId) {
    id
    flowBlock {
      id
    }
    block {
      title
      id
    }
    blockChoice {
      id
    }
    ...FlowBlock_flowBlock
  }
}

fragment BlockAmount_blockAmount on Block {
  title
}

fragment BlockChoices_blockChoices on Block {
  blockChoices {
    id
    title
    slug
  }
}

fragment BlockInput_blockInput on Block {
  title
}

fragment BlockTypeAssigner_blockTypeAssigner on Block {
  id
}

fragment FlowBlock_flowBlock on FlowBlock {
  block {
    id
    blockType
    ...BlockChoices_blockChoices
    ...BlockAmount_blockAmount
    ...BlockInput_blockInput
    ...BlockTypeAssigner_blockTypeAssigner
  }
}
```

I called 
```golang
fields := graphql.CollectAllFields(ctx)
	for _, f := range fields {
		fmt.Println(f)
	}
```

It returned:
```
id
flowBlock
block
blockChoice
```
### What did you expect?

I expected at least 
```
flowBlock
block
block.blockChoices
blockChoice
```
And even better something like this:
```
flowBlock
flowBlock.id
block
block.blockType
block.blockChoices
block.blockChoices.id
block.blockChoices.title
block.blockChoices.slug
block.id
block.title
blockChoice
blockChoice.id
```

### Minimal graphql.schema and models to reproduce
```graphql
type Block implements Node {
  id: ID!
  title: String!
  description: String
  slug: String
  blockType: String!
  blockChoices: [BlockChoice!]
}

type BlockChoice implements Node {
  id: ID!
  block: Block
  title: String!
  description: String
  slug: String
  imgUrl: String
}

type Flow implements Node {
  id: ID!
  title: String!
  description: String
  expression: String
  flowBlocks: [FlowBlock!]
}

type FlowBlock implements Node {
  id: ID!
  flowBlock: FlowBlock
  block: Block
  blockChoice: BlockChoice
  flow: Flow!
  flowBlocks: [FlowBlock]!
}

interface Node {
  id: ID!
}

type RootQueryType {
  node(id: ID!): Node
}

type Query {
[...]

PR #1053, opened 2020-02-20
Add practical example of getting all the requested fields

RichardLindhout:
Based on this https://github.com/99designs/gqlgen/issues/954 was tagged as 'need documentation'


vektah:
Thanks!

PR #1054, opened 2020-02-20
suppress golint messages

vvakame:
I tried new resolver layout. It's pretty nice!
In my app, golint emit messages to resolver files.
```
package comment should be of the form "Package graphqlapi ..."
exported method Resolver.CircleExhibitInfo should have comment or be unexported
```
I wanna suppress it.
gqlgen maintenance files are manually edited & generated file both.
I can't ignore these messages...

I have:

### DIFF
--- a/codegen/testserver/resolver.go
+++ b/codegen/testserver/resolver.go
@@ -1,5 +1,6 @@
-// THIS CODE IS A STARTING POINT ONLY. IT WILL NOT BE UPDATED WITH SCHEMA CHANGES.
 package testserver
+
+// THIS CODE IS A STARTING POINT ONLY. IT WILL NOT BE UPDATED WITH SCHEMA CHANGES.
 
 import (
 	"context"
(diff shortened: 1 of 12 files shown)


codegen/testserver/resolver.go
```diff
@@ -306,17 +307,38 @@ func (r *userResolver) Friends(ctx context.Context, obj *User) ([]*User, error)
...
+// OverlappingFields returns OverlappingFieldsResolver implementation.
 func (r *Resolver) OverlappingFields() OverlappingFieldsResolver { return &overlappingFieldsResolver{r} }
-func (r *Resolver) Panics() PanicsResolver                       { return &panicsResolver{r} }
-func (r *Resolver) Primitive() PrimitiveResolver                 { return &primitiveResolver{r} }
-func (r *Resolver) PrimitiveString() PrimitiveStringResolver     { return &primitiveStringResolver{r} }
-func (r *Resolver) Query() QueryResolver                         { return &queryResolver{r} }
-func (r *Resolver) Subscription() SubscriptionResolver           { return &subscriptionResolver{r} }
-func (r *Resolver) User() UserResolver                           { return &userResolver{r} }
+
+// Panics returns PanicsResolver implementation.
```
vektah: :nauseated_face: 


If you can build them out of a template then its not giving any context or new information to the reader. This is just pure noise.

can we tell golint to ignore these semi-generated files instead?

vvakame: No, We can't... 😿 I agreed about its noisy comment. but I need suppress golint error messages...
golint can't configure that same as gofmt.
golint only have `-min_confidence` options.

vektah: I highly recommend https://github.com/golangci/golangci-lint.

It will speed up your builds, provide consistent formatting and give you the ability to skip linters consistently on a line/function/file basis based on consistent comments - https://github.com/golangci/golangci-lint#false-positives

vvakame: yes, I know... but Out project can't it immediately.
and, I don't think gqlgen should not limit what tools users should use.
this change is little bit noisy, but useful for golint user I think.

vvakame:
thanks! 👍 🙇

PR #960, opened 2019-12-17
urfave/cli upgrade

RichardLindhout:
I've got erros when trying to run 
```
go run github.com/99designs/gqlgen --verbose
```

These are due the urfave/cli package has been changed on master and the default is the new v2 api.

Describe your PR and link to any relevant issues. 
https://github.com/99designs/gqlgen/pull/960#issue-354126694
I have:
 - [ ] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [ ] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

PR #1057, opened 2020-02-21
Upgrade to github.com/urfave/cli/v2

RichardLindhout:
Follow-up feedback from pull request: https://github.com/99designs/gqlgen/pull/960

I got errors when trying to run (I like to work without go modules)
```
go run github.com/99designs/gqlgen --verbose
```

These are due the urfave/cli package has been changed on master and the default is the new v2 api.


vektah:
> I like to work without go modules

That's a really bad idea with gqlgen, the runtime and generated code need to stay in sync, or you risk breaking things in subtle ways. What is pushing you away from go modules?

go.mod
```diff
@@ -1,32 +1,30 @@
...
 
 require (
-	github.com/agnivade/levenshtein v1.0.3 // indirect
-	github.com/go-chi/chi v3.3.2+incompatible
-	github.com/gogo/protobuf v1.0.0 // indirect
-	github.com/gorilla/context v0.0.0-20160226214623-1ea25387ff6f // indirect
-	github.com/gorilla/mux v1.6.1 // indirect
-	github.com/gorilla/websocket v1.2.0
-	github.com/hashicorp/golang-lru v0.5.0
+	github.com/go-chi/chi v4.0.3+incompatible
```
vektah: Theres too much changing in here. Can you revert these mod/sum changes and just update urfave via
```
go get github.com/urfave/cli/v2
go mod tidy
```

RichardLindhout:
@vektah I do use go modules but not when I'm developing custom plugins for gqlgen I'd like them to be directly updated and sometimes even after I push to Github I can't seem to update the module. I have to put the latest tag manually in the go.mod only then it fetches the latest version.

go.mod
RichardLindhout: Done!

vektah:
When working with multiple repos use [replace](https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive) and you will get the same behaviour as before.

eg at the bottom of your mod file
```
replace github.com/my/plugin-repo => ../plugin-repo
```

cmd/gen.go
```diff
@@ -7,24 +7,25 @@ import (
...
 	},
-	Action: func(ctx *cli.Context) {
+	Action: func(ctx *cli.Context) error {
 		var cfg *config.Config
 		var err error
 		if configFilename := ctx.String("config"); configFilename != "" {
 			cfg, err = config.LoadConfig(configFilename)
 			if err != nil {
 				fmt.Fprintln(os.Stderr, err.Error())
 				os.Exit(1)
```
vektah: return after exit doesn't do anything

RichardLindhout: I know maybe I will try to remove os.Exit() and just return the err 
and use OnUsageError. 
```
OnUsageError: func(c *cli.Context, err error, isSubcommand bool) error {
          fmt.Fprintf(c.App.Writer, "for shame\n")
          return err
        },
```

vektah: I think the exit and print is automatic when returning an error, maybe do a quick test?

`OnUsageError` is for when the arg parser fails I think?

RichardLindhout:
@vektah replace works like a charm! I did not understand it at the time of pull request :)

PR #1061, opened 2020-02-22
Fix mismatching documentation of Todo struct

halvdan:
Mismatch between the code and the getting started documentation.

Just tried the tutorial at https://gqlgen.com/getting-started/ and stumbled across a minor error between the documentation and the code.


vektah:
Thanks

PR #1085, opened 2020-03-04
Allow custom resolver filenames using `filename_template` option

tobias-kuendig:
This PR adds a new `filename_template` option to the `resolver` section.

This allows us to get more control over the generated filenames when using the `follow-schema` layout.

The new option is a string that is passed to `fmt.Sprintf` and receives the resolver's name as a parameter.

If the option is empty, the format `%s.resolvers.go` is used.

Before updating the relevant documentation I wanted to get feedback about the implementation and possible other solutions.


tobias-kuendig:
Regarding the failed test:

It looks like removing the original `testdata/followschema/out/schema.resolvers.go` re-generates the file now with new data, so the assertions won't match anymore.

This also happens on the master branch if the file gets removed before running the tests.

This feels like a test that should have been updated?

vektah:
Placeholders are a little more future proof and self describing, eg 
```yaml
filename_template: "{name}.custom.go"
```

that said, whats the use case here? It would be very easy to break gqlgens output using this by not including the name part, or colliding it with another file in the dir by leaving the suffix off. Without a use case its hard judge the tradeoff.

tobias-kuendig:
Thank you for your feedback. 

The current naming convention of `%s.resolvers.go` is very opinionated. Working on a bigger app, the gqlgen files are the only filenames that contain a suffix. This makes them feel out of place. In my case, I wanted to remove the suffix completely, so the files are just called `{name}.go`. 

Let me know if I should work on a second iteration of this PR. Otherwise feel free to close it!

vektah:
Ok, lets switch to a placeholder rather than %s and lets :shipit:

tobias-kuendig:
Great! I did replace the `fmt.Sprintf` call with a call to `strings.ReplaceAll` to get simple placeholder support.

raphaelvigee:
Any update on getting this merged ?

vektah:
I've squash rebase merged this manually in https://github.com/99designs/gqlgen/commit/ad675f0092bc8973dba5114ffa168379647bde45 to fix the broken tests

PR #1086, opened 2020-03-04
Use GitHub Actions

mtibben:
Replaces CircleCI and Appveyor with GitHub Actions.

This is a very direct translation of the CircleCI + Appveyor config, using exactly the same environments, steps and scripts.


vektah:
Oh nice! No more appveyor!

PR #1098, opened 2020-03-08
Update getting started guide with includes

jorgson:
Adding model include and math/rand to prevent new user from errors

Missing Model import will result in undefined: model
Missing math/rand import will result in undefined: rand


vektah:
Thanks for the PR, buit I think the docs should assume a basic knowledge of go, and expect users to be running `goimports`, or a similar tool in their editor, otherwise we are going to need this everywhere.

PR #1100, opened 2020-03-12
Gnerate to fast by exec codegen.GenerateCode before plugin GenerateCode

sonatard:
I fixed to only called once during generation in #1079 but it slowed down generation.
I found to fast by exec codegen.GenerateCode before plugin GenerateCode.

Benchmark in my code.

```shell
// v0.11.2
$ time gqlgen generate
gqlgen generate  20.41s user 15.53s system 30% cpu 1:59.05 total

// this PR
$ time gqlgen generate
gqlgen generate  6.15s user 4.54s system 383% cpu 2.791 total
```

I don't know why...

### DIFF
--- a/api/generate.go
+++ b/api/generate.go
@@ -74,6 +74,10 @@ func Generate(cfg *config.Config, option ...Option) error {
 		return errors.Wrap(err, "merging type systems failed")
 	}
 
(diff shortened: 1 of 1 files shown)


vektah:
Ok, lets revert until we figure it out

PR #1104, opened 2020-03-15
Update Query Complexity Documentation

oshalygin:
- This pass at the documentation updates the
  appropriate section regarding query complexity,
  specifically in the way that the http.Handler
  is created.
- The deprecated handler.GraphQL calls were replaced
  with NewDefaultServer.
- Instead of passing along the fixed query complexity
  as a second argument to the now deprecated handler.GraphQL
  func, extension.FixedComplexityLimit is used instead.

I have:

### DIFF
--- a/docs/content/reference/complexity.md
+++ b/docs/content/reference/complexity.md
@@ -1,8 +1,8 @@
 ---
-title: 'Preventing overly complex queries'
+title: "Preventing overly complex queries"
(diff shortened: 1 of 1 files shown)


vektah:
Thanks!

matheuss-leonel:
Please, update the website. The submenu item "[Query Complexity](https://gqlgen.com/reference/complexity/)" still uses the old version

Issue #1113, opened 2020-03-19
Custom scalar validation


bowd:
### What happened?

When handing a mutation Input that has custom scalars validation of input happens before unmarshaling and results in errors with correct path information, but unmarshaling errors lose the path information and stop at the mutation.

### What did you expect?

I was expecting that errors that occur as a result of unmarshalling fields also act as a form of validation and preserve the input path. See the example in the next section.

### Minimal graphql.schema and models to reproduce

```graphql
input NewContractInput {
    network: Network!
    address: EthAddress!
}

scalar EthAddress
enum Network {
    Mainnet
    Rinkeby
    Ropsten
    Kovan
}

```

```golang
func (value *EthAddress) UnmarshalGQL(rv interface{}) error {
	input, ok := rv.(string)
	if !ok {
		return fmt.Errorf("EthAddress expects a string value")
	}
	if !common.IsHexAddress(input) {
		return fmt.Errorf("not a valid ethereum address: %s", input)
	}

	*value = EthAddress(common.HexToAddress(input))
	return nil
}
```

If I provide a malformed `network` for example I get the error:
```json
{
  "error": {
    "errors": [
      {
        "message": "Mainnest is not a valid Network",
        "path": [
          "variable",
          "input",
          "network"
        ],
        "extensions": {
          "code": "GRAPHQL_VALIDATION_FAILED"
        }
      }
    ],
    "data": null
  }
}
```

But if I provide a malformed EthAddress I get the error:
```json
{
  "errors": [
    {
      "message": "not a valid ethereum address: 0x123",
      "path": [
        "contract",
        "create"
      ]
    }
  ],
  "data": null
}
```

Might be loosely related to #772 
Also from the looks of the code, I can't see a way to discern path based on this. The data would need to be attached somewhere in the generated unmarshal code.

What I want to achieve is to be able to use scalars and for the client to be able to tie errors to particular form fields. But it seems like that's impossible in the current implementation. The way I see it we have two solutions:
- Add another expected interface method to scalars like `ValidateGQL` that's used by the gqlparser at validation time. The downside is that in some cases you might do duplicate work to validate and unmarshal a scalar.
- Wrap errors that happen as a result of unmarshaling parts of the Input and preserve the path information (preferred).  

### versions
 - `v0.11.1`
 - `1.13`
 - `go modules`

bowd:
[...]

PR #1115, opened 2020-03-20
Add input path in unmarshaling errors

bowd:
> Related to #1113 

### Problem summary 

When relying on custom scalar, especially in mutations, you offload an aspect of validation to the unmarshaling step. This means that you ensure these custom types have the right structure in one place and then you can rely on them in the next layers. 

The problem with the current implementation is that errors that happen as a result of scalar unmarshaling only have path information up to the mutation that they occur in. And it's impossible to reconstruct what field in the input is responsible for the error if you have multiple instances of the scalar in a nested input structure. And makes it impossible for the frontend to link the error to a form field, for example.

### Solution

I've added a new context type `FieldInputContext` which acts similarly to `FieldContext`  but is responsible for keeping track of the nesting inside of the arguments received by queries or mutations. This context is then used during unmarshaling to wrap the error returned into a `gqlerror.Error` if it's not already or just inject the `path` if it is. This lets developers return `gqlerror.Error` from their unmarshaling logic in order to inject additional data in `extensions`

See the [updated docs](https://github.com/99designs/gqlgen/pull/1115/files#diff-c9dfeb0cd4ed270d66024b3dbb09de96) for additional details.

__Codegen touchpoints__:
- `args.gotpl` - created nested input context for each argument
- `input.gotpl` - created nested input context for each field in an input
- `type.gotpl` - treat all unmarshaling scenarios

I have:


bowd:
I was thinking now that this is, in a way, a breaking change because some implementations might be expecting the old path format somewhere downstream. So this feature could be enabled through configuration so you can switch between the partial field path or the full path for the errors.

bowd:
An extension to this that I would like to make and I didn't handle in this PR would be to not fail on the first unmarshal error in the arg tree, but collect all errors in the tree using a `multierror` package or something similar to handle merging of errors. This way we could reduce round-trips by returning _all_ errors that we can find in the first call.

vektah:
Thanks, this has annoyed me for a long time. :tada: 

I've added a small commit to drop the custom context names to be consistent with the rest of the context usage (eg field context).

PR #1116, opened 2020-03-20
Do not use pointers on named map types

### LINKED ISSUE #878: map[string]interface has issue when generating in v0.10.0 or v0.10.1
### What happened?

map[string]interface{} generation issue

`example/todo/generated.go:2938:17: cannot use v (type Test) as type *Test in argument to ec._Test`

### What did you expect?
the error does not happen in v0.9.3

I just use example/todo to test the map[string]interface{}

add in example/todo/todo.go

`type Test map[string]interface{}`

and add Test type and also add to any existing types, lets say Todo

```
type Todo {
    id: ID!
    text: String!
    done: Boolean! @hasRole(role: OWNER) # only the owner can see if a todo is done
    test: Test
}

type Test {
  key: String
  value: Any
}
```

then add Test into gqlgen.yml

```
 Test:
    model: github.com/99designs/gqlgen/example/todo.Test
```

run `go run github.com/99designs/gqlgen` in example/todo
and then `go run ./example/todo/server/server.go`

### versions
 - `gqlgen version`? v0.10.0 and v0.10.1
 - `go version`? v1.13
 - dep or go modules? modules

bastianccm:
Named map types such as
type Foo map[string]interface{}
are nilable and named, which means they should not be references as a
pointer in generated code.

I am unsure if the test code is in the proper location.

This resolves #878 

I have:

### DIFF
--- a/codegen/object.go
+++ b/codegen/object.go
@@ -82,15 +82,7 @@ func (b *builder) buildObject(typ *ast.Definition) (*Object, error) {
 }
 
 func (o *Object) Reference() types.Type {
-	switch v := o.Type.(type) {
-	case *types.Named:
-		if _, isInterface := v.Underlying().(*types.Interface); isInterface {
-			return o.Type
-		}
-		if _, isMap := v.Underlying().(*types.Map); isMap {
-			return o.Type
-		}
-	case *types.Pointer, *types.Slice, *types.Map:
+	if config.IsNilable(o.Type) {
 		return o.Type
 	}
 
--- a/codegen/object.go
+++ b/codegen/object.go
@@ -84,8 +84,10 @@ func (b *builder) buildObject(typ *ast.Definition) (*Object, error) {
 func (o *Object) Reference() types.Type {
 	switch v := o.Type.(type) {
 	case *types.Named:
-		_, isInterface := v.Underlying().(*types.Interface)
-		if isInterface {
+		if _, isInterface := v.Underlying().(*types.Interface); isInterface {
+			return o.Type
+		}
+		if _, isMap := v.Underlying().(*types.Map); isMap {
 			return o.Type
 		}
 	case *types.Pointer, *types.Slice, *types.Map:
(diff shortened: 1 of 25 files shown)


codegen/object.go
```diff
@@ -84,8 +84,10 @@ func (b *builder) buildObject(typ *ast.Definition) (*Object, error) {
 func (o *Object) Reference() types.Type {
 	switch v := o.Type.(type) {
 	case *types.Named:
-		_, isInterface := v.Underlying().(*types.Interface)
-		if isInterface {
+		if _, isInterface := v.Underlying().(*types.Interface); isInterface {
```
vektah, line 87: I think this should be calling https://github.com/99designs/gqlgen/blob/14dccc57885df5d5ca4ef347c1b80f5f3648719a/codegen/config/binder.go#L413 (and make it public) to stop these falling out of sync again in the future?

bastianccm:
Hey @vektah,

in general I agree, however updating it I stumbled upon another issue: the same issue this change fixes for maps applies to slices as well.
To have correct generated code I would use
```go
func (o *Object) Reference() types.Type {
	if config.IsNilable(o.Type) {
		return o.Type
	}
	return types.NewPointer(o.Type)
}
```
for the object generation (the change you suggested), and update `IsNilable` to include a check for `types.Slice` as well
```go
_, isSlice := t.(*types.Slice)
return isPtr || isMap || isInterface || isSlice
```

However this breaks `TestIntrospection/enabled_by_default` and `TestSlices/nulls_vs_empty_slices`.
I can fix the nulls_vs_empty_slices tests by using
```go
return &Slices{
	Test1: nil,
	Test2: nil,
	Test3: make([]*string, 0),
	Test4: make([]string, 0),
}, nil
```
to mark explicitly what is null and what is empty.
However for the introspection I am unsure if I am on the right path doing these changes.

I pushed this as a new commit, with the failing tests, so you have a chance to see what I did.

Maybe you have an idea/opinion on this topic?

lwc:
Replaced with https://github.com/99designs/gqlgen/pull/1242 thanks for the contribution!

PR #1117, opened 2020-03-25
Add support for OBJECT directive

s-ichikawa:
I propose the implementation of the execution of directives for OBJECT.
When resolving an object that has a directive, the directive function can insert some process before and after executing resolver.

In the case of setting the same directive at field and object, it executes field directive before object directive.

If directive for OBJECT should do other behaviors or in other execute timing, I will close this PR and rewrite.

I have:


vektah:
Sweet, this looks super clean :tada:

PR #1121, opened 2020-03-28
Do not require a resolver for "empty" extended types.

csilvers:
Summary:
If our schema has a field with a type defined in another service, then
we need to define an "empty extend" of that type in this service, so
this service knows what the type is like.  But the graphql-server will
never ask us to actually resolve this "empty extend", so we don't
require a resolver function for it.  Example:
```
   type MyType {
      myvar: TypeDefinedInOtherService
   }

   // Federation needs this type, but it doesn't need a resolver for
   // it!  graphql-server will never ask *us* to resolve a
   // TypeDefinedInOtherService; it will ask the other service.
   extend TypeDefinedInOtherService @key(fields: "id") {
      id: ID @extends
   }
```

Test Plan:
I manually tested this on a service (`assignments`) that we have that
fell afoul of this problem.  But I had a hard time adding tests inside
gqlgen because the error happens at validation-time, and the
federation tests are not set up to go that far down the processing
path.

### DIFF
--- a/plugin/federation/federation.go
+++ b/plugin/federation/federation.go
@@ -95,11 +95,13 @@ func (f *federation) InjectSourceLate(schema *ast.Schema) *ast.Source {
 		}
 		entities += e.Name
 
-		resolverArgs := ""
-		for _, field := range e.KeyFields {
-			resolverArgs += fmt.Sprintf("%s: %s,", field.Field.Name, field.Field.Type.String())
+		if e.ResolverName != "" {
+			resolverArgs := ""
+			for _, field := range e.KeyFields {
+				resolverArgs += fmt.Sprintf("%s: %s,", field.Field.Name, field.Field.Type.String())
+			}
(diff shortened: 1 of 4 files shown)


csilvers:
@marwan-at-work what do you think of this idea?

See also https://github.com/apollographql/apollo-server/issues/3852

csilvers:
I'd like to reopen this -- it's been working well for us for several months now.  @marwan-at-work are you up for taking a look?

csilvers:
(I don't know how to reopen this --- I probably don't have the powers.)

csilvers:
Yeah, I tried to add tests but couldn't figure out how to do it.  I'll see if I can figure something out but if you have any pointers that might help get started!

csilvers:
@mtibben or other maintainers: are you comfortable merging this with the meager tests it has now, or would you like to see more like marwan mentioned?  I'm not really sure how to do what he's asking for so it may be a while before I can get to it.

lwc:
Let's merge it. Thanks for the contribution @csilvers & the 👀  @marwan-at-work

PR #1134, opened 2020-04-05
Add LoadDefaultConfig to load the schema by default

seriousben:
When `gqlgen gen` falls back to the `DefaultConfig`, the generated code does not have any models, queries or mutations.

The reason is that the default `schema.graphql` is never loaded. `api.Generate` is called with an empty list of `ast.Source`.

This PR changes the `gen` command to properly generate code based on the default config.

I have:
 - Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))
    - There was no documentation to update that I could see.


vektah:
Thanks, looks good :+1:

PR #1137, opened 2020-04-07
Replace ~ with א in package names

### LINKED ISSUE #1136: Special characters in module path cause codegen errors
### What happened?

```
go mod init https://git.sr.ht/~example/foo.bar
go run github.com/99designs/gqlgen init
go build # many errors
```

### What did you expect?

These characters are valid for use in module URLs, so gqlgen should rewrite them to something suitable for its identifiers.

 - `gqlgen version`? 0.11.3
 - `go version`? 1.14.1 linux/amd64

ddevault:
Fixes #1136


vektah:
Nice, symbol looks symbolic.

PR #1151, opened 2020-04-15
fix add waitgroup

nzlov:
Update the wait count once


mtibben:
> Update the wait count once 

Why?

Please add a rationale or tests to demonstrate incorrect behaviour

Issue #746, opened 2019-06-14
CollectAllFields doesn't include nested fields 


Baggerone:
### What happened?
This is my code ...
```
	requested := graphql.CollectAllFields(ctx)
	fmt.Printf("\nRequested Fields:   %v\n", requested)
```
This is the graphql request ...
```
{user(id:"1") {name, nickname}, cars{registration}}
```
Just the top level fields were returned ...
```
Requested Fields:   [name nickname]
```

### What did you expect?
That even the nested fields would be returned
```
Requested Fields: name, nickname, cars, registration
```

### Minimal graphql.schema and models to reproduce

```
type User {
  id: ID
  name: String!
  nickname: String
  cars: [Car!]
}

type Car {
  id: ID
  registration: String!
}
```

### versions
 - `gqlgen version`?   dev
 - `go version`? g01.11.10
 - dep or go modules? dep

mathewbyrne:
This is intentional, as it is a convenience wrapper around `CollectFieldsCtx` to account for the most common case, where you just want to limit querying your data source by the actual fields requested.  If you have a more complex use-case, use that instead.

See also our [field collection docs](https://gqlgen.com/reference/field-collection/).

Baggerone:
@mathewbyrne,

Thank you for your reply.  Using my graphql request (above) and CollectFieldsCtx, I still can't see cars.registration getting returned.  (I still only see user.name and user.nickname.) It still seems like those Collect functions don't look inside the nested part of the graphql request.

Maybe I'm missing something.  

E.g. If I have ... 
```
	requested := []string{}
	satisfiesFields := graphql.CollectFieldsCtx(ctx, []string{"registration"})
	fmt.Printf("\nSatisfies Fields:   \n")
	for _, f := range satisfiesFields {
		requested = append(requested, f.Name)
		fmt.Printf("   %v\n", f.Name)
	}
```
I get ...
```
Satisfies Fields:   
   name
   nickname
```

Also, if I replace "registration", with "car", "Car,  "cars" or "Cars", I still get that result.

Can you give me a concrete example of what I would need to do to get the car.registration field included in the results?

Thank you for your help.

RichardLindhout:
I have the same use-case.

My query:
```
 flowBlocks {
            id
            block {
              id
              title
              type
              choices {
                id
                title
                description
                slug
              }
            }
          }
        }
```

Code
```
	requested := graphql.CollectFieldsCtx(ctx, []string{"block", "choices"})
[...]

PR #1161, opened 2020-04-21
Fix duplicate fields in preloads

nitin302:
This is a documentation change, where we are looping over Selections Twice, which is causing duplicates. I've removed one nested preload to removed these duplicates.

### DIFF
--- a/docs/content/reference/field-collection.md
+++ b/docs/content/reference/field-collection.md
@@ -94,8 +94,6 @@ func GetNestedPreloads(ctx *graphql.RequestContext, fields []graphql.CollectedFi
 		prefixColumn := GetPreloadString(prefix, column.Name)
 		preloads = append(preloads, prefixColumn)
 		preloads = append(preloads, GetNestedPreloads(ctx, graphql.CollectFields(ctx, column.SelectionSet, nil), prefixColumn)...)
-		preloads = append(preloads, GetNestedPreloads(ctx, graphql.CollectFields(ctx, column.Selections, nil), prefixColumn)...)
-
 	}
 	return
 }


vektah:
Thoughts @RichardLindhout ?

RichardLindhout:
There are cases where this is needed I think with fragments, maybe we need to remove duplicates but there was a case when something was inside fragments we need this. I think this issue was posted here. Maybe this is resolved now with newer versions?

Schermafbeelding 2020-04-24 om 11 06 52


My issue back then:
https://github.com/99designs/gqlgen/issues/954

More on that issue here:
https://github.com/99designs/gqlgen/issues/746

Maybe we could do a function where we only append if not exist yet. I did not see this problem before. Do you have a query which result in duplicates?

I generate type-safe preloads here:
https://github.com/web-ridge/gqlgen-sqlboiler/blob/master/examples/social-network/helpers/preload.go

Latest helpers:
https://github.com/web-ridge/gqlgen-sqlboiler/blob/master/helper/preload.go

RichardLindhout:
I think if you don't include it. Selections it won't resolve nested fragment fields. However removing duplicates would be a good idea.

nitin302:
`query Search($input: NewSearch!) {
  search(input: $input) {
    chats {
      id
      lastMessage
      messagestats{
        noOfMessages
      }
  }
}
`
This returns an array of Chats in a Search Response.

I was seeing duplicates, when I fetched the fields, as shown in documentation. Every nested field was duplicated.

RichardLindhout:
Don't have time to test a.t.m.
@vektah Has something changed in the SelectionSet lately or in the new parser?


Maybe this instead of appending
```
func appendPreloadIfMissing(slice []string, v string) []string {
    for _, ele := range slice {
        if ele == v {
            return slice
        }
    }
    return append(slice, i)
}
```

RichardLindhout:
Field are indeed duplicated but for me only column.Selections returned the right fields for me with fragments. SelectionSet did not

RichardLindhout:
This should fix it: https://github.com/99designs/gqlgen/pull/1189

Issue #1090, opened 2020-03-05
Apollo Federation: 422: Unprocessable Entity when no entity defined


aaad:
### What happened?

There is a service which only contains a mutation which has only input types defined:

```
extend type Mutation {
  addSurveyResult(result: SurveyResult!): ID
}

input SurveyResultTuple {
     key: String
     value: String
}

input SurveyResult {
  userId: ID!

  type: String!

  elements: [SurveyResultTuple!]!
}
```

when I try to federate the service by a gateway the following errors occure:

Service:
>unknown field "_service"
>
> goroutine 9 [running]:
> runtime/debug.Stack(0x1, 0x0, 0x0)
>         /usr/local/go/src/runtime/debug/stack.go:24 +0x9d
> runtime/debug.PrintStack()
>         /usr/local/go/src/runtime/debug/stack.go:16 +0x22
> github.com/99designs/gqlgen/graphql.DefaultRecover(0xc0c740, 0xc0002c8690, 0x9fe620, 0xc0001b8de0, 0x44900f, 0x18)
>         /home/simon/go/pkg/mod/github.com/99designs/gqlgen@v0.11.1/graphql/recovery.go:16 +0xa7
> github.com/99designs/gqlgen/graphql/handler.(*Server).ServeHTTP.func1(0xc0000d5e40, 0xc00013fbd0, 0xc0ae40, 0xc0002922a0)
>         /home/simon/go/pkg/mod/github.com/99designs/gqlgen@v0.11.1/graphql/handler/server.go:124 +0xa7
> panic(0x9fe620, 0xc0001b8de0)
>         /usr/local/go/src/runtime/panic.go:679 +0x1b2
> recommender-survey-archive/cmd/graph/generated.(*executionContext)._Query(0xc0001b8da0, 0xc0c740, 0xc0002c8ff0, 0xc0001b8b40, 0x1, 0x1, 0x40dcc6, 0xc0002c8fc0)
>         /home/simon/go/src/recommender-survey-archive/cmd/graph/generated/generated.go:1527 +0x6b4
> recommender-survey-archive/cmd/graph/generated.(*executableSchema).Exec.func1(0xc0c740, 0xc0002c8fc0, 0x10)
>         /home/simon/go/src/recommender-survey-archive/cmd/graph/generated/generated.go:99 +0x7b
[...]

PR #1192, opened 2020-05-14
federation: generate _entities / _service root queries regardless of …

alexsn:
…entity count

Generate a valid federation schema extention if no object uses @key
directive.

On this case the extension will be defined as follows:
```graphql
union _Entity

type _Service {
  sdl: String
}

extend type Query {
  _entities(representations: [_Any!]!): [_Entity]!
  _service: _Service!
}
```

I have:

### DIFF
--- a/example/federation/accounts/graph/generated/federation.go
+++ b/example/federation/accounts/graph/generated/federation.go
@@ -17,7 +17,6 @@ func (ec *executionContext) __resolve__service(ctx context.Context) (fedruntime.
 	}
 
 	var sdl []string
-
 	for _, src := range sources {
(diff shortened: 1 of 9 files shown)


mlaflamm:
This seems to fix issue #1090. We this haven't been merged?

alexsn:
This sat in review for a long time without any comments. I assumed maintainers were not interested.

saminahbab:
I think this should be reopened, there are legitimate reasons for federating servers even if no object has a @key and so this should be default behaviour

kecci:
i got 422 Unprocessable Entity, is this will be the solution ?

PR #1238, opened 2020-07-03
fix #1237 remove unreachable types from generated code

### LINKED ISSUE #1237: remove unreachable types/resolvers from generated code
### What happened?

I have a number of graphql files that hold shared type definitions from multiple backends. Not every backend uses every type, but it's convenient to be able to deduplicate them in cases where they are shared.

Unfortunately, currently, this results in every backend being forced to generate resolvers for every type.

### What did you expect?

I expected that unreachable types don't need to be implemented and don't affect the output of gqlgen.

https://github.com/vikstrous/unreachable-repro

### versions
 - `gqlgen version`: master 
 - `go version`: 1.14.4
 - dep or go modules: go modules

### Other

Note that this is not just a slight annoyance. In order to split up our graphql schema and move towards federation, we need to be able to take chunks out of the main common schema one at a time. With multiple different services re-implementing the same graphql resolvers, we would have to duplicate our schema 5 times as the first step towards federation. It's very hard to do this at scale and having this feature in gqlgen would really simplify our transition.

vikstrous:
This PR removes unreachable objects, inputs, interfaces, unions, scalars, and enums from the schema right after parsing it. It doesn't try to remove unused directives.

Existing tests did a reasonable job at covering the changes, but this does need more testing. I added `unreachable.graphql` with an object, an input, an interface, an union, a scalar, and an enum.

I need feedback on what to do with built in types like Int and if this warning message makes sense:
```
Warning: unreachable types: AIt, AbIt, Content_Child, Content_Post, Content_User, EmbeddedDefaultScalar, EmbeddedPointer, LoopA, LoopB, Map, UnreachableEnum, UnreachableInput, UnreachableInterface, UnreachableObject, UnreachableScalar, UnreachableUnion, XXIt, XxIt, asdfIt, iIt
```

I could use some help with the diff in the generated resolver.go code. There's some formatting issue?

I have:

### DIFF
--- a/api/generate.go
+++ b/api/generate.go
@@ -39,7 +39,7 @@ func Generate(cfg *config.Config, option ...Option) error {
 		}
 	}
 
-	if err := cfg.LoadSchema(); err != nil {
+	if err := cfg.LoadSchema(false); err != nil {
 		return errors.Wrap(err, "failed to load schema")
 	}
 
@@ -52,7 +52,7 @@ func Generate(cfg *config.Config, option ...Option) error {
 	}
 
 	// LoadSchema again now we have everything
-	if err := cfg.LoadSchema(); err != nil {
+	if err := cfg.LoadSchema(true); err != nil {
 		return errors.Wrap(err, "failed to load schema")
 	}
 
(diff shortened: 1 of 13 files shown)


frederikhors:
Nope

vikstrous:
continued in #2139

Issue #1193, opened 2020-05-17
Upload file causes memory leak


arkhvoid:
### What happened?
After a file being uploaded, memory usage is dramatically bumped.
### Minimal graphql.schema and models to reproduce
`type Mutation {
    addVideo(video: Upload!): Video
}`
### versions
 - `v0.11.3`
 - `1.14.2`
 - basic net/http only
### Resolver
`func (r *mutationResolver) AddVideo(ctx context.Context, video graphql.Upload) (*model.Video, error) {
	return nil, nil
}`
### pprof
I have try to use [net/http/pprof](https://golang.org/pkg/net/http/pprof/) to check memory usage, and this is some of the result of them.
- heap

> (pprof) top
Showing nodes accounting for 768MB, 99.80% of 769.50MB total
Dropped 13 nodes (cum <= 3.85MB)
Showing top 10 nodes out of 16
      flat  flat%   sum%        cum   cum%
     768MB 99.80% 99.80%      768MB 99.80%  bytes.makeSlice
         0     0% 99.80%      768MB 99.80%  bytes.(*Buffer).ReadFrom
         0     0% 99.80%      768MB 99.80%  bytes.(*Buffer).grow
         0     0% 99.80%      768MB 99.80%  github.com/99designs/gqlgen/graphql/handler.(*Server).ServeHTTP
         0     0% 99.80%      768MB 99.80%  github.com/99designs/gqlgen/graphql/handler/transport.MultipartForm.Do
         0     0% 99.80%      768MB 99.80%  io.Copy (inline)
         0     0% 99.80%      768MB 99.80%  io.CopyN
         0     0% 99.80%      768MB 99.80%  io.copyBuffer
         0     0% 99.80%      768MB 99.80%  main.Header.func1
         0     0% 99.80%      768MB 99.80%  mime/multipart.(*Reader).ReadForm (inline)

- alloc

> (pprof) top
Showing nodes accounting for 2GB, 99.83% of 2GB total
Dropped 32 nodes (cum <= 0.01GB)
Showing top 10 nodes out of 16
      flat  flat%   sum%        cum   cum%
       2GB 99.83% 99.83%        2GB 99.83%  bytes.makeSlice
         0     0% 99.83%        2GB 99.83%  bytes.(*Buffer).ReadFrom
         0     0% 99.83%        2GB 99.83%  bytes.(*Buffer).grow
[...]

PR #1246, opened 2020-07-12
Fix typo cause memory problem on upload

arkhvoid:
Fix potentially typo that make unusual memory usage when uploaded a large file

The issue [#1193](https://github.com/99designs/gqlgen/issues/1193) created months back

Noticed that [line 67 of /graphql/handler/transport/http_form.go](https://github.com/99designs/gqlgen/blob/master/graphql/handler/transport/http_form.go#L67) has potentially typo of using `MaxUploadSize` instead of `MaxMemory`

[`func (r *Request) ParseMultipartForm(maxMemory int64) error`](https://golang.org/pkg/net/http/#Request.ParseMultipartForm) required max memory

This will cause problem when max upload size set to a large number


lwc:
Thanks!

PR #1248, opened 2020-07-13
close the connection when run returns

### LINKED ISSUE #1245: Leaking File Descriptors when client abruptly disconnects
### What happened?
Our dev server was periodically dying with the error `http: Accept error: accept tcp [::]:3000: accept4: too many open files; retrying in 1s`
### What did you expect?
Connections should be explicitly closed in all cases so sockets aren't just left hanging in a CLOSE_WAIT state
### versions
 - `gqlgen version` 0.11.3
 - `go version` 1.13
 - dep or go modules?

Hi we're seeing an issue of leaking websocket connections. Our dev server was crashing periodically because it would run out of file descriptors. After a lot of digging I found that a refresh on the client side was leaving connections just hanging in a CLOSE_WAIT state. I believe this is because connections aren't explicitly being closed on defer in the websocket run function

sotoslammer:
Closes #1245 

The case of client disconnecting without sending a termination message (ie Browser refresh) wasn't be handled.  This would lead to sockets just hanging open in a CLOSE_WAIT state.  This fix handles this case by always closing the connection when run returns.

I have:

### DIFF
--- a/graphql/handler/transport/websocket.go
+++ b/graphql/handler/transport/websocket.go
@@ -136,7 +136,10 @@ func (c *wsConnection) run() {
 	// We create a cancellation that will shutdown the keep-alive when we leave
 	// this function.
 	ctx, cancel := context.WithCancel(c.ctx)
-	defer cancel()
+	defer func() {
+		cancel()
+		c.close(websocket.CloseAbnormalClosure, "unexpected closure")
+	}()
 
 	// Create a timer that will fire every interval to keep the connection alive.
 	if c.KeepAlivePingInterval != 0 {


RobertoOrtis:
it's not cleaning up the resolvers. 

For example, lets suppose I have a resolver where I store temporal data about a user, when the user abruptly disconnects, the resolver is not cleaned up:
- The following code is never called
```
go func() {
	<-ctx.Done()
	r.mu.Lock()
	delete(r.UserData, user.ID)
	r.mu.Unlock()
}()
```

I wrote a solution a while back. I will be posting it soon.

RobertoOrtis:
Here is the suggested fix:

Edit file `/graphql/handler/transport/websocket.go`
Function: `write`
```
func (c *wsConnection) write(msg *operationMessage) {
	if msg.Type == "data" {
		c.mu.Lock()
		c.conn.WriteJSON(msg)
		c.mu.Unlock()
	}
}
```

Function `sendConnectionError`
```
func (c *wsConnection) sendConnectionError(format string, args ...interface{}) {
	for i, observer := range c.active {
		select {
		case <- c.ctx.Done():
			b, err := json.Marshal(&gqlerror.Error{Message: fmt.Sprintf(format, args...)})
			if err != nil {
				panic(err)
			}

			c.write(&operationMessage{Type: connectionErrorMsg, Payload: b})
		default:
			c.mu.Lock()
			closer := c.active[i]
			c.mu.Unlock()
			c.conn.Close()
			closer()
		}
	}
}
```

Function `readOp`
```
func (c *wsConnection) readOp() *operationMessage {
	_, r, err := c.conn.NextReader()
	if websocket.IsCloseError(err, websocket.CloseNormalClosure, websocket.CloseNoStatusReceived) {
		return nil
	} else if err != nil {
		mes := operationMessage{}
		e := c.active
		c.sendConnectionError("invalid json: %T %s", err, err.Error())
		return nil
	}
	message := operationMessage{}
	if err := jsonDecode(r, &message); err != nil {
		c.sendConnectionError("invalid json")
		return nil
	}

	return &message
}
```

mtibben:
@RobertoOrtis can you create a PR?

RobertoOrtis:
Sure!

RobertoOrtis:
@mtibben Done! https://github.com/99designs/gqlgen/pull/1583

PR #1263, opened 2020-07-25
errors: Add support for wrapped gqlerror.Error in DefaultErrorPresenter

alexsn:
This adds the support for wrapped `*gqlerror.Error` being presented to `DefaultErrorPresenter`. Note that this required go 1.13+.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,20 +33,20 @@ jobs:
   integration:
     runs-on: ubuntu-latest
     timeout-minutes: 3
-    container: alpine:3.10
+    container: golang:1.13-alpine
     steps:
       - uses: actions/checkout@v1
-      - run: apk add --no-cache --no-progress nodejs npm go musl-dev git bash
+      - run: apk add --no-cache --no-progress nodejs npm git bash
       - run: go mod download
       - run: cd integration ; npm install
       - run: .github/workflows/check-integration
 
   federation:
     runs-on: ubuntu-latest
-    container: alpine:3.10
+    container: golang:1.13-alpine
     steps:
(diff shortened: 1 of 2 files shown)


vektah:
I don't think we are ready to break support for 1.12 yet, I've been avoiding introducing wraps for a while.

alexsn:
@vektah will introducing build flags be acceptable?

mtibben:
@vektah is there a reason you want to support go 1.12? Go's official support is for the last 2 major versions, so Go doesn't even support Go 1.12 any longer :)

vektah:
Switching to using error wrapping will require more work to consistently do throughout gqlgen, work I don't have time to tackle right now. Perhaps a concrete argument for switching would help? 

It's worth noting that the proposed changes are only in the `DefaultErrorPresenter` which already can be swapped out by projects that need it, and most projects should already have their own implementation as error handling logic is highly app specific.

> Note that this required go 1.13+.

Not sure I follow, none of this is required to use go 1.13, it's only required if you want to use As and Is.

> @vektah will introducing build flags be acceptable?

I don't think this would help, I don't want to have to maintain two sets of error handling and docs. 

> @vektah is there a reason you want to support go 1.12? Go's official support is for the last 2 major versions, so Go doesn't even support Go 1.12 any longer :)

People can easily run 1.14 without using the new error handling, getting continued security coverage without the added churn. The same was true for go mod, and there are heaps of people that still haven't made the jump.

lwc:
I think https://github.com/99designs/gqlgen/issues/1291 and https://github.com/99designs/gqlgen/pull/1312 present a concrete argument, with the latter being my solve :)

Issue #1259, opened 2020-07-22
Cannot map custom scalars to third party pointer types 


albrow:
### What happened?

Following the [guide here](https://gqlgen.com/reference/scalars/#custom-scalars-with-third-party-types) I am trying to implement a custom scalar type `BigNumber` which maps to `*big.Int` in the standard library. This results in compiler errors in the generated code.

```
validation failed: packages.Load: /Users/alex/programming/gomod/gqlgen-todos/graph/generated/generated.go:1909:9: cannot use scalars.UnmarshalBigNumber(v) (value of type *big.Int) as big.Int value in return statement
/Users/alex/programming/gomod/gqlgen-todos/graph/generated/generated.go:1913:34: cannot use v (variable of type big.Int) as *big.Int value in argument to scalars.MarshalBigNumber
```

Here is the problematic code:

```go
func (ec *executionContext) unmarshalOBigNumber2mathᚋbigᚐInt(ctx context.Context, v interface{}) (big.Int, error) {
	return scalars.UnmarshalBigNumber(v)
}

func (ec *executionContext) marshalOBigNumber2mathᚋbigᚐInt(ctx context.Context, sel ast.SelectionSet, v big.Int) graphql.Marshaler {
	return scalars.MarshalBigNumber(v)
}
```

<img width="1180" alt="Screen Shot 2020-07-22 at 1 53 07 PM" src="https://user-images.githubusercontent.com/800857/88227631-bb352c80-cc22-11ea-94fb-86afacf85481.png">

<img width="1027" alt="Screen Shot 2020-07-22 at 1 52 57 PM" src="https://user-images.githubusercontent.com/800857/88227594-b1abc480-cc22-11ea-838d-c5bad54a964a.png">

In my real application, we need to use `*big.Int` and cannot use `big.Int`. All functions and methods in the `math/big` package expect `*big.Int` so if my models use `big.Int` it makes it extremely hard to work with.

### What did you expect?

Since I have defined custom marshaler and unmarshaler functions, I should be able to map `BigNumber` to `*big.Int` types in my models.

The models are actually being generated correctly. Here's what __model/models_gen.go__ looks like:

```go
type Todo struct {
	Number *big.Int `json:"number"`
}
```

The problem just lies in the generated code in __graph/generated/generated.go__. It looks like this can be resolved by changing the generated code so that it uses the correct types when calling `scalars.UnmarshalBigNumber` and `scalars.MarshalBigNumber`. I would suggest something like this:

```go
[...]

PR #1277, opened 2020-08-12
Support pointers in un/marshal functions

lwc:
## Background
When updating protobuf to new v2 API we started to see a bunch of linting errors like:
```
web/graph/types/timestamp.go:13:25: MarshalTimestamp passes lock by value: 
google.golang.org/protobuf/types/known/timestamppb.Timestamp 
contains google.golang.org/protobuf/internal/impl.MessageState 
contains sync.Mutex
```

We use [custom scalars for third party types](https://gqlgen.com/reference/scalars/#custom-scalars-with-third-party-types) to bind protobuf timestamps, upon updating the un/marshal functions to act on pointers, I ran into the issue outlined in #1259.

Currently gqlgen assumes that all concrete binding can happen against value types and pointers will delegate through to the underlying value type.

## This PR
Updates binding code to bind to support binding to pointers directly, supporting binding between pointers and values in both directions.

Fixes: #1259 

I have:


lwc:
This branch has been running 99designs production workloads without issue so I'm inclined to merge it.

vektah:
> It also removes a special case for binding to named types that wrap strings. This special case existed for a long since removed approach to supporting graphql enums, and the interaction between Target and CastType was proving too hard to reason about.

I hate to break it to you but I think this is going to break the prisma guys, judging from [the issue](https://github.com/99designs/gqlgen/issues/595) that came up last time I tried to remove it :joy:

lwc:
> I hate to break it to you but I think this is going to break the prisma guys, judging from the issue that came up last time I tried to remove it 😂

Bummer 😅 

Oh well, after cleaning up my codegen changes it was easy enough to re-instate (I think)

codegen/config/binder.go
```diff
@@ -167,6 +167,7 @@ type TypeReference struct {
 	Definition  *ast.Definition
 	GQL         *ast.Type
 	GO          types.Type
+	Target      types.Type
```
vektah, line 170: Could we leave a comment that tries to explain the difference between Go and Type?

The target is the type of the physical field in the struct, and GO is the derived type when you combine it with the pointeryness required by the ast?

codegen/config/binder.go
```diff
@@ -345,16 +349,27 @@ func (b *Binder) TypeReference(schemaType *ast.Type, bindTarget types.Type) (ret
...
 			ref.GO = fun.Type().(*types.Signature).Params().At(0).Type()
 			ref.Marshaler = fun
 			ref.Unmarshaler = types.NewFunc(0, fun.Pkg(), "Unmarshal"+typeName, nil)
-		case hasMethod(obj.Type(), "MarshalGQL") && hasMethod(obj.Type(), "UnmarshalGQL"):
+		} else if hasMethod(obj.Type(), "MarshalGQL") && hasMethod(obj.Type(), "UnmarshalGQL") {
 			ref.GO = obj.Type()
 			ref.IsMarshaler = true
-		default:
+		} else if underlying := basicUnderlying(obj.Type()); def.IsLeafType() && underlying != nil && underlying.Kind() == types.String {
+			// Special case for named types wrapping strings. Used by default enum implementations.
```
vektah: :+1:

lwc: Done. TBH I'd still like to remove this case before 1.0.

codegen/config/binder.go
lwc: Done - it's not even that, it's the difference between how to bind to the type, and how to traverse to the type from a field on another type, where sometimes that is due to gql AST requirements.

I wonder if there's an abstraction that represents both pointer juggling _and_ casting sorta like ` “x is satisfiable by y if i do a, b and c”` 🤔

PR #1285, opened 2020-08-14
Resolve requests for federation entities in parallel

### LINKED ISSUE #1278: Federation __resolve_entities should be non-blocking (async)
### What happened?
I generated service with federation enabled and got some code, generated inside method `__resolve_entities`. Basically all it doest it's just fetching product by id (implemented `FindProductByID` method inside `entity.resolvers.go`). When i run service and using dataloader for products to load batch of products instead of fetching one by one, dataloader blocks each call of `FindProductByID` because generated `executionContext.__resolve_entities` method calls it synchronously for each of `representation`.

### What did you expect?
That `executionContext` runs `Find{Entity}By{Key}` in async mode.

```
// inside products service
extend type Query {
    products: [Product!]
}

type Product implements Node @key(fields: "id") {
    id: ID!
    name: String!
}
```

```
// inside other service, extending products
extend type Product @key(fields: "id") {
    id: ID! @external
    testField: String!
}

// And make query 
query {
  products {
    id
    testField
    name
  }
}
```

### versions
 - `gqlgen version` v0.11.0
 - `go version` 1.14
 - dep or go modules - go modules

### What code is being generated inside `federation.go` now:
```
func (ec *executionContext) __resolve_entities(ctx context.Context, representations []map[string]interface{}) ([]fedruntime.Entity, error) {
	list := []fedruntime.Entity{}
	for _, rep := range representations {
		typeName, ok := rep["__typename"].(string)
		if !ok {
			return nil, errors.New("__typename must be an existing string")
		}
		switch typeName {

		case "Product":
			id0, err := ec.unmarshalNID2string(ctx, rep["id"])
			if err != nil {
				return nil, errors.New(fmt.Sprintf("Field %s undefined in schema.", "id"))
			}

			entity, err := ec.resolvers.Entity().FindProductByID(ctx,
				id0)
			if err != nil {
				return nil, err
			}

			list = append(list, entity)

		default:
			return nil, errors.New("unknown type: " + typeName)
		}
	}
	return list, nil
}
```

### What i suggest:
```
func (ec *executionContext) __resolve_entities(ctx context.Context, representations []map[string]interface{}) ([]fedruntime.Entity, error) {
	var (
		wg         sync.WaitGroup
                list = []fedruntime.Entity{}
		errCh      = make(chan error)
		entitiesCh = make(chan fedruntime.Entity)
                done = make(chan bool)
		err  error
	)

	for _, rep := range representations {
		wg.Add(1)
		go func(rep map[string]interface{}) {
			defer wg.Done()
			typeName, ok := rep["__typename"].(string)
			if !ok {
				errCh <- errors.New(fmt.Sprintf("Field %s undefined in schema.", "id"))
				return
			}
			switch typeName {

			case "Product":
				id0, err := ec.unmarshalNID2string(ctx, rep["id"])
				if err != nil {
					errCh <- errors.New(fmt.Sprintf("Field %s undefined in schema.", "id"))
					return
				}

				entity, err := ec.resolvers.Entity().FindProductByID(ctx,
					id0)
				if err != nil {
					errCh <- err
					return
				}

				entitiesCh <- entity

			default:
				errCh <- errors.New("unknown type: " + typeName)
			}
		}(rep)
	}

	go func() {
		wg.Wait()
		done <- true
	}()

	for {
		select {
		case err = <-errCh:
			return nil, err
		case e := <-entitiesCh:
			list = append(list, e)
		case <-done:
			return list, nil
		default:
		}
	}
}

```

benjaminjkraft:
In apollo federation, we may be asked for data about a list of entities.
These can typically be resolved in parallel, just as with sibling fields
in ordinary GraphQL queries.  Now we do!

I also changed the behavior such that if one lookup fails, we don't
cancel the others.  This is more consistent with the behavior of other
resolvers, and is more natural now that they execute in parallel.  This,
plus panic handling, required a little refactoring.

The examples probably give the clearest picture of the changes. (And the
clearest test; the changed functionality is already exercised by
`integration-test.js` as watching the test server logs will attest.)

Fixes #1278.

I have:
 - [x] Added tests covering the bug / feature: federation example integration test covers this
 - [x] Updated any relevant documentation: none applicable

### DIFF
--- a/plugin/federation/federation.gotpl
+++ b/plugin/federation/federation.gotpl
@@ -2,6 +2,7 @@
 {{ reserveImport "errors"  }}
 {{ reserveImport "fmt"  }}
 {{ reserveImport "strings"  }}
+{{ reserveImport "sync"  }}
 
 {{ reserveImport "github.com/99designs/gqlgen/plugin/federation/fedruntime" }}
 
@@ -25,12 +26,20 @@ func (ec *executionContext) __resolve__service(ctx context.Context) (fedruntime.
 }
 
 {{if .Entities}}
-func (ec *executionContext) __resolve_entities(ctx context.Context, representations []map[string]interface{}) ([]fedruntime.Entity, error) {
-	list := []fedruntime.Entity{}
-	for _, rep := range representations {
+func (ec *executionContext) __resolve_entities(ctx context.Context, representations []map[string]interface{}) []fedruntime.Entity {
+	list := make([]fedruntime.Entity, len(representations))
+	resolveEntity := func(ctx context.Context, i int, rep map[string]interface{}) (err error) {
+		// we need to do our own panic handling, because we may be called in a
+		// goroutine, where the usual panic handling can't catch us
+		defer func () {
+			if r := recover(); r != nil {
+				err = ec.Recover(ctx, r)
+			}
+		}()
+
 		typeName, ok := rep["__typename"].(string)
 		if !ok {
-			return nil, errors.New("__typename must be an existing string")
+			return errors.New("__typename must be an existing string")
 		}
 		switch typeName {
 		{{ range .Entities }}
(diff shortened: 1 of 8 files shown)


plugin/federation/federation.gotpl
```diff
@@ -27,43 +28,64 @@ func (ec *executionContext) __resolve__service(ctx context.Context) (fedruntime.
...
-				{{ end }}
-				list = append(list, entity)
-			{{ end }}
-		{{ end }}
-		default:
-			return nil, errors.New("unknown type: "+typeName)
-		}
-	}
-	return list, nil
+    resolveEntity := func(ctx context.Context, rep map[string]interface{}) error {
```
csilvers, line 31: github is showing that these are spaces rather than tabs.

benjaminjkraft: egad, they are!  fixed.

benjaminjkraft:
FYI, this turns out not to handle panics the proper way.  I'll add that, possibly not until later this month.

benjaminjkraft:
Ok, I've now handled panics.  I also changed things so that the entities resolver will report but continue on errors, which seems to be more of the intended behavior and more consistent with other resolvers, even moreso once said resolvers are in parallel.

ToonvanStrijp:
@benjaminjkraft when can we expect this to be released? I'm kinda waiting for this one 🎉  (Great work btw!)

StevenACoffman:
@MichaelJCompton  Can you review this for inclusion?

grzn:
Thanks for merging this.
Any plans to release 0.14.1 with this?

grzn:
Thanks for merging this.
Any plans to release 0.14.1 with this?

grzn:
Hi again,
Any plans to release 0.14.1 with this?

grzn:
@lwc  hi, can you please release v0.14.1?

PR #1289, opened 2020-08-16
Fix panic in template

marvinkite:
With `v0.12.1-dev` the code generator panic in type.gotpl at Line49 `var res = new({{ $type.Elem.GO | ref }})` if Elem returns nil.

With Named types that's not Pointer it happens with the following code.

```yaml
models:
    ID:
        model:
          - github.com/a/b/model.UUID
```

```golang
package model

import (
	"fmt"
	"io"

	"github.com/pborman/uuid"
)

type (
	UUID            uuid.UUID
)

//MarshalGQL implements the graphql.Marshaler interface
func (c UUID) MarshalGQL(w io.Writer) {
	data, _ := uuid.UUID(c).MarshalText()
	if data != nil {
		_, _ = io.WriteString(w, `"`)
		_, _ = w.Write(data)
		_, _ = io.WriteString(w, `"`)
	}
}

// UnmarshalGQL implements the graphql.Marshaler interface
func (c *UUID) UnmarshalGQL(v interface{}) (err error) {
	if data, ok := v.(string); !ok {
		return fmt.Errorf("ID must be strings")
	} else {
		id := uuid.Parse(data)
		if id == nil {
			return fmt.Errorf("invalid UUID value")
		}
		*c = UUID(id)
	}
	return
}
```

### DIFF
--- a/codegen/config/binder.go
+++ b/codegen/config/binder.go
@@ -188,6 +188,19 @@ func (ref *TypeReference) Elem() *TypeReference {
 		}
 	}
 
+	if p, isNamed := ref.GO.(*types.Named); isNamed {
+		return &TypeReference{
+			GO:          p.Underlying(),
+			Target:      ref.Target,
+			GQL:         ref.GQL,
+			CastType:    ref.CastType,
+			Definition:  ref.Definition,
+			Unmarshaler: ref.Unmarshaler,
+			Marshaler:   ref.Marshaler,
+			IsMarshaler: ref.IsMarshaler,
+		}
+	}
+
 	if ref.IsSlice() {
 		return &TypeReference{
 			GO:          ref.GO.(*types.Slice).Elem(),


marvinkite:
The PR does not help because it generates a code which is not compile.
```golang
func (ec *executionContext) unmarshalOID2githubᚗcomᚋa/bᚋmodelᚐUUID(ctx context.Context, v interface{}) (model.UUID, error) {
	if v == nil {
		return nil, nil
	}
	var res = new(model.UUID) // Returns *model.UUID
	err := res.UnmarshalGQL(v)
	return res, graphql.WrapErrorWithInputPath(ctx, err)
}
```
The correct code would be: 

```golang
func (ec *executionContext) unmarshalOID2githubᚗcomᚋa/bᚋmodelᚐUUID(ctx context.Context, v interface{}) (*model.UUID, error) {
	if v == nil {
		return nil, nil
	}
	var res = new(model.UUID)
	err := res.UnmarshalGQL(v)
	return res, graphql.WrapErrorWithInputPath(ctx, err)
}
```

This is a special type `UUID = uuid.UUID = []byte` which is  IsNilable() is true but IsSlice() is false. The proper fix should be somewhere else to set the `$type.GO` to `*model.UUID` and not `model.UUID`

```
{{- range $type := .ReferencedTypes }}
	{{ with $type.UnmarshalFunc }}
		func (ec *executionContext) {{ . }}(ctx context.Context, v interface{}) ({{ $type.GO | ref }}, error) {
```

PR #1292, opened 2020-08-17
fix: same field name in fragments on different types

### LINKED ISSUE #1280: Same field in two different fragments is getting ignored in function collectFields.
### What happened?
If we have field with same name in two different types which implement a common interface, and then while using fragments for the two different types to fetch that field inside the query for the common interface, only one field is returned.
It's happening in `getOrCreateAndAppendField` inside `collectFields` function.

**Query:**
```

query {
getStuff(id: "0x7536") {
things {
  ...on ThingOne { creator { id } }
  ...on ThingTwo { creator { id } }

          }
     }
}
```

### What did you expect?
`collectFields` should return below query, where one creator is of type ThingOne and another one  is of type ThingTwo.
```

query {
getStuff(id: "0x7536") {
things {
        creator { id } 
        creator { id } 

          }
     }
}
```

Reason for this is, currently in `getOrCreateAndAppendField` function, we are not checking  type of the field.
Instead we are just iterating over all the collected fieds and comparing names.
If  field is already present then returning it, even if they belong to different types,  in this case to different fragments.

**Schema:**
```
type User {
     id: ID!
     name:String!
}

interface Things {
    id: ID!
    name: String!
}

type Stuff {
    id: ID!
    things: [Things!]
}

type ThingOne implements Things {
    creator: User!
}

type ThingTwo implements Things {
    creator: User!
}


```

### versions
 - `gqlgen version`? gqlgen@v0.11.0
 - `go version`? go1.14.7
 - dep or go modules?

abhimanyusinghgaur:
Fixes #1280

I have:

### DIFF
--- a/graphql/context_operation_test.go
+++ b/graphql/context_operation_test.go
@@ -78,4 +78,32 @@ func TestCollectAllFields(t *testing.T) {
 		s := CollectAllFields(ctx)
 		require.Equal(t, []string{"fieldA", "fieldB"}, s)
 	})
+
+	t.Run("collect fragments with same field name on different types", func(t *testing.T) {
+		ctx := testContext(ast.SelectionSet{
+			&ast.InlineFragment{
+				TypeCondition: "ExampleTypeA",
+				SelectionSet: ast.SelectionSet{
+					&ast.Field{
+						Name:             "fieldA",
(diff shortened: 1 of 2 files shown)


abhimanyusinghgaur:
Bad Request, wanted for a different Repo. Closing.

PR #1303, opened 2020-08-21
Update README.md

kunalpowar:
Added ability to use custom models and forcing resolvers using inline config with directives.


vektah:
Thanks, theres a fair bit of overlap between this and the config doc now but more visibility is better.

Issue #1265, opened 2020-07-27
[Proposal] Split generated stubs into multiple files


kevinmbeaulieu:
# Problem

#377 introduced the problem that gqlgen's output file can get quite large (in our case, it's over 4MB / 100k lines and we've only been building on GraphQL for ~2 years- it will likely grow further as our schema continues to mature). Such large file sizes can still be problematic for a few reasons:
- Merely opening the file can take seconds to load.
- It's near impossible to find what you're looking for in a file that big.
- It's hard to get a high-level understanding of the generated code's structure because each semantic section might be thousands/tens of thousands of lines long.
- Some code review systems can be configured with a file size limit, which the current output file may exceed, forcing teams like ours to write custom post-processing scripts to break up the output of gqlgen into smaller files.

Based on the upvotes/comments on that original issue, it seems to be a relatively common problem in the community, so I would like to propose an actionable solution (which I can tentatively plan to implement if it's accepted).

# Proposed solution

gqlgen already splits up the templates for the generated code into `args.gotpl`, `directives.gotpl`, `field.gotpl`, `generated!.gotpl`, `input.gotpl`, `interface.gotpl`, `object.gotpl`, and `type.gotpl`. The code generation process essentially generates code based on each of these templates, adds a header, and concatenates the results into the file specified by `templates.Options.Filename`.

I propose that `templates.Options` be updated to support passing a directory into `Filename`. If `Filename` is a single file, behavior would be unchanged. If `Filename` is a directory, it would output files for each template as separate files. 

For instance, the following `gqlgen.yaml`
```
...
exec:
    package: schema
    filename: ./generated/
...
```
would yield the file structure:
```
generated/
    args.generated.go
    directives.generated.go
    field.generated.go
    generated!.generated.go
    input.generated.go
    interface.generated.go
    object.generated.go
    type.generated.go
```

The header for each `*.generated.go` would follow the same format as the single output file's header, so each output file would look like:
```
[If cfg.GeneratedHeader] // Code generated by github.com/99designs/gqlgen, DO NOT EDIT. 

[cfg.PackageDoc]

package [cfg.PackageName]

[cfg.FileNotice]

import (
    ...
)

[Code generated by `*.gotpl` template for this region]
```

## How

In `templates.Render()`, update the loop over `roots` to write to its own file, rather than a single buffer. Also move the header lines into the loop so they get added to each file.

## Open questions

[...]

PR #1309, opened 2020-08-26
Add follow-schema layout for exec

kevinmbeaulieu:
Implements #1265. See the issue for more details, but to summarize:

When `ExecLayout` is set to `follow-schema`, output generated code to a directory instead of a single file. Each file in the output directory will correspond to a single `*.graphql` schema file (plus a `root!.generated.go` file containing top-level definitions that are not specific to a single schema file).

`ExecLayout` defaults to `single-file`, which is the current behavior, so this new functionality is opt-in.

These layouts expose similar functionality to the `ResolverLayout`s with the same name, just applied to `exec` instead of `resolver`.

I have:


kevinmbeaulieu:
Status update: still working on adding more tests to this and fixing some bugs with some edge cases

kevinmbeaulieu:
Status update: I started adding more unit tests for this change, but found a couple corner cases where my current implementation is failing. I've had to put this on hold for now to focus on other work, but I'll try to get back to it when I do have more time. Otherwise, someone else is welcome to pick it up where I've left off too! :)

kevinmbeaulieu:
@vektah I finally got back to this and got everything working as far as I can tell. It looks like a big change so I broke it up into a few commits to make it easier to review:
1. Add Visual Studio Code settings files to .gitignore (alongside IntelliJ IDEA)
2. Create a separate `ExecConfig` type based on `PackageConfig` but with additional fields to support the new layouts
3. Implement two layouts: `single-file`, which corresponds to the baseline functionality of outputting all code to a single `generated.go` file, and `follow-schema`, which implements the functionality described in #1265.
4. Add unit tests to verify that the code generated by `single-file` and `follow-schema` is functionally equivalent (i.e. defines all the same functions, types, etc., just organized differently). This was the main challenge that took months to figure out since there weren't any existing tools that quite served this need, so I developed a new tool [eq-go](https://github.com/kevinmbeaulieu/eq-go/) to handle comparing the packages generated by the two layouts. As part of this change, I basically created two copies of the code in `codegen/testserver`, in `codegen/testserver/singlefile` and `codegen/testserver/followschema` (and then deleted the original copy). So each of these directories tests the corresponding layout, and then there's a `codegen/testserver/generated_test.go` that runs eq-go to ensure the two directories are equivalent.

There seems to be one CI check still failing that I'm not sure how to fix, so any advice there would be appreciated.

kevinmbeaulieu:
@MichaelJCompton @lwc I'd been talking to @vektah last year about this change, but haven't heard from him so not sure if he's still active with gqlgen, so tagging you since you both look like you've been merging changes recently. Would love if one of you could take a look (or tag someone if there's someone specific that should review it). See #1265 for more context on the proposed change.

asif-mahmud:
any plan to merge this pr ?

kevinmbeaulieu:
@asif-mahmud I would love to, if one of the maintainers can review it! 😄

kevinmbeaulieu:
Seeing a few remaining test failures that look like they are related to my changes- will take a look tomorrow

kevinmbeaulieu:
@StevenACoffman -  Just updated the tests and they all pass locally now, if you want to trigger a final build and then merge. Thanks for your help getting this over the finish line!

StevenACoffman:
There were some test failures, so I rebased and regenerated so they passed. I hope that's ok. Thanks!

codegen/generate.go
```diff
@@ -14,3 +39,176 @@ func GenerateCode(data *Data) error {
...
+		SubscriptionRoot: data.SubscriptionRoot,
+		AllDirectives:    data.AllDirectives,
+	}
+}
+
+// Root file contains top-level definitions that should not be duplicated across the generated
+// files for each schema file.
+func generateRootFile(data *Data) error {
+	dir := data.Config.Exec.DirName
+	path := filepath.Join(dir, "root!.generated.go")
```
rthille, line 130: We're seeing issues with our tooling due to the '!' in this filename. Any reason why it has to have an exclamation point in it?  Quoting with "" doesn't even help, so it can be tricky to deal with.
```
bash$ echo "!root"
bash: !root: event not found
bash$ echo '!root'
!root
bash$
```

kevinmbeaulieu: My original intention was to minimize potential naming conflicts in case people had schema files named `root.graphql` in their project, but no reason it needs to be an exclamation point- could be changed to another symbol or just accept the possibility of conflicts

yar00001: Thank you @kevinmbeaulieu - just sent a PR to change the name to `root_.generated.go`
https://github.com/99designs/gqlgen/pull/1701

Issue #1291, opened 2020-08-17
INPUT_FIELD_DEFINITION directive handler returned errors are clobbered


dixieflatline76:
### What happened?
When using an INPUT_FIELD_DEFINITION directive, if the implemented directive handler returns a custom error (meant for leveraging custom error presenter to deliver tailored error graph style and info), it is overwritten with a gqlerror with only the error string copied over. This is a break introduced in v0.12.x, same code runs fine in v0.11.x and v0.10.x.

### What did you expect?
Errors returned in any directive handlers should be propagated without lost of information until the custom error presenters are run.

### Minimal graphql.schema and models to reproduce
directive @titleField(fieldName: String) on INPUT_FIELD_DEFINITION
directive @genresField(fieldName: String) on INPUT_FIELD_DEFINITION

input NewGame {
    title: String! @titleField # Add directive to input type field to pattern validation and string case normalization
    genres: [Genre!]! @genresField(fieldName:"genres") # Refer to demoserver.go to see we check for min 1 and max of 2 genres, and they must be unique
    platforms: [NewGamePlatform!]!
}

### versions
 - `gqlgen version` v0.12.1
 - `go version`? 1.14.2
 - dep or go modules?

lwc:
@dixieflatline76 can you confirm that v0.13.0 fixes this for you, after swapping from type assertions to `errors.As`?

dixieflatline76:
@lwc thanks, I will test this out.

dixieflatline76:
@lwc just curious, looks like the default error handler just does a hard cast no test on the incoming error (i noticed it cause it crashed one of my unit tests). Wouldn't it be better to either just change the signature graphql.ErrorPresenterFunc and the default error presenter to accept a gqlerror?

dixieflatline76:
> @dixieflatline76 can you confirm that v0.13.0 fixes this for you, after swapping from type assertions to `errors.As`?

I can confirm that the issue I reported is fixed with v0.13.0. Thank you!

PR #1312, opened 2020-08-28
Always wrap user errors

lwc:
This is an alternative to #1305 that leans into error wrapping instead of away from it.

Requires use of go 1.13 error unwrapping.

On measure I think I prefer this approach, even though it's a bigger BC break:
- There's less mutex juggling
- It has never felt right to me that we make the user deal with path when overriding the error presenter
- The default error presenter is now incredibly simple

### Questions:
- Are we comfortable with supporting 1.13 and up?
- Should we change the signature of `ErrorPresenterFunc` to `func(ctx context.Context, err *gqlerror.Error) *gqlerror.Error`?
    - It always is now, and breaking BC will force users to address the requirement for `errors.As`

Fixes: #1291

I have:


mtibben (approved):
Yes this looks like a superior approach to me comparing to #1305 

> Are we comfortable with supporting 1.13 and up?

+1 from me. Personally I think supporting the just last 2 versions of Go is totally justifiable, Go has high focus on backwards compatibility and if you really want to use an older go, you can use an older gqlgen release too

> Should we change the signature of ErrorPresenterFunc to func(ctx context.Context, err *gqlerror.Error) *gqlerror.Error?

Yes I think so. Lets make the BC break obvious

Issue #1293, opened 2020-08-17
v0.12.1 crashes inside gotpl where v0.11.3 gives error messages


jwatte:
I'm running gqlgen on a schema that has some errors.
v0.12.1 gives a very unhelpful error message:

```
[21:20:52] dev@devenv-55d5ff879f-9blcj:~/observe/code/go/src/observe/meta/metagql$ go generate
generating core failed: type.gotpl: template: type.gotpl:49:28: executing "type.gotpl" at <$type.Elem.GO>: nil pointer evaluating *config.TypeReference.GOexit status 1
resolver.go:1: running "go": exit status 1

[gqlgen-crash.zip](https://github.com/99designs/gqlgen/files/5086900/gqlgen-crash.zip)

```
v0.11.3 is more helpful:

```
[21:26:17] dev@devenv-55d5ff879f-9blcj:~/observe/code/go/src/observe/meta/metagql$ go generate
validation failed: packages.Load: /home/dev/observe/code/go/src/observe/meta/metagql/gql_dataset.go:700:33: cannot use ret (variable of type []*metatypes.Dataset) as *metatypes.Dataset value in argument to updateDatasetNilsToEmptyArrays
/home/dev/observe/code/go/src/observe/meta/metagql/progressive.go:443:24: cannot use results (variable of type []*metaparser.TaskResult) as []metaparser.TaskResult value in argument to canSendResults
/home/dev/observe/code/go/src/observe/meta/metagql/progressive.go:493:23: cannot use results (variable of type []*metaparser.TaskResult) as []metaparser.TaskResult value in argument to canSendResults
/home/dev/observe/code/go/src/observe/meta/metagql/progressive.go:517:21: cannot use results (variable of type []*metaparser.TaskResult) as []metaparser.TaskResult value in argument to canSendResults
exit status 1
resolver.go:1: running "go": exit status 1

```

I expect gqlgen to not crash with an error message that tells me nothing about where the cause for the error is, instead I expect it to give me good error messages.

(I think 0.11.3 actually generated the code and the errors come from the go compile command, so maybe this reduces to "v0.12.1 shouldn't crash, period!")

Attaching schema.graphql and gqlgen.yml files

tsaikd:
I faced the issue when upgrading to v0.12.2 from v0.11.3. I guess it's a bug that assume Nilable Marshaler will have Elem at the same time. The following patch can work around, but I',m not sure any other edge cases will happen.

```diff
[...]

PR #1317, opened 2020-09-01
codegen: ensure Elem present before using

tmc:
This addresses the problem described in https://github.com/99designs/gqlgen/issues/1293


DAddYE:
I'm hitting the same issue, and this seems to fix it.

ikesyo:
The issue is a blocker for us to upgrade this library.  Could you please take a look? @mtibben @lwc

vektah:
Sorry, I started trying to reverse engineer a test case for this but havent had any luck yet. Because this is exercising binder errors its going to require a little thinking on how we can make these tests maintainable and prevent future regressions. I'm hoping to set some time aside this weekend.

hardlyknowem:
@vektah Is there any way the restrictions on merging without a test can be relaxed in this instance? This has been blocking us and other users from moving past version 11 for quite a while now. (See #1293 and #1391). There's been a similar fix (see #1392) that has also been stalled due to lack of testing.

While I absolutely understand and respect hesitation to proceed without new tests, it seems, from your last comment here, that actually writing a test for this case would require a significant amount of effort and/or refactoring. Would it be possible to apply this as it is a known fix and then raise a separate issue about the lack of test coverage?

PR #1347, opened 2020-10-02
Add ability to generate int enums

### LINKED ISSUE #366: Enum with int values
Is it possible to generate enum with int values rather than string values?  Something that looks like the below code? 

``` 
type Op int32

const (
	Op_NULL   Op = 0
	Op_CREATE Op = 1
	Op_MODIFY Op = 2
	Op_DELETE Op = 3
)

var Op_name = map[int32]string{
	0: "NULL",
	1: "CREATE",
	2: "MODIFY",
	3: "DELETE",
}
var Op_value = map[string]int32{
	"NULL":   0,
	"CREATE": 1,
	"MODIFY": 2,
	"DELETE": 3,
} 

```
```
enum Op {
        Op_NULL 
	Op_CREATE
	Op_MODIFY
	Op_DELETE
}
```
Generates 
```
type Op string

const (
	OpOpNull   Op = "Op_NULL"
	OpOpCreate Op = "Op_CREATE"
	OpOpModify Op = "Op_MODIFY"
	OpOpDelete Op = "Op_DELETE"
)

func (e Op) IsValid() bool {
	switch e {
	case OpOpNull, OpOpCreate, OpOpModify, OpOpDelete:
		return true
	}
	return false
}

func (e Op) String() string {
	return string(e)
}

func (e *Op) UnmarshalGQL(v interface{}) error {
	str, ok := v.(string)
	if !ok {
		return fmt.Errorf("enums must be strings")
	}

	*e = Op(str)
	if !e.IsValid() {
		return fmt.Errorf("%s is not a valid Op", str)
	}
	return nil
}

func (e Op) MarshalGQL(w io.Writer) {
	fmt.Fprint(w, strconv.Quote(e.String()))
}

```
 Is there any way to generate int values rather than string values for enums? Perhaps a custom resolver?

holykol:
This PR adds `@intEnum` directive, which, when applied, changes enums to be generated with int values.
Sample generated code: https://gist.github.com/holykol/c8056f4bb80c0b17e3478a85cbd27be2

I have:
~~Updated any relevant documentation~~ Any suggestions for where to put documentation for this new feature?

closes #366

### DIFF
--- a/codegen/testserver/models-gen.go
+++ b/codegen/testserver/models-gen.go
@@ -231,14 +231,13 @@ func (e EnumTest) String() string {
 }
 
 func (e *EnumTest) UnmarshalGQL(v interface{}) error {
-	str, ok := v.(string)
+	value, ok := v.(string)
 	if !ok {
 		return fmt.Errorf("enums must be strings")
 	}
-
-	*e = EnumTest(str)
+	*e = EnumTest(value)
 	if !e.IsValid() {
-		return fmt.Errorf("%s is not a valid EnumTest", str)
+		return fmt.Errorf("%v is not a valid EnumTest", value)
 	}
 	return nil
 }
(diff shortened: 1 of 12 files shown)


tarrencev:
would love this!

mpieczaba:
Looking forward to the merge...

kateile:
Any progress on this?

FlymeDllVa:
I think what's missing is a directive to set the digits for the value by itself. For the same proto, we can set numbers randomly. This would be useful along with using the goModel directive on proto types

PR #1356, opened 2020-10-10
Chore: update Chat example

maapteh:
Im just new in learning Go. GraphQL and React are not new for me. So i started with the Chat demonstration and thought lets update Apollo Client and use hooks in the application, so deps are new.

I also added some small tweaks to scroll to below on new data, and if needed i can make it look even nicer. 

Screenshot 2020-10-12 at 15 51 51

### DIFF
--- a/example/chat/src/App.js
+++ b/example/chat/src/App.js
@@ -2,7 +2,7 @@ import React, { useState } from 'react';
 import styled from 'styled-components';
 import { Room } from './Room';
 
-export const Input = styled.div`
+const Input = styled.div`
     padding: 4px;
     margin: 0 0 4px;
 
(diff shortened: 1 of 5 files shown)


maapteh:
@vektah if you find some time please let me know

example/chat/src/App.js
```diff
@@ -1,28 +1,33 @@
-import React, { Component } from 'react';
-import Room from './Room';
+import React, { useState } from 'react';
+import styled from 'styled-components';
+import { Room } from './Room';
 
-class App extends Component {
-    constructor(props) {
-        super(props);
+export const Input = styled.div`
```
vektah, line 5: This probably doesnt need to be exported

```suggestion
const Input = styled.div`
```

vektah:
Thanks, this was long overdue for an update.

PR #1360, opened 2020-10-15
Update introspection.md

Captain-K-101:
updated disabling introspect query

The introspect Disable was not modified as now we have to use `AroundOperations`


appleboy (approved):

vektah:
Thanks, I've updated the rest of the docs to use the new interfaces too.

PR #1365, opened 2020-10-20
add uint, uint64, uint32 types in graphql pkg

frederikhors:
I added `uint`, `uint32` and `uint64` in `graphql` package.

I hope you'll merge because I love Open Source Software and I think it is better to have this code here so it is checked daily by multiple eyes.

This is **especially useful** now that with Gorm 2 we have the [default model](https://github.com/go-gorm/gorm/blob/master/model.go) which is:

```go
type Model struct {
	ID        uint `gorm:"primarykey"`
	CreatedAt time.Time
	UpdatedAt time.Time
	DeletedAt DeletedAt `gorm:"index"`
}
```

I have:

### DIFF
--- a/graphql/uint.go
+++ b/graphql/uint.go
@@ -1,0 +1,81 @@
+package graphql
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"strconv"
+)
+
+func MarshalUint(i uint) Marshaler {
+	return WriterFunc(func(w io.Writer) {
+		_, _ = io.WriteString(w, strconv.FormatUint(uint64(i), 10))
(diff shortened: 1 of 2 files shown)


Matts966:
@frederikhors 
Is this mergeable?

icellan:
This seems fairly simple to integrate. Is this still going to happen? Would really want to use this.

frederikhors:
I think we should start an active fork.

carldunham:
Agreed. Happy to participate.

mtibben:
Thanks @frederikhors and sorry for the delay

PR #1374, opened 2020-10-24
Fix small typo in file upload docs

rudylee:
This PR fixes small typo in file upload documentation. It should be `Boolean` instead of `Bool`

### DIFF
--- a/docs/content/reference/file-upload.md
+++ b/docs/content/reference/file-upload.md
@@ -59,9 +59,7 @@ That invokes the following operation:
 {
   query: `
     mutation($file: Upload!) {
-      singleUpload(file: $file) {
-        id
-      }
+      singleUpload(file: $file)
     }
   `,
   variables: {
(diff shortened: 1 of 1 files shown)


docs/content/reference/file-upload.md
```diff
@@ -40,15 +40,15 @@ type Query {
...
+    singleUpload(file: Upload!): Boolean!
 }
 ```
 
 cURL can be used the make a query as follows:
 
 ```
 curl localhost:4000/graphql \
-  -F operations='{ "query": "mutation ($file: Upload!) { singleUpload(file: $file) { id } }", "variables": { "file": null } }' \
+  -F operations='{ "query": "mutation ($file: Upload!) { singleUpload(file: $file) }", "variables": { "file": null } }' \
```
rudylee, line 51: The mutation returns a boolean so there is no `id` field

PR #1389, opened 2020-11-12
👷  add dependabot.yml

uta-mori:
I have:
 - [x] Added dependency management ci feature 

Add dependabot,yml. 
This automatically creates a pull request for dependency updates.
I think it will make development a little bit easier.
[About Dependabot version updates - GitHub Docs](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/about-dependabot-version-updates)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


mtibben:
While bumping deps constantly might be common in other language ecosystems, I don't think it's well suited for the Go "minimal version selection" approach. Stability is more important

PR #1392, opened 2020-11-12
fix(codegen): adress case of nullable types which have no known elements (e.g. external scalar type)

### LINKED ISSUE #1391: Codegen broken since v0.12 using external map scalar types
### What happened?

* I have a schema with imported scalar types.
* works fine with gqlgen =< v0.11.3
* breaks start with v0.12:
```
codegen: generating core failed: type.gotpl: template: type.gotpl:49:28: executing "type.gotpl" at <$type.Elem.GO>: nil pointer evaluating *config.TypeReference.GO
```

The faulty template code in `codegen/type.gotpl` is:
```
@@ -46,7 +46,21 @@
                                        return v.(map[string]interface{}), nil
                                {{- else if $type.IsMarshaler }}
                                        {{- if $type.IsNilable }}
-                                               var res = new({{ $type.Elem.GO | ref }})
```

Investigating further, the template fails when looping over my scalar types.

In my case, scalar types are maps, e.g. `map[string]interface{}`, `map[string]string`, `map[string]float64`.
This hits the case of a nullable type with no (known) element type (there is one, but it is not introspected, as only slices are introspected for the type of their element).

### What did you expect?
Successful codegen

 (TODO)

### versions
 - `gqlgen version`? v0.12.x, v0.13.0, current master
 - `go version`? 1.15.3

fredbi:
This PR covers the case when the type is nullable, but element is not defined.

* fixes #1391

Signed-off-by: Frederic BIDON <fredbi@yahoo.com>

I have:

### DIFF
--- a/codegen/type.gotpl
+++ b/codegen/type.gotpl
@@ -46,8 +46,12 @@
 					return v.(map[string]interface{}), nil
 				{{- else if $type.IsMarshaler }}
 					{{- if $type.IsNilable }}
-						var res = new({{ $type.Elem.GO | ref }})
-					{{- else}}
+                      {{- if $type.Elem }}
+						  var res = new({{ $type.Elem.GO | ref }})
+                      {{- else }}
+						  var res {{ $type.GO | ref }}
+                      {{- end }}
+                    {{- else}}
 						var res {{ $type.GO | ref }}
 					{{- end }}
 					err := res.UnmarshalGQL(v)


vektah:
We cant accept codegen fixes without a test that exercises the new code.

Please add something to https://github.com/99designs/gqlgen/tree/master/codegen/testserver that fails without this change.

fredbi:
@vektah ok will do. That was expected. Just wanted to know if my change was going in the right direction

oiime:
@fredbi Still fails with []string 

`generating core failed: type.gotpl: template: type.gotpl:49:28: executing "type.gotpl" at <$type.Elem.GO>: nil pointer evaluating *config.TypeReference.GO`

for: 

`type Texts []string`

fredbi:
@StevenACoffman after upgrading to v0.17 the need for this disappeared and map scalars are properly generated. Closing this now.

PR #1394, opened 2020-11-15
bugfix: Default Recover func should return gqlerror.Error

j2gg0s:
`DefaultRecover` should return gqlerror.Error, since `DefaultErrorPresenter` expected that.


Sytten:
I hit this bug too, it is fairly annoying that the default error handlers causes a panic. Can this be merged ASAP @vektah? Thanks

lwc (approved):

PR #1399, opened 2020-11-25
Dont hold error lock when calling into error presenters

vektah:
This can result in a deadlock if the error handling code calls back to GetErrors.

Instead, only hold the lock while mutating the slice.


lwc (approved):

kaddy-tom:
Hello, is there any chance you could bump the version number to `0.13.1` for this bug fix?

PR #1404, opened 2020-11-30
feat: generate resolvers for inputs if fields are missing

### LINKED ISSUE #1403: Invalid generated code for inputs when go model field type does not match the graphql field type
Reopen #663

### What happened?
Invalid go code was generated.
### What did you expect?
I expect valid go code to be generated. The actual generated code is not building for the following reason:
`gql/exec.go:2298:17: cannot assign []*proj.MapEntry to it.Map (type map[string]string) in multiple assignment`

This is the generated go code part that is failing:
```go
case "map":
	var err error
	it.Map, err = ec.unmarshalOMapEntry(ctx, v) // failing here
	if err != nil {
		return it, err
	}
```

--- schema.graphqls
```graphql
type Query {
    foo(in: Bar): Boolean
}

input Bar {
    map: [MapEntry]
}
input MapEntry {
    key: String
    val: String
}
```

--- gqlgen.yaml
```yaml
schema:
- schema.graphqls
exec:
  filename: exec.go
model:
  filename: models.go
models:
  Bar:
    model:
    - github.com/company/proj.Bar
  MapEntry:
    model:
    - github.com/company/proj.MapEntry
```
--- models.go
```go
type Bar struct {
    Map map[string]string
}

type MapEntry struct {
    Key string
    Val string
}
```

I tried the same usecase with graphql `type` and an additional resolver was added that has as parameter `Bar` and returns `MapEntry`.

I am aware of the fact that graphql types and inputs are totally different concepts and why the generated code was broken but I think it would be very handy to have some kind of generated resolvers similar to the ones for the type or some workaround for similar usecases.
Maybe show an error at generation time?

danielvladco:

Generate resolvers for input type fields in case the user defined type is missing those fields.

fixes #1403

I have:
 - [  ] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [  ] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

### DIFF
--- a/codegen/field.go
+++ b/codegen/field.go
@@ -452,6 +452,9 @@ func (f *Field) GoNameUnexported() string {
 }
 
 func (f *Field) ShortInvocation() string {
+	if f.Object.Kind == ast.InputObject {
+		return fmt.Sprintf("%s().%s(ctx, &it, data)", f.Object.Definition.Name, f.GoFieldName)
+	}
 	return fmt.Sprintf("%s().%s(%s)", f.Object.Definition.Name, f.GoFieldName, f.CallArgs())
 }
 
@@ -472,6 +475,13 @@ func (f *Field) ResolverType() string {
 }
 
 func (f *Field) ShortResolverDeclaration() string {
+	if f.Object.Kind == ast.InputObject {
+		return fmt.Sprintf("(ctx context.Context, obj %s, data %s) error",
+			templates.CurrentImports.LookupType(f.Object.Reference()),
+			templates.CurrentImports.LookupType(f.TypeReference.GO),
+		)
+	}
+
 	res := "(ctx context.Context"
 
 	if !f.Object.Root {
--- a/codegen/generated!.gotpl
+++ b/codegen/generated!.gotpl
@@ -35,6 +35,11 @@ type ResolverRoot interface {
 		{{$object.Name}}() {{$object.Name}}Resolver
 	{{ end }}
 {{- end }}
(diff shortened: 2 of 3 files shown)


mojtabacazi:
@StevenACoffman Looks like this is broken, the generated code expects a resolver for inputs now but it fails to implement the resolver. Also, the model that was already matching using `@gomodel` directive is no longer matching:

```
type Bar struct {
   a, b string
}

type FooType []Bar
type Baz struct {
    Foo FooType
}


input SomeInput @goModel(model: "Baz") {
   foo: [Bar!]! # this used to get mapped to the go struct, but now it doesn't. 
}

input Bar @goModel(model: "Bar") {
   a: String!
   b: String!
}


```

eric-yoo:
> @StevenACoffman @danielvladco Looks like this is broken, the generated code expects a resolver for inputs now but it fails to implement the resolver. Also, the model that was already matching using `@gomodel` directive is no longer matching:
> 
> ```
> type Bar struct {
>    a, b string
> }
> 
> type FooType []Bar
> type Baz struct {
>     Foo FooType
> }
> 
> 
> input SomeInput @goModel(model: "Baz") {
>    foo: [Bar!]! # this used to get mapped to the go struct, but now it doesn't. 
> }
> 
> input Bar @goModel(model: "Bar") {
>    a: String!
>    b: String!
> }
> ```

I'm facing the same issue. Did you come across any workaround?

PR #1405, opened 2020-12-03
subscriptions: send complete message on resolver panic

alexsn:
Open subscriptions will not send 'complete' message nor remove resolver from `.active` map on resolver panic. I simply moved the termination block to the defer functions which is called regardless of panics.

### DIFF
--- a/graphql/handler/transport/websocket.go
+++ b/graphql/handler/transport/websocket.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 	"net/http"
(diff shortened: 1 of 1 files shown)


graphql/handler/transport/websocket.go
```diff
@@ -232,25 +232,27 @@ func (c *wsConnection) subscribe(start time.Time, message *operationMessage) {
 	go func() {
 		defer func() {
 			if r := recover(); r != nil {
-				userErr := rc.Recover(ctx, r)
-				c.sendError(message.ID, &gqlerror.Error{Message: userErr.Error()})
+				gqlerr := &gqlerror.Error{}
+				if err := rc.Recover(ctx, r); err != nil {
+					gqlerr.Message = err.Error()
```
lwc, line 237: Is there a reason you're throwing `err` away here? it's already a `gqlerror` at this point, and will have potentially useful information like `Path` already set.

alexsn: The existing code was doing the same, anyway I've update the PR to keep existing gqlerr in place.

PR #1417, opened 2020-12-12
Use mutation instead of query in 'Changesets' doc example

RicCu:

The Changesets doc has an example query `updateUser` that would modify data in the db, so it should be a mutation. The technical implementation of the example doesn't change, but a beginner might get the wrong idea/get confused about the separation between queries and mutations.

I have:


mtibben:
Thanks @RicCu

PR #1420, opened 2020-12-14
Add server json serialization.

ainilili:
Allows consumers to configure their own JSON serialization tools:
```go
srv := handler.NewDefaultServer(todo.NewExecutableSchema(todo.New()))
srv.SetSerialization(serial.Default())
```


ainilili:
@StevenACoffman Done.

mtibben:
I'm very concerned about the test coverage for a change like this which touches so many functions. Each function that now uses a serialiser really does now need to include tests for multiple serialisers. This feature seems like it would break very easily without that sort of coverage. Are there alternative approaches to implement this?

frederikhors:
Please, let's take these kinds of changes seriously. There are thousands of projects built with gqlgen and we're not even 1.0.0.

Issue #1274, opened 2020-08-09
The sample code in getting started cannot be compiled


ksoda:
### What happened?

The following command failed. This happens when you implement the code according to [Implement the resolvers](https://gqlgen.com/getting-started/#implement-the-resolvers)

```console
❯ go run server.go
graph/schema.resolvers.go:18:36: not enough arguments in call to "crypto/rand".Int
        have ()
        want (io.Reader, *big.Int)
graph/schema.resolvers.go:18:36: multiple-value "crypto/rand".Int() in single-value context
```

### What did you expect?

The server is successfully started.

### Minimal graphql.schema and models to reproduce

<details><summary>graphql.schema and models</summary>

```graphql
# GraphQL schema example
#
# https://gqlgen.com/getting-started/

type Todo {
  id: ID!
  text: String!
  done: Boolean!
  user: User!
}

type User {
  id: ID!
  name: String!
}

type Query {
  todos: [Todo!]!
}

input NewTodo {
  text: String!
  userId: String!
}

type Mutation {
  createTodo(input: NewTodo!): Todo!
}

```

```go
// Code generated by github.com/99designs/gqlgen, DO NOT EDIT.

package model

type NewTodo struct {
	Text   string `json:"text"`
	UserID string `json:"userId"`
}

type Todo struct {
	ID   string `json:"id"`
	Text string `json:"text"`
	Done bool   `json:"done"`
	User *User  `json:"user"`
}

type User struct {
	ID   string `json:"id"`
	Name string `json:"name"`
}
```

</details>

### versions
 - `gqlgen version`: v0.11.3
 - `go version`: go version go1.13.8 linux/amd64
 - dep or go modules?

ksoda:
This patch makes it boot up.

```diff
diff --git a/server-next/graph/schema.resolvers.go b/server-next/graph/schema.resolvers.go
index d14d5d1..ffa99c8 100644
--- a/server-next/graph/schema.resolvers.go
+++ b/server-next/graph/schema.resolvers.go
@@ -7,15 +7,18 @@ import (
        "context"
        "crypto/rand"
        "fmt"
+       "math/big"

        "github.com/ksoda/todo-app/graph/generated"
        "github.com/ksoda/todo-app/graph/model"
 )

 func (r *mutationResolver) CreateTodo(ctx context.Context, input model.NewTodo) (*model.Todo, error) {
+       rand, _ := rand.Int(rand.Reader, big.NewInt(100))
[...]

PR #1441, opened 2021-01-19
Fixed CreateTodo method

Surfoo:


MichaelJCompton:
Hi, thanks for submitting the PR.

I'm not clear on why this is a 'fix' for the method.  Please supply some context.

Surfoo:
Hello,

Did you really do the whole tutorial?
It's  `rand.Int` from the `crypto` pkg in the generated code, not from `math` pkg.

`rand.Int` on https://golang.org/pkg/crypto/rand/#Int needs `rand io.Reader`, `max *big.Int` in input and returns a `big.Int` and an `error` in output. The actual code in the tutorial doesn't compile.
I assumed we don't care about the error, it's a tutorial :)

And there are still other mistakes inside..

Surfoo:
https://github.com/99designs/gqlgen/issues/1274

MichaelJCompton:
Hi, I think the answer on the issue you linked is what you are looking for:

https://github.com/99designs/gqlgen/issues/1274#issuecomment-734899691

It doesn't say in the tutorial to use `crypto`, but I'll update the text to make it mention `math.rand`. 

I'll close this issue and please report the other errors as new issues so I can track them, thanks.

PR #1449, opened 2021-01-27
feat(codegen): handle (v, ok) methods

### LINKED ISSUE #1445: Explore model compatibility with Prisma
### What happened?

The [Prisma Go client](https://github.com/prisma/prisma-client-go) is currently not perfectly compatible with gqlgen.

The simple reason for that is that in order to work better with pointers, we decided to give the following abstraction to our users:

```go
user, err := prisma.User.FindFirst().Exec(ctx)

if name, ok := user.Name(); ok {
  log.Printf("user name is set! value: %s", name)
} else {
  log.Printf("name is null")
}

// if you really need to use the internal pointer value:
doSomething(user.InnerUser.Email)
```

That means our generated structs look as follows:

```go
// prisma_gen.go
type User struct {
  InnerUser
}

func (v User) Name() (string, bool) { // optional field
  if v.InnerName == nil {
    return "", false
  }
  return *v.InnerName, true
}

type InnerUser struct {
  Email string `json:"email"` // required field
}
```

If you're wondering why we chose this, there are two reasons:
1) Using pointers for nullability is a workaround, and is abusing pointers since there's nothing better around.
2) The shape of (value, ok) feels like Go and is used all over the standard packages.
3) A disadvantage of pointers is that you won't get warned for potential nil pointers, while with (value, ok) the "unused variable" compile time error check forces to you to check (or explicitly ignore) the boolean `ok` result.

### What did you expect?

It would be nice if this would work out of the box when you import Prisma models from gqlgen.yml.

However, right now it doesn't, since those methods are expected to be resolvers in shape of (interface{}, error):

```
validation failed: packages.Load: /Users/luca/projects/ws-api/gqlgen/gql_gen.go:1611:10: cannot use obj.Desc() (value of type bool) as error value in return statement: missing method Erro
```

in generated code:

image

```go
// ...
func (ec *executionContext) _User_desc(ctx context.Context, field graphql.CollectedField, obj *prisma.UserModel) (ret graphql.Marshaler) {
	defer func() {
		if r := recover(); r != nil {
			ec.Error(ctx, ec.Recover(ctx, r))
			ret = graphql.Null
		}
	}()
	fc := &graphql.FieldContext{
		Object:     "User",
		Field:      field,
		Args:       nil,
		IsMethod:   true,
		IsResolver: false,
	}

	ctx = graphql.WithFieldContext(ctx, fc)
	resTmp, err := ec.ResolverMiddleware(ctx, func(rctx context.Context) (interface{}, error) {
		ctx = rctx // use context from middleware stack in children
		return obj.Desc() // <============== HERE ================
	})
	if err != nil {
		ec.Error(ctx, err)
		return graphql.Null
	}
	if resTmp == nil {
		return graphql.Null
	}
	res := resTmp.(string)
	fc.Result = res
	return ec.marshalOString2string(ctx, field.Selections, res)
}
// ...
```

Now to my question:

How can I solve it?

I see a few different ways:

1) Workaround (that's what we do right now). Whenever you pass something to gqlgen, you simply refer to the underlying data structure, such as `user.InnerUser`.
   It's not super beautiful, but it works.

2) Add a special handling for methods of shape (interface{}, bool)
   This would need to be added to gqlgen in some way. The question is, I guess, whether that feature would make sense. FWIW, I’d b e happy to do the work and send a PR should we decide to do this.

  I guess the field shape would need to be checked and then in the template an extra if could be added to add this into the resolver middleware callback:

  ```go
  resTmp, err := ec.ResolverMiddleware(ctx, func(rctx context.Context) (interface{}, error) {
    ctx = rctx // use context from middleware stack in children
    if v, ok := obj.Desc(); ok {
      return v, nil
    }
    return nil, nil
  })
  ```

3) Same as 2, but somehow implemented that as a custom plugin/adapter. Not sure how that would look like or if it's even possible right now (I think it's not).

Does anyone have other suggestion? I also thought of ways to workaround it, but I couldn’t find any.

### versions
- `gqlgen version`?
- `go version`?
- dep or go modules?

steebchen:
This adds support for resolver fields which are in shape of (v interface{}, ok bool).

When:

- ok is false, nil is returned
- ok is true, v is returned

fix #1445 

I have:

### DIFF
--- a/codegen/field.go
+++ b/codegen/field.go
@@ -25,6 +25,7 @@ type Field struct {
 	Args             []*FieldArgument // A list of arguments to be passed to this field
 	MethodHasContext bool             // If this is bound to a go method, does the method also take a context
 	NoErr            bool             // If this is bound to a go method, does that method have an error as the second argument
+	VOkFunc          bool             // If this is bound to a go method, is it of shape (interface{}, bool)
 	Object           *Object          // A link back to the parent object
 	Default          interface{}      // The default value
 	Stream           bool             // does this field return a channel?
@@ -152,6 +153,8 @@ func (b *builder) bindField(obj *Object, f *Field) (errret error) {
 		sig := target.Type().(*types.Signature)
 		if sig.Results().Len() == 1 {
 			f.NoErr = true
+		} else if s := sig.Results(); s.Len() == 2 && s.At(1).Type().String() == "bool" {
+			f.VOkFunc = true
 		} else if sig.Results().Len() != 2 {
 			return fmt.Errorf("method has wrong number of args")
 		}
--- a/codegen/field.gotpl
+++ b/codegen/field.gotpl
@@ -112,7 +112,13 @@ func (ec *executionContext) _{{$object.Name}}_{{$field.Name}}(ctx context.Contex
 			return nil, fmt.Errorf("unexpected type %T for field %s", v, {{ .Name | quote}})
 		}
 	{{- else if .IsMethod -}}
-		{{- if .NoErr -}}
+		{{- if .VOkFunc -}}
+			v, ok := {{.GoReceiverName}}.{{.GoFieldName}}({{ .CallArgs }})
+			if !ok {
+				return nil, nil
+			}
+			return v, nil
+		{{- else if .NoErr -}}
 			return {{.GoReceiverName}}.{{.GoFieldName}}({{ .CallArgs }}), nil
 		{{- else -}}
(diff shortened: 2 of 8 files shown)


tarrencev:
Would love to see this merged

MichaelJCompton:
> Would love to see this merged

Done

MichaelJCompton:
Will cut a new release soon too.

steebchen:
Thank you so much @MichaelJCompton! This especially makes integration with https://github.com/prisma/prisma-client-go extremely smooth and fun :)

PR #1457, opened 2021-02-07
getting started: making `gqlgen generate` a runnable command

aaronArinder:
Making `gqlgen generate` a runnable command to help reduce friction for folks going through the example.


MichaelJCompton:
Thanks.  I like the idea, just replacing a bit more text https://github.com/99designs/gqlgen/pull/1487

PR #1458, opened 2021-02-07
getting started: making the resolver fn section clearer

aaronArinder:
I tripped over these lines while going through the getting started example. What confused me was not having a clear indicator that we were switching back to the original file; so, I made it clearer (to me at least) what we're doing in the example, but also the purpose of the `resolvers.go` file


MichaelJCompton:
LGTM.

Sorry, you tripped up on this.

Issue #462, opened 2018-12-03
Introspection and complexity


saward:
Is it possible to specify a complexity limit that doesn't apply to schema introspection?  Or to manually set the complexity cost for schema introspection?

I took a look through the generated code, but didn't see any hooks relating to this (I may very well have missed it).

vektah:
We currently strip them, what's the use case?

saward:
I was having troubles with the introspection until I removed or increased the complexity limit, so I was assuming it was applying there as well.  I'd prefer not to have the complexity limit apply to introspection.

If you already do that, perhaps there's some change I've made that's impacting here?

rakete:
I am having a similar problem, when I set a complexity limit, and reload my graphiql page, I get this error:

    {
      "errors": [
        {
           "message": "operation has complexity 179, which exceeds the limit of 5"
        }
      ],
      "data": null
    }

I looked in generated.go if I could find something that I could use to exclude the schema introspection from the complexity limitation, but didn't find anything.

When I remove or increase the complexity limit to >179, the above error goes away.

qhenkart:
interesting. I get the same issue except I get a constant 422 response from the server. Increasing the value like rakete suggested solved the problem

0xApe:
+1, set up the query complexity, but I still want to the user can fetch the schema with low complexity value.

saward:
I am still interested in this, so would like the issue to be kept open if that's ok.

saward:
Still interested

jensneuse:
I've solved the problem using a directive that makes the walker skip part of the tree.
https://github.com/jensneuse/graphql-go-tools/blob/db8e785954af348f8c594a819162b60e0b0f6ab7/pkg/middleware/operation_complexity/operation_complexity_test.go#L229

ivishnevs:
Instead of using `FixedComplexityLimit` try to build `ComplexityLimit` by yourself.
```
srv.Use(&extension.ComplexityLimit{
		Func: func(ctx context.Context, rc *graphql.OperationContext) int {
			if rc != nil && rc.OperationName == "IntrospectionQuery" {
				return 200
			}

			return 100
		},
})
```

should do the trick

PR #1492, opened 2021-03-18
Don't impose complexity limit on introspection queries

DSchmidlin:
Don't impose complexity limits on Introspection Queries

Resolves: https://github.com/99designs/gqlgen/issues/462


bofia:
This is not really safe because `operationName` can be passed in arbitrarily by the user, i.e. a user can name a malicious query `IntrospectionQuery`

lwc:
Dupe: https://github.com/99designs/gqlgen/pull/1581

Issue #1490, opened 2021-03-15
gqlgen init not working with g1.16


MichaelJCompton:
### What happened?

There's been some changes in go1.16 that mean go.mod and go.sum aren't always up to date (https://golang.org/issue/40728).  That means running `gqlgen init` in go1.16 will succeed but throw some validation errors.

### versions
 - `gqlgen version`? 
 - `go version`? 1.16+
 - dep or go modules?

### Work around

Running (after `gqlgen init`)

```
go mod tidy
```

cleans everything up.  (will fix in `gqlgen init` soon)

lwc:
Looks likely related: https://github.com/99designs/gqlgen/issues/1483

fabulias:
I have a similar issue, I did the following steps...

1- `go env -w GO111MODULE=auto`
2- Inside my `$GOPATH` or wherever I create a project folder `mkdir myproject`
3- `go mod init`
4- `go get github.com/99designs/gqlgen`
5- `go.mod` file is updated to

```
module github.com/fabulias/myproject

go 1.16

require github.com/99designs/gqlgen v0.13.0 // indirect
```

6- I tried with `go run github.com/99designs/gqlgen init`

And a lot of messages appear...

```
validation failed: packages.Load: go [-e -json -compiled=false -test=false -export=false -deps=false -find=true -- github.com/fabulias/myproject/graph/generated]: exit status 1: go: github.com/fabulias/myproject/graph/generated: package github.com/vektah/gqlparser/v2 imported from implicitly required module; to add missing requirements, run:
        go get github.com/vektah/gqlparser/v2@v2.1.0
go: github.com/fabulias/myproject/graph/generated: package github.com/vektah/gqlparser/v2/ast imported from implicitly required module; to add missing requirements, run:
        go get github.com/vektah/gqlparser/v2/ast@v2.1.0

go [-e -json -compiled=false -test=false -export=false -deps=false -find=true -- github.com/fabulias/myproject/graph/generated]: exit status 1: go: github.com/fabulias/myproject/graph/generated: package github.com/vektah/gqlparser/v2 imported from implicitly required module; to add missing requirements, run:
        go get github.com/vektah/gqlparser/v2@v2.1.0
go: github.com/fabulias/myproject/graph/generated: package github.com/vektah/gqlparser/v2/ast imported from implicitly required module; to add missing requirements, run:
        go get github.com/vektah/gqlparser/v2/ast@v2.1.0

[...]

PR #1501, opened 2021-03-30
Run go mod tidy after code generation

lwc:
Fixes: #1490

Also added an integration test covering `gqlgen init`.

## Questions
- Is this the best way to run `go mod tidy`?
- Is this the only place we need to run tidy?
- What happens in non-module mode? Do we officially support that mode of operation anymore?

I have:

### DIFF
--- a/.github/workflows/check-init
+++ b/.github/workflows/check-init
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -euo pipefail
+
+cd example/init
+
+go get github.com/99designs/gqlgen
+
+if { go run github.com/99designs/gqlgen init 2>&1 >&3 3>&- | grep '^' >&2; } 3>&1; then
+    echo "gqlgen init failed validation"
+    exit 125
+fi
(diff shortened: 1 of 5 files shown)


MichaelJCompton:
Looks fine.

If we needed not to run it, we could add a config option, like for validation.

asterikx:
@MichaelJCompton is there an option for not running `go mod tidy`?

Running `go mod tidy` as part of `gqlgen generate`  removes entries from `go.sum` that subsequent executables (in my case [`github.com/google/wire/cmd/wire`](github.com/google/wire)) rely on.

raoulk:
We have also a problem with this integrated `go mod tidy`. Locally, it works fine, but in our GitHub Actions build we get an error:
```
go run github.com/99designs/gqlgen generate
tidy failed: go mod tidy failed: exit status 1
```
Unfortunately, we don't have a clue what's causing this failure.

As we are using vendoring, we are pretty much sure that we don't need `go mod tidy` and would like to have an option to disable it.

yar00001:
Now you can set `SkipModTidy` in config to bypass `go mod tidy`.

zhixinwen:
@yar00001 is the config still working, getting:
```
unable to parse config: yaml: unmarshal errors:
  line 35: field skip_mod_tidy not found in type config.Config
```

zhixinwen:
nvm, seems to be not released

PR #1511, opened 2021-04-11
codegen/config: restore current working directory after changing it

a8m:
Before this commit, a call to `config.LoadConfigFromDefaultLocations` changed the working directory to the directory that contains the gqlgen config file.

This commit changes the implementation to restore the working directory after loading the config.

I have:

### DIFF
--- a/codegen/config/config.go
+++ b/codegen/config/config.go
@@ -70,8 +70,8 @@ func LoadDefaultConfig() (*Config, error) {
 
 // LoadConfigFromDefaultLocations looks for a config file in the current directory, and all parent directories
 // walking up the tree. The closest config file will be returned.
-func LoadConfigFromDefaultLocations() (*Config, error) {
-	cfgFile, err := findCfg()
+func LoadConfigFromDefaultLocations() (cfg *Config, err error) {
+	cfgFile, cwd, err := findCfg()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


a8m:
cc @lwc, @vektah

MichaelJCompton:
Hi @a8m, I've reverted this.  Changing this will break existing setups because generated files will end up in the different spots to how it previously worked.  That looks like a big breaking change to me.  

I can't see an issue attached.  Maybe you can get back in touch with a reasoning for this change and we could thing about what behaviour you are trying to support.

You can use the config file to specify were generated code goes https://gqlgen.com/config/, so maybe what you are looking for is already supported?

a8m:
Hey @MichaelJCompton, as mentioned in the GoDoc, the `config.LoadConfigFromDefaultLocations` loads the config file, and should not change the CWD (current working directory). I assumed the current behavior is a bug because you don't expect that a third-party package will change your CWD if you didn't ask for this explicitly (it took me some time to figure out who's changing my CWD in my Go programs 😄).

Now, I get your point that some parts rely on this side-effect, and it makes sense to revert it atm. But maybe I can re-submit this change and fix all the other parts with it? 

Other suggestions are to add this side-effect to the GoDoc and maybe add another function like `LoadConfigFromDefaultLocations` that doesn't change the CWD. WDYT?

a8m:
Hey @MichaelJCompton, please let me know if you prefer that I'll open an issue for it. Thanks

frederikhors:
@a8m is totally right!

> maybe I can re-submit this change and fix all the other parts with it?

@a8m can you please re-arrange a PR to fix this totally wrong behavior?

frederikhors:
This is an issue when launching multiple parallel tasks in a project: https://github.com/ent/ent/issues/2762.

Issue #1455, opened 2021-02-06
Data race occurred when input fields have default values with latest gqlparser


skaji:
### What happened?

Data race occurred when input fields have default values with latest gqlparser (which includes bugfix around default values https://github.com/vektah/gqlparser/pull/122).

### What did you expect?

No data race.

### Steps to reproduce the issue

Please look at https://github.com/skaji/go-gqlgen-sample/tree/main/issue, which reproduces this issue.
Also please note that that uses `github.com/vektah/gqlparser/v2 v2.1.1-0.20201022021006-7e475a735fea`.

#### Step 1. Setup server with `-race` flag
```
git clone https://github.com/skaji/go-gqlgen-sample
cd go-gqlgen-sample/issue
go run -race server/main.go
2021/02/06 22:20:06 connect to http://localhost:8080/ for GraphQL playground
```

#### Step 2. Access the server
```
curl 'http://localhost:8080/query' -H 'Accept-Encoding: gzip, deflate, br' -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Connection: keep-alive' -H 'DNT: 1' -H 'Origin: http://localhost:8080' --data-binary '{"query":"query test($filter: Filter!) {\n  users {\n    friends(filter: $filter) {\n      name\n    }\n  }\n}","variables":{"filter":{}}}' --compressed
```

#### Step3. You'll see DATA RACE
```
WARNING: DATA RACE
Read at 0x00c000203560 by goroutine 11:
  runtime.mapaccess2_faststr()
      /Users/skaji/env/go/src/runtime/map_faststr.go:107 +0x0
  github.com/skaji/go-gqlgen-sample/issue.(*executionContext).unmarshalInputFilter()
      /Users/skaji/src/github.com/skaji/go-gqlgen-sample/issue/generated.go:1553 +0x87
  github.com/skaji/go-gqlgen-sample/issue.(*executionContext).unmarshalNFilter2githubᚗcomᚋskajiᚋgoᚑgqlgenᚑsampleᚋissueᚐFilter()
      /Users/skaji/src/github.com/skaji/go-gqlgen-sample/issue/generated.go:1926 +0x70
  github.com/skaji/go-gqlgen-sample/issue.(*executionContext).field_User_friends_args()
      /Users/skaji/src/github.com/skaji/go-gqlgen-sample/issue/generated.go:203 +0x2d2
  github.com/skaji/go-gqlgen-sample/issue.(*executionContext)._User_friends()
[...]

PR #1514, opened 2021-04-13
Bump gqlparser to v2.2.0

lwc:
## Added
- Support for directives on variable definitions
- Support for repeated directives, as per latest GraphQL spec
- gqlerror.List now meets the go 1.13 error interfaces for Is and As 
- Support for coercing single value list inputs to lists of one element
- Support for VARIABLE_DEFINITION executable directive

## Updated
- Improved formatting of block comments on args
- Aligned root type inference more closely to the spec

## Fixed
- Ensure directives are only used in accordance with their definitions
- Fixed issue where default values were not considered when validating fields 

See more https://github.com/vektah/gqlparser/releases/tag/v2.2.0


skaji:
@lwc I believe gqlparser v2.2.0 introduces some data race.
I really appreciate it if you could take a look at #1455 / #1456.

Issue #1510, opened 2021-04-09
Time Marshaling & Unmarshaling are asymmetrical


frankywahl:
### What happened?

Marshaling & Unmarshaling are asymmetrical with time. Golang's build time parsing will work with nanosecond resolution. 
We therefore lose information and symmetry with provided timestamp.

### What did you expect?

Symmetry in the timestamp handling

### Fix
The fix is simple, change https://github.com/99designs/gqlgen/blob/master/graphql/time.go#L16 to `time.RFC3339Nano`. I would happily submit a PR if valuable. 

Sample Test Code to highlight the issues:

```go
package main_test

import (
	"bytes"
	"io"
	"strconv"
	"testing"
	"time"

	"github.com/99designs/gqlgen/graphql"
)

func NanoSecondMarshalTime(t time.Time) graphql.Marshaler {
	if t.IsZero() {
		return graphql.Null
	}

	return graphql.WriterFunc(func(w io.Writer) {
		io.WriteString(w, strconv.Quote(t.Format(time.RFC3339Nano)))
	})
}

func TestMarshaler(t *testing.T) {
	initialTime := time.Now()
	testCases := []struct {
		name      string
		marshaler func(t time.Time) graphql.Marshaler
	}{
		{
			name:      "Provided Marshaler",
			marshaler: graphql.MarshalTime,
		},
		{
			name:      "NanoSecond Marshaler",
			marshaler: NanoSecondMarshalTime,
		},
	}

	for _, tc := range testCases {
		tc := tc
		t.Run(tc.name, func(t *testing.T) {
			bi := bytes.NewBuffer([]byte{})
			tc.marshaler(initialTime).MarshalGQL(bi)

			str, err := strconv.Unquote(string(bi.Bytes()))
			if err != nil {
				t.Fatalf("could not unquote: %v", err)
			}
			newTime, err := graphql.UnmarshalTime(str)
			if err != nil {
				t.Fatalf("could not unmarshal: %v", err)
			}
			if !initialTime.Equal(newTime) {
				t.Errorf("expected times %v and %v to equal", initialTime, newTime)
			}
		})
	}
}

PR #1515, opened 2021-04-13
Marshaling & Unmarshaling time return initial value

frankywahl:
There was a lack of symmetry that would prevent times for being
symmetrical. That is because time.Parse actually parses an RFC3339Nano
implicitly, thereby allowing nanosecond resolution on unmarshaling a
time. Therefore we now marshal into nanoseconds, getting more
information into GraphQL times when querying for a time, and restoring
the symmetry

resolve: https://github.com/99designs/gqlgen/issues/1510

I have:


milesich:
This is a breaking change. Having nano part when it was not there before.

Issue #511, opened 2019-01-24
Support Directives on Input Types


sdalezman:
### Expected Behaviour

When defining a graphql schema you should be able to define a directive on an entire input object or an input field

### Actual Behavior

Can't define any directive on a input field 

### Minimal graphql.schema and models to reproduce

```graphql
directive @hello on INPUT_OBJECT

type Query {
   repeatLogName(n: Name!): String
}

input Name {
   name: String!
}

extend input Name @hello
```

where the directive evaluates to the input from the logging example: 

``` golang
func InputLogging(ctx context.Context, obj interface{}, next graphql.Resolver) (res interface{}, err error) {
	rc := graphql.GetResolverContext(ctx)
	log.Printf("input object logging: %v, %s, %T, %+v", rc.Path(), rc.Field.Name, obj, obj)
	return next(ctx)
}
```

Input logging will never be called. I think this has to do with how FieldMiddleware calls directives, but this is only from a cursory glance at the generated code

sdalezman:
@mathewbyrne the reason I thought it was supported was because in the example of the type system extension there's an `INPUT_OBJECT`: https://github.com/99designs/gqlgen/blob/master/example/type-system-extension/schemas/input-object-extension.graphql

Searching through the generated code it didn't seem to ever be called, but thought it might've been my lack of understanding the generated code entirely

I understand your comments about when/how it should be called. For us, adding in some sort of input property/input object directive would be great. It would allow us to not have to add function calls to ensure that a given user has permission to filter/query by various fields (from a security perspective that adds onus to a developer to ensure they remember to add those checks as well as the reviewer to ensure that they're obvious- we do that and have tests for those flows, but reducing the mental burden of rapidly iterating on api's is always a plus in my book). 

That way we can have a handful of types and know that they can always be used securely and reduce the exposure of surface areas

sdalezman:
@vektah @mathewbyrne our team would be happy to tackle this issue. If you can provide a bit more design guidance so we don't go down the wrong rabbit whole. We were thinking that you can have two types of input directives: 

1. Input directive that's called before any resolvers are called
2. Input directive that's called after resolvers are called

Which I think would cover 80% of use cases directives for inputs. If your team agrees on that we'd be happy to start working on this (and save us a tremendous headache at the moment)

mathewbyrne:
[...]

PR #1525, opened 2021-04-30
support input object directive

Code-Hex:

Currently, we cannot use INPUT_OBJECT directive in gqlgen. (actually we can use as graphql schema syntax, but it does not use directive hooks)
So I supported this feature.

- https://github.com/99designs/gqlgen/issues/511

I have:

### DIFF
--- a/codegen/field.go
+++ b/codegen/field.go
@@ -89,6 +89,11 @@ func (b *builder) bindField(obj *Object, f *Field) (errret error) {
 			if err != nil {
 				errret = err
 			}
+			for _, dir := range obj.Directives {
+				if dir.IsLocation(ast.LocationInputObject) {
+					dirs = append(dirs, dir)
+				}
+			}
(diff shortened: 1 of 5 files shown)


Code-Hex:
@mtibben I need this PR. Please review 🙏

Dan6erbond:
I just came across this PR and wanted to use this feature to implement a validation directive, however it seems like input objects come in as `map[string]any` in the directive. Is this intended? I would prefer if the `obj` had the generated type instead.

Sample directive:

```go
	config.Directives.Validate = func(ctx context.Context, obj interface{}, next graphql.Resolver) (res interface{}, err error) {
		fmt.Println("--- VALIDATE ---")
		fmt.Println(obj)

		if err := validate.StructCtx(ctx, obj); err != nil {
			return nil, err
		}

		return next(ctx)
	}
```

Thanks!

PR #1562, opened 2021-06-24
[CARESERV-928] Renaming Entity in gqlgen to avoid type name conflict

AndrewRayCode:
[CARESERV-928](https://grandrounds.atlassian.net/browse/CARESERV-928)


AndrewRayCode:
Whoops meant to open this against my fork ignore

Issue #1256, opened 2020-07-20
Unable to add type "Service" in federation


josemarluedke:
I'm declaring a GraphQL type called "Service" in a federated gqlgen. gqlgen fails to generate because of "redeclared service type".

Reproduction: https://github.com/josemarluedke/gqlgen-fed-service-type
GraphQL: https://github.com/josemarluedke/gqlgen-fed-service-type/blob/master/graph/schema.graphqls

### What happened?
```
➜ go generate ./...
validation failed: packages.Load: /Users/josemarluedke/code/tmp/gqlgen-fed-service-type/graph/generated/generated.go:62:2: Service redeclared
/Users/josemarluedke/code/tmp/gqlgen-fed-service-type/graph/generated/generated.go:58:2: 	other declaration of Service
/Users/josemarluedke/code/tmp/gqlgen-fed-service-type/graph/generated/generated.go:135:27: e.complexity.Service.SDL undefined (type struct{ID func(childComplexity int) int} has no field or method SDL)
/Users/josemarluedke/code/tmp/gqlgen-fed-service-type/graph/generated/generated.go:139:31: e.complexity.Service.SDL undefined (type struct{ID func(childComplexity int) int} has no field or method SDL)
exit status 1
```

Here is the generated code: https://github.com/josemarluedke/gqlgen-fed-service-type/blob/master/graph/generated/generated.go#L58-L64

### What did you expect?

Only one type Service declared. No conflict with the Federation.

### Minimal graphql.schema and models to reproduce

Full repro: https://github.com/josemarluedke/gqlgen-fed-service-type
GraphQL: https://github.com/josemarluedke/gqlgen-fed-service-type/blob/master/graph/schema.graphqls

### versions
 - `gqlgen version` v0.11.3
 - `go version` 1.13.6
 - go modules

frederikhors:
nope

moensch:
I just hit the same thing - glad it got reported. The `ComplexityRoot` ends up with two child structs named `Service`.

One of them represents my actual model, the other looks like this:

```go
	Service struct {
		SDL func(childComplexity int) int
	}
```

zhevron:
I just found this one as well when trying to migrate an existing schema to use federation, so renaming my type is hardly an option without causing breaking changes.

It seems to be caused by an internal model representing the service (as required by Apollo Federation) defined here: https://github.com/99designs/gqlgen/blob/master/plugin/federation/fedruntime/runtime.go#L6

[...]

PR #1563, opened 2021-06-25
Renaming Entity in gqlgen to avoid type name conflict

AndrewRayCode:
Addresses https://github.com/99designs/gqlgen/issues/1256

Mostly copy-pasta'd from @zhevron's https://github.com/zhevron/gqlgen/commit/cecd727168a6a5873c2d5ac4d75a857da341dbe7# which is linked from the ticket

I chose a name that's very unlikely to conflict with end users' structs.

Since Go is a type safe language, introducing a breaking change seems less scary. Multiple people have hit this issue.

I have:
 - [ x ] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [ ] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content)) - I couldn't find any docs that reference Service/Entity, by grepping for those words in *.md files


zhevron:
I wonder how the changes in #1549 affects the underlying issue here. It seems like it will make the code generation output `_Service` and `_Entity` as the type names, which is representative of what they are named in the schema.

The only issue I had with this solution was that it required federation plugin specific changes in the main codegen package.

aanari:
+1

PR #1568, opened 2021-07-02
Update apq.md

DanyHenriquez:
There is no relevant issue. The old APQ has been merged into apollo. so this is an update to the correct link.

I have:

### DIFF
--- a/docs/content/reference/apq.md
+++ b/docs/content/reference/apq.md
@@ -14,7 +14,7 @@ to register query hash with original query on a server.
 ## Usage
 
 In order to enable Automatic Persisted Queries you need to change your client. For more information see
-[Automatic Persisted Queries Link](https://github.com/apollographql/apollo-link-persisted-queries) documentation.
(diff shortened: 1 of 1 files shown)


wilhelmeek:
Perhaps this might be a good opportunity to link to the [Apollo Basics](https://www.apollographql.com/docs/resources/graphql-glossary/#automatic-persisted-queries-apq) section of their documentation.

Linking to the React section may imply that this is a React-client specific feature.

DanyHenriquez:
Maybe you're right.

Can we  do something about the incorrect link in the docs?

wilhelmeek (approved):
Thanks for updating @DanyHenriquez! I'm not a maintainer so my tick doesn't have impact, but this looks good to me 👍

mtibben:
Thanks @DanyHenriquez

DanyHenriquez:
Glad to help. I will check if i can pickup more in the sense of coding as well and go over the docs. I noticed some things are not correctly described there or somewhat unclear :octocat: 

Great project though and love using it!

PR #1604, opened 2021-09-08
fix double indirect bug

carldunham:
Fixes a bug where binding to struct elements with multiple indirection will compile.

May also be useful for some (as it is for us) that are using this technique for parse updates in mutations.

I have:


carldunham:
@mtibben @lwc Can someone take a look at this, and provide some feedback? It doesn't address the model generation side, but it could. Otherwise it can address space update and falsy defaults, if the models are created properly.

PR #1610, opened 2021-09-11
Update to go 1.17

mtibben:
- Updates errors to use the error semantics introduced in go 1.13 (Builds on top of the changes in #1606)
 - gqlgen now runs linting and tests against go 1.17 only. I'd suggest the simplest policy around go versioning is to support only the latest version of go. Those wanting to use an older version of go can also use an older version of gqlgen. Thoughts @wilhelmeek @lwc?
 - Works around some bugs around autobinding and init in the latest version of go. Go is much more strict about what it considers a package in recent versions, so we create a temporary file so we have a valid package before generating models
 - Updates `gqlgen init` to be more strict about what it does. It will only create the config, schema and server if those files do not already exist. It also gives more feedback to the user about what files are being created

I have:


wilhelmeek:
> I'd suggest the simplest policy around go versioning is to support only the latest version of go.

I think this is very sensible 👍

frederikhors:
I would suggest the latest two but we can maybe start this policy from 1.17 because this is a `package` and `mod` revolution version.

a8m:
> gqlgen now runs linting and tests against go 1.17 only. I'd suggest the simplest policy around go versioning is to support only the latest version of go. Those wanting to use an older version of go can also use an older version of gqlgen. Thoughts @wilhelmeek @lwc?


I'll suggest supporting the last 2 versions of Go. This is the same thing we do in [ent](https://github.com/ent/ent), and we do it because we found that it takes some time for some companies to upgrade the Go version.

cmd/init.go
```diff
@@ -109,90 +111,92 @@ var initCmd = &cli.Command{
...
-			if err := initConfig(configFilename, pkgName); err != nil {
-				return err
-			}
+		if fileExists(serverFilename) {
+			return fmt.Errorf("%s already exists\n", serverFilename)
+		}
+		_, err := config.LoadConfigFromDefaultLocations()
+		if err == nil {
+			return fmt.Errorf("gqlgen.yml already exists in a parent directory\n")
 		}
```
a8m, line 141: Some suggestions for improvement. Also, I think the `\n` belongs to the place the error is printed and should not be part of the message. 

```suggestion
		for _, name := range []string{configFilename, schemaFilename, serverFilename} {
			if fileExists(name) {
				return fmt.Errorf("%s already exists", name)
			}
		}
		if _, err := config.LoadConfigFromDefaultLocations(); err == nil {
			return errors.New("gqlgen.yml already exists in a parent directory")
		}
```

cmd/init.go
```diff
@@ -109,90 +111,92 @@ var initCmd = &cli.Command{
...
+			return err
+		}
 
-func GenerateGraphServer(serverFilename string) {
-	cfg, err := config.LoadConfigFromDefaultLocations()
-	if err != nil {
-		fmt.Fprintln(os.Stderr, err.Error())
-	}
+		// create schema
+		fmt.Println("Creating", schemaFilename)
```
a8m, line 150: Can we enable this printing only on debug/verbose mode?  Some users (like myself), find it more annoying than useful.

cmd/init.go
```diff
@@ -109,90 +111,92 @@ var initCmd = &cli.Command{
...
+		if err := initFile(tmpPackageNameFile, "package model"); err != nil {
+			return err
+		}
+		defer os.Remove(tmpPackageNameFile)
 
-	fmt.Fprintf(os.Stdout, "Exec \"go run ./%s\" to start GraphQL server\n", serverFilename)
-}
+		var cfg *config.Config
+		if cfg, err = config.LoadConfig(configFilename); err != nil {
+			panic(err)
```
a8m, line 165: Do we need the stack trace here? If not, I think `log.Fatal` is preferred.

cmd/init.go
```diff
@@ -109,90 +111,92 @@ var initCmd = &cli.Command{
...
+		}
 
-func configExists(configFilename string) bool {
-	var cfg *config.Config
+		fmt.Println("Creating", serverFilename)
+		fmt.Println("Generating...")
```
a8m, line 169: Same comment about printing above.

Also, I think that most Go/Unix  commands/programs don't print anything on the "happy path".

cmd/init.go
wilhelmeek: I assume the panic is necessary to handle the deferred cleanup of the above files.

a8m: Might be cleaner (also for the end-user) to call `os.Remove` manually before exiting.

mtibben: it's a panic because this should never ever happen - it's our own template we're parsing

mtibben:
> What about minor patches or bug fixes? Don't we want to make it possible for companies to get these patches even though they use the old Go version?

It depends on whether someone is willing to put in (potential) extra work to support that sort of thing. This project has not been moving for a long time, so the purpose of that statement was to remove constraints and allow it to move forward quickly at this point in time. 

So if someone wants to support older versions, I'm not opposed to that at all - just want to ensure it's not a blocker right now

cmd/init.go
mtibben: Yep I'm a big fan of the "[silence is golden](http://www.linfo.org/rule_of_silence.html)" philosophy, and follow it myself all the time for CLI tools - especially in tools like `aws-vault` which I maintain.

Yet in this case I think we should actually be more verbose when running `gqlgen init` for the following reasons
 1. `gqlgen init` is the first command a user will run when using gqlgen. Many users won't be familiar with it or what it's doing. 
 2. `gqlgen init` only gets run once. It's not a command users will regularly use.
 3. `gqlgen init` is a slightly "surprising" command - it only creates boilerplate config and schemas, that should then be immediately edited to fit the use-case. Creating that visibility makes that fact more obvious
 4. It mirrors the behaviour of `go mod init` which lists the files that it creates. `go mod init` is the command that will precede `gqlgen init` most of the time

I think the "silence is golden" philosophy may be better applied to `gqlgen generate` where none of the above is true

cmd/init.go
mtibben: see my comment above

cmd/init.go
mtibben: hmm the newline is there because for some reason github.com/urfave/cli/v2 runs the error message together 🤷

```
$ go run github.com/99designs/gqlgen init
gqlgen.yml already existsexit status 1
```

mtibben:
I've [written a little integration test](https://github.com/99designs/gqlgen/commit/5dfd236de1df837ea3ed6febe5405f9f112fc69f#diff-d771e0afcdec153a565dc10610eb83e9c64c011306e5176f4349fea1d8a76266) for `gqlgen init` but i'm not sure I love the approach - it needs to shell out to`gqlgen` and `go`. Not sure if this is the best approach, but can't work out an alternative. Any thoughts? The problem I'm running into is go's expectations around packages and modules - it won't let me execute gqlg en within the context of a test because of missing types. Perhaps this is better suited to CI where the environment is controlled? I think this what the `check-init` script is doing - is that right @lwc ?

cmd/init.go
mtibben: fixed in 9162c53fc32471c96d874764b47bf71f0c493fce

mtibben:
I'll merge this for now - I think i've addressed your main concerns @a8m but happy to continue the conversation if you think it's warranted.

Issue #1195, opened 2020-05-20
Incorrect handling of Float64: NaN, +Inf and -Inf


adrianduke:
### What happened?
I encoded a `float64` that was set to `math.Inf(1)`, the outputted JSON is invalid:

```json
{
  "someKey": +Inf,
}
```

### What did you expect?
As per Go's `json.Marshal()` I'd expect an error that I can't encode `+Inf`, `-Inf` or `NaN`:

- https://play.golang.org/p/3V_qBPlAeR-
- https://golang.org/src/encoding/json/encode.go?s=17629:17751#L573

JSON doesn't support infinity or nan, see:

- https://www.json.org/
- https://stackoverflow.com/questions/1423081/json-left-out-infinity-and-nan-json-status-in-ecmascript

### Minimal graphql.schema and models to reproduce

```graphql
type FloatExample {
  myFloat: Float
}
```

```go
func NewFloatExample(f float64) *FloatExample {
  return &FloatExample{
    MyFloat: f,
  }
}

type FloatExample struct {
  MyFloat float64
}

NewFloatExample(math.Inf(1))
NewFloatExample(math.Inf(-1))
NewFloatExample(math.NaN())
```

set `myFloat` to either `math.Inf(1)`, `math.Inf(-1)` or `math.NaN()`.

### versions
 - `gqlgen version`? **v0.10.1**
 - `go version`? **go1.13.4**
 - dep or go modules? **go modules**

frederikhors:
Nope.

frederikhors:
No

Issue #743, opened 2019-06-13
Allow return error for scalar marshaling


pkosiec:
I have a case when marshaling a scalar can return an error.

Currently we have the following `Marshaler` and `Unmarshaler` interfaces:

```go
type Marshaler interface {
	MarshalGQL(w io.Writer)
}

type Unmarshaler interface {
	UnmarshalGQL(v interface{}) error
}
```

It'd be great to allow return errors for `Marshaler`, similarly to `Unmarshaler`.

### versions
 - `gqlgen version` 0.9.0

pkosiec:
This issue is still valid.

pkosiec:
Still valid. @vektah if you agree, I can create a PR with this change during my spare time. It would be a breaking change, but I can try to avoid this. Waiting for your feedback. Cheers!

adrianduke:
Implementing this would make resolving issue https://github.com/99designs/gqlgen/issues/1195 simpler.

PR #1652, opened 2021-10-09
ContextMarshaler

### LINKED ISSUE #1645: [proposal] Pass context to scalar Marshal/Unmarshal
### Proposal

Add an alternate interface for `Marshaller` and `Unmarshaller` that includes a context. 

```go
type ContextMarshaler interface {
	MarshalGQLContext(ctx context.Context, w io.Writer) error
}
type ContextUnmarshaler interface {
	UnmarshalGQLContext(ctx context.Context, v interface{}) error
}
```

The method names have changed, so types can implement the old and new interface simultaneously to support current and older versions of gqlgen. If that is done, the Context version is preferred.

The error response has been added to the Marshal function to satisfy #743 with a single signature change.

Additionally add a new form of an external Marshaller/Unmarshaller that uses a similar signature.

```go
func MarshalX(ctx context.Context, val T) error

func UnmarshalX(ctx context.Context, input interface{}) (T, error)
```

Since multiple external Marshal/Marshall functions for the same type can coexist with different names, there is no reason for a special prefix. The code generator can detect the updated signature and use it appropriately.

### Use Case

I want to define a scalar for various units that have regional-specific preferences. Since they will be utilizing `big.Float` I don't want to do unit conversions in the front-end. These types will be used in many places throughout the schema, so I'd like to do the formatting in the scalar type when doing the Marshal to a string.

To my knowledge, there is no way to `Marshal`/`Unmarshal` with an interface/custom type while passing it a context. This would allow me to pass the relevant context of which unit the scalar should format as. If I try to pass the context in a similar pattern as the automatically collected resolver methods, generates incorrect code.

```go
(r Radioactivity) MarshalGQL(ctx context.Context, w io.Writer)
// vs.
MarshalRadioactivity(ctx context.Context, r Radioactivity) graphql.Marshaler
```
yields unsatisfied interfaces or mistaking `context.Context` as the type being marshaled/unmarshaled respectively

duckbrain:
Adds a new interface that a type may implement for receiving a context when marshaling/unmarshaling. The new interface may return an error for marshaling. The built-in Float scalar will error using the new interface.

- [x] Add new interface and wrapper to minimize impact on existing code-base.
- [x] Generate appropriate code to pass the context or wrap the `ContextMarshaler` with the standard `Marshaler`.
  - [x] write test
- [x] Add new marshaler set for Float that returns error for -Inf, Inf, and NaN; set that to be the default
  - [x] write test
- [x] Write test for new interface to show context being passed
- [x] Document the new alternative interface

Closes #1645, #743, and #1195

I have:


wilhelmeek:
Thanks @duckbrain 🙏

PR #1660, opened 2021-10-15
Update directives doc page

johnmaguire:
I am new to GraphQL and gqlgen but excited about both.

While going through the docs I got hung up on a few things. I figured I would make some PR's to try to help another newcomer to both GraphQL and gqlgen.

This type of thing is subjective, so please merge or modify any bits you think might be valuable and discard anything/everything if it's not helpful.


StevenACoffman:
Thanks so much for the improvement!

PR #1418, opened 2020-12-12
Update GQLgen test client to work with multipart form data

Sonna:
When attempting to perform a file upload with this GQLgen test client in tests (see [fileupload README](https://github.com/99designs/gqlgen/blob/master/example/fileupload/readme.md) for an example) it fails with the following message, mentioning `multipart/form-data;` is not supported

```
panic: unsupported encodingmultipart/form-data; boundary=033fcc25416def96ea04cd9f81bb71a256a18eeb8b3aab15eacfd22c13f9
```

This updates test client to support `multipart/form-data;` requests.

I have also provided an `client.Option` named `WithFiles()` to convert an existing outgoing request to `multipart/form-data;` encoding and any `os.File` variables to a generic `Content-Type: application/octet-stream` part within the body of the outgoing request

PR #1661, opened 2021-10-15
Update GQLgen test client to work with multipart form data (take 2)

Sonna:
Redoes previous PR #1418, however updates flaky tests to handle race conditions; like with the order of file inputs within a GQL request; which failed after being merged previously (see PR #1659).

```go
// for example, in a test the input with files expected vs result
`{"0":["variables.req.files.0"],"1":["variables.req.files.1"],"2":["variables.req.foo.bar"]}`
// vs
`{"0":["variables.req.foo.bar"],"1":["variables.req.files.0"],"2":["variables.req.files.1"]}`
```

---

When attempting to perform a file upload with this GQLgen test client in tests (see [fileupload README](https://github.com/99designs/gqlgen/blob/master/example/fileupload/readme.md) for an example) it fails with the following message, mentioning `multipart/form-data;` is not supported

```
panic: unsupported encodingmultipart/form-data; boundary=033fcc25416def96ea04cd9f81bb71a256a18eeb8b3aab15eacfd22c13f9
```

This updates test client to support `multipart/form-data;` requests.

Also provides an `client.Option` named `WithFiles()` to convert an existing outgoing request to `multipart/form-data;` encoding and any `os.File` variables to a generic `Content-Type: application/octet-stream` part within the body of the outgoing request

### DIFF
--- a/client/withfilesoption_test.go
+++ b/client/withfilesoption_test.go
@@ -1,7 +1,6 @@
 package client_test
 
 import (
-	"fmt"
 	"io"
 	"io/ioutil"
 	"mime"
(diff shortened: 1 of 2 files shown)


StevenACoffman:
Yay! Thanks!

PR #1680, opened 2021-10-29
Rename `@extraTag` directive to `@goTag` and make repeatable

wilhelmeek:
- Rename for consistency with `@goField` and `@goModel`
- Allow for repeatable use to bind multiple tag pairs

**edit:** added a comment down below, but contemplating whether it makes sense to have this directive at all given the addition of field-level hooks cc @JohnMaguire

I have:
 - [ ] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [ ] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

PR #1682, opened 2021-11-01
Reimplement goTag using FieldMutateHook

tprebs:
This PR reimplements the extraTags/goTag directive as a fieldMutateHook and applies it as the default hook for the modelgen Plugin. This has been done as a request off the back of #1680. 

I did notice that some functionality did change during the renaming of `@extraTag` to `@goTag`. e.g. if the value argument is not passed the value of the tag is set to the field name. e.g. ` testField String @goTag( key:"yaml")` sets the tag `json:"testField" yaml:"testField"`.  Though this was not mentioned in the PR, I assumed that this is the intended behaviour and this has also been implemented in the FieldMutateHook. 

I have:

### DIFF
--- a/codegen/config/config.go
+++ b/codegen/config/config.go
@@ -239,10 +239,6 @@ func (c *Config) injectTypesFromSchema() error {
 		SkipRuntime: true,
 	}
 
-	c.Directives["goTag"] = DirectiveConfig{
-		SkipRuntime: true,
-	}
-
 	for _, schemaType := range c.Schema.Types {
 		if schemaType == c.Schema.Query || schemaType == c.Schema.Mutation || schemaType == c.Schema.Subscription {
 			continue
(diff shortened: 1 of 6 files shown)


StevenACoffman (approved):
Thanks! This looks good to me, but I'm also wondering about fleshing out the hooks documentation a bit to promote this technique a bit more.

wilhelmeek (approved):
> Though this was not mentioned in the PR, I assumed that this is the intended behaviour and this has also been implemented in the FieldMutateHook.

Yes it was intended. Apologies for not mentioning in the PR, but felt that that was a sensible fallback.

wilhelmeek:
Thanks for putting this together @tprebs! 🙏

tprebs:
> Thanks! This looks good to me, but I'm also wondering about fleshing out the hooks documentation a bit to promote this technique a bit more.

I agree. Though there are some examples in the recipes section, the plugins documentation in the reference section has not been updated in a couple of years

StevenACoffman:
I didn't want to hold up merging this PR so I merged it as it was, but I'm sorry if you weren't finished. If you get the time to polish any documentation or examples up more, I would be very grateful and glad to merge them as a separate PR.

tprebs:
I'm Happy to create another PR updating the plugin reference documentation and hooks. The thoughts I had were 
1. Show the goTag implementation as a hook in the recipes, replacing the current constraint validation example
2. Add ModelMutateHook and FieldMutateHook documentation to the plugin reference documentation, explaining the scope of what each is used for and what it has access to giving reference to the example implementation which is goTag.
3. Outline the modelgen hooks lifecycle and when the hooks are applied

Is there anything else you can think of that we should add?

syssam:
any update?

Issue #1670, opened 2021-10-21
Add to Apollo subgraph compatibility test suite


kdawgwilk:
Would be great to have compatibility determined by the new test suite from Apollo.  I have a PR up adding an example subgraph and would love any feedback to improve it

https://github.com/apollographql/apollo-federation-subgraph-compatibility/pull/42

PR #1684, opened 2021-11-02
Support for multiple @key directives in federation

carldunham:

Adds support for multiple @key directives for OBJECT types, which is allowed for in the Apollo spec. Can be useful in cases where some subgraphs have different pieces of data for an external type that can be resolved in multiple ways.

Should also open the door for #1670 to be implemented, and remove another * from https://github.com/apollographql/apollo-federation-subgraph-compatibility/pull/42.

I have:


StevenACoffman:
Thanks for this!

PR #1697, opened 2021-11-08
Entity resolver tests

MiguelCastillo:
The tests work by sending `_entities` queries with `representation` variables directly to the mocked server, which will allow us to test generated federation code end to end.  For context, the format of the entity query is something like:

```
query($representations:[_Any!]!){_entities(representations:$representations){ ...on Hello{secondary} }}
```

And `representations` are the list of federated keys for the entities being resovled, and they look like

```
representations: [{
   "__typename": "Hello",
   "name":       "federated key value 1",
}, {
   "__typename": "Hello",
   "name":       "federated key value 2",
}]
```

The entity resolver tests are in `plugin/federation/federation_entityresolver_test.go` and they rely on `plugin/federation/testdata/entityresolver`.

To run the tests:
1. Build the entityresolver testdata
  - From plugin/federation, run `go run github.com/99designs/gqlgen --config testdata/entityresolver/gqlgen.yml`
2. Run the tests with `go test ./...` or similar

I have:
 - [x] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [x] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

PR #1698, opened 2021-11-08
Revert "Support for multiple @key directives in federation"

StevenACoffman:
Reverts 99designs/gqlgen#1684

@carldunham We found that this has been causing a number of problems that were not well covered with existing tests. We are going add more tests, which highlight some of the issues. We would still very much like to have this feature, but it's not obvious how to solve it without causing these problems so we think it needs a little more work. Please see #1697 for more details


carldunham:
OK, will take a look

PR #1699, opened 2021-11-09
Ignore generated files from test coverage

StevenACoffman:
We were not ignoring a number of generated files from the test coverage reports, giving a false impression of lower coverage.


MiguelCastillo (approved):
looks good.

wilhelmeek:
Nice!

PR #1702, opened 2021-11-11
Optimize performance for binder, imports and packages

sbalabanov:
This PR consists of multiple commits and is pure refactoring. It should maintain full backwards compatibility with the origin.

- `imports.goModuleRoot` is called for every type resolution. It tries to find go.mod file in the directory tree making a gazillion of requests to open file. While someone more familiar with the application design can come up with a better architecture to avoid this altogether, I added a caching layer on top of this function to only traverse file system once.
- `binder.FindObject` traverses all definitions in a package and for all needed types it is effectively O(n^2). Make it to O(n) by caching all package top-level definitions in a hashtable.
- `packages.Load` calls 'packages.LoadAll' which allocates a slice from the heap even when a single package is required. Added a fast path by checking a cache map before calling to `packages.LoadAll` to avoid that. A possibly better solution is to refactor  common parts of `packages.Load` and `packages.LoadAll` into a separate function.

For the targeted use case we achieved ~20% speed optimizations after those improvements.

Because functionality did not change, existing tests should cover it all so no new tests added.

### DIFF
--- a/codegen/config/binder.go
+++ b/codegen/config/binder.go
@@ -112,45 +112,47 @@ func (b *Binder) FindObject(pkgName string, typeName string) (types.Object, erro
 	if pkgName == "" {
 		return nil, fmt.Errorf("package cannot be nil")
 	}
-	fullName := typeName
-	if pkgName != "" {
-		fullName = pkgName + "." + typeName
-	}
 
 	pkg := b.pkgs.LoadWithTypes(pkgName)
 	if pkg == nil {
 		err := b.pkgs.Errors()
 		if err != nil {
-			return nil, fmt.Errorf("package could not be loaded: %s: %w", fullName, err)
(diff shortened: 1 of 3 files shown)


StevenACoffman:
@sbalabanov  Thank you for your contribution! It looks like there are some linting and test failures. Can you make sure your files are formatted using gofmt (or gofumpt), and you've regenerated everything (including the examples) and ran (and passed) all the tests?

```
gofmt -s -w .
go generate ./... && cd example/ && go generate ./... && cd ..
go test -race ./... && cd example && go test -race ./... && cd ..
```

sbalabanov:
Can I please have an approval to run workflows on PRs? On the fork I am getting lint error which seems unrelated.
Test failures seem to be fixed now.

StevenACoffman:
Merged in #1711

PR #1706, opened 2021-11-12
Fix #1704: handle @required nested fields as in @key

### LINKED ISSUE #1704: Nested @requires generates incorrectly in generated federation file
### What happened?

I'm using the recently implemented nested @requires directives to try to pass through an id of a subtype to a field extension, and the generated file is not able to correctly unmarshall the nested id field, resulting in a null id being passed through.

### Schema 

```
extend type User @key(fields:"id"){
  id: ID! @external
  name: String!
  subType: SubType!
  extension: String @requires(fields:"subType {id }")
}

extend type SubType @key(fields:"id"){
  id: ID! @external
}
```

### Generated federation output

	
```
case "User":
	id0, err := ec.unmarshalNID2string(ctx, rep["id"])
	if err != nil {
		return errors.New(fmt.Sprintf("Field %s undefined in schema.", "id"))
	}

	entity, err := ec.resolvers.Entity().FindUserByID(ctx,
		id0)
	if err != nil {
		return err
	}

	entity.SubType.ID, err = ec.unmarshalNID2string(ctx, rep["subTypeID"])
	if err != nil {
		return err
	}

	list[i] = entity
	return nil
```

### Expected generated output

```
case "User":
	id0, err := ec.unmarshalNID2string(ctx, rep["id"])
	if err != nil {
		return errors.New(fmt.Sprintf("Field %s undefined in schema.", "id"))
	}

	entity, err := ec.resolvers.Entity().FindUserByID(ctx,
		id0)
	if err != nil {
		return err
	}

	entity.SubType.ID, err = ec.unmarshalNID2string(ctx, rep["subType"].(map[string]interface{})["id"])
	if err != nil {
		return err
	}

	list[i] = entity
	return nil
```

### What this does

When resolving User.extension, subType.ID = "null"

### versions
 - gqlgen: commit ed2d699, https://github.com/99designs/gqlgen/pull/1701
 - go: 1.16.3

carldunham:
A small change missed in previous PR, should fix #1704 

Included entity resolver test, and tossed in a test for interfaces as well. Happy to pull that, or move to a different PR.

I have:

### DIFF
--- a/plugin/federation/federation.gotpl
+++ b/plugin/federation/federation.gotpl
@@ -97,7 +97,7 @@ func (ec *executionContext) __resolve_entities(ctx context.Context, representati
 				}
 
 				{{ range .Requires }}
-					entity.{{.Field.JoinGo `.`}}, err = ec.{{.Type.UnmarshalFunc}}(ctx, rep["{{.Name}}"])
+					entity.{{.Field.JoinGo `.`}}, err = ec.{{.Type.UnmarshalFunc}}(ctx, rep["{{.Field.Join `"].(map[string]interface{})["`}}"])
 					if err != nil {
 						return err
 					}
--- a/plugin/federation/federation_entityresolver_test.go
+++ b/plugin/federation/federation_entityresolver_test.go
@@ -231,6 +231,47 @@ func TestEntityResolver(t *testing.T) {
 		require.Equal(t, resp.Entities[1].Name, "mars")
 		require.Equal(t, resp.Entities[1].Diameter, 10)
 	})
+
+	t.Run("PlanetRequiresNested entities with requires directive having nested field", func(t *testing.T) {
+		representations := []map[string]interface{}{
+			{
+				"__typename": "PlanetRequiresNested",
+				"name":       "earth",
+				"world": map[string]interface{}{
+					"foo": "A",
+				},
+			}, {
+				"__typename": "PlanetRequiresNested",
+				"name":       "mars",
+				"world": map[string]interface{}{
+					"foo": "B",
+				},
+			},
+		}
+
+		var resp struct {
+			Entities []struct {
+				Name  string `json:"name"`
+				World struct {
+					Foo string `json:"foo"`
+				} `json:"world"`
+			} `json:"_entities"`
+		}
+
+		err := c.Post(
+			entityQuery([]string{
+				"PlanetRequiresNested {name, world { foo }}",
+			}),
+			&resp,
+			client.Var("representations", representations),
+		)
+
+		require.NoError(t, err)
+		require.Equal(t, resp.Entities[0].Name, "earth")
+		require.Equal(t, resp.Entities[0].World.Foo, "A")
+		require.Equal(t, resp.Entities[1].Name, "mars")
+		require.Equal(t, resp.Entities[1].World.Foo, "B")
+	})
 }
 
 func TestMultiEntityResolver(t *testing.T) {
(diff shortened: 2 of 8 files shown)


MiguelCastillo:
@carldunham changes look good to me! Thank you for fixing. It would be a good idea to add tests here [1]. The schema file is this one [2].  Take a look here for more details for running them [3].

Happy to provide more specific details, just let me know.

1. https://github.com/99designs/gqlgen/blob/master/plugin/federation/federation_entityresolver_test.go
2. https://github.com/99designs/gqlgen/blob/master/plugin/federation/testdata/entityresolver/schema.graphql
3. https://github.com/99designs/gqlgen/blob/master/plugin/federation/readme.md

carldunham:
Sounds good. I have some tests in my reverted PR that I am revamping, I can drop those in here.

MiguelCastillo:
> Sounds good. I have some tests in my reverted PR that I am revamping, I can drop those in here.

Yeah sorry about that.  We just found two issues that were not straight forward to fix.

I just made a PR adding a few tests, including one for testing `requires` directive. https://github.com/99designs/gqlgen/pull/1708.  That should help with your tests as well.

plugin/federation/federation_test.go
```diff
@@ -77,6 +77,12 @@ func TestNoEntities(t *testing.T) {
 	require.Len(t, f.Entities, 0)
 }
 
+func TestInterfaces(t *testing.T) {
+	require.Panics(t, func() {
```
MiguelCastillo, line 81: Can you say more about the purpose of this test?  Are we just looking to verify that we can load a configuration that includes a schema that has interfaces?  Could this be an entity resolver test instead to verify that interfaces work?

plugin/federation/federation_entityresolver_test.go
```diff
@@ -231,6 +231,47 @@ func TestEntityResolver(t *testing.T) {
 		require.Equal(t, resp.Entities[1].Name, "mars")
 		require.Equal(t, resp.Entities[1].Diameter, 10)
 	})
+
+	t.Run("PlanetRequiresNested entities with requires directive having nested field", func(t *testing.T) {
```
MiguelCastillo, line 235: Thank you for writing these!

plugin/federation/federation.gotpl
```diff
@@ -97,7 +97,7 @@ func (ec *executionContext) __resolve_entities(ctx context.Context, representati
 				}
 
 				{{ range .Requires }}
-					entity.{{.Field.JoinGo `.`}}, err = ec.{{.Type.UnmarshalFunc}}(ctx, rep["{{.Name}}"])
+					entity.{{.Field.JoinGo `.`}}, err = ec.{{.Type.UnmarshalFunc}}(ctx, rep["{{.Field.Join `"].(map[string]interface{})["`}}"])
```
MiguelCastillo, line 100: Good catch.

MiguelCastillo (commented):
Changes look great! Thank you for fixing the issue and adding tests for it.  The only thing is around the interfaces tests.

plugin/federation/federation_test.go
carldunham: So this is a negative test to verify that we panic out when attempting to apply @key to an interface. Happy to move this to another PR, but it seems related.

carldunham:
OK, I removed that commit and will submit another PR for that

StevenACoffman:
Thanks so much! Looking forward to the next one!

PR #1717, opened 2021-11-17
Fix plugin addition

erwin-k:
As posted in Discord, the api.AddPlugin function appends plugins to the existing array of plugins, this causes the default plugins to run after the custom plugins, overwriting the result. (the modelgen hook example in the documentation does not work)

This fix makes it prepend the new plugin. May or may not be a breaking change for some.


StevenACoffman:
Thanks! I can't imagine that anyone was relying on the old behavior, but we should advertise this change for existing custom plugin authors.

Issue #1031, opened 2020-02-15
[Federation] Allow multiple entity keys.


BitPhinix:
### What happened?
GQLGen only generates one entity resolver function for an entity with 2 (or more) keys.

### What did you expect?
2 (or more) Entity resolver functions.

### Minimal graphql.schema and models to reproduce
```
type Product @key(fields: "upc") @key(fields: "sku") {
  upc: UPC!
  sku: SKU!
  name: String
}
```
[See Apollo Documentation](https://www.apollographql.com/docs/apollo-server/federation/federation-spec/#query_entities)

### versions
 - `gqlgen version`? 
 - `go version`?
 - dep or go modules?

Also, thanks for the awesome work :+1:

vektah:
Yeah, this is a known limitation for now. We will continue to iterate, but this wont make 0.11.

jclyons52:
It looks like this was resolved by https://github.com/99designs/gqlgen/pull/1723

PR #1723, opened 2021-11-21
Support for multiple @key directives in federation (reworked)

carldunham:
Reworked support for multiple @key and @requires directives in federation.

Would like feedback in general, and specifically for real use cases.

I have:


StevenACoffman (approved):
Looks pretty good to me. Do you feel like you're ready for review at this point? I can ask @MiguelCastillo if he has any more suggestions before merging.

carldunham:
So @MiguelCastillo had some specific issues he found in the last one, so happy to wait for his input. I pinged him on discord about it. I'll remove the "draft" status and we can go from there.

plugin/federation/federation.go
```diff
@@ -195,16 +201,29 @@ func (f *federation) GenerateCode(data *codegen.Data) error {
 		}
 		for _, e := range f.Entities {
 			obj := data.Objects.ByName(e.Def.Name)
+			e.Type = types.TypeString(obj.Type, func(i *types.Package) string {
+				// TODO: somehow this must recognize case of `code.ImportPathForDir(s.destDir) == path`
```
carldunham, line 205: pulling this comment, it's been addressed

plugin/federation/federation.go
```diff
@@ -195,16 +201,29 @@ func (f *federation) GenerateCode(data *codegen.Data) error {
...
-			for _, keyField := range e.KeyFields {
-				if len(keyField.Field) == 0 {
-					fmt.Println("skipping key field " + keyField.Definition.Name + " in " + e.Def.Name)
-					continue
+			for _, r := range e.Resolvers {
+				// fill in types for key fields
+				//
+				for _, keyField := range r.KeyFields {
+					if len(keyField.Field) == 0 {
+						fmt.Println("skipping field " + keyField.Definition.Name + " in " + r.ResolverName + " in " + e.Def.Name)
```
carldunham, line 221: Is this output needed?

plugin/federation/federation.gotpl
```diff
@@ -6,6 +6,8 @@
 
 {{ reserveImport "github.com/99designs/gqlgen/plugin/federation/fedruntime" }}
 
+{{ .Imports }}
```
carldunham, line 9: This seems a little hacky. Trying to deal with imports in the context of code generation. Is there a better way?

MiguelCastillo:
> So @MiguelCastillo had some specific issues he found in the last one, so happy to wait for his input. I pinged him on discord about it. I'll remove the "draft" status and we can go from there.

Thanks for the ping! I will give it a look later today.

plugin/federation/federation.gotpl
MiguelCastillo: Yeah I think this has the potential to introduce tricky circular dependencies, which I see you worked around so that at least autgen models work.  That was actually one of the issues I saw before.  I think this complexity is from wrapping the resolver in the self executing function.

example/federation/products/graph/generated/federation.go
```diff
@@ -75,36 +76,52 @@ func (ec *executionContext) __resolve_entities(ctx context.Context, representati
...
 		case "Product":
-			id0, err := ec.unmarshalNString2string(ctx, rep["manufacturer"].(map[string]interface{})["id"])
-			if err != nil {
-				return errors.New(fmt.Sprintf("Field %s undefined in schema.", "manufacturerID"))
-			}
-			id1, err := ec.unmarshalNString2string(ctx, rep["id"])
-			if err != nil {
-				return errors.New(fmt.Sprintf("Field %s undefined in schema.", "id"))
+			entity, err := func() (*model.Product, error) {
+				id0, err := ec.unmarshalNString2string(ctx, rep["manufacturer"].(map[string]interface{})["id"])
```
MiguelCastillo, line 100: This is the other issue.  If the query entity was for `FindProductByUpc` and there is no `manufacturer` representation, than this will fail because we would be trying to read `id` from `nil`.

https://go.dev/play/p/yLWsp2g0gyR

plugin/federation/federation.go
MiguelCastillo: i'd think so.  I think it would be useful to log why we are skipping a `@key`.  tho I would treat this more as an error, but leaving as is for now is good idea.

plugin/federation/testdata/entities/nokey.yml
```diff
@@ -4,3 +4,6 @@ exec:
   filename: testdata/entities/generated/exec.go
 federation:
   filename: testdata/entities/generated/federation.go
+
+autobind:
+  - "github.com/99designs/gqlgen/plugin/federation/test_data/model"
```
MiguelCastillo, line 9: It would be good to have a separate test with autobind.  Maybe have a different `nokey-autobind.yml` so that we can ensure these tests work with model autogen as well as autobinding.  WDYT?

MiguelCastillo:
Thanks for tackling this complicated issue.  The code itself to enable this looks good to me.  The issues are more around the generated code:

1. Wrapping the resolver in a function needs us to have `import` logic that make this complicated.  Ultimately I think we want to remove this, which I have some thoughts about below.
2. The process of determining which resolver to call needs some tweaking so that we are not building the arguments for the resolvers and also calling the resolver in that same self executing function.

I think that we can address for issues together tho!  What if we generated a function that for a given type and representation, it would determine what resolver to call.

```
func entityResolverForType(typename string, rep map[string]interface{}) (string, error) {
   switch typename:
   case "World":
      // Do the logic to determine rep based on the representations
      // what resolver to call.  This is the complicated part that you
      // are currently doing in the self executing functions.  But instead
     // of marshaling, we are just figuring what function to call.
      return "findHelloByUpc", nil
}
```

Then in `resolveEntity` the switch statement would look something more like:

```
   case "World":
      switch entityResolverForType(typename, rep)
        case "findHelloByUpc":
           // the same old marshaling logic we currently create without
          // wrapping in self executing function. Especially keeping the
          // marshaling logic linear without the nesting we are doing
          // in this PR.
```

carldunham:
Smart. Let me take a run at this and see where it goes. Getting rid of the imports would definitely clean things up, and this seems like a good way to do it.

StevenACoffman:
So... @duckbrain are you still working on this, or do you want a re-review as is?

carldunham:
I'm still working on this between other things, should have something to review in the next day or two.

carldunham:
@MiguelCastillo @StevenACoffman (and @duckbrain, if you care 🙂), I have updated this based on comments. Still missing a couple of things, noted in the commit comments. Happy to continue, or address in a follow-on. Also willing to squash and clean up before you merge.

Thanks!

carldunham:
@MiguelCastillo @StevenACoffman (and @duckbrain, if you care 🙂), I have updated this based on comments. Still missing a couple of things, noted in the commit comments. Happy to continue, or address in a follow-on. Also willing to squash and clean up before you merge.

Thanks!

StevenACoffman:
(Oh so sorry duckbrain... I just typed `@`, the first suggestion was you for some reason, and I just tab completed without thinking).

MiguelCastillo:
@carldunham i am planning on taking another look either today or tomorrow. thank you for your patience and your work on this!

lkendrickd:
@MiguelCastillo Thank you for looking into this.  We just adopted gqlgen and I was wondering when I had multiple @key described no entity resolvers would be created after the first initial @key.  I believe this also has to do with this open issue from last year.
https://github.com/99designs/gqlgen/issues/1031

superbeeny:
Looking forward to this PR getting merged. Thanks!

LockedThread:
@MiguelCastillo I would appreciate it a lot if you could continue this. This is a huge factor of gqlgen. If you're not able to could you tell me what else needs to be done so I could contribute?

plugin/federation/federation.go
MiguelCastillo: @carldunham let's add the word `key` in the message again.  It makes it more clear what it is we are skipping.  Below in line 217 we include `requires` to disambiguate.

plugin/federation/federation.go
```diff
@@ -243,100 +246,102 @@ func (f *federation) setEntities(schema *ast.Schema) {
 		}
 		if schemaType.Kind == ast.Object {
 			keys := schemaType.Directives.ForNames("key")
-			if len(keys) > 1 {
-				// TODO: support multiple keys -- multiple resolvers per Entity
-				panic("only one @key directive currently supported")
+			if len(keys) == 0 {
```
MiguelCastillo, line 249: Food for thought perhaps for later PRs.  One of the things that makes this code kinda hard to grok unless you are very familiar with federation internals is the lack of human readable conditionals.  Like I'd love to have a helper function here that's perhaps called something is `isFederatedEntity` or something.  And that checks `schemeType.Kind == ast.Object` and ensure that there are federated keys.

But yeah not for this PR.  Just thoughts I have had about how to make this code more readable and accessible to folks.

plugin/federation/federation.go
```diff
@@ -243,100 +246,102 @@ func (f *federation) setEntities(schema *ast.Schema) {
...
+			e := &Entity{
+				Name:      schemaType.Name,
+				Def:       schemaType,
+				Resolvers: nil,
+				Requires:  nil,
+			}
+
+			// Let's process custom entity resolver settings.
+			dir := schemaType.Directives.ForName("entityResolver")
+			if dir != nil {
```
MiguelCastillo, line 262: Same sort of idea here.  Like a helper function like `configureEntityResolverDirective`...  And `configureKeyDirective`...  These are TODOs in my head.  Hoping to get more cycles for this over the break.

plugin/federation/federation.go
```diff
@@ -243,100 +246,102 @@ func (f *federation) setEntities(schema *ast.Schema) {
...
 					}
 
-					e.InputType = resolverFieldsToGo + "Input"
+					e.Resolvers = append(e.Resolvers, &EntityResolver{
+						ResolverName: resolverName,
+						KeyFields:    keyFields,
+						InputType:    resolverFieldsToGo + "Input",
+					})
+
+					e.Requires = []*Requires{}
```
MiguelCastillo, line 325: I think we want to process `requires` directives outside of the loop for processing all the `keys` to avoid reprocessing them.

plugin/federation/federation.gotpl
```diff
@@ -78,41 +84,66 @@ func (ec *executionContext) __resolve_entities(ctx context.Context, representati
...
+					resolverName, err := ec.entityResolverNameFor{{.Def.Name}}(ctx, rep)
 					if err != nil {
-						return err
+						return fmt.Errorf(`finding resolver for Entity "{{.Def.Name}}": %w`, err)
 					}
-				{{ end }}
-
-				list[idx[i]] = entity
-				return nil
+					{{ if eq (len .Resolvers) 1 }}
```
MiguelCastillo, line 95: I think it should be OK to always generate the switch statement. There isn't a huge runtime performance penalty and it keep this already giant template more manageable and less error prone since you have to make any template changes in multiple places.

Or is there a reason you'd prefer to having a different generated code when there is only 1 entity vs multiple?

plugin/federation/federation.gotpl
```diff
@@ -163,7 +195,8 @@ func (ec *executionContext) __resolve_entities(ctx context.Context, representati
 					return nil
 				{{ end }}
 			{{ end }}
-		{{- end }}
+		{{ end }}
+	{{- end }}
```
MiguelCastillo, line 199: 😅   templates tags are so hard to match up!!!  Did an `{{ end }}`  sneak in or is there a misalignment that makes it harder to spot where the opening tag is?

plugin/federation/federation.gotpl
```diff
@@ -6,6 +6,12 @@
 
 {{ reserveImport "github.com/99designs/gqlgen/plugin/federation/fedruntime" }}
 
+var (
+	ErrUnknownType = errors.New("unknown type")
+	ErrTypeNotFound = errors.New("type not found")
+	ErrUnableToResolveEntity = errors.New("unable to resolve Entity")
```
MiguelCastillo, line 12: Is ErrUnableToResolveEntity being used?  It seems like it isn't...

plugin/federation/federation.gotpl
```diff
@@ -217,4 +249,64 @@ func (ec *executionContext) __resolve_entities(ctx context.Context, representati
 		return list
 	}
 }
+
+{{- /* Make sure the required fields are in the given entity representation and return the name of the proper resolver. */ -}}
+
+{{ range $_, $entity := .Entities }}
+	{{- if .Resolvers }}
+
+		func (ec *executionContext) entityResolverNameFor{{$entity.Name}}(ctx context.Context, rep map[string]interface{}) (string, error) {
```
MiguelCastillo, line 258: From a mental model standpoint, I think it would be a lil better if `ec` was passed in as an argument to `entityResolverNameFor`...  Not a major thing, so if you feel strongly about keeping it as is then no problem.  But in this generated code the execution only really care about being able to call `__resolve__service` and `__resolve_entities` as well as providing an interface for marshaling fields. 

Tho - what do you think about not calling any marshaling functions in the generated `entityResolverNameFor` functions?  We are primarily interested in matching up the shapes of `rep` to determine what resolver to call.  We don't care as much about marshaling the values.  Especially since `rep` values are getting marshaled again before calling the entity resolvers.   That would do two things: 1. Avoid double marshaling for values. 2. Can allow us to not even pass `ec` into these generated function.

Thoughts?

MiguelCastillo:
@carldunham this is absolute awesome!  Im really excited for this changes to land, and so are lots of other folks. 👏 
I left you a couple of thoughts for cleaning up.  But nothing really major.  Overall the flow and the generated is pretty awesome.  Let me know what you think of the comments I left.

carldunham:
Thanks, @MiguelCastillo. Great feedback! I'll attack this over the weekend, if not today. It looks like mostly moving and removing, should clean things up nicely!

MiguelCastillo:
> @MiguelCastillo I would appreciate it a lot if you could continue this. This is a huge factor of gqlgen. If you're not able to could you tell me what else needs to be done so I could contribute?

Thank you for your offer @LockedThread!  💟     Happily I was able to squeeze in some cycles today for this.  I think we are very close to landing.  And it's always helpful to have other folks eyes on PRs, so if you go thru the changes and spot anything - call it out!  It can only help :)

LockedThread:
> > @MiguelCastillo I would appreciate it a lot if you could continue this. This is a huge factor of gqlgen. If you're not able to could you tell me what else needs to be done so I could contribute?
> 
> 
> 
> Thank you for your offer @LockedThread!  💟     Happily I was able to squeeze in some cycles today for this.  I think we are very close to landing.  And it's always helpful to have other folks eyes on PRs, so if you go thru the changes and spot anything - call it out!  It can only help :)

Alright, I'll stay up-to-date on your changes. Good luck!

plugin/federation/federation.gotpl
carldunham: Only that it made the generated code cleaner. But you're right, it comes at the cost of a more complicated template, which will probably need to be comprehended more often. I'm removing the ==1 case.

plugin/federation/federation.gotpl
carldunham: Hopefully, it's a little more aligned and clear now. I think indenting at the case statements might make things clearer? But not doing that yet.

plugin/federation/federation.gotpl
carldunham: Not any more :)

plugin/federation/federation.gotpl
carldunham: The only weird cases would be where an invalid representation in the map[string]interface{} would result in that resolver being chosen, rather than another that had proper reps. Of course, the reverse could happen also, and not be caught. I'll modify it and we can see how it looks.

plugin/federation/federation.go
carldunham: Went ahead and added this. Hopefully returning the keys isn't more confusing.

plugin/federation/federation.go
carldunham: Left this for you 🙂

plugin/federation/federation.go
carldunham: Yeah, dumb oversight on my part!

plugin/federation/federation.gotpl
carldunham: There were two places where this could be done, hopefully strikes a good balance.

carldunham:
@MiguelCastillo @LockedThread ready for re-review!

plugin/federation/federation.go
```diff
@@ -281,67 +293,69 @@ func (f *federation) setEntities(schema *ast.Schema) {
...
+		f.Entities = append(f.Entities, e)
 	}
 
 	// make sure order remains stable across multiple builds
 	sort.Slice(f.Entities, func(i, j int) bool {
 		return f.Entities[i].Name < f.Entities[j].Name
 	})
 }
+
+func isFederatedEntity(schemaType *ast.Definition) ([]*ast.Directive, bool) {
```
MiguelCastillo, line 342: Ah thank you!  Yeah this is is great.  We can tweak the function name as we use it more.

MiguelCastillo (approved):
Woohoo!  This is great, @carldunham.  Thank you for all your work with this.

plugin/federation/federation.gotpl
MiguelCastillo: yay!

plugin/federation/federation.go
MiguelCastillo: all good! these changes were very tricky.  and hard to catch in unit tests since the generated should still work correctly. 🤷

plugin/federation/federation.go
MiguelCastillo: Thank you :)

plugin/federation/federation.go
MiguelCastillo: Yeah I think its great.  The only nit is that functions that start with `is` usually are just predicate functions.  I would have called this instead perhaps `getFederatedEntityKeys` or something and that can return they keys and ok...

but this is perfect.  We can always iterate on this nits.

PR #1725, opened 2021-11-22
Don't overwrite field arguments when none match

### LINKED ISSUE #1319: Method name collision on model results in buggy resolver
### What happened?

When a model has a method named after a field it should resolve but has the incorrect signature, and the field has parameters, the resulting generated resolver has the incorrect signature. In particular the input parameters are missing.

### What did you expect?

The input parameters to be part of the generated resolver signature. If not possible then second best option is for the generation to error.

Schema:
```
type Todo {
  working(in: Int): Int
  broken(in: Int): Int
}

type Query {
  todo: Todo
}
```

Model:
```
package model

import "context"

type Todo struct{}

func (*Todo) Broken(ctx context.Context) string {
	return ""
}
```

Resulting generated resolvers:
```
func (r *todoResolver) Working(ctx context.Context, obj *model.Todo, in *int) (*int, error) {
	panic(fmt.Errorf("not implemented"))
}

func (r *todoResolver) Broken(ctx context.Context, obj *model.Todo) (*int, error) {
	panic(fmt.Errorf("not implemented"))
}
```

Expected generated resolver:
```
func (r *todoResolver) Working(ctx context.Context, obj *model.Todo, in *int) (*int, error) {
	panic(fmt.Errorf("not implemented"))
}

func (r *todoResolver) Broken(ctx context.Context, obj *model.Todo, in *int) (*int, error) {
	panic(fmt.Errorf("not implemented"))
}
```

### versions
 - `gqlgen version` - `v0.12.2`
 - `go version` - `1.13`
 - dep or go modules - `modules`

### extra

You can find all the code related to this issue as generated by walking through the getting started guide  and adapted accordingly in [this repo](https://github.com/matiasanaya/gqlgen-scratch/tree/main/shadow-resolver).

edigaryev:
Problem: during GraphQL field binding, when there's matching Golang method, yet without any arguments besides `ctx`, a resolver with no arguments from the original GraphQL schema will be generated.

Solution: don't overwrite field's arguments until we're sure we've got a perfect match.

Resolves #1319.

### DIFF
--- a/codegen/args.go
+++ b/codegen/args.go
@@ -73,7 +73,7 @@ func (b *builder) buildArg(obj *Object, arg *ast.ArgumentDefinition) (*FieldArgu
 	return &newArg, nil
 }
 
-func (b *builder) bindArgs(field *Field, params *types.Tuple) error {
+func (b *builder) bindArgs(field *Field, params *types.Tuple) ([]*FieldArgument, error) {
 	var newArgs []*FieldArgument
 
 nextArg:
@@ -83,7 +83,7 @@ nextArg:
 			if strings.EqualFold(oldArg.Name, param.Name()) {
 				tr, err := b.Binder.TypeReference(oldArg.Type, param.Type())
 				if err != nil {
-					return err
+					return nil, err
 				}
 				oldArg.TypeReference = tr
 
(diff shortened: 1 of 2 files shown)


StevenACoffman:
Thanks!

Issue #1715, opened 2021-11-15
How to completely remove dataloaden from go.mod?


frederikhors:
### What happened?

I removed `dataloaden` from my project ([because of this](https://github.com/99designs/gqlgen/issues/1714)) but I'm still having `dataloden` in my `go.mod` as an `indirect` dependency. **Why?**

```
go 1.17

require (
  ...
  github.com/99designs/gqlgen v0.14.0
  github.com/graph-gophers/dataloader/v6 v6.0.0
  github.com/vektah/gqlparser/v2 v2.2.0
  ...
)

require (
  ...
  github.com/vektah/dataloaden v0.2.1-0.20190515034641-a19b9a6e7c9e // indirect
  ...
)
```

[I searched in project for "dataloaden"](https://github.com/99designs/gqlgen/search?q=dataloaden) but nothing appears except for `example` dir. **Is this the problem?**

I also tried with:

`go mod why github.com/vektah/dataloaden`

in my project, which shows:

```
# github.com/vektah/dataloaden
myProject/lib/gqlgen
github.com/99designs/gqlgen
github.com/vektah/dataloaden
```

### What did you expect?

To remove it completely.

### versions
 - `gqlgen version` v0.14.0 
 - `go version` 1.17.3

duckbrain:
From `go help packages`:

> Directory and file names that begin with "." or "_" are ignored
> by the go tool, as are directories named "testdata".

You could rename `example` to `_examples`. That _should_ get all of the Go tooling to ignore the whole directory. My testing on my machine indicates that the projects inside of `_examples` would still function normally if directly compiled. ie: `go install ./_examples/project-dir` works but `go install ./_examples/...` will have "matched no packages"

frederikhors:
Thanks @duckbrain.

PR #1734, opened 2021-11-30
rename "example" dir to "_examples"

frederikhors:
Should fix https://github.com/99designs/gqlgen/issues/1715. 

I have:


StevenACoffman:
So I think you also need to change the github actions and documentation to match to get the test failures to go away?

frederikhors:
@StevenACoffman what this test fails?

```
--- FAIL: TestRewriter (0.11s)
    --- FAIL: TestRewriter/default (0.11s)
        rewriter_test.go:27: 
            	Error Trace:	rewriter_test.go:27
            	Error:      	"[]" should have 2 item(s), but has 0
            	Test:       	TestRewriter/default
```

duckbrain:
@frederikhors  It looks like `internal/rewrite/rewriter_test.go` should be reverted. It was referencing a file named `example.go`, which was not renamed (and is already in a directory named `testdata`, so it doesn't need to be).

frederikhors:
@duckbrain done. What is the problem now?

@StevenACoffman can you please merge this as soon as we understand the remaining problem? I had to fix a lot of conflicts from the master branch.

StevenACoffman:
Sure. If we can get it passing, I'd like to merge it sooner rather than later.

frederikhors:
@StevenACoffman I do not understand what these 2 errors are.

duckbrain:
The Lint/Lint one probably needs running go generate on each of the examples. Because of the rename, go generate ./... will no longer generate the examples, so you'll need to refer to them individually.

The integration/federation one is likely referring to an old path in start.sh. need to look into it more.

frederikhors:
I fixed lint step.

Now only this is the problem:

image

I cannot understand the reason!

mtibben:
```
.github/workflows/check-federation: line 7: ./start.sh: Permission denied
```

Were the permissions for `start.sh` changed? A `chmod 755 start.sh` should fix it

frederikhors:
> ```
> .github/workflows/check-federation: line 7: ./start.sh: Permission denied
> ```
> 
> Were the permissions for `start.sh` changed? A `chmod 755 start.sh` should fix it

Oh! Wow! I'm on Windows. Can you please `chmod` it?

mtibben:
hmmm you should be able to use [WSL](https://docs.microsoft.com/en-us/windows/wsl/)?

frederikhors:
> hmmm you should be able to use [WSL](https://docs.microsoft.com/en-us/windows/wsl/)?

I'm not really in that things... 😢

duckbrain:
@mtibben Good catch.

@frederikhors I don't use Windows, does  `git update-index --chmod=+x _examples/federation/start.sh` do it?

Source https://www.scivision.dev/git-windows-chmod-executable/

StevenACoffman:
FTFY.

StevenACoffman:
One concern is that it is worth highlighting in the directions for contributing code how to run all these checks locally. When they fail, it's not obvious from the github action failures what the problem is.

frederikhors:
Can we merge now?

PR #1737, opened 2021-12-01
fix: automatically register built-in directive goTag

### LINKED ISSUE #1736: Default/builtin `goTag` directive requires implementation in `DirectiveRoot`
### What happened?

I learned about the inline `goTag` directive from this documentation: https://github.com/99designs/gqlgen/blob/37a4e7eefa241dd7e5eeab7bbca56b1677d37daf/docs/content/config.md#inline-config-with-directives

After applying it to my schema, regenerating code, and running my GQL server my GQL client started reporting these errors:

```
Error: directive goTag is not implemented
```

Doing more research I found the following documentation about directives and the need to implement them: https://github.com/99designs/gqlgen/blob/8b9737179d3ba08405ab6422fb8c4883dd8e720c/docs/content/reference/directives.md#implement-the-directive

I was able to implement the directive on `DirectiveRoot` (`config.Directives`) as a no-op/pass through directive like so and eliminated the error:

```
resolver := resolver.CreateAndInitResolver()

config := server.Config{Resolvers: resolver}

config.Directives.GoTag = func(ctx context.Context, obj interface{}, next graphql.Resolver, key string, value *string) (res interface{}, err error) {
	// No-op/pass through directive
	return next(ctx)
}

graphql := handler.NewDefaultServer(server.NewExecutableSchema(config))
```

### What did you expect?

The aforementioned config documentation states (https://github.com/99designs/gqlgen/blob/37a4e7eefa241dd7e5eeab7bbca56b1677d37daf/docs/content/config.md#inline-config-with-directives):

> gqlgen ships with some builtin directives that make it a little easier to manage wiring.

It then lists the following builtin directives: `goModel`, `goField`, `goTag`

After reading the documentation I would expect any use of builtin directives to not require me to need to implement the directive on `DirectiveRoot` (`config.Directives`).

If an implementation is required then I would have expected:
* (Preferred :smile:) The Go code generation process to define it for more
* (Not Preferred :vomiting_face:) The documentation to state that I need to implement it with example

`Schema`
```
directive @goTag(
    key: String!
    value: String
) on INPUT_FIELD_DEFINITION | FIELD_DEFINITION

type Profile {
  email: String! @goTag(key: "validate", value: "required,email")
}

type Query {
  profile(email: String!): Profile!
}

type Mutation {
  createProfile(email: String!): Profile!
}
```

`Model`
```
type Profile struct {
	Email string `json:"email" validate:"required,email"`
}
```

Note the use of `validator` library/tags in conjunction with `goTag`: https://github.com/go-playground/validator

I wouldn't expect it to matter (i.e. you can change the key and value to `goTag` to anything and still get the same error.

### versions
 - `gqlgen version`? 
`github.com/99designs/gqlgen v0.14.1-0.20211122181954-4581fccdb67e`

I'm actually pinned to `master` at this commit: `https://github.com/99designs/gqlgen/commit/4581fccdb67e05937ba862bf601a18393731bccc`

 - `go version`?
```
$ go version
go version go1.17.3 linux/amd64
```
 
 - dep or go modules?

`go modules`

tprebs:
Fixes #1736 

The built-in directive `goTag` was not automatically registered as a runtime directive causing a directive to be generated when it is not required. This change adds the directive as a runtime directive as default. 

I have:

### DIFF
--- a/codegen/config/config.go
+++ b/codegen/config/config.go
@@ -239,6 +239,10 @@ func (c *Config) injectTypesFromSchema() error {
 		SkipRuntime: true,
 	}
 
+	c.Directives["goTag"] = DirectiveConfig{
+		SkipRuntime: true,
+	}
+
 	for _, schemaType := range c.Schema.Types {
 		if schemaType == c.Schema.Query || schemaType == c.Schema.Mutation || schemaType == c.Schema.Subscription {
 			continue
--- a/docs/content/config.md
+++ b/docs/content/config.md
@@ -115,3 +115,16 @@ type User @goModel(model: "github.com/my/app/models.User") {
 		@goTag(key: "yaml")
 }
 ```
+
+The builtin directives `goField`, `goModel` and `goTag` are automatically registered to `skip_runtime`. Any directives registered as `skip_runtime` will not exposed during introspection and are used during code generation only. 
+
+If you have created a new code generation plugin using a directive which does not require runtime execution, the directive will need to be set to `skip_runtime`.
+
+e.g. a custom directive called `constraint` would be set as `skip_runtime` using the following configuration
+```yml
+# custom directives which are not exposed during introspection. These directives are
+# used for code generation only
+directives:
+  constraint:
+    skip_runtime: true
+```


Giners:
Thanks so much for this @tprebs! :smile: 

Might I recommend at least indicting in the documentation that describes the built-in directives that the `gqlgen.yml` needs to have this added to it?

```
directives:
  goTag:
    skip_runtime: true
```

Even if you have it so the config automatically generates then if there are modifications to the config/or something else might cause the config to not automatically generate properly there will at least be documentation describing how to manually update it so it isn't registered.

Also do we need to automatically register `goField` and `goModel`?

Here are is the doc in question: https://github.com/99designs/gqlgen/blob/37a4e7eefa241dd7e5eeab7bbca56b1677d37daf/docs/content/config.md#inline-config-with-directives

tprebs:
Adding that block to `gqlgen.yml` was just a temporary work around. Once this change is merged it is not required as the `goTag` directive is automatically registered. This is how `goField` and `goModel` already behave.

Reading back though the documentation you are right, there is no documentation explaining runtime directives in either 

https://github.com/99designs/gqlgen/blob/37a4e7eefa241dd7e5eeab7bbca56b1677d37daf/docs/content/config.md#inline-config-with-directives

or 

https://github.com/99designs/gqlgen/blob/37a4e7eefa241dd7e5eeab7bbca56b1677d37daf/docs/content/reference/directives.md

I will update the  inline config with directives documentation.

StevenACoffman:
Thanks so much!

Issue #1732, opened 2021-11-29
List input coercion not work with enum provided as variable


anton-kabysh-spire:
### What happened?

When you send query graphql is able to coerce single value to an array:

```graphql
# WOKS!
query vessels {
  vessels(first: 3, shipType: [CAR_CARRIER]) {
    nodes {
      id
    }
  }
}
```

```graphql
# WORKS!
query vessels {
  vessels(shipType: CAR_CARRIER) {
    nodes {
      id
    }
  }
}
```

But when enum value submitted via variables, the coercion does not work:

```graphql
# FAILS!
query vessels($shipType: [ShipType!]) {
  vessels(shipType: $shipType) {
    nodes {
      id
    }
  }
}
```

with variables:
```json
{
 	"shipType": "CAR_CARRIER"
}
```

with message `"enums must be strings"`

### What did you expect?

Array coercion works for variables.

### Minimal graphql.schema and models to reproduce

```graphql
type Query {
  vessels(shipType: [ShipType!]!): [Vessel]
}

enum ShipType {
  CAR_CARRIER
}

```

### versions
 - `gqlgen version` - v0.14.0 
 - `go version` - go 1.17.2

tprebs:
I ran into the same issue today. 

List coercion is currently managed by the generated gqlgen unmarshallers. A recent change to `vektah/gqlparser` (https://github.com/vektah/gqlparser/pull/138) has added coercion into the variable validation process which is used by gqlgen to validate all variables. This method of coercion conflicts with `gqlgen`'s implementation causing it to fail. 

The gqlgen unmarshallers still work as they did before for non-variable coercion but variable coercion no longer works for any type of variable, not just enums. 

The unmarshallers apply coercion by checking the type of the input. if the input is not already a slice of interfaces (the decoding of variables converts all slices into a slice of interfaces) then it is assumed that coercion is required and it creates a new slice and wraps the input. the PR mentioned above to gqlparser now applies its own coercion creating a slice with the correct type, in your case `[]string`, but this could also be `[]map[string]interface{}`, `int[]`,`float[]`,`int64[]`... etc. 

This causes the `v.([]interface{})` type conversion to fail and ultimately the unmarshalling to fail.

I do have a PR which I have nearly finished with which generates a better type check for these unmarshallers but I am not sure if that is the best place to resolve this. I am not 100% convinced that the validator in gqlparser should be applying the coercion in the first place.

Issue #1369, opened 2020-10-23
List input coercion not work when using variables


NateScarlet:
### What happened?

frontend code results `must be an array` error after i switch from apollo-server to gqlgen

![image](https://user-images.githubusercontent.com/21083014/96971515-cb251b00-1547-11eb-93e6-a045e85d6f86.png)

#244 added support for literal values but not variables

### What did you expect?

Pass single value as array variable should not error.

https://spec.graphql.org/June2018/#sec-Type-System.List

### Minimal graphql.schema and models to reproduce

schema:
```graphql
type Todo {
  name: String!
}

type Query {
  todos(name: [String!]): [Todo!]
}
```
query:
```grapqhl
query todos($name: [String!]) {
  name
}
```
variables:
```json
{"name": "foo"}
```

### versions
 - `gqlgen version`?: 0.13.0
 - `go version`?: 1.14
 - dep or go modules?: go modules

NateScarlet:
https://github.com/vektah/gqlparser/blob/b778db6115bc4ae40c63fa4a9c2e9805344ce182/validator/vars_test.go#L154-L160

This is a upstream issue

NateScarlet:
after upgrade gqlparser to v2.2.0

the error changed to `[]string is not a string` because `[]string` is not assignable to `[]interface{}`

need update
https://github.com/99designs/gqlgen/blob/3f68ea27a1a9fea2064caf877f7e24d00aa439e6/codegen/type.gotpl#L13-L17

NateScarlet:
This seems fix the issue
```golang
func normalizeVariableValue(v interface{}) interface{} {
	var rv = reflect.ValueOf(v)
	switch rv.Type().Kind() {
	case reflect.Array:
		fallthrough
	case reflect.Slice:
		var ret = make([]interface{}, rv.Len())
		for i := range ret {
			ret[i] = rv.Index(i).Interface()
		}
		return ret
	}
	return v

}

// Issue1369Fix https://github.com/99designs/gqlgen/issues/1369
type Issue1369Fix struct{}

func (Issue1369Fix) ExtensionName() string {
	return "Issue1369Fix"
}
func (Issue1369Fix) Validate(schema graphql.ExecutableSchema) error {
	return nil
}

func (Issue1369Fix) MutateOperationContext(ctx context.Context, rc *graphql.OperationContext) *gqlerror.Error {
	for k, v := range rc.Variables {
		rc.Variables[k] = normalizeVariableValue(v)
[...]

PR #1740, opened 2021-12-06
Fix list coercion when using graphql variables

tprebs:

This PR enables support for list type coercion when using variables. This currently works fine without variables but as of gqlparser 2.2, this does not work when using variables due to gqlparser sending back typed slices e.g. `[]string` instead of `[]interface{}` which is what the generated code is expecting. This PR resolves this issue by type checking against all the possible types which may be sent by `gqlparser`. 

There are 2 open issues currently tracking this which are #1369 and #1732. 

I have:


StevenACoffman:
Thanks!

PR #1743, opened 2021-12-13
Downgrade to Go 1.16

StevenACoffman:


MiguelCastillo:
ah good idea.  I think there are some things to work out in terms of process of which versions we support and when to update go.mod.  But I think 1.16 is a good starting point.

Im happy to merge and if others have objections or thoughts we can easily adjust.

PR #1760, opened 2022-01-12
Avoid problems with `val` being undefined in the federation template.

csilvers:
When running gqlgen over our schema, we were seeing errors like:
```
assignments/generated/graphql/service.go:300:4: val declared but not used
```

The generated code looks like this:
```
func entityResolverNameForMobileNavigation(ctx context.Context, rep map[string]interface{}) (string, error) {
        for {
                var (
                        m   map[string]interface{}
                        val interface{}
                        ok  bool
                )
                m = rep
                if _, ok = m["kaid"]; !ok {
                        break
                }
                m = rep
                if _, ok = m["language"]; !ok {
                        break
                }
                return "findMobileNavigationByKaidAndLanguage", nil
        }
        return "", fmt.Errorf("%w for MobileNavigation", ErrTypeNotFound)
}
```

Looking at the code, it's pretty clear that this happens when there
are multiple key-fields, but each of them has only one keyField.Field
entry.  This is because the old code looked at `len(keyFields)` to
decide whether to declare the `val` variable, but looks at
`len(keyField.Field)` for each keyField to decide whether to use the
`val` variable.

The easiest solution, and the one I do in this PR, is to just declare
`val` all the time, and use a null-assignment to quiet the compiler
when it's not used.

We have added tests to ensure that this fixes the issue, which you can run:
1. Go to `plugins/federation`
2. Run the command `go run github.com/99designs/gqlgen --config testdata/entityresolver/gqlgen.yml`
3. Verify no errors are generated.

I have:

### DIFF
--- a/plugin/federation/testdata/entityresolver/generated/federation.go
+++ b/plugin/federation/testdata/entityresolver/generated/federation.go
@@ -104,6 +104,30 @@ func (ec *executionContext) __resolve_entities(ctx context.Context, representati
 				list[idx[i]] = entity
 				return nil
 			}
+		case "HelloMultiSingleKeys":
+			resolverName, err := entityResolverNameForHelloMultiSingleKeys(ctx, rep)
+			if err != nil {
+				return fmt.Errorf(`finding resolver for Entity "HelloMultiSingleKeys": %w`, err)
+			}
+			switch resolverName {
+
+			case "findHelloMultiSingleKeysByKey1AndKey2":
+				id0, err := ec.unmarshalNString2string(ctx, rep["key1"])
+				if err != nil {
+					return fmt.Errorf(`unmarshalling param 0 for findHelloMultiSingleKeysByKey1AndKey2(): %w`, err)
+				}
+				id1, err := ec.unmarshalNString2string(ctx, rep["key2"])
+				if err != nil {
+					return fmt.Errorf(`unmarshalling param 1 for findHelloMultiSingleKeysByKey1AndKey2(): %w`, err)
+				}
+				entity, err := ec.resolvers.Entity().FindHelloMultiSingleKeysByKey1AndKey2(ctx, id0, id1)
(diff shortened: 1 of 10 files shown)


csilvers:
I don't seem to able to suggest reviewers, but @carldunham you're probably the best person to review this  Also cc-ing @StevenACoffman and @MiguelCastillo for landing this when the time comes.

carldunham:
Good catch @csilvers! Could you add some unit tests that would fail without this fix?

StevenACoffman (approved):
LGTM. Seems a little difficult to make a meaningful unit test for.

MiguelCastillo:
> LGTM. Seems a little difficult to make a meaningful unit test for.

It shouldn't be too bad.  We can create a unit test based on how @csilvers found it in the first place.  Let me see if I can add one.  But happy to do after in a separate PR.

MiguelCastillo:
Ok I added a test. I verified the issue happens without these changes and that the issue is fixed with these changes.  Also - this is a very clever solution, Craig!  I would definitely tried adjusting the logic.  This seems good enough but may circle back if a more clear fix comes to mind.

plugin/federation/testdata/entityresolver/generated/federation.go
```diff
@@ -384,12 +410,35 @@ func entityResolverNameForHello(ctx context.Context, rep map[string]interface{})
...
 }
 
+func entityResolverNameForHelloMultiSingleKeys(ctx context.Context, rep map[string]interface{}) (string, error) {
+	for {
+		var (
+			m   map[string]interface{}
+			val interface{}
+			ok  bool
+		)
+		_ = val
```
MiguelCastillo, line 420: This is where the fix does its thing.

csilvers:
Thanks for the quick review everyone!  And extra thank you for adding the test @MiguelCastillo !

PR #1763, opened 2022-01-14
Fix #1762: Reload packages before merging type systems

### LINKED ISSUE #1762: generated models across loaded packages causes error merging type systems
### What happened?
If models are generated in a package that has already been loaded, and
that package refers to another package that has already been loaded, we
can find ourselves in a position where it appears that a GQL `union` is
not satisfied.

https://github.com/99designs/gqlgen/pull/1020 introduces the
code that fails in this scenario. This was supposed to be addressed in https://github.com/99designs/gqlgen/pull/1491 by @wendorf
but per @ipfans https://github.com/99designs/gqlgen/pull/1491#issuecomment-1012046943 it appears to still be occurring.

This repository currently reproduces the issue reliably:
https://github.com/ipfans/gqlgen-error

and is currently getting this error:
```
merging type systems failed: unable to bind to interface: xxx/src/business/entity.A does not satisfy the interface gitlab.com/hookactions/gqlgen-relay/relay.Node
```

ipfans:
Fixes #1762  `cfg.ReloadAllPackages()` should be called before merging type systems.

I have:

### DIFF
--- a/codegen/data.go
+++ b/codegen/data.go
@@ -54,6 +54,9 @@ func (d *Data) Directives() DirectiveList {
 }
 
 func BuildData(cfg *config.Config) (*Data, error) {
+	// We reload all packages to allow packages to be compared correctly.
+	cfg.ReloadAllPackages()
+	
 	b := builder{
 		Config: cfg,
 		Schema: cfg.Schema,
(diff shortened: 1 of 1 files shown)


StevenACoffman:
Thanks for the fix!

PR #1839, opened 2022-01-24
add PrependPlugin

Code-Hex:
This PR should only be merged after https://github.com/99designs/gqlgen/pull/1838 is merged. #1838 reverts 99designs/gqlgen#1717

- Some plugins that worked in versions prior to v0.15, which generated code based on a previously generated model, no longer work above v0.15.
  - https://github.com/99designs/gqlgen/blob/8b25c9e005c44ae3730eca83445fa7f7223481d1/api/generate.go#L21-L32
- I think we shouldn't implicit breaking changes should be made
- If we want to achieve the behavior @erwin-k claims, we should create a new API.
  - e.g. `func PrependPlugin`

I have:

### DIFF
--- a/api/option.go
+++ b/api/option.go
@@ -19,6 +19,13 @@ func AddPlugin(p plugin.Plugin) Option {
 	}
 }
 
+// PrependPlugin prepends plugin any existing plugins
+func PrependPlugin(p plugin.Plugin) Option {
+	return func(cfg *config.Config, plugins *[]plugin.Plugin) {
+		*plugins = append([]plugin.Plugin{p}, *plugins...)
+	}
+}
+
 // ReplacePlugin replaces any existing plugin with a matching plugin name
 func ReplacePlugin(p plugin.Plugin) Option {
 	return func(cfg *config.Config, plugins *[]plugin.Plugin) {
--- a/api/option_test.go
+++ b/api/option_test.go
@@ -53,3 +53,16 @@ func TestReplacePlugin(t *testing.T) {
 		require.EqualValues(t, expectedPlugin, pg[2])
 	})
 }
+
+func TestPrependPlugin(t *testing.T) {
+	modelgenPlugin := modelgen.New()
+	pg := []plugin.Plugin{
+		modelgenPlugin,
+	}
+
+	expectedPlugin := &testPlugin{}
+	PrependPlugin(expectedPlugin)(config.DefaultConfig(), &pg)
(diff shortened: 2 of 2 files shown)


StevenACoffman:
Thanks for your work on this. Can you please add test coverage for your new PrependPlugin?

It seems like to make the modelgen documentation example work we **_do_** need AppendPlugin functionality as mentioned by @erwin-k in #1717.
Which of those two behaviors "AddPlugin"  should default to is more of an open question. Breaking changes are to be avoided, but I would like to know which known plugins require which behavior. Do you have examples of either?

erwin-k:
@Code-Hex I don't remember it exactly anymore but at the time I pushed this change it seemed that the plugins were being used in reverse order (by gqlgen). The former plugins would overwrite results of latter plugins (in the array), so prepending plugins would give desired behaviour for user using "addPlugin". Atleast that was what I thought was happening.
Maybe it wasn't the case.

@StevenACoffman perhaps it only applied to the modelgen hook or something? I noticed the documentation for modelgen is using "replacePlugin" now.

Code-Hex:
@erwin-k Even if this were the case, it would not be a reason to implicitly change the behavior already provided in other versions.

> The former plugins would overwrite results of latter plugins (in the array), so prepending plugins would give desired behaviour for user using "addPlugin". Atleast that was what I thought was happening.
Maybe it wasn't the case.

Look at the API provided by Go itself. Create new APIs without breaking the logic already provided. Always use the "Deprecated" comment even if you are going to deprecate.

StevenACoffman:
While avoiding breaking compatibility with existing usage is always the goal, mistakes happen. As mentioned in #1717 at the time I deemed it unlikely that anyone was _relying_ on behavior that had changed from even earlier releases. These changes had then caused the documented example to fail. It also seemed that `AddPlugin` was somewhat counterintuitive in that it seemed that the plugins were being applied in reverse order, but this wasn't well documented or explicit.

Thanks to your work, people can now explicitly choose the behavior that they want. We can improve documentation on whether or not the plugins are applied in reverse/forward order (PRs welcome).

Please further consider contributing test cases for any uncovered areas of the code that you would like to avoid us unintentionally breaking compatibility with.

I am sorry for any inconvenience this may have caused you. Please also grant fellow open-source contributors the benefit of assuming only the best of intentions.

I will make a new v0.16.0 release with these changes shortly.

PR #1877, opened 2022-02-01
Add ZIP code to common initialisms

FlymeDllVa:
Add ZIP code to common initialisms.

I have:

### DIFF
--- a/codegen/templates/templates.go
+++ b/codegen/templates/templates.go
@@ -16,7 +16,6 @@ import (
 	"unicode"
 
 	"github.com/99designs/gqlgen/internal/code"
-
 	"github.com/99designs/gqlgen/internal/imports"
 )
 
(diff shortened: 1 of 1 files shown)


codegen/templates/templates.go
```diff
@@ -493,6 +492,7 @@ var commonInitialisms = map[string]bool{
 	"XMPP":  true,
 	"XSRF":  true,
 	"XSS":   true,
+	"ZIP":   true,
```
wilhelmeek, line 495: I worry that `ZIP` might fall into the category of words that are _not_ [highly unlikely to be initialisms](https://github.com/99designs/gqlgen/pull/1877/files#diff-54e0b04129f7e9f7a0d75ae045a49f4c09d5cba9d09fb01fc499719216bb5b86R447). It may also refer to the term in compression and textiles contexts.

Issue #1765, opened 2022-01-15
generating core failed: unable to load github.com/kelvin/gqlgen-todos/graph/model - make sure you're using an import path to a package that exists


kelvinink:
### What happened?
I encountered an issue when following the init instruction to init my example gqlgen project. When I ran `go run github.com/99designs/gqlgen init` it output an error message:

```
Creating gqlgen.yml
Creating graph/schema.graphqls
Creating server.go
Generating...
generating core failed: unable to load github.com/kelvin/gqlgen-todos/graph/model - make sure you're using an import path to a package that exists

Exec "go run ./server.go" to start GraphQL server
```

### What did you expect?
I expect it to generate the models, resolvers. But it didn't.

### versions
 - `gqlgen version`? 
 -  v0.15.0
 - `go version`?
 - go1.16.3
 - dep or go modules?

yash492:
I'm, having the same issue

ManWithA1000Names:
If you want to use v0.15.0, you can **for now** comment out the autobind option in the gqlgen.yml and it should work fine:
```yml
# ...
resolver:
# ...
  autobind:
    - "your module name/graph/model"
```
||
\\/
```yml
# ...
resolver:
# ...
#  autobind:
#    - "your module name/graph/model"
```

ipfans:
Yes, You have some workarounds for this(choose the one you like):

1. comment the `autobind` section so you will not auto load the empty dir.
2. Touch a new empty go file in `graph/model` dir:

```
echo "package model" > graph/model/_tmp_gqlgen_init.go
```

3. Use `gqlgen` cli. I found use cli it works :)

```
go install github.com/99designs/gqlgen@latest
gqlgen init
```

StevenACoffman:
Sorry folks! I would appreciate more code coverage around this area.

domgolonka:
Seems the issue persists, but when I am commenting out autobind, the generate fails to find my custom models:

`modelgen: required package was not loaded: gitlab.com/domgolonka/api/scalars.UUIDmake: *** [generate] Error 1`

ipfans:
@domgolonka Hi, have you tried the master branch yet?

zachsmith1:
@ipfans and @StevenACoffman this is still an issue on 1.16 and on master. I was simply trying out the demo [here](https://www.howtographql.com/graphql-go/1-getting-started/) and it wouldn't work. 

Go version: 1.17

[...]

PR #1949, opened 2022-02-06
remove autobind example from init commands

zachsmith1:
When developers are running `init` for the first time to generate base configs and files, an example is included in the autobind section that points to `- "{{.}}/graph/model`. The problem with this is that developers likely won't already have external types in this directory so any subsequent `generate` commands will fail with the error `make sure you're using an import path to a package that exists`. I've commented this field out from the configTemplate so that developers following simple tutorials online aren't immediately blocked when running `init` followed by `generate`. Leaving this commented out would still show how a developer could reference external/already-defined types elsewhere. 

Should fix errors users are seeing for this: https://github.com/99designs/gqlgen/issues/1765

I have:
 (Not really needed)
 (Not really needed)

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -55,7 +55,7 @@ resolver:
 # gqlgen will search for any type names in the schema in these go packages
 # if they match it will use them, otherwise it will generate them.
 autobind:
-  - "{{.}}/graph/model"
+#  - "{{.}}/graph/model"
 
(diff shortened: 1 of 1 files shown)


frederikhors:
@StevenACoffman I think this is URGENT!

@zachsmith1 THANKS!

Issue #1483, opened 2021-03-10
go generate gives missing go sum entry for module errors


peter9207:
### What happened?

running `go generate ./...` 
gives go package errors
```
../../../../../go/pkg/mod/github.com/99designs/gqlgen@v0.13.0/cmd/gen.go:9:2: missing go.sum entry for module providing package github.com/urfave/cli/v2
../../../../../go/pkg/mod/github.com/99designs/gqlgen@v0.13.0/internal/imports/prune.go:15:2: missing go.sum entry for module providing package golang.org/x/tools/go/ast/astutil
../../../../../go/pkg/mod/github.com/99designs/gqlgen@v0.13.0/internal/code/packages.go:8:2: missing go.sum entry for module providing package golang.org/x/tools/go/packages
../../../../../go/pkg/mod/github.com/99designs/gqlgen@v0.13.0/internal/imports/prune.go:16:2: missing go.sum entry for module providing package golang.org/x/tools/imports
graph/resolver.go:3: running "go": exit status 1
```

if i run `go get` on each of these dependencies the error goes away and i can generate files just fine, but they come back after 1-2 days. so its quite annoying to keep having to run these.

I've spoken to a couple of people who used gqlgen before and none of them had any problems so I'm sure its not a super big deal. But just looking to get some help to get rid of these errors.

### What did you expect?
go generate to generate the files without these errors
### Minimal graphql.schema and models to reproduce
this happens with even the default generated schema
### versions
 - `gqlgen version`? v0.13.0
 - `go version`?  go1.16beta1 darwin/arm64
 - dep or go modules? go modules

jaanreinok:
In my case I get next error in gitlab runner:
`../.cache/pkg/mod/github.com/99designs/gqlgen@v0.13.0/cmd/gen.go:9:2: missing go.sum entry for module providing package github.com/urfave/cli/v2 (imported by github.com/99designs/gqlgen/cmd); to add:
	go get github.com/99designs/gqlgen/cmd@v0.13.0`
What I have done without any results:
cleaned gitlab runner cache. Upped version gqlgen version, ran go mod tidy.  
go 15

peter9207:
upon further debugging it seems that when i manually run `go get` it adds them to my go.mod file. which is why the error disappears. but when i run `go mod tidy` it cleans up these dependencies cuz i dont use them anywhere. and i get the error again.

jaanreinok:
[...]

PR #1958, opened 2022-02-09
Cleanup main

mtibben:
It's simpler to navigate code for the main executable when it's in one file


skaji:
FYI: As a workaround for https://github.com/99designs/gqlgen/issues/1483, I have

```
 import _ "github.com/99designs/gqlgen/cmd"
```

in my go project. This PR broke this use case.

PR #1964, opened 2022-02-14
Allow other formats when unmarshalling time

ItalyPaleAle:
This PR makes the method that unmarshalls times a bit more flexible.

Currently, only dates that are in full RFC339 format are supported, strictly. This adds a couple more formats relaxing the parser:

- `2006-01-02 15:04:05.999999999` -> allows things like `2022-02-10 10:20:30` (nanoseconds are optional)
- `2006-01-02` -> allowing things like `2022-02-10` (which defaults to midnight)

When a time zone is not specified, it defaults to UTC (as the default for `time.Parse`)

I have:

### DIFF
--- a/graphql/time.go
+++ b/graphql/time.go
@@ -18,8 +18,18 @@ func MarshalTime(t time.Time) Marshaler {
 }
 
 func UnmarshalTime(v interface{}) (time.Time, error) {
-	if tmpStr, ok := v.(string); ok {
-		return time.Parse(time.RFC3339Nano, tmpStr)
+	formats := []string{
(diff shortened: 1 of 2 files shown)


giautm:
I think you should mapping Time to your custom scalar instead.

giautm:
This PR changes the format to accept date-only layout (`2006-01-02`) can break other applications that don't expect it.

PS: Time layout format was a very-old issue in the Golang repository. :(

xsadegh (approved):

Boxtell (approved):

hsblhsn (approved):

PR #1973, opened 2022-02-17
Websocket i/o timeout fix

RobinCPel:
## Description

When configuring a ping-pong interval on a web socket, it requires each web socket connection to send pings, and receive pongs. To make sure that gqlgen receives the pongs in time, a read deadline is set on the websocket connection. This functions fine as long as the `graphql-transport-ws` sub protocol is used. However, when using the `graphql-ws` sub protocol, it causes i/o timeout errors. 

The reason that there are timeout errors when `graphql-ws` is used with a configured ping-pong interval is quite simple; [`graphql-ws` does not support ping-pong messages](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md).

So, to fix this, we just disable the whole ping-pong mechanism when using `graphql-ws`!

## What does this PR exactly do?

* Fixes a typo (`pingMesageType` => `pingMessageType`)
* Slightly refactors the code in the `websocket_graphqlws.go` file to look more like the code in the `websocket_graphql_transport_ws.go` file (for the sake of consistency)
* Added two extra `if` statements in the websocket code that make sure that the `keepAlive` messages, as well as the ping messages, only get send when their corresponding sub protocol is selected (keep-alive => `graphql-transport-ws`, ping => `graphql-ws`)

## Some extra details

The full server error is `read tcp x.x.x.x:x->x.x.x.x:x: i/o timeout`, and the full client error is `invalid close code`.

---

I have:

Not needed:

### DIFF
--- a/graphql/handler/transport/websocket.go
+++ b/graphql/handler/transport/websocket.go
@@ -150,8 +150,9 @@ func (c *wsConnection) run() {
 		c.close(websocket.CloseAbnormalClosure, "unexpected closure")
 	}()
 
-	// Create a timer that will fire every interval to keep the connection alive.
-	if c.KeepAlivePingInterval != 0 {
+	// If we're running in graphql-ws mode, create a timer that will trigger a
+	// keep alive message every interval
+	if (c.conn.Subprotocol() == "" || c.conn.Subprotocol() == graphqlwsSubprotocol) && c.KeepAlivePingInterval != 0 {
 		c.mu.Lock()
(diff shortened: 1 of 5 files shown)


soenkehahn:
We also just (independently of @RobinCPel) ran into this exact bug. Thanks a lot for putting up the PR, @RobinCPel! We would love to see this merged!

StevenACoffman:
Thanks!

PR #1975, opened 2022-02-18
Websocket Error Handling

RobinCPel:
## Description

I would like to be able to change the behaviour of my program depending on whether a web socket connection was closed with, or without an error. This also enables users of this library to detect errors that were previously "under the radar".

## Side note

The i/o timeout bug that #1973 fixes, was found because I was experimenting with this error handler.

---

I have:

Not needed:


StevenACoffman:
Thanks for this!

PR #2008, opened 2022-03-01
fix introspection for description to be nullable

Code-Hex:
description field must be `description may return a String or null.`

spec: https://spec.graphql.org/October2021/#sec-Schema-Introspection

I have:

### DIFF
--- a/integration/schema-expected.graphql
+++ b/integration/schema-expected.graphql
@@ -34,7 +34,6 @@ input ListCoercion {
   strVal: [String]
 }
 
-""""""
 scalar Map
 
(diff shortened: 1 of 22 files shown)


integration/schema-expected.graphql
```diff
@@ -34,7 +34,6 @@ input ListCoercion {
   strVal: [String]
 }
 
-""""""
```
Code-Hex, line 37: We can fix this using this PR

Code-Hex:
@StevenACoffman Could you check this?

Issue #1849, opened 2022-01-25
reloading module info generating core failed:


kokutas:
### What happened?
gqlgen --config ./gqlgen.yml 
```
reloading module info
generating core failed: unable to load github.com/kokutas/graphic/graph/model - make sure you're using an import path to a package that exists
```
gqlgen.yml
```
# Where are all the schema files located? globs are supported eg  src/**/*.graphqls
schema:
  - graph/*.graphqls

# Where should the generated server code go?
exec:
  filename: graph/generated/generated.go
  package: generated

# Uncomment to enable federation
# federation:
#   filename: graph/generated/federation.go
#   package: generated

# Where should any generated models go?
model:
  filename: graph/model/models_gen.go
  package: model

# Where should the resolver implementations go?
resolver:
  layout: follow-schema
  dir: graph
  package: graph

# Optional: turn on use `gqlgen:"fieldName"` tags in your models
# struct_tag: json

# Optional: turn on to use []Thing instead of []*Thing
# omit_slice_element_pointers: false

# Optional: set to speed up generation time by not performing a final validation pass.
# skip_validation: true

# gqlgen will search for any type names in the schema in these go packages
# if they match it will use them, otherwise it will generate them.
autobind:
  - "github.com/kokutas/graphic/graph/model"

# This section declares type mapping between the GraphQL and go type systems
#
# The first line in each type will be used as defaults for resolver arguments and
# modelgen, the others will be allowed when binding to fields. Configure them to
# your liking
models:
  ID:
    model:
      - github.com/99designs/gqlgen/graphql.ID
      - github.com/99designs/gqlgen/graphql.Int
      - github.com/99designs/gqlgen/graphql.Int64
      - github.com/99designs/gqlgen/graphql.Int32
  Int:
    model:
      - github.com/99designs/gqlgen/graphql.Int
      - github.com/99designs/gqlgen/graphql.Int64
      - github.com/99designs/gqlgen/graphql.Int32

```
schema.graphs
```
# GraphQL schema example
#
# https://gqlgen.com/getting-started/

type Todo {
  id: ID!
  text: String!
  done: Boolean!
  user: User!
}

type User {
  id: ID!
  name: String!
}

type Query {
  todos: [Todo!]!
}

input NewTodo {
  text: String!
  userId: String!
}

type Mutation {
  createTodo(input: NewTodo!): Todo!
}

```
### versions
 - `gqlgen version`? 
 -  v0.14.0-dev
 - `go version`?
[...]

PR #2014, opened 2022-03-01
Revert "Update quickstart"

mtibben:
Reverts 99designs/gqlgen#1850 as the original issue #1849 can no longer be reproduced on master (see https://github.com/99designs/gqlgen/issues/1849#issuecomment-1055236969 and https://github.com/99designs/gqlgen/pull/2010#discussion_r816600473)

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,6 @@ Still not convinced enough to use **gqlgen**? Compare **gqlgen** with other Go g
 3. Initialise gqlgen config and generate models
 
        go run github.com/99designs/gqlgen init
-       printf 'package model' | gofmt > graph/model/doc.go
 
 4. Start the graphql server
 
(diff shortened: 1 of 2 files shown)


mtibben:
May be worth waiting until just before a release before merging, as v16.0.0 still has this issue and the README on master is what many will use

StevenACoffman:
When following the directions, users got an error because gqlgen cannot autobind to an empty models directory.
The workaround was to either comment out `autobind` in the config or put an empty model file in there. We did both, which is like wearing both a belt and braces (suspenders / サスペンダ ?). Reverting the prtinf is fine as long as the autobind config is still commented out.

Code-Hex:
off topic

Yes, right "サスペンダー" == suspenders in Japanese.

> suspenders / サスペンダ ?

Today I learned phrase like "wear both a belt and braces".  means "being careful". we can translate「石橋を叩いて渡る」in Japanese.

Thanks

PR #2015, opened 2022-03-01
Clean up docs to clarify how to use a particular version

mtibben:
Clean up docs to clarify how to use a particular version


mtibben:
But I understand this line might be misleading for someone who is blindly copying pasting commands, so perhaps there's a better way to phrase this

StevenACoffman:
Great! Thanks!

Issue #1144, opened 2020-04-11
Query variables are not collected in GetRequestContext for complex type.


dtrckd:
### What happened?

When trying to obtain the variables from a request context Object (with `graphql.GetRequestContext(ctx).Variables` with the following query: 
```
query MyQuery {
  queryNode(filter:{isRoot:true}) {
  name
  }
}
```
The variable fields is an empty map

### What did you expect?

a ` map[filter: map[isRoot:true]]`

Furthermore, I get the desired result if I use 
```
query MyQuery($args: MyArgs!) {
  queryNode(filter:$args) {
  name
  }
}
```

{ "args": {"isRoot": true}}
```


### versions
 - glqgen : v0.11.3
 - go: 1.14.1
- gqlparser: v2/v2.0.1

**Comment by dtrckd** [comment]
Hop,
I apology, I think I misinterpreted the `ctx.Variables` object semantic, and thus, in the example above, there is nothing abnormal when this field is empty because no variables (graphql data) was provided.

Thus, allows me to rephrase my issue as follow: 

Is it possible to access the arguments of a query within the RequestContext Object ?

**Comment by isuhendro** [comment]
> Is it possible to access the arguments of a query within the RequestContext Object ?

@dtrckd I did not do it from `RequestContext`, instead of that from the 2nd input argument (in below example is `input`)

```
func (r *mutationResolver) CreateContact(ctx context.Context, input model.NewContact) (*model.Contact, error) {
```
```
input NewContact {
  firstName: String!
  lastName: String!
  accountId: String!
}

type Mutation {
  createContact(input: NewContact!): Contact!
}
```

dtrckd:
Tnaks @isuhendro.  Yes I can access the input argument from the resolvers but the point is that I need to reconstruct th  query as a bytes/string to re-send it to an other Graphql layer. I found it inappropriate to use those inputs because there are typed variable which I need to know before Marshalling them. The difficulties arise because I use a general interface share by several resolver to reconstruct the query, and checking/asserting the type everytimes although I do the same operation for all types which is: "reconstructing the  query as a string".

I am currently using `graphql.GetRequestContext(ctx).RawQuery` as a workaround bur I am not satisfied as I may need to modify the input query in the Business Logic Layer.

ravisastryk:
[...]

PR #2062, opened 2022-03-22
graphql: add FieldContext.Child field function and enable it in codegen

a8m:
Pasting here the discussion from gqlgen Discord:

In Ent, we have a nice workaround for the N+1 problem when working with gqlgen (see https://entgo.io/docs/tutorial-todo-gql-field-collection#ent-solution), and we do it by reading the query fields recursively and constructing one Ent query for the GraphQL query.

However, we have a problem when a field accepts arguments as parameters - we can't get them today with the current API. We use "fields collection" to get the "raw arguments", but that's not enough as we need the "processed version" that exists in the `FieldContext` in order to evaluate the Ent query. 

This PR adds a new field function to the `graphql.FieldContext` named `Child` that gets a `graphql.CollectedField` object and returns the processed  `graphql.FieldContext` for the collected field. This can also solve https://github.com/99designs/gqlgen/issues/1144.

### DIFF
--- a/_examples/chat/generated.go
+++ b/_examples/chat/generated.go
@@ -6,6 +6,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"strconv"
 	"sync"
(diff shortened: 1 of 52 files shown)


StevenACoffman:
This is great! Thanks so much!

frederikhors:
Do we need to update docs for these changes?

StevenACoffman:
I rebased it for you to fix the conflicts

a8m:
Thanks, @StevenACoffman.

@frederikhors, I'll extend the godoc for it before landing it.

frederikhors:
> I'll extend the godoc for it before landing it

Yes, please. Code without docs == no code at all.

frederikhors:
> I'll extend the godoc for it before landing it

Yes, please. Features without docs == no features at all.

StevenACoffman (approved):
I'll let you land it yourself.

a8m:
Added godoc and merged. Thanks for the review @StevenACoffman and @frederikhors.

PR #2063, opened 2022-03-22
AWS AppSync "start_ack" response on subscribe success

nathanhoult:
Added a "start_ack" response when a subscription is successful

### DIFF
--- a/graphql/handler/transport/websocket_graphqlws.go
+++ b/graphql/handler/transport/websocket_graphqlws.go
@@ -14,6 +14,7 @@ const (
 	graphqlwsConnectionInitMsg      = graphqlwsMessageType("connection_init")
 	graphqlwsConnectionTerminateMsg = graphqlwsMessageType("connection_terminate")
 	graphqlwsStartMsg               = graphqlwsMessageType("start")
+	graphqlwsStartAckMsg            = graphqlwsMessageType("start_ack")
 	graphqlwsStopMsg                = graphqlwsMessageType("stop")
 	graphqlwsConnectionAckMsg       = graphqlwsMessageType("connection_ack")
 	graphqlwsConnectionErrorMsg     = graphqlwsMessageType("connection_error")
@@ -152,7 +153,7 @@ func (m *graphqlwsMessage) fromMessage(msg *message) (err error) {
 	case connectionCloseMessageType:
 		m.Type = graphqlwsConnectionTerminateMsg
 	case startMessageType:
-		m.Type = graphqlwsStartMsg
+		m.Type = graphqlwsStartAckMsg
 	case stopMessageType:
 		m.Type = graphqlwsStopMsg
 	case dataMessageType:
--- a/graphql/handler/transport/websocket.go
+++ b/graphql/handler/transport/websocket.go
@@ -15,6 +15,7 @@ import (
 	"github.com/99designs/gqlgen/graphql"
 	"github.com/99designs/gqlgen/graphql/errcode"
 	"github.com/gorilla/websocket"
+	"github.com/vektah/gqlparser/v2/ast"
 	"github.com/vektah/gqlparser/v2/gqlerror"
 )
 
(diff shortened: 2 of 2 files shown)


graphql/handler/transport/websocket_graphqlws.go
```diff
@@ -14,6 +14,7 @@ const (
 	graphqlwsConnectionInitMsg      = graphqlwsMessageType("connection_init")
 	graphqlwsConnectionTerminateMsg = graphqlwsMessageType("connection_terminate")
 	graphqlwsStartMsg               = graphqlwsMessageType("start")
+	graphqlwsStartAckMsg            = graphqlwsMessageType("start_ack")
```
a8m, line 17: I don't see "start_ack" in the protocol mentioned in the comment. Maybe @jordanabderrachid (author of this file) can review it?

RobinCPel:
What would be the use case for this behaviour? By quickly googling I assume it has something to do with AWS AppSync?

I think that if we want to add support for AWS AppSync behaviour, we should maybe create a separate sub-protocol for it (or something alike). Because the `start_ack` message is not part of the modified protocol's spec... (which can be found [here](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md).)

nathanhoult:
Thanks for the feedback! @RobinCPel is correct and it was to mimic AWS AppSync and I thought you may want a sub-protocol but was avoiding doing that. I can work on another sub-protocol and open a PR, though I am not able to find a nice spec for AppSync so it will be some reverse enginering.

Since we already have mobile apps using AppSync I would like to add an optional configuration option to default the sub-protocol since setting a header will require an update (apparently end users are not too happy with that).

Should I just close this PR and open another with a whole sub-protocol?

graphql/handler/transport/websocket_graphqlws.go
nathanhoult: Since everything else was working I was trying to just add one more message that AWS AppSync does. I will make a sub-protocol instead since as you said it isn't part of the spec for this protocol.

RobinCPel:
I believe [this page](https://docs.aws.amazon.com/appsync/latest/devguide/real-time-websocket-client.html) contains the whole spec. It seems like AWS only adds the `start_ack` message, nothing else (for now). 

I think a separate sub-protocol would be the cleanest option. That also means that the web socket clients have to specify it in their `Sec-WebSocket-Protocol` header once it is added. 

Regarding the configurable default sub-protocol, I'm not against it, but in my opinion it is a way better choice to update the client's headers.

Side note; [AWS does say that they use `graphql-ws`](https://docs.aws.amazon.com/appsync/latest/devguide/real-time-websocket-client.html#handshake-details-to-establish-the-websocket-connection) while they go out of spec... 🙄
Screenshot 2022-03-25 at 16 53 39

nathanhoult:
Thank you for the AWS pages (not sure why I couldn't find them, maybe I was looking for a git repo).

So this is embarrassing but just I talked to the app developers and they use the [Apollo client](https://www.apollographql.com/docs/) to talk to AppSync and were considering going to the [AppSync SDK](https://github.com/awslabs/aws-mobile-appsync-sdk-android) so I asked them of they could not (or maybe just don't use [notifySubscriptionStarted()](https://github.com/awslabs/aws-mobile-appsync-sdk-android/blob/main/aws-android-sdk-appsync/src/main/java/com/amazonaws/mobileconnectors/appsync/WebSocketConnectionManager.java) callback). I believe I can work around my personal `legacy API <-> GraphQL` service for using the `start_ack` response by just assuming the subscription worked unless I get an error (no news is good news). I thought some subscription success was a good idea and still a little surprised it is not part of the GraphQL spec.

If I find I need `start_ack` some day I know what work needs to be done but adding another sub-protocol no one needs right now seems like a bad idea. Hopefully this PR discussion will help someone looking to get off AppSync and trying to figure out why the AppSync SDK callback isn't working. If for some reason you really like the idea of a AppSync sub-protocol I can still do the work but otherwise I think we can just close this PR.

Thanks for your time and all the good work everyone has done on gqlgen, I really appreciate it!

nathanhoult:
Not needed at this time

PR #2066, opened 2022-03-23
codegen: allow binding methods with optional variadic arguments

a8m:
In some cases, as we have in Ent, the same method used for GraphQL binding can accept optional parameters.


StevenACoffman:
Nice! Thanks!

Issue #2068, opened 2022-03-24
proposal: Add fieldalignment to generated structs


switchupcb:
# What is fieldalignment?

## fieldalignment in gqlgen

The following code block contains output from `fieldalignment -fix` on gqlgen structs.
```go
models.gqlgen.go:17:14: struct with 184 pointer bytes could be 176
models.gqlgen.go:31:31: struct with 48 pointer bytes could be 40
models.gqlgen.go:84:23: struct with 64 pointer bytes could be 56
models.gqlgen.go:117:25: struct with 72 pointer bytes could be 56
models.gqlgen.go:148:25: struct with 88 pointer bytes could be 80
models.gqlgen.go:172:22: struct with 40 pointer bytes could be 32
models.gqlgen.go:294:18: struct with 40 pointer bytes could be 32
models.gqlgen.go:327:12: struct with 80 pointer bytes could be 72
models.gqlgen.go:342:16: struct with 56 pointer bytes could be 48
models.gqlgen.go:362:14: struct with 80 pointer bytes could be 64
models.gqlgen.go:377:12: struct with 40 pointer bytes could be 32
models.gqlgen.go:385:11: struct with 48 pointer bytes could be 40
models.gqlgen.go:393:14: struct with 88 pointer bytes could be 80
models.gqlgen.go:409:18: struct with 64 pointer bytes could be 56
models.gqlgen.go:468:23: struct with 80 pointer bytes could be 72
models.gqlgen.go:504:25: struct with 88 pointer bytes could be 72
models.gqlgen.go:518:23: struct with 40 pointer bytes could be 24
models.gqlgen.go:528:22: struct with 48 pointer bytes could be 40
models.gqlgen.go:538:25: struct with 112 pointer bytes could be 104
models.gqlgen.go:575:11: struct with 96 pointer bytes could be 88
```

176 bytes saved. 20 misaligned models. You don't use these models in every request. So for each model used in 1 MILLION requests, _that's..._

**8.8 MB saved per misaligned model**.

Along with better [Garbage Collection](https://agrim123.github.io/posts/go-garbage-collector.html) performance due to a lower amount of peaks.

## Previous Mentions

https://github.com/99designs/gqlgen/issues/1710 (https://github.com/99designs/gqlgen/discussions/1895)

## Steps

- [x] Add code to field align the written file. 
- [ ] Add conditional statement to fieldalign when enabled (unless it's encouraged to always fieldalign).
[...]

PR #2069, opened 2022-03-24
add fieldalignment to Codegen

switchupcb:
https://github.com/99designs/gqlgen/issues/2068

### DIFF
--- a/codegen/templates/templates.go
+++ b/codegen/templates/templates.go
@@ -6,6 +6,7 @@ import (
 	"go/types"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"reflect"
 	"runtime"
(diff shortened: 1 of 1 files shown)


switchupcb:
Feel free to add the conditional check for the `fieldalignment` option, along with the option's placement in the `.yml` (or other form of configuration).

codegen/templates/templates.go
```diff
@@ -167,7 +168,30 @@ func Render(cfg Options) error {
...
+	// change the current working directory to ensure the file is in the current module.
+	cwd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+	if err := os.Chdir(fileDir); err != nil {
+		return err
+	}
+
+	cmd := exec.Command("fieldalignment", "-fix", cfg.Filename)
```
a8m, line 184: For cases like `fieldalignment` (or for example, `addlicense`), I add another line to `go generate` file and run it separately. I recommend not adding it to gqlgen, but if others think differently, it should be executed as a package (see [example](https://cs.opensource.google/go/x/tools/+/refs/tags/v0.1.10:go/analysis/passes/fieldalignment/cmd/fieldalignment/main.go)), because some users may not have it installed.

switchupcb: Great! Didn't see the exported `Run: run` for some reason. I will change this if the idea is approved. Is there any reason you recommend not adding it as an option? I plan to add the boolean option here below `autobind`: https://gqlgen.com/config/

a8m: Hey, @switchupcb, as a package maintainer of [ent](https://github.com/ent/ent), I can share my 2 cents why I didn't add it to the project, even not as an option. 

1\. For some packages, like gqlgen and ent, I never used `fieldalignment` because I prefer the order of the fields in the generated code to follow the same order they are defined in the graphql.schema or ent/schema. In most cases, I prefer code readability and maintainability over fine-tuning optimization. This is why it's not enabled by default.

2\. The reason it's not an option in ent, is because for users, adding the following line to the `go generate` process is simpler than all other options. Also, this gives them the option to choose any version they want (of `fieldalignment` for example) and may run it on multiple packages at once. 

```diff
//go:generate go run entgo.io/ent/cmd/ent generate ./ent/schema
//go:generate go run -mod=mod github.com/99designs/gqlgen
+//go:generate go run golang.org/x/tools/go/analysis/passes/fieldalignment/cmd/fieldalignment -fix <packages...>
```

switchupcb: That seems more practical. You _can_ run it programmatically _(not using cmd)_, but its not straightforward.

PR #2079, opened 2022-03-31
Add instructions for enabling autobinding

jpwilliams:
The _Getting Started_ guide doesn't mention that you need to enable autobinding to follow the final steps.

The `gqlgen.yml` file generated using the `init` command [has autobinding disabled by default](https://github.com/99designs/gqlgen/blob/master/init-templates/gqlgen.yml.gotmpl#L35-L38) instead of enabled as the docs previously stated.

This tiny PR adjusts the guide to address this point.

I have:
 - [ ] ~Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))~

### DIFF
--- a/docs/content/getting-started.md
+++ b/docs/content/getting-started.md
@@ -193,7 +193,17 @@ This example is great, but in the real world fetching most objects is expensive.
 todo unless the user actually asked for it. So lets replace the generated `Todo` model with something slightly more
 realistic.
 
-Create a new file called `graph/model/todo.go`
+First let's enable `autobind`, allowing gqlgen to use your custom models if it can find them rather than generating them. We do this by uncommenting the `autobind` config line in `gqlgen.yml`:
+
+```yml
+# gqlgen will search for any type names in the schema in these go packages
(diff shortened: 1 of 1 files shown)


StevenACoffman:
So... the difficulty here is that if you have previously generated no models, then there will be an error when you run gqlgen with autobind enabled since it won't be able to bind to those nonexistant models. That's why we disabled autobind for when people run init. Can you carefully verify that these updated directions work without any issue when you start completely fresh?

jpwilliams:
@StevenACoffman Mm I suppose that could cause some problems if someone were to skip around the _Getting Started_ guide rather than following every step. We could add a short note about _when_ to enable autobinding (i.e. enable it if you've done the above step of creating a `graph/model/todo.go` file, otherwise don't)?

Have just tested again with a fresh repo - works perfectly with the new instruction.

StevenACoffman:
Ok, I'm going to merge this as is, but if we start to get a lot of problems, then we can add a note.

PR #2096, opened 2022-04-15
Change the error message to be consumer targeted

telemenar:
The existing "must not be null" message if read by the caller of a gql api could be confusing because nothing they provided was null.

Changed the message from "must not be null" to "the requested element is null which the schema does not allow"

I have:

### DIFF
--- a/_examples/chat/generated.go
+++ b/_examples/chat/generated.go
@@ -3500,7 +3500,7 @@ func (ec *executionContext) marshalNBoolean2bool(ctx context.Context, sel ast.Se
 	res := graphql.MarshalBoolean(v)
 	if res == graphql.Null {
 		if !graphql.HasFieldError(ctx, graphql.GetFieldContext(ctx)) {
-			ec.Errorf(ctx, "must not be null")
+			ec.Errorf(ctx, "the requested element is null which the schema does not allow")
 		}
 	}
 	return res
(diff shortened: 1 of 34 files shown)


StevenACoffman:
@RobinCPel Would you mind reviewing this websocket PR?

telemenar:
Oh, I need to re-run generate after cherry-picking this from a v0.17.2. And it looks like array ordering for the error messages isn’t stable as the failing test passes on my infra. I’ll fix both of these on Monday.

telemenar:
Sorry for the noise with this I messed up my rebase and ended up with reparented commits I didn't intend to reparent. 

To simplify review a little bit since this makes a bunch of noise in generated file the main file with explicit changes is:

codegen/type.gotpl


And then these test files:

codegen/testserver/followschema/nulls_test.go
codegen/testserver/followschema/time_test.go
codegen/testserver/singlefile/nulls_test.go
codegen/testserver/singlefile/time_test.go

StevenACoffman:
Huh, so is the array order for the error message stable now and this is ok to merge as is? Or is there something else to be done to make the tests not flaky?

StevenACoffman:
Thanks so much! I appreciate the wording improvement.

frederikhors:
@telemenar I really appreciated this change.

Now that I updated my project's gqlgen version I noticed the diff in new generated files and found out 84 times the string: `"the requested element is null which the schema does not allow"`.

Can we use a `const` for it?

Since I'm not familiar with the gqlgen code could you create a PR for this? What do you think?

Maybe something smarter would be: "if there are at least two occurrences of this string create a const".

PR #2101, opened 2022-04-17
Use Github API to update the docs

frankywahl:
Instead of a hard-coded version of the docs we want to realease, this
uses the Github API to get the last 20 versions and publish those. This
will allow any script invoking this to make sure to always have the
latest version of the docs

### DIFF
--- a/docs/build.sh
+++ b/docs/build.sh
@@ -3,25 +3,13 @@
 # This was adapted from https://github.com/dgraph-io/dgraph/blob/master/wiki/scripts/build.sh
 #
 
-set -e
-
 GREEN='\033[32;1m'
 RESET='\033[0m'
 HOST=https://gqlgen.com
 
-VERSIONS_ARRAY=(
-    'v0.16.0'
-    'v0.15.1'
-    'origin/master'
-    'v0.15.0'
-    'v0.14.0'
-    'v0.13.0'
-    'v0.12.2'
-    'v0.11.3'
-    'v0.10.2'
-    'v0.9.3'
-    'v0.8.3'
-)
+IFS=$'\n' read -r -d '' -a VERSIONS_ARRAY < <(curl -s -H "Accept: application/vnd.github.v3+json" "https://api.github.com/repos/99designs/gqlgen/releases?per_page=20" | jq -r '.[].tag_name')
+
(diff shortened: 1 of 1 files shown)


docs/build.sh
```diff
@@ -3,25 +3,13 @@
...
-    'v0.15.0'
-    'v0.14.0'
-    'v0.13.0'
-    'v0.12.2'
-    'v0.11.3'
-    'v0.10.2'
-    'v0.9.3'
-    'v0.8.3'
-)
+IFS=$'\n' read -r -d '' -a VERSIONS_ARRAY < <(curl -s -H "Accept: application/vnd.github.v3+json" "https://api.github.com/repos/99designs/gqlgen/releases?per_page=20" | jq -r '.[].tag_name')
```
frankywahl, line 10: Couple of things to note: 
1. This limits to the last 20 versions. I think it's probably enough. We could have more, but those versions risk being very old. 
2. This assumes `jq` & `curl` has been installed on the machine running this script. I suspect that is the case though I am not sure how [render.com](render.com) works
3. This will also change the position of `master` to being at the bottom of the list all the time.

docs/build.sh
```diff
@@ -3,25 +3,13 @@
 # This was adapted from https://github.com/dgraph-io/dgraph/blob/master/wiki/scripts/build.sh
 #
 
-set -e
```
frankywahl, line 6: For some reason, this would not work if I left it in. Don't think it's a huge issue

StevenACoffman: > `-e `    When this option is on, if a simple command fails for any of the reasons listed in Consequences of
       Shell  Errors or returns an exit status value >0, and is not part of the compound list following a
       while, until, or if keyword, and is not a part of an AND  or  OR  list,  and  is  not  a  pipeline
       preceded by the ! reserved word, then the shell shall immediately exit.

frankywahl: Yes - I understand that - I'm just unsure what the error is because I don't get an output. I'll put it a commit that reinstates it. It may help clarify

StevenACoffman:
This will be great. I'm going to merge this as is, but if it fails on master, then I'll look into changing it to `set -ex` to and add some printf message to see what's going on.

Issue #2114, opened 2022-04-22
Federation v2 support


superbeeny:
Previous issue was closed

We've also found this as gqlgen does not handle the @shareable directive which is a key component of federation 2.

@StevenACoffman i think this will need to be addressed, [according to Apollo's federation 2 compatibility table](https://github.com/apollographql/apollo-federation-subgraph-compatibility) `gqlgen` is not compatible with the new federation spec.

_Originally posted by @jakemhiller in https://github.com/99designs/gqlgen/issues/1695#issuecomment-1098277141_

shellfu:
This has also been a blocker for us to move to federation 2. I would love to see this implemented!!

frederikhors:
Can you both PR?

lleadbet:
@superbeeny - it was just merged, but likely needs a new version cut; @StevenACoffman can speak to next release timing

StevenACoffman:
Hi. I would appreciate it if you pinned to the latest commit and then ran the [Apollo Federation tests](https://github.com/apollographql/apollo-federation-subgraph-compatibility):
```
go get -d github.com/99designs/gqlgen@2a2a3dcb67c7d713e41476eac47e20ab0e21fba7
```

lleadbet:
Ran the tests and will get the updated version in the Apollo tests, but the updated results are:

| Language | Library | Federation 1 Support | Federation 2 Support |
| --- | --- | --- | --- |
| Go | [gqlgen](https://gqlgen.com) | <table><tr><th>_service</th><td>✅</td></tr><tr><th>@key (single)</th><td>✅</td></tr><tr><th>@key (multi)</th><td>✅</td></tr><tr><th>@key (composite)</th><td>✅</td></tr><tr><th>@requires</th><td>✅</td></tr><tr><th>@provides</th><td>✅</td></tr><tr><th>@ftv1</th><td>❌</td></tr></table> | <table><tr><th>@link</th><td>✅</td></tr><tr><th>@shareable</th><td>✅</td></tr><tr><th>@tag</th><td>✅</td></tr><tr><th>@override</th><td>✅</td></tr><tr><th>@inaccessible</th><td>✅</td></tr></table> |

Fork: https://github.com/lleadbet/apollo-federation-subgraph-compatibility
PR: https://github.com/apollographql/apollo-federation-subgraph-compatibility/pull/87

StevenACoffman:
[...]

PR #2115, opened 2022-04-22
Feature: Adds Federation 2 Support

lleadbet:
This PR adds in support for Apollo's Federation 2 standard, as requested per issue #2114. 

This adds a new configuration value to set the federation version manually as well as a method to detect federation 2 schemas using a small regex string. 

Largely, this doesn't change much about the generated code, but instead allows for `graphql-parser` to not error by adding the new Federation 2 directives before loading new schemas. Previously, this was panicing due to a missing directive. 

By default, this still uses Federation 1 unless detected by aforementioned regex or set by the user to avoid breaking changes.

I have:


lleadbet:
@frederikhors - Just a small question, as I can't seem to quite understand why it's erroring out on tests/using it. 

I have added a Link interface here: plugin/federation/fedruntime/runtime.go, however referencing within the federation file such as:

```go
		builtins["link__Import"] = config.TypeMapEntry{
			Model: config.StringList{"github.com/99designs/gqlgen/fedruntime.Link"},
		}
```

Still doesn't find the interface.

frederikhors:
@lleadbet we need to wait @StevenACoffman for this answer.

docs/content/recipes/federation.md
```diff
@@ -19,6 +19,16 @@ federation:
...
 ```
 
+### Federation 2
+
+If you are using Apollo's Federation 2 standard, your schema should automatically be upgraded so long as you include the required `@link` directive within your schema. If you want to force Federation 2 composition, the `federation` configuration supports a `version` flag to override that. For example:
+
+```yml
+federation:
+  filename: graph/generated/federation.go
+  package: generated
```
kdawgwilk, line 29: Did you want to add a `version: 2` key here?

lleadbet: We can implicitly pull that from the schema, hence why I left it out; might be worthwhile, however.

plugin/federation/fedruntime/runtime.go
```diff
@@ -11,3 +11,6 @@ type Service struct {
 type Entity interface {
 	IsEntity()
 }
+
+// Used for the Link directive
+type Link interface{}
```
StevenACoffman, line 16: I'm idly wondering if the Link interface should have functions to introspectively answer:
1. What version was imported (currently only "2.0" possible)
2. What imports out of the list: `["@key","@shareable", "@provides", "@external", "@tag", "@extends", "@override", "@inaccessible"]`

StevenACoffman:
The test error root cause is here: https://github.com/99designs/gqlgen/blob/cddbf02d494e3aeaac3f60d1708b25facc5b767d/codegen/config/binder.go#L149

plugin/federation/federation.go
```diff
@@ -51,6 +56,13 @@ func (f *federation) MutateConfig(cfg *config.Config) error {
 			Model: config.StringList{"github.com/99designs/gqlgen/graphql.Map"},
 		},
 	}
+
+	if f.Version == 2 {
+		builtins["link__Import"] = config.TypeMapEntry{
+			Model: config.StringList{"github.com/99designs/gqlgen/fedruntime.Link"},
```
StevenACoffman, line 62:
```suggestion
			Model: config.StringList{"github.com/99designs/gqlgen/plugin/federation/fedruntime.Link"},
```

StevenACoffman:
Applying my suggested fix removes one test failure and moves the second to:
```
=== Failed
=== FAIL: api TestGenerate/federation2 (7.45s)
    generate_test.go:55: Generate() error = validation failed: packages.Load: /Users/steve/Documents/git/gqlgen/api/testdata/federation2/graph/generated/generated.go:4118:17: ec.unmarshalInputlink__Import undefined (type *executionContext has no field or method unmarshalInputlink__Import)
        /Users/steve/Documents/git/gqlgen/api/testdata/federation2/graph/generated/generated.go:4126:12: ec._link__Import undefined (type *executionContext has no field or method _link__Import)
        , wantErr false
    --- FAIL: TestGenerate/federation2 (7.45s)

=== FAIL: api TestGenerate (12.63s)
```

StevenACoffman:
I modified that [federation.gotpl](https://github.com/99designs/gqlgen/blob/master/plugin/federation/federation.gotpl) to have this chunk to get the tests to succeed:
```

{{ reserveImport "github.com/vektah/gqlparser/v2/ast" }}

{{ reserveImport "github.com/99designs/gqlgen/plugin/federation/fedruntime" }}
{{ reserveImport "github.com/99designs/gqlgen/graphql" }}

var (
	ErrUnknownType = errors.New("unknown type")
	ErrTypeNotFound = errors.New("type not found")
)

func (ec *executionContext) __resolve__service(ctx context.Context) (fedruntime.Service, error) {
	if ec.DisableIntrospection {
		return fedruntime.Service{}, errors.New("federated introspection disabled")
	}

	var sdl []string

	for _, src := range sources {
		if src.BuiltIn {
			continue
		}
		sdl = append(sdl, src.Input)
	}

	return fedruntime.Service{
		SDL: strings.Join(sdl, "\n"),
	}, nil
}

func (ec *executionContext) unmarshalInputlink__Import(ctx context.Context, v  interface{}) (string, error) {
	fmt.Println("Yahoo-hoo-hooey")
	return "", nil
}

func (ec *executionContext) _link__Import(ctx context.Context, sel ast.SelectionSet, v  interface{}) graphql.Marshaler {
	fmt.Println("Yahoo-hoo-hooey")
	return nil
}
```

These are silly, but something better should go there or have them not be required.

lleadbet:
> These are silly, but something better should go there or have them not be required.

Agreed- was copying Apollo's material, but since those imports are just strings, decided to convert it into `[String!]` instead of the custom scalar.

lleadbet:
> https://github.com/99designs/gqlgen/runs/6159890288?check_suite_focus=true#step:6:27

I have, however it still fails; would be curious on the fix here.

StevenACoffman:
You need to run `cd ~/_examples/; go generate ./...` i think

StevenACoffman:
Thanks!

PR #2119, opened 2022-04-23
use "embed" in generated code

vikstrous:
This is a proof of concept. It shows what the generated.go files could look like without the literal strings from graphql schema files.

Unfortunately, this PR is not mergable yet. To make this mergable, we have to check whether or not each file is relative to the current working directory. If it is, it can be embedded. If it's not, it has to be literally included. If any files can be embedded, the FS must be created. (it also needs a test)

This reduces the size of the generated files and can reduce merge conflicts when working on a large schema.

This idea can be taken further and include all `*.graphql(s)` files in the current directory and take some further steps to minimize potential conflicts in generated code.

I'm opening this PR just to get some early feedback and see if anyone is interested in seeing this finished. I'm also open to ideas about how to do the check described above. I might have to augment the data structures passed into the template or make a new function available that can check whether or not a path is relative.

I have:
 - [n/a] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))


StevenACoffman:
This is a great idea, but we definitely need more coverage around it. We support Go 1.16 and newer, and we haven't really taken advantage of embedding up until now.

vikstrous2:
Alright, here's how we get this mergable:
* [ ] Schema filenames are currently relative to the place gqlgen is run, not the generated code directory, so we need to change the filenames to make them relative to the correct directory
* [ ] The filesystem variable and function have to be conditionally included if there are any embeddable sources - so we basically need a list of embeddable and non-embeddable sources
* [ ] Some of the examples have files that are not embeddable, so that's nice, but maybe we should add an extra test for schema files that are entirely outside the go module to catch any unexpected regressions
* [ ] We should consider making the sources for "builtins" references to gqlgen library constants rather than including them in the generated code... It would be a nice bonus.

ramezanius (approved):

StevenACoffman:
Hey, I'm waiting on you to tell me when you think this is mergeable to re-review it

codegen/data.go
```diff
@@ -147,6 +170,30 @@ func BuildData(cfg *config.Config) (*Data, error) {
 		// otherwise show a generic error message
 		return nil, fmt.Errorf("invalid types were encountered while traversing the go source code, this probably means the invalid code generated isnt correct. add try adding -v to debug")
 	}
+	aSources := []AugmentedSource{}
+	for _, s := range cfg.Sources {
+		wd, err := os.Getwd()
```
vikstrous, line 175: I'm not happy with having this here, but it seemed like the most reliable way to build the full path to the sources. Is there a better way?

vikstrous:
The PR is reviewable now. Sorry for the delay. This has been a weekend project :)
I didn't do the extra work to make "builtin" sources references, but someone can add that as a follow up if it's useful for them.

frederikhors:
I'm having trouble with this. I think we'll have A LOT of issues for this.

@vikstrous can you please PR ASAP for docs instruction for this to work?

For example I'm having coming from 0.17.5:

```
panic: codegen problem: schema\schema.graphql not availalbe
```

THIS IS A HUGE PROBLEM.

frederikhors:
Maybe this is a problem with Windows, @vikstrous?

I see path like: `//go:embed "schema\\schema.graphql"`.

frederikhors:
@StevenACoffman this is a huge problem.

I agree with the need to have it - even soon - but it should be tested better and maybe used with a feature flag maybe in the config file.

In projects like the one where I'm applying it (where there are several source of *.graphql files) it has now become a mess and nothing works.

Basically it's a COMPLICATED breaking change that came with a patch.

I propose to release 0.17.7 as soon as possible with this commit turned off and then we can see how to release it properly.

If we don't release 0.17.7 we can't use the other important changes from 0.17.6.

Please.

vikstrous:
Thanks for letting me know about the bug. I created issues for you: https://github.com/99designs/gqlgen/issues/2189 https://github.com/99designs/gqlgen/issues/2190

vikstrous:
@frederikhors Let me know if this fixes the issue: https://github.com/99designs/gqlgen/pull/2191 I don't have windows and I don't have a way to test it.

frederikhors:
I renamed manually all the `\\` to `/` and it works.

So I think we should fix it this way.

mtibben:
Hey @vikstrous @StevenACoffman 

I'm a bit taken aback by the scale of this change with minimal discussion on it's merits. The PR description talks about this as a Proof of Concept, yet it has been merged.

The PR description doesn't explain what the purpose of this change is, who or what it benefits and why.

Have I missed a discussion that happened somewhere else?

mtibben:
Hey @vikstrous @StevenACoffman 

I'm a bit taken aback by the scale of this change with minimal discussion on it's merits. The PR description talks about this as a Proof of Concept, yet it has been merged.

The PR description doesn't explain what the purpose of this change is, who or what it benefits and why.

Have I missed a discussion that happened somewhere else?

vikstrous:
@mtibben You are not missing any other discussion, but note that the PR description hasn't been updated after the proof of concept was turned into a mergable PR.

Sorry for the issues this caused. TBH I didn't expect anyone to be using this on windows, but I see that with popular open source projects, you never know what people will do with your code.

I can add more context about the motivation. I work in a codebase that uses gqlgen heavily for a large graphql monolith. There are 384k lines of code in the files generated by gqlgen. We check in all these files into git be because it takes a long time to generate them (and because it's a lot simpler to work that way). The motivation behind this PR was to reduce the size of the generated files and reduce the chance of having to resolve PR conflicts.

I think that the complexity of the change is fairly low and there is no down side to introducing it. I made sure it has reasonable test coverage. Clearly I hadn't considered windows support, but that's just a bug that could have been introduced by any PR regardless of what it's changing. This didn't seem like a major change because it's backwards compatible and doesn't change any behaviour (I wasn't the one that decided to include it in a patch relaese, but I agree with the decision).

mtibben:
Cool thanks for the explanation @vikstrous. Do you think you could update the PR description with this summary and an accurate description of the changes, so that future devs can understand the change and motivations fully?

frederikhors:
I agree with @vikstrous this is not a big issue (the Windows issue is fixed in master and ready to be released with next patch release).

Perhaps a more conservative approach would have avoided some stress: releasing more often (using minor releases maybe for thing like this) is useful to reduce issues like the one I had on Windows.

Anyway @vikstrous did you get to see https://github.com/99designs/gqlgen/issues/2198?

StevenACoffman:
@mtibben This PR worked flawlessly in our production codebase (on linux and locally on MacOS), and was a backwards compatible change (with a bug on windows that went unnoticed because we weren't running tests there).

Backwards compatible changes seem ok to put in patch releases to me, but I'm open to discussing it. I don't think I've ever articulated my approach, so it's good to talk about it.

While semver implies `Major.Minor.Patch` => `Breaking.Feature.Fix` because of how go modules work, we aren't releasing a gqlgen V2, V3, V4, nor are we staffed to maintain more than whatever the latest release is. The V0 implies gqlgen may occasionally (and with great hesitation) break compatibility in minor ways between releases.

mtibben:
@StevenACoffman yeah it would be great to get visibility of production systems testing these changes. I'd love for our systems here at 99 to also be on bleeding edge to build confidence (this is what google does with go), this would be something our @99designs/delivery-systems team may be able to think about the practicalities of.

As mentioned above I also think our PR descriptions could just be levelled up to provide accuracy. The PR description here still says this is proof of concept, when it seems to have moved beyond that - I'll try to fix that up now

mtibben:
@vikstrous I have updated the PR description now, can you confirm that it is accurate?

vikstrous:
Thanks! I removed the last part. Looks good now

vikstrous:
Yeah, sorry for the delay. This is more of a weekend project :) I'll clean
it up this weekend

vikstrous:
Please don't merge yet! This is not ready. It's still very rough

PR #2120, opened 2022-04-24
shorten some generated code

vikstrous:
I was reading some templates and saw that this code can be shorter, so I shortened it. Less code is better than more, right?

I have:
 - [n/a] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [n/a] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))


StevenACoffman:
Thanks for working on improving things! It looks like `.github/workflows/check-generate` is failing, but that might just mean a test needs to be updated or something. I'm away from a computer right now, so it's hard to tell from my phone.

PR #2131, opened 2022-04-27
feat: added graphql.UnmarshalInputFromContext

giautm:
This allows unmarshaling the input object from a context.

This PR is the support for ent/contrib#297: entgql: Unmarshal whereInput using graphql package

I have:

### DIFF
--- a/codegen/generated!.gotpl
+++ b/codegen/generated!.gotpl
@@ -135,6 +135,13 @@
 	func (e *executableSchema) Exec(ctx context.Context) graphql.ResponseHandler {
 		rc := graphql.GetOperationContext(ctx)
 		ec := executionContext{rc, e}
+		inputUnmarshalMap := graphql.BuildMap(
+			{{- range $input := .Inputs -}}
+			{{ if not $input.HasUnmarshal }}
+			ec.unmarshalInput{{ $input.Name }},
+			{{- end }}
+			{{- end }}
+		)
 		first := true
 
 		switch rc.Operation.Operation {
@@ -142,6 +149,7 @@
 			return func(ctx context.Context) *graphql.Response {
 				if !first { return nil }
 				first = false
+				ctx = graphql.WithUnmarshalerMap(ctx, inputUnmarshalMap)
 				{{ if .Directives.LocationDirectives "QUERY" -}}
 					data := ec._queryMiddleware(ctx, rc.Operation, func(ctx context.Context) (interface{}, error){
 						return ec._{{.QueryRoot.Name}}(ctx, rc.Operation.SelectionSet), nil
@@ -162,6 +170,7 @@
 			return func(ctx context.Context) *graphql.Response {
 				if !first { return nil }
 				first = false
+				ctx = graphql.WithUnmarshalerMap(ctx, inputUnmarshalMap)
 				{{ if .Directives.LocationDirectives "MUTATION" -}}
 					data := ec._mutationMiddleware(ctx, rc.Operation, func(ctx context.Context) (interface{}, error){
 						return ec._{{.MutationRoot.Name}}(ctx, rc.Operation.SelectionSet), nil
--- a/codegen/generated!.gotpl
+++ b/codegen/generated!.gotpl
@@ -135,11 +135,11 @@
 	func (e *executableSchema) Exec(ctx context.Context) graphql.ResponseHandler {
 		rc := graphql.GetOperationContext(ctx)
 		ec := executionContext{rc, e}
-		inputUnmarshalMap := graphql.BuildMap(
+		inputUnmarshalMap := graphql.BuildUnmarshalerMap(
 			{{- range $input := .Inputs -}}
-			{{ if not $input.HasUnmarshal }}
-			ec.unmarshalInput{{ $input.Name }},
-			{{- end }}
+				{{ if not $input.HasUnmarshal }}
+					ec.unmarshalInput{{ $input.Name }},
+				{{- end }}
 			{{- end }}
 		)
 		first := true
--- a/graphql/input.go
+++ b/graphql/input.go
@@ -27,28 +27,28 @@ func WithUnmarshalerMap(ctx context.Context, maps map[reflect.Type]reflect.Value
 }
 
 // UnmarshalInputFromContext allows unmarshaling input object from a context.
-func UnmarshalInputFromContext(ctx context.Context, raw, inputObj interface{}) error {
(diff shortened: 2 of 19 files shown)


graphql/input.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"context"
+	"errors"
+	"reflect"
+)
+
+const unmarshalInputCtx key = "unmarshal_input_context"
+
+func BuildMap(unmarshaler ...interface{}) map[reflect.Type]reflect.Value {
```
a8m, line 11: Please, add godoc.

Also, what do you think about returning an error instead of panic? 
```suggestion
func BuildUnmarshalerMap(unmarshaler ...interface{}) map[reflect.Type]reflect.Value {
```

graphql/input.go
```diff
@@ -0,0 +1,53 @@
...
+	maps := make(map[reflect.Type]reflect.Value)
+	for _, v := range unmarshaler {
+		ft := reflect.TypeOf(v)
+		if ft.Kind() != reflect.Func {
+			panic("unmarshaler must be a function")
```
a8m, line 16:
```suggestion
			panic(fmt.Sprintf("unmarshaler must be a function, got: %s", ft.Kind()))
```

codegen/generated!.gotpl
```diff
@@ -135,13 +135,21 @@
 	func (e *executableSchema) Exec(ctx context.Context) graphql.ResponseHandler {
 		rc := graphql.GetOperationContext(ctx)
 		ec := executionContext{rc, e}
+		inputUnmarshalMap := graphql.BuildMap(
+			{{- range $input := .Inputs -}}
+			{{ if not $input.HasUnmarshal }}
+			ec.unmarshalInput{{ $input.Name }},
+			{{- end }}
+			{{- end }}
```
a8m, line 143:
```suggestion
			{{- range $input := .Inputs -}}
				{{ if not $input.HasUnmarshal }}
					ec.unmarshalInput{{ $input.Name }},
				{{- end }}
			{{- end }}
```

graphql/input.go
```diff
@@ -8,15 +8,15 @@ import (
...
-func BuildMap(unmarshaler ...interface{}) map[reflect.Type]reflect.Value {
+// BuildUnmarshalerMap returns a map of unmarshal functions of the ExecutableContext
+// to use with the WithUnmarshalerMap function.
+func BuildUnmarshalerMap(unmarshaler ...interface{}) map[reflect.Type]reflect.Value {
 	maps := make(map[reflect.Type]reflect.Value)
 	for _, v := range unmarshaler {
 		ft := reflect.TypeOf(v)
-		if ft.Kind() != reflect.Func {
-			panic("unmarshaler must be a function")
+		if ft.Kind() == reflect.Func {
```
StevenACoffman, line 17: I prefer this avoiding of panics, but if the kind is not a function... what now happens when nothing is added to the maps here?

StevenACoffman, line 17: I prefer this avoiding of panics, but if the kind is not a function... what now happens when nothing is added to the maps here?

giautm: We can handle it late, I think we should return the error in `UnmarshalInputFromContext` went the map is empty or the mapping is not found. @a8m

giautm:
This function can leak to another issue, that is resolver/directives can be execute twice times during query phrase. The first time by `UnmarshalInputFromContext`, then the second by gqlgen runtime.

graphql/input.go
```diff
@@ -0,0 +1,54 @@
...
+			maps[ft.Out(0)] = reflect.ValueOf(v)
+		}
+	}
+
+	return maps
+}
+
+func WithUnmarshalerMap(ctx context.Context, maps map[reflect.Type]reflect.Value) context.Context {
+	return context.WithValue(ctx, unmarshalInputCtx, maps)
+}
```
a8m, line 27: Add godoc please.

a8m (approved):

graphql/input.go
```diff
@@ -0,0 +1,54 @@
...
+		ft := reflect.TypeOf(v)
+		if ft.Kind() == reflect.Func {
+			maps[ft.Out(0)] = reflect.ValueOf(v)
+		}
+	}
+
+	return maps
+}
+
+func WithUnmarshalerMap(ctx context.Context, maps map[reflect.Type]reflect.Value) context.Context {
```
giautm, line 25:
```suggestion
// WithUnmarshalerMap returns a new context with InputMaps attached
func WithUnmarshalerMap(ctx context.Context, maps map[reflect.Type]reflect.Value) context.Context {
```

StevenACoffman:
> This function can lead to another issue, that is resolver/directives can be executed twice times during query phrase. The first time by `UnmarshalInputFromContext`, then the second by gqlgen runtime.

What is the symptom of this and how can we work around this problem?

graphql/input.go
```diff
@@ -0,0 +1,55 @@
...
+		ft := reflect.TypeOf(v)
+		if ft.Kind() == reflect.Func {
+			maps[ft.Out(0)] = reflect.ValueOf(v)
+		}
+	}
+
+	return maps
+}
+
+// WithUnmarshalerMap returns a new context with InputMaps attached
```
a8m, line 25:
```suggestion
// WithUnmarshalerMap returns a new context with a map from input types to their unmarshaler functions.
```

giautm:
> What is the symptom of this and how can we work around this problem?

For example, the below directive will be called twice times. But I think it won't have any critical effects (because the most directive are stateless)
```graphql
input TodoInput {
  blahBlah: String! @directiveCount
}
```

a8m:
> > This function can lead to another issue, that is resolver/directives can be executed twice times during query phrase. The first time by `UnmarshalInputFromContext`, then the second by gqlgen runtime.
> 
> What is the symptom of this and how can we work around this problem?

@StevenACoffman, currently, ent executes the resolvers unmarshaling "ahead of time" and resolves all fields (in all levels) in the root query. However, gqlgen runtime does not know that and calls all fields resolvers even though they were resolved - ent returns the cached result in this case. We plan to send additional patches to skip this and improve performance for ent users.

StevenACoffman:
Ok, I can merge this now, but I would ask for a fast follow with patches to have gqlgen runtime skip the redundant resolving so other users of this functionality do not have a problem that seems like it would be very hard to see the cause of.

For example, I can imagine someone who wanted to build a GraphQL federatation gateway using gqlgen using this same functionality with non-idempotent directives.

PR #2135, opened 2022-04-29
Use MultipartReader to parse file uploads

emersion:
Use a streaming MultipartReader to parse requests with file
uploads. The GraphQL multipart request specification guarantees
that the operations and map form fields will come first.

There are two reasons motivating this change:

- This allows for file uploads without specifying a specific
  filename.
- This avoids unnecessary copies for requests with more than one
  file. Go's ParseForm already copies the request's body into
  memory or on disk. We were also doing this manually as a second
  step.

I have:
 - [x] (N/A) Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [x] (N/A) Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

### DIFF
--- a/graphql/handler/transport/http_form.go
+++ b/graphql/handler/transport/http_form.go
@@ -7,7 +7,6 @@ import (
 	"mime"
 	"net/http"
 	"os"
-	"strings"
 
 	"github.com/99designs/gqlgen/graphql"
 )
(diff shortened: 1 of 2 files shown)


robertmarsal:
FYI after this PR the underlying type of `File` in `graphql.Upload` changed from `multipart.sectionReadCloser` to `*transport.bytesReader`. This is not a problem because the new type still implements the `io.Reader` type however the old type was more convenient because it implemented `io.Seeker` so the file could be sent directly to S3 for example. 

Leaving the comment for anyone wondering why their tests are suddenly failing.

emersion:
Yeah, because the HTTP request body cannot be seeked. To restore the previous behavior, you can copy to a temporary file.

PR #2140, opened 2022-05-03
Update getting-started.md

alan890104:
function rand.Int requires two parameters and returns two value in golang version 1.18.1.

I have:
 - [ ] Fixed document error in golang 1.18.1. (see [docs](https://github.com/99designs/gqlgen/blob/master/README.md))

### DIFF
--- a/docs/content/getting-started.md
+++ b/docs/content/getting-started.md
@@ -139,9 +139,10 @@ Returning to `graph/schema.resolvers.go`, let's implement the bodies of those au
 
 ```go
 func (r *mutationResolver) CreateTodo(ctx context.Context, input model.NewTodo) (*model.Todo, error) {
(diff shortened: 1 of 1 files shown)


StevenACoffman:
Are you sure this is correct? https://pkg.go.dev/math/rand#Rand.Int

alan890104:
Oh I see.

I don't know that the tutorial is using math/rand package, so I use Rand.Int() in package [crypto/rand](https://pkg.go.dev/crypto/rand#Int)  instead.

Sorry for misunderstanding.

StevenACoffman:
I really appreciate you taking the time to report it, as it means probably others were tripped up by this as well. I know I have accidentally substituted one of those for the other in the past. What do you think of how I reworded it to highlight the package?

StevenACoffman:
Thanks!

PR #2141, opened 2022-05-03
Fix docs bug in field collection

V1def:
Corrected an error in the documentation - [link](https://gqlgen.com/reference/field-collection/#practical-example).


StevenACoffman:
Thank you so much for your contribution, and we are all hoping that you and yours are safe.

PR #2142, opened 2022-05-03
Allow absolute URLs to the GraphQL playground

marcusirgens:
We sometimes want to use absolute URLs to the playground. This adds support for this by attempting to parse the playground URL using `url.Parse()`, and if there's a scheme, the URL is used as provided instead of being constructed by the client.

I have:

### DIFF
--- a/graphql/playground/playground.go
+++ b/graphql/playground/playground.go
@@ -3,6 +3,7 @@ package playground
 import (
 	"html/template"
 	"net/http"
+	"net/url"
 )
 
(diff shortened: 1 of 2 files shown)


marcusirgens:
I'm not sure how happy I am with the "`endpointIsAbsolute`" naming, but the code works. Happy to make changes.

StevenACoffman:
Thanks. I couldn't think of a dramatically better naming, but if you do get inspired before the next release, make a PR and I'll fast track it!

PR #2147, opened 2022-05-04
Update eagerly fetching section

theverything:
This fixes the docs so that the generate command will correctly generate the User resolver.

Discussions about this can be found here https://github.com/99designs/gqlgen/discussions/2146 and here https://github.com/99designs/gqlgen/discussions/2144

I have:

### DIFF
--- a/docs/content/getting-started.md
+++ b/docs/content/getting-started.md
@@ -211,7 +211,7 @@ type Todo struct {
 	ID     string `json:"id"`
 	Text   string `json:"text"`
 	Done   bool   `json:"done"`
-	User   *User  `json:"user"`
+	UserID string `json:"user"`
 }
 ```
 
@@ -226,7 +226,7 @@ func (r *mutationResolver) CreateTodo(ctx context.Context, input model.NewTodo) 
 	todo := &model.Todo{
 		Text:   input.Text,
 		ID:     fmt.Sprintf("T%d", rand.Int()),
-		User:   &model.User{ID: input.UserID, Name: "user " + input.UserID},
+		UserID: input.UserID,
 	}
 	r.todos = append(r.todos, todo)
 	return todo, nil


frederikhors:
This is not a solution. Please read carefully the docs and other issues and questions.

frederikhors:
This is not an issue at all.

theverything:
@frederikhors I'm not sure what you mean. The docs in the https://gqlgen.com/getting-started are incorrect, specifically the https://gqlgen.com/getting-started/#dont-eagerly-fetch-the-user. If you compare code snippets to the code in the completed tutorial here https://github.com/vektah/gqlgen-tutorials/tree/master/gettingstarted you can see that they are clearly different.

StevenACoffman:
Can you compare your fix with #2157 and see if it addresses your concern fully?

theverything:
@StevenACoffman it does not. The PR does not address the issue properly nor does it address all of the issues.

Adding the resolver section to the gqlgen config is not necessary because once you update the `Todo` struct with the `UserID` instead of the `User` field, gqlgen will create the resolver for you.

This is not needed.

```
 Todo:
    fields:
      user:
        resolver: true
```

Its also missing the update the the `CreateTodo` function that is included in my PR.

frederikhors:
@theverything, in gqlgen there are two different ways to do this. And they are both in docs.

theverything:
@frederikhors @StevenACoffman After the user finishes the the tutorial their code should look exactly like the code in the link referenced in the tutorial itself. Found here

https://github.com/99designs/gqlgen/blob/master/docs/content/getting-started.md?plain=1#L15

@frederikhors I understand there  is multiple ways to create a resolver but why have the user use both solutions when only one is necessary

scottopherson:
I was confused by the getting-started eager-fetching section as well. I came upon this PR while searching around to see if others were confused as well.

While searching around, [this issue comment](https://github.com/99designs/gqlgen/issues/2301#issuecomment-1192210518) got me to look closer at the github README and see the explanation about the two different eager-fetching solutions. I haven't found this info yet on https://gqlgen.com docs. That particular README section and the link to the completed tutorial code cleared up this topic for me.

PR #2148, opened 2022-05-05
Fix invalid relative playground subscription URL

marcusirgens:
A bug snuck in past the PR in #2142. Sorry!

This fixes it and adds tests for the subscription handler URL. 

cc: @StevenACoffmani


StevenACoffman:
Thanks!

PR #2149, opened 2022-05-09
go: update gqlparser to latest

a8m:
See title


a8m:
Merging as this change is really simple.

Issue #2167, opened 2022-05-14
Memory Leak at Websocket Transport


fletcherist:
### What happened?
There is a memory/goroutine leak at websocket transport

I think this goroutine never closes after client closes was connection. https://github.com/99designs/gqlgen/blob/master/graphql/handler/transport/websocket.go#L374

```go
responses, ctx := c.exec.DispatchOperation(ctx, rc)
for {
       // this never closes even if ctx is done
	response := responses(ctx)
	if response == nil {
		break
	}

	c.sendResponse(msg.id, response)
}
```

### Minimal graphql.schema and models to reproduce
I created a reproduction demo https://github.com/fletcherist/gqlgen-memory-leak

### versions
 - `go run github.com/99designs/gqlgen version` 
 - `go version` go1.17.7 darwin/amd64

Here's some stats
<img width="789" alt="memory-leak" src="https://user-images.githubusercontent.com/15146860/168428872-382a3d3b-40de-432b-b734-a96a29cd8e66.png">

![image](https://user-images.githubusercontent.com/15146860/168429025-9428cdb9-0175-4630-87e6-937c4dcb5d3a.png)

I'm working on fix

fletcherist:
https://github.com/99designs/gqlgen/pull/2168

i made a fix, please someone look onto iit

fletcherist:
however i continue investigating the issue, because there's another goroutine leak somewhere

StevenACoffman:
Hey, I really appreciate your persistence on this. Thanks and keep the contributions coming!

fletcherist:
thanks for merging this, i've deployed my production env with master branch's fix, and will monitor, I hope it should do

memory without fix:
<img width="445" alt="Screenshot 2022-05-15 at 5 54 35 PM" src="https://user-images.githubusercontent.com/15146860/168479264-02a06e0a-a855-4224-a819-8819ceab06d7.png">
 
I'll edit this comment with updated data after fix

I continue to investigate the issue:)

fletcherist:
Update: [MoofMonkey](https://github.com/MoofMonkey) found out the root cause of the problem. It turns out that in the [chat demo](https://github.com/99designs/gqlgen/blob/master/_examples/chat/resolvers.go#L135) we don't close that channel on ctx done and it causes goroutines leak. I thought it was handled internally, but it's not. However #2168 added this feature and now it works fine

fletcherist:
[...]

PR #2168, opened 2022-05-14
fix: prevents goroutine leak at websocket transport #2167

fletcherist:
Fixed memory leak described in #2167

I have:


StevenACoffman:
Thanks for hunting this down!

frederikhors:
@StevenACoffman I think we need to release a PATCH ASAP for this.

StevenACoffman:
@frederikhors In the issue, @fletcherist mentions there is another leak that he is working on identifying and fixing, so I would prefer to wait and give him time to do so and then include both in a new point release.

frederikhors:
Ok, but I think this is already an important PATCH to release because we don't know how much time he needs.

PR #2173, opened 2022-05-16
Add support to preload Go roots

### LINKED ISSUE #2169: Unable to support Bazel due to assumption in how module root is found
### What happened?
When trying to build using Bazel, the environment is pretty different from regular `go build`. Specifically, the [goModuleRoot](https://github.com/99designs/gqlgen/blob/master/internal/code/imports.go#L58) function assumes that there will be a `go.mod` file which is not always true.

### What did you expect?
We need a way to set a package root or set of package roots manually.

Are you open to a PR for this?

clayne11:
Doing this means you don't need to rely on assumptions about the project environment (ie every root has a go.mod file).

Fixes #2169

### DIFF
--- a/api/generate.go
+++ b/api/generate.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/99designs/gqlgen/codegen"
 	"github.com/99designs/gqlgen/codegen/config"
+	"github.com/99designs/gqlgen/internal/code"
 	"github.com/99designs/gqlgen/plugin"
 	"github.com/99designs/gqlgen/plugin/federation"
 	"github.com/99designs/gqlgen/plugin/modelgen"
@@ -43,6 +44,19 @@ func Generate(cfg *config.Config, option ...Option) error {
 
 	for _, o := range option {
 		o(cfg, &plugins)
+	}
+
+	for _, p := range plugins {
+		if inj, ok := p.(plugin.GoRootInjector); ok {
+			roots := inj.GoRoots()
+			for _, root := range roots {
+				code.AddGoRoot(code.GoModuleSearchResult{
+					Path:       root.Path,
+					GoModPath:  root.GoModPath,
+					ModuleName: root.ModuleName,
+				})
+			}
+		}
 	}
 
 	for _, p := range plugins {
--- a/internal/code/imports.go
+++ b/internal/code/imports.go
@@ -45,13 +45,21 @@ func NameForDir(dir string) string {
 	return SanitizePackageName(filepath.Base(dir))
 }
 
-type goModuleSearchResult struct {
-	path       string
-	goModPath  string
-	moduleName string
+// GoModuleSearchResult describes a root path
+type GoModuleSearchResult struct {
+	Path       string
+	GoModPath  string
+	ModuleName string
 }
 
-var goModuleRootCache = map[string]goModuleSearchResult{}
+var goModuleRootCache = map[string]GoModuleSearchResult{}
+
+// AddGoRoot enables you to augment the default mobile lookup behavior
+// with default roots. This helps with environments like Bazel where
+// you may not have a go.mod file.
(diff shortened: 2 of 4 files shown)


StevenACoffman:
Hmmm, looks like some things are not quite right.

clayne11:
Fixed the issue, please run again.

api/generate.go
```diff
@@ -45,6 +46,19 @@ func Generate(cfg *config.Config, option ...Option) error {
...
+				code.AddGoRoot(code.GoModuleSearchResult{
+					Path:       root.Path,
+					GoModPath:  root.GoModPath,
+					ModuleName: root.ModuleName,
+				})
+			}
+		}
+	}
+
 	for _, p := range plugins {
```
StevenACoffman, line 62:
```suggestion
		}

```
There's no point in iterating through the plugins twice.

StevenACoffman:
Looks good now, but because both of these plugins execute at the same time, we can avoid redundantly iterating through the plugins a second time.

mtibben:
Hmmmm isn't defining roots what `GOROOT` does? Shouldn't gqlgen just read that env var rather than adding a new API?

StevenACoffman:
@clayne11 You might also check in with some of your fellow Uber co-workers like @sbalabanov-zz @bmonkman @vnj-uber who have contributed to gqlgen in the past.

clayne11:
I do have a Bazel rule working, however it's internal Uber code so I would have to get legal clearance to publish it. I can give you some guidance though:

1. We do not use `gqlgen init` - this isn't particularly important, it's only needed one time to bootstrap a project. You can run this outside of the scope of your project and just copy the files over.
2. `gqlgen generate` uses `packages list` internally which uses the `go build` system under the hood. In order to make this work, or at least to make it fast, you need to provide a `GOPACKAGESDRIVER` and write out the packages you're going to use to a JSON file. You can see the `GOPACKAGESDRIVER` that ships with Bazel go rules [here](https://github.com/bazelbuild/rules_go/tree/master/go/tools/gopackagesdriver). We are not using this one internally but it should work.
3. Due to the `GOPACKAGESDRIVER`, I separated out model generation into a separate step before generating the server. Because model generation creates new Go packages, you need to reset the `GOPACKAGESDRIVER`. We use the model gen rule as a dependency in the server gen rule. After the model gen runs, we write out an intermediary `gqlgen.yml` and use it as an input to the server gen.

Let me know if you have any questions.

jamesthompson:
> I do have a Bazel rule working, however it's internal Uber code so I would have to get legal clearance to publish it. I can give you some guidance though:
> 
>     1. We do not use `gqlgen init` - this isn't particularly important, it's only needed one time to bootstrap a project. You can run this outside of the scope of your project and just copy the files over.
> 
>     2. `gqlgen generate` uses `packages list` internally which uses the `go build` system under the hood. In order to make this work, or at least to make it fast, you need to provide a `GOPACKAGESDRIVER` and write out the packages you're going to use to a JSON file. You can see the `GOPACKAGESDRIVER` that ships with Bazel go rules [here](https://github.com/bazelbuild/rules_go/tree/master/go/tools/gopackagesdriver). We are not using this one internally but it should work.
> 
>     3. Due to the `GOPACKAGESDRIVER`, I separated out model generation into a separate step before generating the server. Because model generation creates new Go packages, you need to reset the `GOPACKAGESDRIVER`. We use the model gen rule as a dependency in the server gen rule. After the model gen runs, we write out an intermediary `gqlgen.yml` and use it as an input to the server gen. Inside of this rule is where we write the "dummy" `go.mod` file.
> 
> 
> Let me know if you have any questions.

@clayne11 Any chance you could check to see if you could get permission to publish this internal rule? I'm new to `GOPACKAGESDRIVER` usage, so any tips you could give on how you use that would be greatly appreciated as well! Thanks a lot for your help!

Issue #1393, opened 2020-11-13
[RFC] Option in config to have return values based on schema nullability


frederikhors:
### What happened?

In #375 we discussed and changed the resolvers return from this:

for `query { user: User! }` we generate `func (r *Query) User(ctx context.Context) (User, error)`
for `query { user: User }` we generate `func (r *Query) User(ctx context.Context) (*User, error)`

to always this:

`query { user: User }` ---> `func (r *Query) User(ctx context.Context) (*User, error)`

### What did you expect?

To have in config something like: `omit_slice_element_pointers: false` (for slices) to have not pointer return values for resolvers.

Does it make sense or am I wrong?

frederikhors:
This is an example of the "issue".

I'm using [@vektah/dataloaden](https://github.com/vektah/dataloaden) with:

```
//go:generate go run github.com/vektah/dataloaden PlayerLoader int demo_app/Player
```

and in my resolver I have to use this code:

```go
func (r *teamResolver) Category(ctx context.Context, obj *Team) (*Category, error) {
	if obj == nil {
		return nil, nil
	}
	res, err := dataloaders.For(ctx).PlayerById.Load(obj.CategoryID)
	if err != nil {
		return nil, err
	}
	return &res, err
}
```

instead of:

```go
func (r *teamResolver) Category(ctx context.Context, obj *Team) (Category, error) {
	if obj == nil {
		return Category{}, nil
	}
	return dataloaders.For(ctx).PlayerById.Load(obj.ExamID)
}
```

ianling:
Can this be revisited @frederikhors @StevenACoffman ? The original issue (#375) that caused this change is flat-out wrong in its assumptions and resulted in a half-baked "solution" that I can't opt out of. It lists the following things as reasons to always return pointers instead of values:

> 
    1. Forces you to return an empty objects with errors, eg return User{}, fmt.Errorf("not found")
    2. Isn't very idiomatic. Most go apis will return (*Thing, error), and return nil when error is set
    3. Fragments function signatures; if you have a database function looking up a user it probably already returns a (*User, error), being able to return it directly would be nice.

#1 is not actually a problem and is perfectly normal in Go code.

#2 is wrong and doesn't cite any examples to support their claim.

#3 also isn't true.

[This comment](https://github.com/99designs/gqlgen/issues/375#issuecomment-500866097) that was left a few months after the change was implemented said about as much, and has a bunch of people agreeing with it, but it was ignored.

[...]

PR #2175, opened 2022-05-19
Config option to make resolvers return values instead of pointers

ianling:
https://github.com/99designs/gqlgen/issues/1393

Adds a config option to make resolvers return values instead of pointers.

For example, given the following schema:
```
type User {
  id: ID!
  name: String!
}

type Query {
  user: User!
  maybeUser: User
}
```

Old behavior before this PR (matches default behavior after this PR)
```
func (r *queryResolver) User(ctx context.Context) (*User, error) {
	panic("not implemented")
}

func (r *queryResolver) MaybeUser(ctx context.Context) (*User, error) {
	panic("not implemented")
}
```

New behavior with `resolvers_always_return_pointers: false` (`*User` vs `User` in return type of first function):

```
func (r *queryResolver) User(ctx context.Context) (User, error) {
	panic("not implemented")
}

func (r *queryResolver) MaybeUser(ctx context.Context) (*User, error) {
	panic("not implemented")
}
```

I have:


ianling:
Holding until my other PR is finished/merged

nikitacrit:
@ianling Thanks, I'm really looking forward to it. Any predictions when it will be merged?

StevenACoffman:
Sorry! I wanted to let the release sit for a few days for any urgent fixes for the websocket changes (since we don't use those in production at Khan Academy, so I always worry about breaking someone's edge case). I merged the other PR though.

ianling:
Fixed merge conflicts.

Don't be afraid of the large change size, the actual change is only like 1 line, the rest is auto-generated models for tests.

ianling:
Not sure why the Windows tests failed.

PR #2184, opened 2022-05-23
Improve operation error handling

Desuuuu:
Currently, operation errors are not treated as protocol errors due to the lack of error code. They look like this:
```json
{"errors":[{"message":"operation badOp not found"}],"data":null}
```

Or, if no operation is provided (e.g. the query is empty):
```json
{"errors":[{"message":"operation  not found"}],"data":null}
```

This change aims to improve on this by:
* Adding the `GRAPHQL_VALIDATION_FAILED` error code when an invalid operation name is provided, making transports treat it as a protocol error.
* Returning a separate error when no operation is provided.

This was previously requested here: https://github.com/99designs/gqlgen/discussions/1786

I also added 2 tests covering the changes to the executor.

### DIFF
--- a/graphql/executor/executor.go
+++ b/graphql/executor/executor.go
@@ -67,7 +67,9 @@ func (e *Executor) CreateOperationContext(ctx context.Context, params *graphql.R
 
 	rc.Operation = rc.Doc.Operations.ForName(params.OperationName)
 	if rc.Operation == nil {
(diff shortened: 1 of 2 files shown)


StevenACoffman:
Thanks! This is great! A lot of folks have struggled with less than helpful error messages in a variety of situations, so I really appreciate efforts to make it easier to understand when things go wrong.

PR #2219, opened 2022-05-26
fix: chat example frontend race condition

MoofMonkey:
This fixes the race condition bug revealed by #2209.
I have also noticed [`this.cache.batch`](https://stackoverflow.com/q/68608132) error that was caused by an incorrectly imported `InMemoryCache` from `apollo-cache-inmemory`, because it has been exported by the `@apollo/client` since v3, and the chat example is using v3.2.3.
It's far from perfect, but at least `this.cache.batch` error and the race condition are gone now.

I have:

### DIFF
--- a/_examples/chat/.gitignore
+++ b/_examples/chat/.gitignore
@@ -5,6 +5,7 @@
 
 # testing
 /coverage
+/corpus
 
 # production
 /build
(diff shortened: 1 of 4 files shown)


StevenACoffman:
So good! Thanks very much!

PR #2223, opened 2022-05-31
Respect options in WebsocketOnce

crossworth:

`WebsocketOnce` for some reason don't pass down the options.

Since is a small change to a package that is used for testing (I use it for testing as well), I have not added tests.


StevenACoffman:
Thanks for catching this!

PR #2231, opened 2022-06-07
fix #1876: Optional Any type should allow nil values

### LINKED ISSUE #1876: Optional Any type should allow nil values
### What happened?

When trying to use `Any` as an optional input parameter, if you omit it or pass in `null`, it panics:

```
interface conversion: interface is nil, not interface {}
```

The reason seems to be similar to the one in #799 - the relevant line in `generated.go`:

```
return ec.resolvers.Mutation().Set(rctx, args["key"].(string), args["value"].(interface{}))
```

The "value" arg should be checked for `nil` before the type assertion. If I make this change directly in the generated file, it works as expected.

### What did you expect?

Expected no error and the argument to be resolved to `nil`.

```
mutation {
    set(key: String!, value: Any): Boolean
}

scalar Any
```

### versions
 - `gqlgen version`? v0.16.0
 - `go version`? 1.17

vediatoni:
This PR addresses and fixes this open issue https://github.com/99designs/gqlgen/issues/1876

Short TLDR of the issue (quoted from the above link):

> When trying to use Any as an optional input parameter, if you omit it or pass in null, it panics.

Error it produces:
`interface conversion: interface is nil, not interface {}`

I added a simple solution that checks whether the argument is of an interface type and if it is, then instead of appending the standard string code `fc.Args[...].(...)`, it will append a stringified anonymous function that will either return an `interface{}` or `nil`. The value returned is decided based on the value of the argument. So if the argument value is `nil`, it won't try to do type assertion to `interface{}` (which makes Go panic), rather it will return `nil`.

This anonymous function can also be converted into a helper function. 

I have:

### DIFF
--- a/codegen/field_test.go
+++ b/codegen/field_test.go
@@ -140,7 +140,7 @@ func TestField_CallArgs(t *testing.T) {
 							Name: "test2",
 						},
 						TypeReference: &config.TypeReference{
-							GO: types.Typ[types.Int],
+							GO: types.Typ[types.String],
 						},
 					},
 				},
@@ -151,7 +151,7 @@ func TestField_CallArgs(t *testing.T) {
 						return nil
 					}
 					return fc.Args["test"].(interface{})
-				}(), fc.Args["test2"].(int)`,
+				}(), fc.Args["test2"].(string)`,
 		},
 		{
 			Name: "Resolver field that isn't root object with single int argument",


vediatoni:
> This seems ok, but I would like to see a unit test for it, which might be easier to do if it was a helper function. What do you think about adding a test?

I added a little unit test that tests the` Field.CallArgs()` function as a whole. Please tell me if this is alright, if it's not I'd love some basic pointers you'd be able to provide me. As you can see this is my first time contributing to this project and I don't have too much time to spend to fully dig in into the code. But we do need this issue fixed since it does impact our project a bit.
Thanks!

PR #2237, opened 2022-06-11
Use exact capitalization from field names overridden in config

### LINKED ISSUE #1447: should be able to override go field capitalization
### What happened?

I've inherited code that currently uses gqlgen v0.7.1.  I'm trying to upgrade to 0.13.0, to take advantage of new features.

There's a change to generated field capitalization which ripples through this inherited code base.  I'm hoping to override gqlgen's current preferred capitalization.  I haven't found any way to do that (short of defining the models manually).

For example, if I define a gql type...

```
type Something {
  somethingDDL: Int!
  somethingGB: Int!
}
```

the generated go type is...

```
type Something struct {
  SomethingDdl int // I want "SomethingDDL" here
  SomethingGb int // I want "SomethingGB" here
}
```

I've tried to set the name using `@goField(name: "SomethingDDL")` directives.  And I've tried to set the field name in the gqlgen.yml file.  Either I haven't done it just right, or the capitalization of generated names cannot be overridden.

### What did you expect?

I'd like to better control the capitalization of generated field names.  If I could, I'd make the generated code capitalize the first letter of the GQL field name, and leave the other letters unchanged.

It would be a nice feature to have, because gqlgen is opinionated when it comes to capitalizing field names.  It insists that "Id" (not an acronym) be spelled "ID", while "GB" (an acronym) be spelled "Gb".  And this has changed sometime between version 0.7 and 0.13.

Sorry I haven't put together a minimal schema.

### versions
 - `gqlgen version`? 
 0.13.0

 - `go version`?
1.15.2

 - dep or go modules?
modules

ianling:
Closes #1447 

Behavior prior to this PR: given the following schema and gqlgen config:

Schema:
```graphql
type RenameFieldTest {
    badName: String!
    otherField: String!
}
```

gqlgen.yml:
```yml
models:
  RenameFieldTest:
    fields:
      badName:
        fieldName: GOODnaME
```

gqlgen generates the following Go struct:

```go
type RenameFieldTest struct {
	GOODnaMe   string `json:"badName"`  // <----- field name should be GOODnaME, but e is lowercase
	OtherField string `json:"otherField"`
}
```

Behavior after this PR (modelgen strictly adheres to field names overridden in the config): 

```go
type RenameFieldTest struct {
	GOODnaME   string `json:"badName"` // <----- the 'e' is now correctly capitalized
	OtherField string `json:"otherField"`
```

Field names that are not overridden in the config will be exactly the same as before this PR, so users will only see different behavior if they had an overridden name that happened to fall into one of these funny capitalization edge cases.

I have:
 - [x] ~~Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))~~ (not necessary)

### DIFF
--- a/plugin/modelgen/models.gotpl
+++ b/plugin/modelgen/models.gotpl
@@ -29,7 +29,7 @@
 			{{- with .Description }}
 				{{.|prefixLines "// "}}
 			{{- end}}
-			{{ $field.Name|go }} {{$field.Type | ref}} `{{$field.Tag}}`
+			{{ $field.GoName }} {{$field.Type | ref}} `{{$field.Tag}}`
 		{{- end }}
 	}
 
--- a/plugin/modelgen/models.go
+++ b/plugin/modelgen/models.go
@@ -48,9 +48,12 @@ type Object struct {
 
 type Field struct {
 	Description string
-	Name        string
-	Type        types.Type
-	Tag         string
+	// Name is the field's name as it appears in the schema
+	Name string
+	// GoName is the field's name as it appears in the generated Go code
+	GoName string
+	Type   types.Type
+	Tag    string
 }
 
 type Enum struct {
@@ -178,7 +181,7 @@ func (m *Plugin) MutateConfig(cfg *config.Config) error {
 					}
 				}
 
-				name := field.Name
+				name := templates.ToGo(field.Name)
 				if nameOveride := cfg.Models[schemaType.Name].Fields[field.Name].FieldName; nameOveride != "" {
 					name = nameOveride
 				}
@@ -192,7 +195,8 @@ func (m *Plugin) MutateConfig(cfg *config.Config) error {
 				}
 
 				f := &Field{
-					Name:        name,
+					Name:        field.Name,
+					GoName:      name,
 					Type:        typ,
 					Description: field.Description,
 					Tag:         `json:"` + field.Name + `"`,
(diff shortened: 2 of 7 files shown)


StevenACoffman:
Thanks! Nice catch!

plugin/modelgen/models.gotpl
```diff
@@ -29,7 +29,7 @@
 			{{- with .Description }}
 				{{.|prefixLines "// "}}
 			{{- end}}
-			{{ $field.Name|go }} {{$field.Type | ref}} `{{$field.Tag}}`
```
carldunham, line 32: So is this because of a bug in the `go` filter?

ianling: Nope, the `go` filter was just being erroneously applied in situations where the field name and its capitalization were being explicitly overridden in the gqlgen config. This was a problem because that filter changes the capitalization of field names that are already valid in Go in a slightly opinionated way, which was unwanted behavior in this case.

carldunham: Got it. Cool.

YogeshLele: @ianling Is there a way (in config or otherwise) to disable this?

ianling: @YogeshLele disable what exactly? If you want to specify a particular capitalization for a field in your schema, this PR made it so you can do so without gqlgen overriding it.

MohitVachhani:
Will this work for the cases of enums ?

PR #2266, opened 2022-07-02
github: Fix CI pipelines

tmc:
This fixes the CI pipelines:

[Integration](https://github.com/tmc/gqlgen/actions/workflows/integration.yml) [Security](https://github.com/tmc/gqlgen/actions/workflows/security.yml) [Lint](https://github.com/tmc/gqlgen/actions/workflows/lint.yml) [Test](https://github.com/tmc/gqlgen/actions/workflows/test.yml)

This modernizes the workflows uses a curl retry to wait for the servers being up.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -5,7 +5,8 @@ jobs:
   integration:
     strategy:
       matrix:
-        go: [1.16]
+        go: [1.16, 1.18]
+        node: [14, 16]
     runs-on: ubuntu-latest
     timeout-minutes: 3
     steps:
@@ -15,7 +16,7 @@ jobs:
           go-version: ${{ matrix.go }}
       - uses: actions/setup-node@v3
         with:
-          node-version: 14
+          node-version: ${{ matrix.node }}
       - run: go mod download
       - run: cd integration ; npm install
       - run: .github/workflows/check-integration
@@ -23,7 +24,8 @@ jobs:
   federation:
     strategy:
       matrix:
-        go: [1.16]
+        go: [1.16, 1.18]
+        node: [14, 16]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
@@ -32,7 +34,7 @@ jobs:
           go-version: ${{ matrix.go }}
       - uses: actions/setup-node@v3
         with:
-          node-version: 14
+          node-version: ${{ matrix.node }}
       - run: go mod download
       - run: cd _examples/federation ; npm install
       - run: .github/workflows/check-federation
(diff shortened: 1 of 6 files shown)


.github/workflows/integration.yml
```diff
@@ -1,35 +1,50 @@
...
-    container: golang:${{ env.GO_VERSION }}-alpine
     steps:
       - uses: actions/checkout@v3
-      - run: apk add --no-cache --no-progress nodejs npm git bash
+      - uses: actions/setup-go@v3
+        with:
+          go-version: ${{ matrix.go }}
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 14
```
StevenACoffman, line 18: Can we make this a root env variable so we can only maintain it in one place?

.github/workflows/integration.yml
```diff
@@ -1,35 +1,50 @@
...
 on: [push, pull_request]
 
-env:
-  GO_VERSION: 1.16
-
 jobs:
   integration:
+    strategy:
+      matrix:
+        go: [1.16, 1.17, 1.18]
```
StevenACoffman, line 8: Testing on all three versions will take longer and cost more. Do you think passing on Go 1.16 is likely to fail on Go 1.17 or Go 1.18?

StevenACoffman (commented):
Did not testing on Go 1.17 or Go 1.18 cause problems?

.github/workflows/integration.yml
tmc: I thought GitHub was pretty generous with the open source use of github actions.

StevenACoffman:
This is great! Thanks so much for helping to avoid the flaky failures we've been experiencing!

.github/workflows/integration.yml
StevenACoffman: 99designs has exceeded the free tier, but it's unclear to me as to how much this repository contributed to that usage, compared to their other open source repositories, or their private GitHub action usage. Since it's not my money, I try to only use what I need. I think your revision where you are testing oldest supported (Go 1.16) and newest supported (Go 1.18) is a great compromise. Thanks!

mtibben: I would prefer the last 2 major versions of Go, to align with the Go EOL policy

StevenACoffman: @mtibben I agree with you that is the goal, but we at Khan are using Go 1.16 until [GCP AppEngine](https://cloud.google.com/appengine/docs/standard/go/release-notes) supports a newer Go version, and I can't help maintain this project if it is beyond that. GCP tells us they plan on skipping Go 1.17 and releasing support for Go 1.18 in H2 of 2022 (so July-Dec 2022) but haven't increased that. As a result, we previously agreed to keep gqlgen supporting the same Go version as Khan uses in prod (and newer).

mtibben: yeah fair enough 👍

PR #2267, opened 2022-07-02
codegen: fix the run order of resolvers

giautm:
We (entgql) have a report from @amrnt about the unexpected result when he want to trying to use the custom Input field to set the value for the mutation's input.

Because with `range` the order of the map's iteration is randomized, so the custom field may run first then overrided by the original input.

You can run the test here few times without cache: https://github.com/ent/contrib/compare/master...amrnt:contrib:custom-input-field

I have:


codegen/input.gotpl
```diff
@@ -14,7 +14,16 @@
 			{{- end}}
 		{{- end }}
 
-		for k, v := range asMap {
+		fieldsInOrder := []string{
+			{{- range $field := .Fields }}
+				{{$field.Name|quote}},
+			{{- end }}
+		}
```
a8m, line 21:
```suggestion
		fieldsInOrder := [...]string{
			{{- range $field := .Fields }}
				{{quote $field.Name}},
			{{- end }}
		}
```

a8m: I also think newline is not needed and you can write it as:

```
fieldsInOrder := [...]string{ {{range .Fields}}{{quote .Name}},{{end}} }
```

a8m (approved):
Looks good.

giautm:
Hello @StevenACoffman, please review this PR. I have added tests and updated the template to generate the input resolver

a8m (approved):
OK! Looks good 😍 

Thanks for addressing this @giautm

PR #2268, opened 2022-07-02
Replace use of strings.Title with cases.Title

tmc:
This removes the use of the deprecated `strings.Title` function and enables go 1.18 in CI while still working on Go 1.16+.

[Integration](https://github.com/tmc/gqlgen/actions/workflows/integration.yml) [Security](https://github.com/tmc/gqlgen/actions/workflows/security.yml) [Lint](https://github.com/tmc/gqlgen/actions/workflows/lint.yml) [Test](https://github.com/tmc/gqlgen/actions/workflows/test.yml)

Depends on #2266 (will rebase once that's in).


a8m (approved):
Looks good to me. My only concern before merging is if we introduce any breaking change by moving from `strings.Title` to `cases.Title`. Can you check please and add a link here? Thanks

StevenACoffman:
Thanks! This was reported to Go in 2013 in https://github.com/golang/go/issues/6801.

The two test cases I have used elsewhere for this are unicode punctuation and language specific capitalization.
Unicode punctuation is like this:
```
import (
    "fmt"
    "strings"
)

func main() {
    a := strings.Title("go.go\u2024go")
    b := "Go.Go\u2024Go"
    if a != b {
        fmt.Printf("%s != %s\n", a, b) // Go.Go․go != Go.Go․Go
    }
}
```
The language specific test case I've used is in Dutch words, "ijsland" should be capitalized as "IJsland", but the `strings.Title("ijsland")` is converted to "Ijsland".
```
caser := cases.Title(language.Dutch)
caser.String("ijsland") // Ijsland
```

StevenACoffman:
`golang.org/x/text/cases` currently has a `go.mod` with a `go` directive that indicates that the module was written assuming the semantics of the Go version 1.17 as here https://cs.opensource.google/go/x/text/+/master:go.mod;l=10 but appears to work fine in Go 1.16 as all gqlgen tests still pass in Go 1.16.

PR #2287, opened 2022-07-12
Hide `github.com/matryer/moq` dependency in `tools.go` from importers

danielwhite:
I have project(s) that use `go mod vendor`. Currently, `github.com/matryer/moq` is vendored despite it not being required at runtime.

Moving `tools.go` to `internal` hides this import from downstream users and avoids `github.com/matryer/moq` being vendored. 

`go generate` of the mocks still works as expected.

I have:
 - [x] ~Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))~
 - [x] ~Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))~


internal/code/imports_test.go
```diff
@@ -40,6 +40,6 @@ func TestNameForDir(t *testing.T) {
 
 	assert.Equal(t, "tmp", NameForDir("/tmp"))
 	assert.Equal(t, "code", NameForDir(wd))
-	assert.Equal(t, "internal", NameForDir(wd+"/.."))
```
danielwhite, line 43: Apologies for failing to sanity check the tests initially.

Turns out  that the assumption behind the this assertion broke due to moving `tools.go`. I've pointed it at a different path that has no Go code. This seems to match the intent behind the original test for the `internal/code/..` path.

I _may_ have misunderstood, so happy for alternative suggestions. :thinking:

StevenACoffman:
Thanks!

Issue #2291, opened 2022-07-14
Websocket Client can't handle KeepAlive message


alexvelea:
### What happened?
I created a stress test for my server similar to the [chat example](https://github.com/99designs/gqlgen/blob/master/_examples/chat/chat_test.go#L14). The stress test is working great until it reaches the `10s` mark when the KeepAlive message is being sent. The 10s is the default KA interval for the server and I'd like to not mess with that if possible. 

When the WebSocket Client receives a KeepAlive message it can't handle it and it just crashes with the following error:
```
expected data message, got client.operationMessage{Payload:json.RawMessage(nil), ID:"", Type:"ka"}
```

### What did you expect?
The KeepAlive message to be abstracted away from user applications. The WebSocket client comes by default with `gqlgen` or transport layer itself to handle the KA logic.

### Background
I've looked into older PRs and there's this one, #2209 which actually implements support for KeepAlive messages in the websocket client in commit 670cf22272b490005d46dc2bee1634de1cd06d68 but it's later reverted in d8309fd2b2e6a0dabb79a502f8a59d48930d3b7e. 

I totally agree that the support for KA messages in WS Client wouldn't've fixed that particular problem but I still argue it's a good feature to have.

### versions
 - `go run github.com/99designs/gqlgen version` 
   - `v0.17.2`
 - `go version`
    - `go version go1.18.3 darwin/arm64`

frederikhors:
Can you PR?

alexvelea:
@frederikhors opened PR #2293

frederikhors:
Technically this is not closed yet.

alexvelea:
Fixed in #2293

PR #1578, opened 2021-08-02
Add support for graphql-transport-ws with duplex ping-pong

zdraganov:
Extending this [PR](https://github.com/99designs/gqlgen/pull/1507), included also the duplex ping-pong functionality.

How to use:
```
srv.AddTransport(transport.Websocket{
    PingPongInterval: 5 * time.Second,
    # Other options ...
})

```
Example Apollo client usage:
```
let activeSocket: unknown = null;
let timedOut: unknown = null;

const wsLink = new WebSocketLink({
  url: REACT_APP_GRAPHQL_WS_ENDPOINT as string,
  keepAlive: 10_000, // ping server every 10 seconds
  on: {
    ping: (received) => {
      if (!received) {
        timedOut = setTimeout(() => {
          if (activeSocket && activeSocket.readyState === WebSocket.OPEN) {
            activeSocket.close(4408, 'Request Timeout');
          }
        }, 5_000); // wait 5 seconds for the pong and then close the connection
      }
    },
    pong: (received) => {
      if (received) {
        clearTimeout(timedOut);
      }
    },
  },
});
```

PR #2293, opened 2022-07-14
Add support for KeepAlive message in websocket client

alexvelea:
Addressing issue #2291 reimplementing commit 670cf22272b490005d46dc2bee1634de1cd06d68 from issue #2209

The default WebSocket client has no support for KeepAlive messages. This is a problem for use cases that take more than 10 seconds (the default keepalive interval).

This PR adds support for KeepAlive messages by ignoring them, so applications don't need to worry about handling KA messages when using the provided WS Client.

I have:

### DIFF
--- a/client/websocket.go
+++ b/client/websocket.go
@@ -109,32 +109,37 @@ func (p *Client) WebsocketWithPayload(query string, initPayload map[string]inter
 			return c.Close()
 		},
 		Next: func(response interface{}) error {
-			var op operationMessage
-			err := c.ReadJSON(&op)
-			if err != nil {
-				return err
-			}
-			if op.Type != dataMsg {
-				if op.Type == errorMsg {
-					return fmt.Errorf(string(op.Payload))
-				} else {
-					return fmt.Errorf("expected data message, got %#v", op)
+			for {
+				var op operationMessage
+				err := c.ReadJSON(&op)
+				if err != nil {
+					return err
 				}
-			}
 
-			var respDataRaw Response
-			err = json.Unmarshal(op.Payload, &respDataRaw)
-			if err != nil {
-				return fmt.Errorf("decode: %w", err)
-			}
+				if op.Type != dataMsg {
+					if op.Type == connectionKaMsg {
+						continue
+					} else if op.Type == errorMsg {
+						return fmt.Errorf(string(op.Payload))
+					} else {
+						return fmt.Errorf("expected data message, got %#v", op)
+					}
+				}
 
-			// we want to unpack even if there is an error, so we can see partial responses
-			unpackErr := unpack(respDataRaw.Data, response)
+				var respDataRaw Response
+				err = json.Unmarshal(op.Payload, &respDataRaw)
+				if err != nil {
+					return fmt.Errorf("decode: %w", err)
+				}
 
-			if respDataRaw.Errors != nil {
-				return RawJsonError{respDataRaw.Errors}
+				// we want to unpack even if there is an error, so we can see partial responses
+				unpackErr := unpack(respDataRaw.Data, response)
+
+				if respDataRaw.Errors != nil {
+					return RawJsonError{respDataRaw.Errors}
+				}
+				return unpackErr
 			}
-			return unpackErr
 		},
 	}
 }


alexvelea:
I'm unsure how to write a test for this without having a `graphql` schema & the generated files associated.
This seems a lot of overhead to me.
The simplest way to add a test would be to make it inside the chat example, but it seems out of scope.

Could someone direct me to a better approach?

StevenACoffman:
Hi @alexvelea ! Thanks for this improvement! Can you rewrite the if-elseif-else to a switch here https://github.com/99designs/gqlgen/pull/2293/files#diff-1bd741acc5f9d4c55a88ea72c713fc540237862bcb0e5185f55ddd0e0fe8f804R120 ?

@telemenar @RobinCPel @fletcherist @zdraganov @alexsn @jordanabderrachid You have all contributed to WebSocket improvements, and I wonder if you can provide guidance on this PR. We don't use WebSockets at Khan, so I'm nervous about changing things without getting perspective from those who are using them daily.

frederikhors:
@StevenACoffman I ask you to please evaluate this merge very well because websockets are always very delicate for everyone.

alexvelea:
Managed to write a test, but it uses the generated files under `_examples/chat` so I'm not able to import them, unfortunately.  

Gonna dump it here since I wouldn't put it on git



```
package client

import (
	"fmt"
	"github.com/99designs/gqlgen/_examples/chat"
	"github.com/99designs/gqlgen/graphql/handler"
	"github.com/99designs/gqlgen/graphql/handler/transport"
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
	"testing"
	"time"
)

func TestWSClientHandlesKA(t *testing.T) {
	srv := handler.New(chat.NewExecutableSchema(chat.New()))

	// reduce KeepAlive interval to 10 MS so the test runs fast
		KeepAlivePingInterval: 10 * time.Millisecond,
	})

	c := New(srv)

	sub := c.Websocket(fmt.Sprintf(
		`subscription @user(username:"vektah") { messageAdded(roomName:"#gophers%d") { text createdBy } }`,
		0,
	))
	defer sub.Close()

	var msg struct {
		resp struct {
			MessageAdded struct {
				Text      string
				CreatedBy string
			}
		}
		err error
	}

	// read WS initialization message, per chat example protocol definition
	msg.err = sub.Next(&msg.resp)
	require.NoError(t, msg.err, "sub.Next")
	require.Equal(t, "You've joined the room", msg.resp.MessageAdded.Text)
	require.Equal(t, "system", msg.resp.MessageAdded.CreatedBy)

	// sleep for 100 millisecond so multiple KeepAline messages get queued up
	time.Sleep(100 * time.Millisecond)

	go func() {
		var resp interface{}
		err := c.Post(`mutation {
					a:post(text:"Hello!", roomName:"#gophers0", username:"vektah") { id }
				}`, &resp)
		assert.NoError(t, err)
	}()

	// sub.Next call skips all queued up KA messages
	msg.err = sub.Next(&msg.resp)
	require.NoError(t, msg.err, "sub.Next")
	require.Equal(t, "Hello!", msg.resp.MessageAdded.Text)
	require.Equal(t, "vektah", msg.resp.MessageAdded.CreatedBy)
}
```

telemenar:
We don't use the packaged client implementation. Given that the README for the client says:

> This client is used internally for testing. I wanted a simple graphql client sent user specified queries.
> 
> You might want to look at: ... 

I'm personally not deeply concerned about client changes. 

At Observe we actually have a similar light weight testing client for the server side unit tests. For go-lang code that uses a graphql client for something other than testing we use https://github.com/Khan/genqlient/

zdraganov:
Hello! We are also not using the package as a GraphQL client.
I looked at the PR and it looks good to me.

One think to consider is the ping-pong mechanism here, as from my past PR, the ping-pong keep alive was released - https://github.com/99designs/gqlgen/pull/1578

If you want the client to be 100% compatible with the features the gqlgen server supports, maybe you have to consider implementing it as well (but it's not required). If you are interested to do so, check how it's handled on the server side here - https://github.com/99designs/gqlgen/blob/v0.17.12/graphql/handler/transport/websocket.go#L268-L271. The Golang websocket client should look the same.

Overall LGTM.

alexvelea:
Thanks a lot, everyone for taking the time to look at this, and esp @zdraganov for the insight
I'll look into the ping-pong mechanism as well and try to add it to this PR

alexvelea:
@zdraganov, I've investigated the ping-pong mechanism & it seems that it needs to be activated on purpose by setting the `Sec-Websocket-Protocol=graphql-transport-ws` header in the request. 

By default, it's not activated, and I would personally create a different client build to support the ping-pong mechanism. As stated in the Readme, this client is meant for testing only, and I'd like to not add a lot of complexity to it.

alexvelea:
I've gone over the docs & testing side of things. The testing side went more over how to generate code & the client doesn't really fit in that group. Also no relevant docs on that.

Ran `go test ./...` and every test passes.

Is there something else needed to do? What should I do to get the PR merged?

PR #2305, opened 2022-07-29
Add hackernews graphql api tutorial to other resources

Glyphack:
Hi, I wrote this tutorial a while back, and it still has some visitors. Add it here if you think it would be helpful for newcomers.

Just a note is that the gqlgen version is not the latest here, but I do update the tutorial from time to time.

I have:


StevenACoffman:
Yay! thanks for writing this and for bringing it to our attention.

PR #2314, opened 2022-08-04
Generate getters for interface fields

### LINKED ISSUE #1469: Generate Getters for Interface Fields
Currently, when you have a schema like this:
```
interface Foo {
  value: String!
}

type FooA {
  bla: String!
}

type FooB {
  blub: String!
}
```

The generated go file will look like this:
```golang
type Foo interface {
   IsFoo()
}

type FooA struct {
   Bla string
   Value string
}

func (f FooA) IsFoo() {}

type FooB struct {
   Blub string
   Value string
}

func (f FooB) IsFoo() {}

```

Additionally, it would be great to also generate getters for all interface fields. This makes using typeswitches obsolet when you're not interested in the underlying type.

```golang
type Foo interface {
   IsFoo()
   
   GetValue() string
}

type FooA struct {
   Bla string
   Value string
}

func (f FooA) IsFoo() {}

func (f FooA) GetValue() string {
   return f.Value
}

type FooB struct {
   Blub string
   Value string
}

func (f FooB) IsFoo() {}

func (f FooB) GetValue() string {
   return f.Value
}
```

neptoess:
Closes #1469 
The Star Wars example has a small hack in it (GetFriendsConnection / GetFriends) due to the custom resolver logic. That may need looked at, since it's no longer a very clear cut example

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@
 *.test
 *.out
 gqlgen
+*.exe
--- a/_examples/selection/models_gen.go
+++ b/_examples/selection/models_gen.go
@@ -8,6 +8,8 @@ import (
 
 type Event interface {
 	IsEvent()
+	GetSelection() []string
+	GetCollected() []string
 }
 
 type Like struct {
@@ -17,7 +19,9 @@ type Like struct {
 	Collected []string  `json:"collected"`
 }
 
-func (Like) IsEvent() {}
+func (Like) IsEvent()                    {}
+func (this Like) GetSelection() []string { return this.Selection }
+func (this Like) GetCollected() []string { return this.Collected }
 
 type Post struct {
 	Message   string    `json:"message"`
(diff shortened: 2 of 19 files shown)


neptoess:
I see a lot of the checks failing. Didn’t realize quite how many tests there were throughout the codebase. Are any safe to ignore, or should I go into each and make sure they all pass?

giautm:
> I see a lot of the checks failing. Didn’t realize quite how many tests there were throughout the codebase. Are any safe to ignore, or should I go into each and make sure they all pass?

There is some issue with the template, you need to fix all issues.
> template.gotpl: template: template.gotpl:39:17: executing "template.gotpl" at <.Interfaces>: error calling Interfaces: not found

neptoess:
The lint step is failing because of a generated code diff. No matter how I try to run `go generate` locally, including running `./.github/workflows/check-generate` directly in my local bash, I can't seem to generate anything different. Not sure how to fix this one.

StevenACoffman:
So the generated code comment has either one or two leading spaces depending on Go 1.18 vs Go 1.19:
```
//  These things are all valid, but without care generate invalid go code
```

Which originates from the GraphQL schema comment:
```
""" These things are all valid, but without care generate invalid go code """
```

Not sure what is different, but that's probably a separate issue from this PR, so it's ok to have manually fudged it here.

StevenACoffman:
Please see #2331

PR #2315, opened 2022-08-05
Leverage (*Imports).LookupType when generating interface field getters

neptoess:
After #2314 was merged, I tried using it as a go module from a different repository, and realized that import paths were being added to the type names. `templates` already has a helper function for this, so I just used that


StevenACoffman:
Thanks! Can you please ping me when you convert this from a draft?

neptoess:
> Thanks! Can you please ping me when you convert this from a draft?

Sorry about that :)

PR #2317, opened 2022-08-09
Include docstrings on interface getters

neptoess:
Continuing the #2314 saga, realized that docstrings weren't getting included for interface getters, so I'm adding those into the template.


StevenACoffman:
Thanks for your diligence!

Issue #2321, opened 2022-08-11
GraphQL Name Collision in overly aggressive "ToGo" conversion


dcarbone:
### What happened?

I have a GraphQL schema with the following Enum:

```graphql
enum MyEnum {
  Value
  value
  TitleValue
  title_value
}
```

And so-on.  This, unfortunately, is forcibly turned into:

```golang
type MyEnum string

const (
  MyEnumValue MyEnum = "Value"
  MyEnumValue MyEnum = "value"
  MyEnumTitleValue MyEnum = "TitleValue"
  MyEnumTitleValue MyEnum = "title_value"
)
```

The obvious issue here is that we have two consts with the same name, but different values.

### What did you expect?

The consts to be defined as

```golang
type MyEnum string

const (
  MyEnumValue MyEnum = "Value"
  MyEnumvalue MyEnum = "value"
  MyEnumTitleValue MyEnum = "TitleValue"
  MyEnumtitle_value MyEnum = "title_value"
)
```

or some derivative thereof.  Perhaps with an underscore in between the `$enum.Name` and `.Name` to make them more readable.

I am working on a PR on this, but I wanted to open this and get some feedback on possible alternative solutions.

### Minimal graphql.schema and models to reproduce

```graphql
enum MyEnum {
  Value
  value
  TitleValue
  title_value
}
```

### versions
 - gqlgen version: `v0.17.13`
 - go version: `go1.18.5`

dcarbone:
closed with #2322

PR #2322, opened 2022-08-12
Avoid GraphQL to Go Naming Collision with "ToGoModelName" func

dcarbone:
This is a first pass attempt at addressing #2321.  There are some edge cases, as described in the documentation, but I believe it to be about the best possible until the generator is altered to take a high-level look at every single possile generated name before templating even starts, and handles resolution there.

I have:

### DIFF
--- a/docs/content/reference/name-collision.md
+++ b/docs/content/reference/name-collision.md
@@ -1,0 +1,188 @@
+---
+title: Handling type naming collisions
+description: Examples of logic used to avoid type name collision
+linkTitle: Name Collision
+menu: { main: { parent: 'reference', weight: 10 }}
+---
+
+While most generated Golang types must have unique names by virtue of being based on their GraphQL `type` counterpart,
+which themselves must be unique, there are a few edge scenarios where conflicts can occur.  This document describes
+how those collisions are handled.
+
+## Enum Constants
+
+Enum type generation is a prime example of where naming collisions can occur, as we build the const names per value
+as a composite of the Enum name and each individual value.
+
+### Example Problem
+
+Currently, enum types are transposed as such:
+
+```graphql
+enum MyEnum {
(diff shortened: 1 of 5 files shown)


docs/content/reference/name-collision.md
```diff
@@ -0,0 +1,188 @@
...
+	MyEnumValue4 MyEnum = "Value_4"
+)
+```
+
+Which immediately leads to compilation errors as we now have three constants with the same name, but different values.
+
+### Resolution
+
+1. Store each name generated as part of a run for later comparison
+2. Try pretty approach first.  Use if no conflicts
```
StevenACoffman, line 78: Can you explain what the pretty approach is here? Thanks!

dcarbone: Hmm, so the pretty approach is to use the raw output from the existing `ToGo` function.  I did not change that logic in this PR, and the logic in there is a bit tricky.  Would it be sufficient to say something like "Attempt to coerce name into CapitalCaseForm", or something?

dcarbone:
If this solution is acceptable, I think I should do the same for the "private" name func, yeah?

docs/content/reference/name-collision.md
StevenACoffman: Yeah, that would be much more clear. Thanks!

StevenACoffman:
Looks like a failing test after this latest update. Thanks for sticking with this!

dcarbone:
@StevenACoffman yep, sorry, yesterday got busy.  Working on it now :)

dcarbone:
@StevenACoffman looks like all the things are happy.  any additional changes you'd like made?

StevenACoffman:
Thanks for sticking with it! This is a nice improvement.

Issue #953, opened 2019-12-11
Can't find docs/examples of Subscriptions


Sequoia:
### What happened?

I want to write a Subscription resolver

* Looked through this repo
* Looked through https://gqlgen.com

### What did you expect?

* Expected to find a section called "Subscriptions" explaining how to write them with gqlgen

### More info

Through some issue searching I found a link to this: https://github.com/99designs/gqlgen/blob/master/example/chat/resolvers.go#L115-L150

That's what I'm working off of currently. It would be useful to have some part of the docs touch on subscriptions.

arddor:
While the chat example is a good starting point, it misses in my opinion two important parts:

1. The channel is not closed in the cleanup. This leads to `gqlgen` never being able to terminate the internal go routine.
```go
go func() {
  <-ctx.Done()
  r.mu.Lock()
  delete(room.Observers, id)
  close(events) // <-
  r.mu.Unlock()
}()
```

2. The lock in this case can get blocked when there are connection issues with one client. The example buffers the channel, but I would argue that at some point the message should just be dropped so that no other clients are blocked from receiving messages.
```go
r.mu.Lock()
for _, observer := range room.Observers {
  if observer.Username == "" || observer.Username == message.CreatedBy {
    observer.Message <- &message
    select {
    case observer.Message <- &message:
    default:
      log.Printf("could not send message to %v\n", observer.Username)
    }
  }
}
r.mu.Unlock()
```

qhenkart:
@vektah is anyone working on this? I need to create a gql subscription and if no one is working on it, I will write some basic documentation once I figure it out

RobertoOrtis:
Yes, it would be nice to have some documentation on subscriptions. I have been trying to figure out how they work.

txbrown:
The only other reference I can find is https://www.freecodecamp.org/news/https-medium-com-anshap1719-graphql-subscriptions-with-go-gqlgen-and-mongodb-5e008fc46451/

RobertoOrtis:
> The only other reference I can find is https://www.freecodecamp.org/news/https-medium-com-anshap1719-graphql-subscriptions-with-go-gqlgen-and-mongodb-5e008fc46451/

I got them working with arsmn's fork for gofiber framework. However, I am having an issue where I get an error whenever a user abruptly closes the subscription by closing the browser window/tab

https://github.com/99designs/gqlgen/issues/1281

Naatan:
Bump

justgook:
[...]

PR #2346, opened 2022-08-28
Add `subscriptions.md` recipe to docs

Unkn0wnCat:
I've recently wanted to add subscriptions to my GraphQL-API and couldn't find anything in the official docs.

I think that's a shame, so I've decided to change it. 😄 

This PR contains `docs/content/recipes/subscriptions.md` which in turn contains my tutorial on subscriptions.

If you read this, please take some time to proof-read the document and if you find a mistake, please do a code-review with a change suggestion. I've tried proof-reading to the best of my ability, but more eyes are always better!

I have:
 - [x] N/A ~Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))~

Closing #953

### DIFF
--- a/docs/content/recipes/subscriptions.md
+++ b/docs/content/recipes/subscriptions.md
@@ -29,7 +29,7 @@ To add the WebSocket transport change your `main.go` by calling `AddTransport(&t
 on your query handler.
 
 **If you are using an external router, remember to send *ALL* `/query`-requests to your handler!**
-**Not just GET requests!**
+**Not just POST requests!**
 
 ```go
 package main
--- a/docs/content/recipes/subscriptions.md
+++ b/docs/content/recipes/subscriptions.md
@@ -1,0 +1,293 @@
+---
+title: "Subscriptions"
+description: Subscriptions allow for streaming real-time events to your clients. This is how to do that with gqlgen.
+linkTitle: "Subscriptions"
+menu: { main: { parent: 'recipes' } }
+---
+
+GraphQL Subscriptions allow you to stream events to your clients in real-time.
+This is easy to do in gqlgen and this recipe will show you how to setup a quick example.
+
+## Preparation
+
+This recipe starts with the empty project after the quick start steps were followed.
+Although the steps are the same in an existing, more complex projects you will need
+to be careful to configure routing correctly.
+
+In this recipe you will learn how to
+
+1. add WebSocket Transport to your server
+2. add the `Subscription` type to your schema
+3. implement a real-time resolver.
+
+## Adding WebSocket Transport
+
+To send real-time data to clients, your GraphQL server needs to have an open connection
+with the client. This is done using WebSockets.
+
+To add the WebSocket transport change your `main.go` by calling `AddTransport(&transport.Websocket{})`
+on your query handler.
+
+**If you are using an external router, remember to send *ALL* `/query`-requests to your handler!**
+**Not just GET requests!**
+
+```go
+package main
+
+import (
+	"log"
+	"net/http"
+	"os"
+
+	"github.com/99designs/gqlgen/graphql/handler"
+	"github.com/99designs/gqlgen/graphql/handler/transport"
+	"github.com/99designs/gqlgen/graphql/playground"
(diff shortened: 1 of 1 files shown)


StevenACoffman:
Sorry for the delay! This looks great. I was trying to figure out some way to scrape this and other tutorials to test that it continues to function, as we have had some problems where we don't always keep the documentation up to date.

Unkn0wnCat:
Thanks @StevenACoffman! I'll look into if I can write some kind of automated unit-test for the tutorial code, although I don't quite know how I'd go about testing everything at once

zdraganov:
I have a question. In blog posts from previous versions, I was looking at people awaiting for ctx.Done() to be invoked, but now I see the examples in the docs are awaiting the default of the select.

Is there something that requires the change?

zdraganov:
Thanks for the quick reply!

I will give you my example that we have to use the subscriptions and using `ctx.Done()` makes it at least more readable.

```
func (r *subscriptionResolver) ChatMessage(ctx context.Context) (<-chan model.ChatMessage, error) {
        user := utils.GetUserFromContext(ctx)
	channel := r.ChatMessagesBroadcast.Subscribe(user.ID) // this returns the <-chan model.ChatMessage

	go func() {
	     <-ctx.Done()
	     r.ChatMessagesBroadcast.Unsubscribe(user.ID)
	}()

	return channel, nil
}
```

I have found couple of examples online following this approach, but in that case we have the control at which point to call the Unsubscribe function to properly release the subscription.

If I follow the example from the docs, my implementation should be something similar to:
```
func (r *subscriptionResolver) ChatMessage(ctx context.Context) (<-chan model.ChatMessage, error) {
        user := utils.GetUserFromContext(ctx)
	channel := r.ChatMessagesBroadcast.Subscribe(user.ID)
	resultChannel := make(<-chan model.ChatMessage)

	go func() {
	    for {
                 msg := <- channel
	         
	         select {
	         case resultChannel <- msg:
	         default:
	              r.ChatMessagesBroadcast.Unsubscribe(user.ID)
	              return
	         }
	    }
	}()

	return resultChannel, nil
}
```

The example above is not only looking more complex, but can lead to a blocking on `msg := <- channel` even if the `resultsChannel` is closed. If we not have a proper `ctx.Done()` it will be harder to implement proper exit mechanism from the subscriptions.

What do you think?

Unkn0wnCat:
@zdraganov

If the implementation which provides the channel / sends to it already handles safe sending your first solution is good, and yes it's cleaner.

However if the send is unsafe (aka. just sending without any `select`-esque checks) a race condition could happen due to other Goroutines running in parallel:

```go
go func() {
  <-ctx.Done()
  // Imagine right here the send is happening -
  // The unsubscribe has not yet happened, even though
  //  the context has been finished
  // The send method will now panic and, if not `recover()`-ed
  //  it will take your entire program go down.
  r.ChatMessagesBroadcast.Unsubscribe(user.ID)
}()
```

The example I wrote for the documentation is built with the `select`-statement to make it as safe as possible out-of-the-box, even for newcomers to channels in Go. If the channel is completely handled somewhere else with a safe send method the `select` statement if not needed.

zdraganov:
After couple of load tests, I can confirm that you should check before sending to the channel. There is really a race condition causing send to closed channels even with the `<-ctx.Done()` approach.

zdraganov:
Actually I found one more issue with the approach in the docs.
If you work with unbuffered channels in the subscriptions and there is in-flight subscription message, you will go in the default block and will cut-off the subscription.

Unkn0wnCat:
That's actually a good call, you're completely right on that!
I've overlooked that when writing the example. Maybe you could do a PR adding the buffer? Otherwise I can do it later-ish.

silicon-ninja:
Hey guys, i've made a graphql subscription able to connect to websocket how do i start receiving the data ? 

Screenshot 2022-11-04 at 2 28 07 PM

tried doing this but it disconnects right after that.. 

Screenshot 2022-11-04 at 2 28 44 PM

Issue #2331, opened 2022-08-19
0.17.14 generate getters for interfaces problem


argoyle:
### What happened?
When upgrading to 0.17.14 my code generation started to fail due to the new generation of getters for interfaces. It's due to interfaces returning other interfaces I think.

### What did you expect?
Code generation to result in valid code

### Minimal graphql.schema and models to reproduce
Reproduction repo: https://github.com/argoyle/gqlgen-implements

### versions
 - `go run github.com/99designs/gqlgen version`? 0.17.14
 - `go version`? 1.19

argoyle:
I can of course work around this by removing a couple of the interfaces (Node, Edge, Connection). Looking at it right now they don't really give me any value. 🙈

StevenACoffman:
@neptoess Can you look at the  https://github.com/argoyle/gqlgen-implements repo and investigate what is happening with your change? I'm traveling and this is hard to look at on a phone.

neptoess:
This is an interesting one. Looking at the schema (https://github.com/argoyle/gqlgen-implements/blob/main/graph/schema.graphqls), we have `UsersConnection implements Connection`, but... it doesn't look like it actually does. The `edges` field on `UsersConnection` is `[UsersEdge!]!`, but the `edges` field in the `Connection` interface is `[Edge!]!`.

Even though `UsersEdge implements Edge` is true, I'm pretty sure `[UsersEdge!]! implements [Edge!]!` isn't, i.e., even if every element in `edges` is a `UsersEdge`, I'm fairly positive the type on `edges` still needs to be `[Edge!]!`.

I can see if the GraphQL spec calls this out.

MichaelMure:
I have a similar issue, but #2332 doesn't solve it. It happens when I try to upgrade gqlgen in https://github.com/MichaelMure/git-bug. Repro repo: https://github.com/MichaelMure/gqlgen-issue

In short:
- I have a bunch of struct in the core of the code. Some of them implement the sign-post that gqlgen generate because why not, it's not hard to do.
- I have a layer that does lazy loading and wrap some of the previous types. It also change a bit the interface and add errors (as the lazy loading can fail)
- I ask gqlgen to bind on the lazy loading types, which also transitively carry some of the normal types
- until 0.17.13, gqlgen was happy with that. With 0.17.14, gqlgen generate an additional getter in the `Authored` interface, which I can't implement in the normal types as that would create a circular dependency (--> lazy --> normal --> lazy).
[...]

PR #2348, opened 2022-08-29
Add omit_getters config option

neptoess:
As discussed in #2331 with @StevenACoffman and @MichaelMure , adding a config option to disable getter generation for interface types

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,16 @@ first model in this list is used as the default type and it will always be used 
 
 There isn't any way around this, gqlgen has no way to know what you want in a given context.
 
+### Why do my interfaces have getters? Can I disable these?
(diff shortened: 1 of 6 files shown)


MichaelMure (commented):
From a gqlgen outsider, that seems reasonable. Thanks!

StevenACoffman:
Thanks!

MichaelMure:
I can confirm that it worked for me. However, that option is not documented on the website, it might make sense to do so.

neptoess:
I added it as the last question in the FAQ section of the README

MichaelMure:
Arguably, I'd expect all the possible options to show up at https://gqlgen.com/config/

PR #2366, opened 2022-09-12
fix: apollo federation tracer was race prone

roeest:
The apollo federation tracer was using a global state that was accessed from different goroutines regardless of the request they are currently handling.
Added req headers to operation context to allow it to be fetched in InterceptOperation

When running the following test (via `go test -race ./...`) i was able to consistently get error messages from go's race detection.
```go
func TestApolloTracing_Concurrent(t *testing.T) {
	h := testserver.New()
	h.AddTransport(transport.POST{})
	h.Use(&apollofederatedtracingv1.Tracer{})
	for i := 0; i < 2; i++ {
		go func() {
			resp := doRequest(h, http.MethodPost, "/graphql", `{"query":"{ name }"}`)
			assert.Equal(t, http.StatusOK, resp.Code, resp.Body.String())
			var respData struct {
				Extensions struct {
					FTV1 string `json:"ftv1"`
				} `json:"extensions"`
			}
			require.NoError(t, json.Unmarshal(resp.Body.Bytes(), &respData))
			tracing := respData.Extensions.FTV1
			pbuf, err := base64.StdEncoding.DecodeString(tracing)
			require.Nil(t, err)

			ftv1 := &generated.Trace{}
			err = proto.Unmarshal(pbuf, ftv1)
			require.Nil(t, err)
			require.NotZero(t, ftv1.StartTime.Nanos)
		}()
	}
}
```
Made some additional fixes to the test so it passes with `-race` flag.

### DIFF
--- a/graphql/context_operation.go
+++ b/graphql/context_operation.go
@@ -3,6 +3,7 @@ package graphql
 import (
 	"context"
 	"errors"
+	"net/http"
 
 	"github.com/vektah/gqlparser/v2/ast"
 )
@@ -15,6 +16,7 @@ type OperationContext struct {
 	Variables     map[string]interface{}
 	OperationName string
 	Doc           *ast.QueryDocument
+	Headers       http.Header
 
 	Operation              *ast.OperationDefinition
 	DisableIntrospection   bool
--- a/graphql/executor/executor.go
+++ b/graphql/executor/executor.go
@@ -58,6 +58,7 @@ func (e *Executor) CreateOperationContext(ctx context.Context, params *graphql.R
 
 	rc.RawQuery = params.Query
 	rc.OperationName = params.OperationName
+	rc.Headers = params.Headers
 
 	var listErr gqlerror.List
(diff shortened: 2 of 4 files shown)


StevenACoffman (approved):
Thanks!

roeest:
Thanks for approving, any idea when the next release will take place? Would love to get this fix in my project

StevenACoffman:
I'm getting test failures on windows with some regularity:
```
--- FAIL: TestApolloTracing (0.01s)
    tracing_test.go:48: 
        	Error Trace:	tracing_test.go:48
        	Error:      	"25[26](https://github.com/99designs/gqlgen/actions/runs/3046701310/jobs/4910136009#step:4:27)05000" is not less than "252605000"
        	Test:       	TestApolloTracing
FAIL
```

I can re-run it, and eventually it seems like it passes, but I wonder if you can look to see if there's something to make it more reliable?

StevenACoffman:
@roeest I went ahead and released https://github.com/99designs/gqlgen/releases/tag/v0.17.17 but I would like you to look at improving your test reliability if possible.

PR #2380, opened 2022-09-16
Fix field merging behavior for fragments on interfaces

### LINKED ISSUE #2379: Bug when merging fields from a fragment on an interface
### What happened?

Given the following schema:

```
type Query {
    dog: Dog
}

interface Animal {
    species: String!
    size: Size!
}

type Size {
    height: Int!
    weight: Int!
}

type Dog implements Animal {
    species: String!
    size: Size!
    dogBreed: String!
}
```

Then the following query:

```
{
  dog {
    size {
      height
    }
    ...AnimalWeight
  }
}
fragment AnimalWeight on Animal  {
  size {
    weight
  }
}
```

returns duplicate keys in the response JSON:

```
{
  "dog": {
    "size": {
      "height": 100
    },
    "size": {
      "weight": 35
    },
  }
}
```

When parsed by a JSON parser, the second `size` key overrides the first, and the `height` field cannot be read by clients.

### What did you expect?

```
{
  "dog": {
    "size": {
      "height": 100,
      "weight": 35
    },   
  }
}
```

Fragment fields should be merged correctly.

See above

### versions
 - `go run github.com/99designs/gqlgen version`? `v0.17.16`, but can be repro'd in newer versions as well
 - `go version`? `go1.19.1 darwin/arm64`

kxfang:
This PR fixes #2379 . The cause appears to be that when merging fields from a fragment spread, the interface equality check needs to be done in both directions, but is only done in one direction in the existing code.

I have:

### DIFF
--- a/codegen/testserver/followschema/interfaces.generated.go
+++ b/codegen/testserver/followschema/interfaces.generated.go
@@ -196,6 +196,53 @@ func (ec *executionContext) fieldContext_Cat_species(ctx context.Context, field 
 	return fc, nil
 }
 
+func (ec *executionContext) _Cat_size(ctx context.Context, field graphql.CollectedField, obj *Cat) (ret graphql.Marshaler) {
+	fc, err := ec.fieldContext_Cat_size(ctx, field)
+	if err != nil {
+		return graphql.Null
+	}
+	ctx = graphql.WithFieldContext(ctx, fc)
+	defer func() {
+		if r := recover(); r != nil {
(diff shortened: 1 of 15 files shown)


StevenACoffman:
Thanks very much for this!

kxfang:
@StevenACoffman I’m glad I could contribute! Any chance of a release in the next day or two?

davidbyttow:
Thank you @StevenACoffman! (and Kev) ;)

PR #2382, opened 2022-09-17
Update go-colorable and x/tools.

tjni:
This picks up a new 2022 version of golang.org/x/sys which is caused by https://github.com/golang/go/issues/49219 and is needed to fix building using Go 1.18 on aarch64-darwin.


StevenACoffman:
Oh! Thanks very much!

PR #2396, opened 2022-09-28
Add json.Number support to UnmarshalString

Desuuuu:
`UnmarshalString` currently supports most basic types (`string`, `int`, `float64`, ...), but not `json.Number`, which is used when decoding float values from JSON.

This PR:
* Adds `json.Number` support to `UnmarshalString`
* Implements `graphql.UnmarshalString` tests which were missing
* Changes the behaviour of `graphql.UnmarshalString` with `float64` inputs to output a string with the smallest number of digits necessary (`graphql.UnmarshalString(float64(123))` will output `"123"` instead of `"123.000000"`)

The last change while not strictly necessary seemed to make more sense when I was writing the tests. I made it in a separate commit so it can be easily discarded.

I have:

### DIFF
--- a/graphql/string.go
+++ b/graphql/string.go
@@ -56,6 +56,8 @@ func UnmarshalString(v interface{}) (string, error) {
 		return strconv.FormatInt(v, 10), nil
 	case float64:
 		return fmt.Sprintf("%f", v), nil
+	case json.Number:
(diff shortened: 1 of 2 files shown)


StevenACoffman:
Thanks!

PR #2421, opened 2022-11-07
replace encoding/json with goccy/go-json

roeest:
goccy/go-json is a more performant version of encoding/json.
Using it seems to improve the amount of memory allocated and ns/op. 
I used the benchmark in _examples/starwars to compare the performance:

encoding/json version:
```
BenchmarkSimpleQueryNoArgs-12              48986             23931 ns/op            8237 B/op        146 allocs/op
```

goccy/go-json version:
```
BenchmarkSimpleQueryNoArgs-12              53048             21194 ns/op            7824 B/op        141 allocs/op
```
 On average i observed about a 5% improvement

### DIFF
--- a/_examples/go.mod
+++ b/_examples/go.mod
@@ -17,6 +17,7 @@ require (
 )
 
 require (
+	github.com/goccy/go-json v0.9.11 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/gorilla/mux v1.8.0 // indirect
 	github.com/opentracing/basictracer-go v1.1.0 // indirect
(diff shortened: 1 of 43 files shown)


mtibben:
With just a 5% improvement in reality this might shave off microseconds from a typical request. Personally I don't think it's worth diverging from the stdlib for tiny gains

roeest:
> @roeest may you check the diff if it goes with https://github.com/json-iterator/go ?

I could do that. I'm also considering expanding the size of the requests and responses and adding another test to the benchmark, I was looking at the doc and the improvement is more significant for larger data. Now I just need to find some time to do it :)

PR #2432, opened 2022-11-19
Check if go.mod exists while init

soeti:
If for something reason we miss go.mod and then run "go run github.com/99designs/gqlgen init" we get cascade errors like:

```
projects/source/go/test/typeparam/issue51836.dir imports
        ./aa: "./aa" is relative, but relative import paths are not supported in module mode
tidy failed: go mod tidy failed: exit status 

```

This simple check prevent this


StevenACoffman:
Thanks!

PR #2434, opened 2022-11-25
Add resolver commit

voslartomas:
I would like to add `Generated by` section into resolvers comments to be able to suppress linters. We have internal linters to have Copyright docs and also there is `type queryResolver struct{ *Resolver }` at the end of the file and our gocritic expects it to be defined before it's use.

If this is not possible, would it be possible to add possibility to add custom docs at the beginning and also fix order of queryResolver? 

I have:


StevenACoffman:
Thanks!

voslartomas:
Thank you, when are you planning to create new release?

PR #1477, opened 2021-03-04
Nullable fields generated with json omit empty tag

Luscha:
Describe your PR and link to any relevant issues. 

Nullable fields should be generated with json omit empty tag.
There are cases where the graphql server accepts only one of the field to perform a mutation (an example could be performing a mutation on an union field which expects only one of the type reference set).

This is the case working in particular with Dgraph, which expects a strict format for inputs.

I have:
 - [X ] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [ ] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

PR #2436, opened 2022-11-25
Add omitempty json tags for nullable fields

### LINKED ISSUE #789: Request for omitempty on json tag for models
Instead of having to manually define models with the 'omitempty' json tag, it would be great if we could pass that through the yaml config file.

StevenACoffman:
Fix #789: Request for `omitempty` on json struct tag for models

### Per @Luscha PR #1477 : Nullable fields generated with json omit empty tag

> Nullable fields should be generated with json omit empty tag.
> There are cases where the graphql server accepts only one of the field to perform a mutation (an example could be performing a mutation on an union field which expects only one of the type reference set).

This is the case working in particular with Dgraph, which expects a strict format for inputs.

I have rebased that PR and regenerated it


StevenACoffman:
Looks like rebasing isn't enough, and it requires a bit of cleanup, but I'm not able to give it attention right now. Help wanted.

plugin/modelgen/models_test.go
```diff
@@ -65,9 +65,9 @@ func TestModelGeneration(t *testing.T) {
 
 		expectedTags := []string{
 			`json:"missing2" database:"MissingTypeNotNullmissing2"`,
-			`json:"name" database:"MissingInputname"`,
-			`json:"missing2" database:"MissingTypeNullablemissing2"`,
-			`json:"name" database:"TypeWithDescriptionname"`,
+			`json:"name, omitempty" database:"MissingInputname"`,
+			`json:"missing2, omitempty" database:"MissingTypeNullablemissing2"`,
+			`json:"name, omitempty" database:"TypeWithDescriptionname"`,
```
pascalmail, line 70:
```suggestion
			`json:"name,omitempty" database:"MissingInputname"`,
			`json:"missing2,omitempty" database:"MissingTypeNullablemissing2"`,
			`json:"name,omitempty" database:"TypeWithDescriptionname"`,
```

pascalmail: In line 84-89, change it to this:
```
		expectedTags := []string{
			`json:"name,omitempty" anotherTag:"tag"`,
			`json:"enum,omitempty" yetAnotherTag:"12"`,
			`json:"noVal,omitempty" yaml:"noVal" repeated:"true"`,
			`json:"repeated,omitempty" someTag:"value" repeated:"true"`,
		}
```

After both of these changes, running:
```
go generate; cd _examples; go generate; cd ..
go test ./...; cd _examples; go test ./...
```

All test will pass.

koenpunt:
I believe `omitempty` will also leave out fields which are explicitly set to `nil`, which isn't correct; in graphql there's a difference between sending in `null` and omitting a field.

StevenACoffman:
Hmmm... I won't get a chance to look at this soon, so if you can make a PR, I would appreciate it.

adomaskizogian:
> I believe `omitempty` will also leave out fields which are explicitly set to `nil`, which isn't correct; in graphql there's a difference between sending in `null` and omitting a field.

Having the same question. Has anyone experienced any issues on live systems?

PR #2437, opened 2022-11-27
feat: support Altair playground

yk-eukarya:
I've added another graphQL client known as [Altair](https://altairgraphql.dev/).

No test was added.
I could not find where should I update the docs!

### DIFF
--- a/graphql/playground/altair_playground.go
+++ b/graphql/playground/altair_playground.go
@@ -1,0 +1,84 @@
+package playground
+
+import (
+	"html/template"
+	"net/http"
+)
+
+var altairPage = template.Must(template.New("altair").Parse(`<!doctype html>
+<html>
+
+<head>
+  <meta charset="utf-8">
+  <title>{{.title}}</title>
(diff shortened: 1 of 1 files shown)


StevenACoffman:
Can you say more as to what is different between Altair and GraphiQL? Also, it looks like you added it, but have not hooked it up to be routable.

yk-eukarya:
@StevenACoffman For me, the main big difference is Altair has a file upload functionality, in addition to a lot of many cool features.
> Also, it looks like you added it, but have not hooked it up to be routable.
sorry I did not get this point, I'm thinking of Altair as another option, not a replacement

graphql/playground/altair_playground.go
```diff
@@ -0,0 +1,84 @@
...
+    window.addEventListener("load", function() {
+      AltairGraphQL.init(altairOptions);
+    });
+  </script>
+</body>
+
+</html>`))
+
+// AltairHandler responsible for setting up the altair playground
+func AltairHandler(title string, endpoint string) http.HandlerFunc {
```
kacperjurak, line 67: `title, endpoint string`

Issue #2430, opened 2022-11-17
Error handling for malformed body


koniuszy:
### What happened?
I cannot intercept the error using [the hooks](https://gqlgen.com/reference/errors/). I always get:
```
{"errors":[{"message":"json body could not be decoded: json: cannot unmarshal array into Go value of type graphql.RawParams"}],"data":null}
```

### What did you expect?
Possibility to overwrite the error. I want to return `Internal server error` or `wrong body/query error` instead of revealing the genuine error.

### Minimal graphql.schema and models to reproduce
any will work,  your request needs to include invalid json, not matching graphql.RawParams, for example:
`--data-raw '[]'`

### versions
 - `go run github.com/99designs/gqlgen version`? **0.17.20**
 - `go version`?: **1.19**

StevenACoffman:
Can you make a small repository that reproduces your problem?
```
package bar

import (
	"context"
	"errors"

	"github.com/vektah/gqlparser/v2/gqlerror"
	"github.com/99designs/gqlgen/graphql"
	"github.com/99designs/gqlgen/graphql/handler"
)

func main() {
	server := handler.NewDefaultServer(MakeExecutableSchema(resolvers))
	server.SetErrorPresenter(func(ctx context.Context, e error) *gqlerror.Error {
		err := graphql.DefaultErrorPresenter(ctx, e)

		var myErr *MyError
		if errors.As(e, &myErr) {
			err.Message = "Eeek!"
		}

		return err
	})
}
```

koniuszy:
@StevenACoffman 
I literally followed your `Quick start` guide. Launched the server and hit the server with
```shell
$ curl --compressed -i 'http://localhost:8080/query' -X POST -H 'Content-Type: application/json' --data-raw '[]'
```

Repo is here:
https://github.com/koniuszy/basic-gqlgen-example

StevenACoffman:
Thanks for your help!

PR #2438, opened 2022-11-28
fix: safe http error response 

koniuszy:
I cannot intercept the error of an invalid request using [the hooks](https://gqlgen.com/reference/errors/#hooks). I always get:
```
{
  "errors": [
    {
      "message": "json body could not be decoded: json: cannot unmarshal array into Go value of type graphql.RawParams"
    }
  ],
  "data": null
}
```

This change removes from the response the details of a decoding error, and instead returns a StatusBadRequest with a more generic decoding error. The more specific error is instead logged.

Before:
```
{"errors":[{"message":"json body could not be decoded: invalid character 'o' in literal null (expecting 'u')"}],"data":null}
```
After:
```
{"errors":[{"message":"json body could not be decoded"}],"data":null}
```

I have:

Solved:
https://github.com/99designs/gqlgen/issues/2430


StevenACoffman:
Hi. Rather than remove the displaying information from those who want it, can we instead look for a way to avoid having the hook be unable to parse it?

Do you have a small repository where you can reproduce the original problem?

koniuszy:
hey @StevenACoffman, thanks for the reply,

Initially I tried to control the error with [the hooks](https://gqlgen.com/reference/errors/#hooks) but to no avail. 
This could be controlled by a config flag or some sort of an error hook. At the moment it's a security leak and in my opinion it should be disabled by default asap.

If it does not persuade you, please point the direction so I could at least try implementing what's requested.

StevenACoffman:
Thanks! So there were actually multiple problems that this uncovered.
The main one was that decoding errors were not dispatched, so the hooks had no ability to intercept errors to present the errors as you wished. There was also a potential nil pointer dereference which would cause a panic, which would also bypass the error presenter hook (although the panic recover hook did catch that one).

Issue #2443, opened 2022-11-29
unnecessary entity resolvers generated for stub types in Federation 2


jdavisclark:
### What happened?
Given the scenario where you have a subgraph that needs to reference a federated type that the local subgraph does not otherwise define or contribute to, its common practice to define a stub type for it with nothing but the @key with resolvable=false, and the field of the type mapping to the key directive. In federation 1 the field mapping to the key directive had to be explicitly marked as `@external`, but in Federation 2 it does not.

With a valid Federation 2 schema, gqlgen is generating entity resolvers for these stub types even when `resolvable: false` is specified in their `@key` directives. This entity resolver is never utilized.

### What did you expect?
no entity resolver to be generated for the stub types.

### Minimal graphql.schema and models to reproduce
User subgraph/service:
```graphql
type User @key(fields: "id") {
  id: ID!
  name: String!
}
```

address subgraph/service:
```graphql
type Address @key(fields: "id") {
  id: ID!
  street1: String!
  city: String!
  state: String!

 user: User!
}

type User @key(fields: "id", resolvable: false) {
  id: ID!
}

type Query {
  address(id: String!): Address
}

```

query:
```graphql
query {
  address(id: "123) {
    street1,
    city,
    user {
      name
    }
  }
}
```

### versions
 - `go run github.com/99designs/gqlgen version`?
   - v0.17.20  
 - `go version`?
   -  go version go1.19.1 darwin/amd64

jdavisclark:
based on a little poking around, it looks like the `allFieldsAreExternal` implementation, or its caller, needs an update to deal with the scenario where federation == 2, the only fields defined on a type map to the `@key` directive, and resolvable==false in the `@key` directive.  

https://github.com/99designs/gqlgen/blob/master/plugin/federation/federation.go#LL326C15-L326C15

jdavisclark:
I stumbled on to #2387 while looking at this, which is related but ultimately different I think

Issue #2387, opened 2022-09-21
Federation: Entity type is not shareable / Resolver generated for non-resolvable @key


nimehner:
When defining two subgraphs implementing the same @key directive on the same type:

type SomeEntity @key(fields: "id", resolvable: false) {
  id: ID!
}

registering the generated schema in apollographql results in an error message:

UNKNOWN: Non-shareable field "Entity.findSomeEntityByID" is resolved from multiple subgraphs: it is resolved from subgraphs "schema1" and "schema2" and defined as non-shareable in all of them

There are two issues with this:
1. Since resolvable is false, no resolver should be generated 
2. If a resolver is generated, it probably should be marked shareable (in https://github.com/99designs/gqlgen/blob/master/plugin/federation/federation.go#L166 )

 - `go run github.com/99designs/gqlgen version`?
 v0.17.4

 - `go version`?
go version go1.19 darwin/arm64

nimehner:
I'd be happy if you could confirm if my understanding of this is correct. If so I could try to come up with a patch to address it.

jdavisclark:
I think I'm experiencing basically the same thing, except I get no errors re schema registry.
given this supergraph:


User subgraph:
```graphql
type User @key(fields: "id") {
  id: ID!
  name: String!
  email: String!
}
```

address subgraph:
```graphql
type Address @key(fields: "id") {
  id: ID!
  street1: String!
  city: String!
  state: String!

 user: User!
}

type User @key(fields: "id", resolvable: false) {
  id: ID!
}

type Query {
  address(id: String!): Address
}

```


gqlgen is creating a `func (r *entityResolver) FindCustomerByID(ctx context.Context, id string) (whatever, error)` entity resolver *in the address service subgraph*, which appears to be unnecessary given `resolvable: false`.

query for context:
```graphql
query {
  address(id: "12345") {
    id,
    city,
    customer: {
      name
    }
  }
}
```

jdavisclark:
ahh, so the deal is it looks like this just isn't 100% optimized for Federation 2 yet. 
in Federation 2 with this example, you **do not** have to declare the `id` of the `User` type as `@external`. However, the gqlgen implementation [only skips generating the entity resolver if all fields on your stub/reference typedef are marked as `@external` ](https://github.com/99designs/gqlgen/blob/master/plugin/federation/federation.go#L326)(which was required in federation 1).

[...]

PR #2449, opened 2022-12-05
Implicit external check

jclyons52:
Fixes https://github.com/99designs/gqlgen/issues/2443

As specified in the [Federation v2 documentation](https://www.apollographql.com/docs/federation/federation-2/moving-to-federation-2) The `@external` directive is no longer required for `@key` fields, but it is still required for [@requires](https://www.apollographql.com/docs/federation/entities-advanced/#contributing-computed-entity-fields) and [@provides](https://www.apollographql.com/docs/federation/entities-advanced/#using-provides). When fields are part of the key they should implicitly be considered `@external`. 

This PR updates the `allFieldsAreExternal` method on the entity to take a federation version parameter, if the verstion is set to 2 it checks if the relevant field is is listed in the `@key` directive. If it is, the field is considered external without the `@external` directive.

This PR also requires the `resolvable` argument to the `@key` directive to be set to false for it to consider any fields implicitly external. This is not entirely in line with [the spec](https://www.apollographql.com/docs/federation/subgraph-spec/) as it seems that if `resolvable` is false the entity should be excluded irrespective of what fields are external. In this PR, a less significant departure from the current behavior is favored over correctness as the change would potentially cause existing schemas to start failing.

This PR does not address reference types that have multiple keys defined. from the documentation it seems like a reference type should only use one key directive. All examples show entities with a single key, but the description is ambiguous:
>A subgraph that [references an entity without contributing any fields](https://www.apollographql.com/docs/federation/entities/#referencing-an-entity-without-contributing-fields) can include the fields of any @key in its stub definition

I have:


jclyons52:
This may also resolve https://github.com/99designs/gqlgen/issues/2387

jclyons52:
Nevermind the lack of multi key support, that seems like it's broadly an issue that will have to be addressed https://github.com/99designs/gqlgen/issues/1031

StevenACoffman:
Thanks so much! We would love any help on #1031 if you can!

jclyons52:
> Thanks so much! We would love any help on #1031 if you can!

No worries, I'll pick that up next 99time, which is every other Monday

jdavisclark:
awesome. thanks for getting this fixed!

PR #2452, opened 2022-12-07
Remove DO NOT EDIT in resolvergen plugin

atzedus:
Sometimes vscode warn about this while editing resolvers code.
Finally the resolver's code is editable and generated at the same time.


atzedus:
Resolves issues
#2452 
#2453

StevenACoffman:
Already in #2452

PR #2454, opened 2022-12-07
Ability to return multiple errors from resolvers raise than add it to stack.

atzedus:
Code detects gqlerror.List type returned from resolver and add all errors to stack.

I have:


StevenACoffman:
There's something dumb about the generation in Go 1.18 that is off by one space from the generation in Go 1.16. I am hoping GCP AppEngine releases a new version this week so we can leave Go 1.16 support behind.

giautm:
> There's something dumb about the generation in Go 1.18 that is off by one space from the generation in Go 1.16. I am hoping GCP AppEngine releases a new version this week so we can leave Go 1.16 support behind.

@StevenACoffman: I just cross the thread on Google Issues Track, They already support `go118` and `go119` runtimes in public preview: https://cloud.google.com/appengine/docs/standard/go/runtime#go-1.19-preview

Issue #2470, opened 2022-12-14
Incorrect response when errors occurred


atzedus:
### What happened?
When querying multiple fields at once and at least one of it returns error, then whole response become null with no any description where it `null` is.

### What did you expect?
List of errors and `data` with fields name queried.

### Minimal graphql.schema and models to reproduce

Schema:

```graphql
type Query {
    Kids: [ID!]
    Martials: [ID!]
}
```

Query:

```graphql
query Lists {
  	Kids
  	Martials
}
```

Result: 

```json
{
  "errors": [
    {
      "message": "some error",
      "path": [
        "Kids"
      ]
    }
  ],
  "data": null
}
```

What expected?: 

```json
{
  "errors": [
    {
      "message": "some error",
      "path": [
        "Kids"
      ]
    }
  ],
  "data": {
       "Kids": null,
       "Martials": [1, 2, 3, 4]
  }
}
```

### versions
 - `go run github.com/99designs/gqlgen v.0.17.22`?
 - `go 1.19`?

P.S. I will send pull request to fix this soon.

jwatte:
This is a major behavior change, breaking a lot of existing code.
Making this part of a bugfix version release is quite surprising.

pksieminski:
@atzedus are you sure that behaviour that you described in your issue is correct for the schema you listed?

I rolled back gqlgen to version `v0.17.22` and tested it with your schema:
```graphql
extend type Query {
  Kids: [ID!]
  Martials: [ID!]
}
```
Returned `nil, errors.New("some error")` in Kids resolver and received this response:
```json
{
  "errors": [
    {
      "message": "some error",
      "path": [
        "Kids"
      ],
    }
  ],
  "data": {
    "Kids": null,
    "Martials": ["1", "2", "3", "4"]
  }
}
```

To receive result as claimed in your issue, Kids list would have to be Non-Null type as follows for `KidsNN`:
```graphql
extend type Query {
  Kids: [ID!]
  KidsNN: [ID!]!
  Martials: [ID!]
}
```
Now with same `nil, errors.New("some error")` return value for KidsNN resolver, you will receive response:
```json
{
  "errors": [
    {
      "message": "some error",
      "path": [
        "Kids"
      ],
    },
    {
      "message": "some error",
      "path": [
        "KidsNN"
      ],
    }
  ],
  "data": null
}
```

And above behaviour is correct according to Graphql specification:
[...]

PR #2471, opened 2022-12-14
Fix issue #2470: Incorrect response when errors occurred

atzedus:
Fix issue #2470: Incorrect response when errors occurred

Old Result: 

```json
{
  "errors": [
    {
      "message": "some error",
      "path": [
        "Kids"
      ]
    }
  ],
  "data": null
}
```

After PR result: 

```json
{
  "errors": [
    {
      "message": "some error",
      "path": [
        "Kids"
      ]
    }
  ],
  "data": {
       "Kids": null,
       "Martials": [1, 2, 3, 4]
  }
}
```


atzedus:
My comment to commit 0855439dfb618a1386780f074bf85129d514e3ca

Why it removed?

resTmp is of type interface{} that returns from functions and then casted to interface{} as underlaying type, except "nil" will never pass check "resTmp == nil" because it interface with underlaying type of nil.

Explanations:

```go
package main

import "fmt"

func main() {
	fmt.Println(interface{}(nil) == nil) // true
	var s *string
	fmt.Println(interface{}(s) == nil) // false
	var s2 []*string
	fmt.Println(interface{}(s2) == nil) // false
	type s3type struct{}
	var s3 s3type
	fmt.Println(interface{}(s3) == nil) // false
	var s4 *s3type
	fmt.Println(interface{}(s4) == nil) // false
}
```

PR #2486, opened 2022-12-23
fix #2485 Defined type from a basic type should not need scalar

### LINKED ISSUE #2485: Defined type from a basic type should not need a mandatory scalar + MarshalFunc and UnmarshalFunc
### What happened?

In an existing Golang project where we are trying to add `gqlgen`, we have defined some types from basic types (string, int, boolean, float) that are being used in other structs. Currently, we need to write a `scalar` in the GraphQL schema and also do a `MarshalFunc` and `UnmarshalFunc` in order to avoid an error message saying `<type> is incompatible with <named_type>`.

image

### What did you expect?

No error message appear and no need to create a `scalar` with related `MarshalFunc` and `UnmarshalFunc` when `go generate`.

In a remote package, create some named basic types:
```go
package namedtypes

type (
	NamedString  string
	NamedInt     int
	NamedInt32   int32
	NamedInt64   int64
	NamedBool    bool
	NamedFloat64 float64
)
```
In another package, create a struct using these named basic types (might need to add this package to autobind):
```go
package usenamedtypes

import "github.com/99designs/gqlgen/integration/namedtypes"

type DummyUser struct {
	NewString  namedtypes.NamedString
	NewInt     namedtypes.NamedInt
	NewInt32   namedtypes.NamedInt32
	NewInt64   namedtypes.NamedInt64
	NewBool    namedtypes.NamedBool
	NewFloat64 namedtypes.NamedFloat64
}
```
In the GraphQL schema, create a type for the struct:
```graphql
type DummyUser {
    NewString: String!
    NewInt: Int!
    NewInt32: Int!
    NewInt64: Int!
    NewBool: Boolean!
    NewFloat64: Float!
}
```

Please note that a PR is coming to fix this.

### versions
 - `go run github.com/99designs/gqlgen version`? v0.17.22-dev
 - `go version`? go1.19.3 darwin/arm64

mstephano:
### Fixes #2485 : Defined type from a basic type should not require a scalar

Before this PR, running `go generate` when using a defined type from a basic type (string, int, boolean, float) that are then used in go struct will display an error message saying `<type> is incompatible with <named_type>`. We would then need to create a `scalar` with its related `MarshalFunc` and `UnmarshalFunc` to remove the error message. This will lead to many copy-pastes of the same MarshalFunc and UnmarshalFunc with only the type being changed.

After this PR gqlgen will support the following cases (see Go spec: https://go.dev/ref/spec#Type_declarations):
- GraphQL scalar with custom marshalling/unmarshalling
- a type definition from a basic type **without** a custom scalar and `MarshalFunc` and `UnmarshalFunc`
- an alias declaration **without** a custom scalar and `MarshalFunc` and `UnmarshalFunc`

I have:

### DIFF
--- a/_examples/scalars/schema.graphql
+++ b/_examples/scalars/schema.graphql
@@ -1,6 +1,6 @@
 type Query {
     user(id: ID!): User
-    search(input: SearchArgs = {location: "37,144", isBanned: false}): [User!]!
+    search(input: SearchArgs = { location: "37,144", isBanned: false }): [User!]!
     userByTier(tier: Tier!, darkMode: DarkMode!): [User!]!
 }
 
@@ -12,10 +12,13 @@ type User {
     valPrefs: DarkMode
     ptrPrefs: DarkMode
     isBanned: Banned!
+    isLoginBanned: Boolean!
+    isQueryBanned: Boolean!
     primitiveResolver: String!
     customResolver: Point!
     address: Address
     tier: Tier
+    carManufacturer: String!
 }
 
 type Address {
--- a/integration/testomitempty.graphql
+++ b/integration/testomitempty.graphql
@@ -2,7 +2,7 @@ type RemoteModelWithOmitempty {
     newDesc: String
 }
 
-type DummyUserWithRemoteNamedBasics {
+type DefinedTypeFromBasics {
     newString: String!
     newInt: Int!
     newInt32: Int!
--- a/integration/testomitempty.graphql
+++ b/integration/testomitempty.graphql
@@ -1,3 +1,25 @@
 type RemoteModelWithOmitempty {
     newDesc: String
 }
+
+type DummyUserWithRemoteNamedBasics {
+    newString: String!
+    newInt: Int!
+    newInt32: Int!
+    newInt64: Int!
+    newBool: Boolean!
+    newFloat64: Float!
+    newID: ID!
+
+    ### BELOW ARE NOT SUPPORTED AS BASIC - MUST CREATE SCALAR WITH MARSHALFUNC AND UNMARSHARLFUNC ###
+    #newInt8: Int!
+    #newInt16: Int!
+    #newFloat32: Float!
+    newUint: Uint! # scalar created for uint
+    #newUint8: Int!
+    #newUint16: Int!
+    #newUint32: Int!
+    #newUint64: Int!
+}
+
+scalar Uint
--- a/_examples/scalars/external/model.go
+++ b/_examples/scalars/external/model.go
@@ -1,3 +1,12 @@
 package external
 
-type ObjectID int
+type (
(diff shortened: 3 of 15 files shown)


_examples/scalars/schema.graphql
```diff
@@ -12,10 +12,13 @@ type User {
     valPrefs: DarkMode
     ptrPrefs: DarkMode
     isBanned: Banned!
+    isLoginBanned: Boolean!
+    isQueryBanned: Boolean!
```
mstephano, line 16: As per Go spec: https://go.dev/ref/spec#Type_declarations:
- isBanned is using scalar with custom marshalling/unmarshalling
- isLoginBanned is a type definition
- isQueryBanned is an alias declaration

Making sure my PR is working fine in any of these situations.

mstephano:
If you want me to rebase all my commits as one commit, let me know.

integration/testomitempty.graphql
```diff
@@ -1,3 +1,25 @@
 type RemoteModelWithOmitempty {
     newDesc: String
 }
+
+type DummyUserWithRemoteNamedBasics {
```
StevenACoffman, line 5: Can you change references like this from "RemoteNamedBasics" to DefinedTypeFromBasics?

integration/testomitempty.graphql
```diff
@@ -1,3 +1,25 @@
...
+    newString: String!
+    newInt: Int!
+    newInt32: Int!
+    newInt64: Int!
+    newBool: Boolean!
+    newFloat64: Float!
+    newID: ID!
+
+    ### BELOW ARE NOT SUPPORTED AS BASIC - MUST CREATE SCALAR WITH MARSHALFUNC AND UNMARSHARLFUNC ###
```
StevenACoffman, line 14: Is this comment still correct?

mstephano: Yes, for some reason, the ones below still has the message `is incompatible with`, and I am not sure why. I could spend some time to investigate if you want.

integration/testomitempty.graphql
mstephano: I updated the code and removed "DummyUserWith", just kept "DefinedTypeFromBasics".

integration/testomitempty.graphql
mstephano: Sorry, I did a quick test and the error message is not `..incompatible with..` but `Undefined type uint64` coming from `gqlparser`.

StevenACoffman:
Thanks so much for your work on this! If you could look into why those types require a scalar, I would appreciate a follow-up.

mstephano:
@StevenACoffman FYI, I found the issue with those types that require a scalar. I will send a PR soon after I finalize the tests.

PR #2488, opened 2022-12-28
plugin/resolvergen: respect named return values

### LINKED ISSUE #2417: Named return values in resolvers
### What happened?
Return values are not named by default and they also cannot be named manually, as stated in issue #1927.
```go
func (r *Resolver) FunctionName(ctx context.Context, input *model.something) (*model.somethingElse, error) {
```

### What did you expect?
Return values to be named possibly via the `gqlgen.yml` (by default they should be named like `output` and `err`).
It would allow the use of `defer` inside the resolver funcions allowing the use for a comprehensive external logging of the inputs, outputs and errors from the resolver function itself.
```go
func (r *Resolver) FunctionName(ctx context.Context, input *model.something) (output *model.somethingElse, err error) {
```

Any

### versions
 - `gqlgen version` v0.17.16
 - `go version` 1.19

a8m:
Fixed #2417

### DIFF
--- a/codegen/field.go
+++ b/codegen/field.go
@@ -3,6 +3,7 @@ package codegen
 import (
 	"errors"
 	"fmt"
+	goast "go/ast"
 	"go/types"
 	"log"
 	"reflect"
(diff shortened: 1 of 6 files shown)


a8m:
Websocket tests are flaky on windows. Previous failures:
https://github.com/99designs/gqlgen/actions/runs/3645986386/jobs/6156627161

Issue #329, opened 2018-09-01
Subscriptions via Server-Sent Events


ssttevee:
I've been working on a PR that adds SSE support for subscriptions, but I noticed that #317 mentions that GET requests should be blocked for subscriptions as well as mutations.

So, I'm wondering, since SSE requests use the GET verb, is this something that was previously decided to be not supported or something that was overlooked when that message was posted?

Sach97:
@ssttevee do you plan to listen to Postgresql notifications over websockets like in this [example](http://coussej.github.io/2015/09/15/Listening-to-generic-JSON-notifications-from-PostgreSQL-in-Go/)?

vektah:
@ssttevee Interesting, Is there a spec for SSE yet?

The http spec is pretty clear on not doing mutations in GET requests https://github.com/APIs-guru/graphql-over-http

ssttevee:
@Sach97 I'm actually using it with events from an MQTT server.

@vektah afaik, there is not, however, I feel like SSE is actually an overall better fit for subscriptions over http(s) since its uni-directional nature matches how graphql subscriptions work. Additional, when combined with http2's multiplexing ability, the whole subscription id situation can be offloaded to the transport level while still using a single connection.

gedw99:
IE and edge don't support SSE, according to the web site you linked to.

Are you aware of this ?

vektah:
I'm going to close this until a standard is built around this. 

You should be able to write your own transport using whatever you like, and ditch the handler package.

frederikhors:
@ssttevee any news on this?

ssttevee:
@frederikhors I'm keeping patched version on my own fork, but I don't think a pr will be accepted

frederikhors:
> @frederikhors I'm keeping patched version on my own fork, but I don't think a pr will be accepted

Maybe a good PR with good config options will...

vektah:
Sorry if I've given off the wrong impression here. gqlgen is first and foremost open source software, if there is something people want, and it cant easily be implemented outside of core we should either make it possible to do or implement it internally.

My only concern is if there isn't a stable spec there isn't any interoperability. This is one of those I just don't know enough to make an informed decision, is https://github.com/CodeCommission/subscriptions-transport-sse the defacto standard? 

Are you implementing your own client code or is this something supported widely?

frederikhors:
> Sorry if I've given off the wrong impression here.

[...]

PR #2498, opened 2023-01-06
Add Server-Sent Events transport

jmic:
Inspired by the feature request [#329](https://github.com/99designs/gqlgen/issues/329) and several [articles](https://hackernoon.com/graphql-subscriptions-using-ssefetch-over-websockets) about GraphQL subscriptions via Server-Sent Events, I started to implement an SSE transport that follows this [protocol specification](https://github.com/enisdenjo/graphql-sse/blob/master/PROTOCOL.md) and is compatible with the [graphql-sse](https://github.com/enisdenjo/graphql-sse) library at the _distinct connections mode_.
The result is this PR. Besides tests and documentation, I added an SSE option to the chat example as well. Please try it.

Remarks: I didn't want to change the `Supports(r *http.Request)` functions of the present transports to add a filter that rejects all `text/event-stream` requests, instead the order in which the transport  `Supports(r *http.Request)` are called is essential. This can of course be solved the other way around if desired.

Many thanks to the developers of gqlgen for this great library!

### DIFF
--- a/graphql/handler/transport/sse.go
+++ b/graphql/handler/transport/sse.go
@@ -1,0 +1,108 @@
+package transport
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/99designs/gqlgen/graphql"
+	"github.com/vektah/gqlparser/v2/gqlerror"
+	"io"
+	"log"
+	"mime"
+	"net/http"
+	"strings"
+)
+
+type SSE struct{}
+
+var _ graphql.Transport = SSE{}
+
+func (t SSE) Supports(r *http.Request) bool {
+	if !strings.Contains(r.Header.Get("Accept"), "text/event-stream") {
+		return false
+	}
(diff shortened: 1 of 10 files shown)


StevenACoffman:
Thanks so much! This is a great alternative to websockets for subscriptions!

graphql/handler/transport/sse.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (t SSE) Supports(r *http.Request) bool {
+	if !strings.Contains(r.Header.Get("Accept"), "text/event-stream") {
+		return false
+	}
+	mediaType, _, err := mime.ParseMediaType(r.Header.Get("Content-Type"))
+	if err != nil {
+		return false
+	}
+	return r.Method == http.MethodPost && mediaType == "application/json"
```
derekwilling, line 29: The graphql-sse protocol states that the method for SSE subscriptions should be PUT.

https://github.com/enisdenjo/graphql-sse/blob/master/PROTOCOL.md#making-reservations

Issue #2502, opened 2023-01-09
Proposal: Add omit_complexity config flag to skip generation of the body of ComplexityRoot struct and Complexity function


mariuscraciunescu:
### What happened?
When GraphQL schema is really large (~60.000+ lines), the code generated with **gqlgen** fails to compile with the following error:

`<autogenerated>:1: internal compiler error: NewBulk too big: nbit=13244 count=1469951 nword=414 size=608559714`
More details about this compiler issue can be found here: https://github.com/golang/go/issues/48375, https://github.com/golang/go/issues/33437

This compilation error is caused by:
`type ComplexityRoot struct { ... }`
`func (e *executableSchema) Complexity(typeName, field string, childComplexity int, rawArgs map[string]interface{}) (int, bool) { ... }`

### What did you expect?
Code should compile successfully, regardless of the GraphQL schema size.

### versions
 - v0.17.22

### Proposed solution
Introduce a new config flag `omit_complexity`, which when set to `true` will skip the generation of the body of these constructs.

The resulting generated code will be:
```
type ComplexityRoot struct {
  // empty struct
}
```

```
func (e *executableSchema) Complexity(typeName, field string, childComplexity int, rawArgs map[string]interface{}) (int, bool) {
  ec := executionContext{nil, e}
  _ = ec

  // switch statement is not generated

  return 0, false
}
```

This proposal is backwards compatible, won't introduce any breaking changes.

PR #2504, opened 2023-01-12
Add omit_complexity config option for issue #2502

mariuscraciunescu:
Full description in issue #2502 

Introduced a new config flag `omit_complexity`, which when set to `true` will skip the generation of the body of the following:
```
type ComplexityRoot struct {
  // empty struct
}
```

```
func (e *executableSchema) Complexity(typeName, field string, childComplexity int, rawArgs map[string]interface{}) (int, bool) {
  ec := executionContext{nil, e}
  _ = ec

  // switch statement is not generated

  return 0, false
}
```

Fixes the following compiler issue when GraphQL schema is too large: 
`<autogenerated>:1: internal compiler error: NewBulk too big: nbit=13244 count=1469951 nword=414 size=608559714`


StevenACoffman:
Thanks!

PR #2508, opened 2023-01-14
fix #2485 for some types requiring a scalar

### LINKED ISSUE #2485: Defined type from a basic type should not need a mandatory scalar + MarshalFunc and UnmarshalFunc

mstephano:
Following PR # [2486](https://github.com/99designs/gqlgen/pull/2486), some types require a scalar to work properly. This PR fixes some of them except `float32` and `uint64`. The reasons are as follow:

- float32: the precision changed when marshalling/unmarshalling from float64 back to float32. As an example, value of 5.76 comes out as 5.760000228881836. Anyone knows how to fix this? You can see this [branch](https://github.com/mstephano/gqlgen/tree/fix/float32) for an example, or precisely this [commit](https://github.com/mstephano/gqlgen/commit/9ee30a5d9897a82b3a9db7fb752504530e05613d).
- uint64: it is bigger than int64

I have:

### DIFF
--- a/codegen/config/binder.go
+++ b/codegen/config/binder.go
@@ -257,6 +257,30 @@ func (t *TypeReference) IsUnderlyingBasic() bool {
 	return isUnderlyingBasic
 }
 
+func (t *TypeReference) IsUnusualBasic() bool {
+	if basic, isBasic := t.GO.(*types.Basic); isBasic {
+		switch basic.Kind() {
+		case types.Int8, types.Int16, types.Uint, types.Uint8, types.Uint16, types.Uint32:
+			return true
+		default:
+			return false
+		}
+	}
+	return false
+}
+
+func (t *TypeReference) IsUnderlyingUnusualBasic() bool {
+	if basic, isUnderlyingBasic := t.GO.Underlying().(*types.Basic); isUnderlyingBasic {
+		switch basic.Kind() {
+		case types.Int8, types.Int16, types.Uint, types.Uint8, types.Uint16, types.Uint32:
+			return true
+		default:
+			return false
+		}
+	}
+	return false
+}
+
 func (t *TypeReference) IsScalarID() bool {
 	return t.Definition.Kind == ast.Scalar && t.Marshaler.Name() == "MarshalID"
 }
--- a/_examples/scalars/external/model.go
+++ b/_examples/scalars/external/model.go
@@ -3,6 +3,7 @@ package external
 type (
 	ObjectID     int
 	Manufacturer string // remote named string
+	Count        uint32 // remote named uint32
 )
 
 const (
--- a/_examples/scalars/generated.go
+++ b/_examples/scalars/generated.go
@@ -61,6 +61,8 @@ type ComplexityRoot struct {
 	User struct {
 		Address           func(childComplexity int) int
 		CarManufacturer   func(childComplexity int) int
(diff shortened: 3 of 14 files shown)


codegen/config/binder.go
```diff
@@ -257,6 +257,30 @@ func (t *TypeReference) IsUnderlyingBasic() bool {
 	return isUnderlyingBasic
 }
 
+func (t *TypeReference) IsUnusualBasic() bool {
```
mstephano, line 260: I named these as `Unusual` but a suggestion for a better name is welcomed.

mojtabacazi:
This commit actually broke our app, it generates invalid code when trying to unmarshal typed scalars. It evens breaks `time.Duration` types to name a few. Here is an example:

```
func (ec *executionContext) unmarshalNInt642timeᚐDuration(ctx context.Context, v interface{}) (time.Duration, error) {
	res, err := graphql.UnmarshalInt64(v)
}
```
which causes this error:

```
cannot use res (variable of type int64) as type time.Duration in return statement
```

Here is another example with a pointer type

```
func (ec *executionContext) unmarshalOBoolean2DummyFilter(ctx context.Context, v interface{}) (*dummy.DummyFilter, error) {
	if v == nil {
		return nil, nil
	}
	res, err := graphql.UnmarshalBoolean(v)
	return &res, graphql.ErrorOnPath(ctx, err)
}
```

StevenACoffman:
@mojtabacazi Did #2528 fix your issue?

StevenACoffman:
also related: #2486

mojtabacazi:
> @mojtabacazi Did #2528 fix your issue?

Nope

StevenACoffman:
@mojtabacazi I reverted this, so please see if v0.17.27 resolves your issue!

PR #2519, opened 2023-01-23
remove extra call to packages.Load fix #2505

### LINKED ISSUE #2505: "reloading module info" doesn't work
### What happened?
I see "reloading module info" 3 times when we generate our config, so I did a little test to see if it actually helps with anything. It doesn't appear to. It seems to just cause unnecessary slowness.

```diff
diff --git a/internal/code/packages.go b/internal/code/packages.go
index c800d3d8..bb92c802 100644
--- a/internal/code/packages.go
+++ b/internal/code/packages.go
@@ -71,6 +71,9 @@ func (p *Packages) LoadAll(importPaths ...string) []*packages.Package {
                        fmt.Println("reloading module info")
                        pkgs, err = packages.Load(&packages.Config{Mode: mode}, missing...)
                }
+               if !p.checkModuleLoaded(pkgs) {
+                       fmt.Println("waste of time")
+               }
 
                if err != nil {
                        p.loadErrors = append(p.loadErrors, err)
```

### What did you expect?
```
reloading module info
reloading module info
reloading module info
```

What I saw:
```
reloading module info
waste of time
reloading module info
waste of time
reloading module info
waste of time
```

our config is massive, sorry...

### versions
 - `go run github.com/99designs/gqlgen version`? 0.17.22
 - `go version`? 1.19

Removing this "reloading" reduces the run time for us from 38s to 26s

vikstrous:
This removes a slow and seemingly unnecessary call to packages.Load

I don't know how to test this because it's not clear what it's solving and how to reproduce the issue. In our codebase nothing bad happens, it just runs faster.

I have:
 - [N/A] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

### DIFF
--- a/internal/code/packages.go
+++ b/internal/code/packages.go
@@ -38,15 +38,6 @@ func (p *Packages) ReloadAll(importPaths ...string) []*packages.Package {
 	return p.LoadAll(importPaths...)
 }
 
-func (p *Packages) checkModuleLoaded(pkgs []*packages.Package) bool {
-	for i := range pkgs {
-		if pkgs[i] == nil || pkgs[i].Module == nil {
-			return false
-		}
-	}
-	return true
-}
-
 // LoadAll will call packages.Load and return the package data for the given packages,
 // but if the package already have been loaded it will return cached values instead.
 func (p *Packages) LoadAll(importPaths ...string) []*packages.Package {
@@ -65,13 +56,6 @@ func (p *Packages) LoadAll(importPaths ...string) []*packages.Package {
 	if len(missing) > 0 {
 		p.numLoadCalls++
 		pkgs, err := packages.Load(&packages.Config{Mode: mode}, missing...)
-
-		// Sometimes packages.Load not loaded the module info. Call it again to reload it.
-		if !p.checkModuleLoaded(pkgs) {
-			fmt.Println("reloading module info")
-			pkgs, err = packages.Load(&packages.Config{Mode: mode}, missing...)
-		}
-
 		if err != nil {
 			p.loadErrors = append(p.loadErrors, err)
 		}


StevenACoffman:
@vikstrous Can you verify that this does not cause an issue when following the example documentation and doing the `go run github.com/99designs/gqlgen init` ? 

This code was added to fix https://github.com/99designs/gqlgen/issues/1765 
in https://github.com/99designs/gqlgen/pull/1773 

it is possible that it is not still necessary but I would like you to verify that. If it is necessary, perhaps there is a better way to fix the original problem that doesn't cause the redundant load for every other situation.

vikstrous2:
The [quick start](https://gqlgen.com/#quick-start) steps work with and without this PR.

If I comment out the autobind line and/or add one that points to a package that doesn't exist, `generate` fails. With and without this PR.

This is the error:
```
reloading module info
generating core failed: unable to load example/graph/model - make sure you're using an import path to a package that exists
exit status 1
```

There was a test added in the linked PR and all tests are passing on this PR. I don't know which test suite it's in and there are a lot of them, but I'm guessing that whatever tests were added there are still running.

I'm really curious about "or just call packages.Load() twice" from https://github.com/99designs/gqlgen/pull/1767. I don't understand how calling load twice changes anything. It seems like @ipfans knows something I don't?

---
Re: the other reloads... It seems like all the extra calls to load are related to changes to packages caused by mixing up input and output packages. I want to dig into this more and see if I can remove more calls to load. Source: https://github.com/99designs/gqlgen/issues/1762 Maybe just making all calls to "ReloadAllPackages" more specific will do what I want. I'll look at that next.

StevenACoffman:
Great! I appreciate your digging into this!

PR #2525, opened 2023-01-25
fix similar basic kind bytes edge case

### LINKED ISSUE #2524: BasicTypes "Byte" Binding for Scalars fails
### What happened?

Since https://github.com/99designs/gqlgen/pull/2486 some common types don't get binded correctly

### What did you expect?
Bytes to be binded to []bytes.

What happens that in this new function introduced in #2486 
https://github.com/99designs/gqlgen/blob/9d22d98c792ba7214dc1aad4366e3f7eba0299f7/internal/code/compare.go#L167

Converts some basic kinds to "similar" types which makes sense in some cases as gqlgen converts Int into multiple types of int.

But in this case where we do the check in :

https://github.com/99designs/gqlgen/blob/9d22d98c792ba7214dc1aad4366e3f7eba0299f7/internal/code/compare.go#L43

It will fail as similar will convert byte -> int64 != byte.

```
scalar Bytes

type SomeBytesType {
   value: Bytes
}
```

And we add to the gqlgen the scalar + Unmarshal

### versions
 - `go run github.com/99designs/gqlgen version`? 0.17.24
 - `go version`? 1.18

roneli:
Found and issue from v0.17.24+ with binding bytes type

Fixes #2524

I currently went for a quick fix / show as draft for what the issue is as describe in #2524. 
Another solution is to add a check in the compare to check for simple kind before going for the similar. This approach is cleaner as I less liked adding the if bytes return check in the function. Added a test to show the failure.

```go
if actualBasic.Kind() != expected.Kind() &&  similarBasicKind(actualBasic.Kind()) != expected.Kind() {
		return fmt.Errorf("basic kind differs, %s != %s", expected.Name(), actualBasic.Name())
	}
```

@mstephano if you can weigh in, as you wrote the original fix, which is great! so we don't miss any edge case.

I have:

### DIFF
--- a/internal/code/compare_test.go
+++ b/internal/code/compare_test.go
@@ -63,6 +63,25 @@ func TestCompatibleTypes(t *testing.T) {
 	}
 }
 
+func TestSimilarBasicKind(t *testing.T) {
+	valid := []struct {
+		name     string
+		expected types.BasicKind
+		actual   types.BasicKind
+	}{
+		{name: "string=string", expected: types.String, actual: types.String},
+		{name: "int8=int", expected: types.Int64, actual: types.Int8},
+		{name: "byte=byte", expected: types.Byte, actual: types.Byte},
+	}
+
+	for _, tc := range valid {
+		t.Run(tc.name, func(t *testing.T) {
+			require.Equal(t, tc.expected, similarBasicKind(tc.actual))
+
+		})
+	}
+}
+
 func parseTypeStr(t *testing.T, s string) types.Type {
 	t.Helper()
 
--- a/internal/code/compare.go
+++ b/internal/code/compare.go
@@ -165,6 +165,9 @@ func CompatibleTypes(expected types.Type, actual types.Type) error {
 }
 
 func similarBasicKind(kind types.BasicKind) types.BasicKind {
+	if kind == types.Byte {
+		return kind
+	}
 	switch kind {
 	case types.Int8, types.Int16:
 		return types.Int64


mstephano:
Hi @roneli, I will investigate your issue right away, brb.

internal/code/compare_test.go
```diff
@@ -63,6 +63,25 @@ func TestCompatibleTypes(t *testing.T) {
...
+	}
+
+	for _, tc := range valid {
+		t.Run(tc.name, func(t *testing.T) {
+			require.Equal(t, tc.expected, similarBasicKind(tc.actual))
+
+		})
+	}
+}
+
```
mstephano, line 84: I upgraded your test with the following to include all `BasicKind` defined by [Go](https://github.com/golang/go/blob/a11b8e37652ff60302b4a4a55a8a43db6a066ecf/src/go/types/basic.go#L10), and because `Byte` is an alias to `Uint8`, your code in `similarBasicKind()` is failing in test `Uint8=Int64`. I am trying to find a way to make it work.

```Go
func TestSimilarBasicKind(t *testing.T) {
	valid := []struct {
		name     string
		expected types.BasicKind
		actual   types.BasicKind
	}{
		// invalid
		{name: "Invalid=Invalid", expected: types.Invalid, actual: types.Invalid},

		// predeclared types
		{name: "Bool=Bool", expected: types.Bool, actual: types.Bool},
		{name: "Int=Int", expected: types.Int, actual: types.Int},
		{name: "Int8=Int64", expected: types.Int64, actual: types.Int8},
		{name: "Int16=Int64", expected: types.Int64, actual: types.Int16},
		{name: "Int32=Int32", expected: types.Int32, actual: types.Int32},
		{name: "Int64=Int64", expected: types.Int64, actual: types.Int64},
		{name: "Uint=Int64", expected: types.Int64, actual: types.Uint},
		{name: "Uint8=Int64", expected: types.Int64, actual: types.Uint8},
		{name: "Uint16=Int64", expected: types.Int64, actual: types.Uint16},
		{name: "Uint32=Int64", expected: types.Int64, actual: types.Uint32},
		{name: "Uint64=Int64", expected: types.Uint64, actual: types.Uint64},
		{name: "Uintptr=Uintptr", expected: types.Uintptr, actual: types.Uintptr},
		{name: "Float32=Float32", expected: types.Float32, actual: types.Float32},
		{name: "Float64=Float64", expected: types.Float64, actual: types.Float64},
		{name: "Complex64=Complex64", expected: types.Complex64, actual: types.Complex64},
		{name: "Complex128=Complex128", expected: types.Complex128, actual: types.Complex128},
		{name: "String=String", expected: types.String, actual: types.String},
		{name: "UnsafePointer=UnsafePointer", expected: types.UnsafePointer, actual: types.UnsafePointer},

		// types for untyped values
		{name: "UntypedBool=UntypedBool", expected: types.UntypedBool, actual: types.UntypedBool},
		{name: "UntypedInt=UntypedInt", expected: types.UntypedInt, actual: types.UntypedInt},
		{name: "UntypedRune=UntypedRune", expected: types.UntypedRune, actual: types.UntypedRune},
		{name: "UntypedFloat=UntypedFloat", expected: types.UntypedFloat, actual: types.UntypedFloat},
		{name: "UntypedComplex=UntypedComplex", expected: types.UntypedComplex, actual: types.UntypedComplex},
		{name: "UntypedString=UntypedString", expected: types.UntypedString, actual: types.UntypedString},
		{name: "UntypedNil=UntypedNil", expected: types.UntypedNil, actual: types.UntypedNil},

		// aliases
		{name: "Byte=Byte", expected: types.Byte, actual: types.Byte},
		{name: "Rune=Rune", expected: types.Rune, actual: types.Rune},
	}

	for _, tc := range valid {
		t.Run(tc.name, func(t *testing.T) {
			require.Equal(t, tc.expected, similarBasicKind(tc.actual))
		})
	}
}
```

roneli: Thanks for the reply, yes that would be perfect to check them all, but I think that my suggestion to use this check in the comparison before the method:

```go
if actualBasic.Kind() != expected.Kind() &&  similarBasicKind(actualBasic.Kind()) != expected.Kind() {
		return fmt.Errorf("basic kind differs, %s != %s", expected.Name(), actualBasic.Name())
	}
```

would be simpler and wouldn't require the test.

mstephano: Yes, I am testing your suggestion by adding some tests in `_examples/scalars`. Can you provide me with your `MarshalBytes/UnmarshalBytes` functions?

mstephano: Sorry, just found out about this [file](https://github.com/99designs/gqlgen/blob/master/codegen/testserver/singlefile/bytes.go), so I will use it for my tests.

mstephano:
@roneli I just pushed a branch that includes some tests on `byte` and `rune` following your proposition (thank you for that). You can view this [branch](https://github.com/mstephano/gqlgen/tree/fix/byte) or this [commit](https://github.com/mstephano/gqlgen/commit/359223d46cc9168da4931614c7bacc2d0616ab84).

Feel free to copy over to your PR, or let me know if you just want me to create a PR using my branch.

roneli:
Closing in favor of #2528. I'll review it as well + verify it fixes the issues.

I will note as well, will verify on your branch that defining for example
```
scalar StringMap # map[string]string
```

Will also fail on basic types as the generated.go requires a casting of the secondary value to the simple kind, I overcame this by forcing a resolver.

PR #2528, opened 2023-01-26
fix #2524 basic alias Byte was not binded properly

### LINKED ISSUE #2524: BasicTypes "Byte" Binding for Scalars fails

mstephano:
Issue from v0.17.24+ with binding bytes type

Fixes #2524 for basic [alias](https://github.com/golang/go/blob/a11b8e37652ff60302b4a4a55a8a43db6a066ecf/src/go/types/basic.go#L45) `Byte` (proposition from @roneli). Also added tests to make sure binding works properly for:
- basic alias `Rune`
- defined type for []byte
- defined type for []rune

I have:

### DIFF
--- a/_examples/scalars/.gqlgen.yml
+++ b/_examples/scalars/.gqlgen.yml
@@ -22,3 +22,7 @@ models:
     model: "github.com/99designs/gqlgen/codegen/testserver/singlefile.Bytes"
   Runes:
     model: "github.com/99designs/gqlgen/_examples/scalars/model.Runes"
+  DefinedTypeBytes:
+    model: "github.com/99designs/gqlgen/_examples/scalars/external.Bytes"
+  DefinedTypeRunes:
+    model: "github.com/99designs/gqlgen/_examples/scalars/external.Runes"
--- a/_examples/scalars/.gqlgen.yml
+++ b/_examples/scalars/.gqlgen.yml
@@ -18,3 +18,7 @@ models:
     model: github.com/99designs/gqlgen/_examples/scalars/model.Banned
   DarkMode:
     model: github.com/99designs/gqlgen/_examples/scalars/model.Preferences
+  Bytes:
+    model: "github.com/99designs/gqlgen/codegen/testserver/singlefile.Bytes"
+  Runes:
+    model: "github.com/99designs/gqlgen/_examples/scalars/model.Runes"
--- a/_examples/scalars/external/model.go
+++ b/_examples/scalars/external/model.go
@@ -3,7 +3,7 @@ package external
 type (
 	ObjectID     int
 	Manufacturer string // remote named string
-	Count        uint32 // remote named uint32
+	Count        uint8  // remote named uint8
 )
 
 const (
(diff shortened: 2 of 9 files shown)


mstephano:
Should I add some integration tests too to increase the coverage?

roneli (commented):
Looks great,  will add as per my comment if we define something like this:

```graphql
scalar StringMap

type SomeStruct {
  field: StringMap
}
```

```go
func MarshalStrMap(val map[string]string) graphql.Marshaler {
	return graphql.WriterFunc(func(w io.Writer) {
		err := json.NewEncoder(w).Encode(val)
		if err != nil {
			panic(err)
		}
	})
}

func UnmarshalStrMap(v interface{}) (map[string]string, error) {
	if m, ok := v.(map[string]string); ok {
		return m, nil
	}

	return nil, fmt.Errorf("%T is not a map", v)
}
```

And we have a struct that is 

```go
type NamedStringType string

type SomeStruct struct{
   Field map[string]NamedStringType
}

```

The generated go will try to use the unrmarshal of the string map but it won't work, since we need to convert the named type to pass to the unmarshaler.

I am trying to figure out in the generator template how to check and fix it.

We can maybe do this in a different PR / I will open an issue about this later today.

mstephano:
@roneli I just tried your code, can you replace `map[string]string` by `map[string]NamedStringType` in your `marshal/unmarshal` func and let me know if that works.

roneli:
> @roneli I just tried your code, can you replace `map[string]string` by `map[string]NamedStringType` in your `marshal/unmarshal` func and let me know if that works.

Yes that will work, but then I need to create an un/marshal for every named type. What I did was force a resolver on the field and convert there. Basically this error will occur if its an external map and you auto bind. (I am doing a lot of auto binding with many many go structs). In that case my simple StringMap isn't enough and I need to create a special map per type.

mstephano:
@roneli Can you create a new issue for the map and describe the improvement you need with code example? I will take a look into it and we can keep our conversation there.

This PR is really to fix for `Byte` for anyone who use it.

StevenACoffman:
Thanks for the fix, and I really appreciate you continuing to work on these issues!

PR #2584, opened 2023-03-15
POST transport: missing return and unnecessary logs

Desuuuu:
The HTTP POST transport was missing a return in the first error case.

I also removed the two `log.Printf` calls which seem to have been left there unintentionally.

I have:


StevenACoffman:
Thanks!

Issue #1416, opened 2020-12-12
Final solution for the eternal null VS undefined problem


frederikhors:
I think this topic could win the award as the most debated topic on web all along
(since that damn day `null` was invented, at least :smile:)!

I would like to summarize here everything I have read and tried on the subject, **in the hope that Santa Claus will give us the robust solution we deserve** this year (at least for gqlgen).

### THE PROBLEM

As a developer I would like a quick way to identify in my resolvers if the user has deliberately chosen `null` as the value of the field (string/int/time/...) or if he is using sparse updates and has omitted those fields entirely.

### PROPOSALS READ AROUND

1. the official solution proposed by Gqlgen team (@vektah at least), use `map[string]interface{}` (https://gqlgen.com/reference/changesets/); but of course there are [several counter-arguments](https://github.com/99designs/gqlgen/issues/866#issuecomment-737413495)

1. nullable types (proposed many times, [one for all here](https://github.com/99designs/gqlgen/issues/505#issuecomment-457000904)); but as @jszwedko said [it seems to not work either](https://github.com/99designs/gqlgen/issues/505#issuecomment-475163747)

1. checking ArgumentMap from context [as suggested here]()https://github.com/99designs/gqlgen/issues/505#issuecomment-586584763; the solution I'm using although it is very verbose and I need things like code generation

1. [methods for setters](https://github.com/99designs/gqlgen/issues/506) proposed by the amazing @danilobuerger which is still a WIP

### THREADS ON THE TOPIC

1. https://github.com/99designs/gqlgen/issues/505

1. https://github.com/99designs/gqlgen/issues/506

1. https://github.com/99designs/gqlgen/issues/977

1. https://github.com/99designs/gqlgen/issues/866

1. https://github.com/99designs/gqlgen/issues/646

### WE ARE NOT ALONE

1. https://webdevstation.com/post/Work-with-nullable-strings-in-gqlgen-GraphQL-Server-0x4e28

1. https://www.calhoun.io/how-to-determine-if-a-json-key-has-been-set-to-null-or-not-provided/

1. https://github.com/guregu/null/issues/39

1. https://github.com/graphql-go/graphql/pull/401

1. https://github.com/graphql-go/graphql/issues/178

1. https://github.com/datastax/cassandra-data-apis/issues/6

1. https://github.com/graph-gophers/graphql-go/issues/210

1. https://github.com/graphql-rust/juniper/issues/183

1. https://github.com/graphql-rust/juniper/issues/108

[...]

PR #2585, opened 2023-03-18
Omittable input fields

Desuuuu:
This PR adds the `Omittable` type to the `graphql` package. Its purpose is distinguishing between omitted and null input fields and it has the following API:

```go
type Omittable[T any] struct {
	// ...
}

func (o Omittable[T]) Value() T {
	if !o.set {
		var zero T
		return zero
	}
	return o.value
}

func (o Omittable[T]) ValueOK() (T, bool) {
	if !o.set {
		var zero T
		return zero, false
	}
	return o.value, true
}

func (o Omittable[T]) IsSet() bool {
	return o.set
}
```

The type is meant to be bound with nullable input fields.

Model generation is able to generate omittable fields by:
* Using the `@goField(omittable: true)` directive on a nullable input field.
* Setting the `nullable_input_omittable` configuration option to `true`. This causes nullable input field to be omittable by default.

The `goField` directive always has priority over the `nullable_input_omittable` option meaning you can have nullable input fields default to being omittable, and override it with `@goField(omittable: false)`.

Relevant issue: https://github.com/99designs/gqlgen/issues/1416

I have:


StevenACoffman:
Cool! I would love to get some other people to look it over and take it for a spin.

carldunham:
This is great! Would love to see some tests using `input` fields in addition to `String`. There may be some more complexity there around using `omitempty` in the json tags or something. Didn't debug, but was playing around some and saw some issues: https://github.com/carldunham/gqlgen-optional-test/tree/cd/gqlgen-2585/testing

carldunham:
Ah, got it. Would still like to see those tests :)

StevenACoffman:
I generally prefer `*Optional` over `Optional` because I prefer pointers for every struct type, but would be open to hearing arguments the other way.

carldunham:
For small structs like this, meant to be embedded, it can be nice to avoid the indirection. But I would prefer nil meaning “not set” over having a bool field for that. CarlOn Mar 21, 2023, at 13:34, Steve Coffman ***@***.***> wrote:﻿
I generally prefer *Optional over Optional because I prefer pointers for every struct type, but would be open to hearing arguments the other way.

oiime:
@Desuuuu 

> The name is not in any way linked to GraphQL, it's simply a Go type that fields can be bound to. It will not appear anywhere in the schema. Maybe we can find a better name for a value that can be undefined, null or set though?

Yeah, that's what I mean, it can get conflated with something else, that is, people don't read docs in depth and might confuse this as a way to have "optional" input fields, maybe `Undefined` or `Omittable` might work better, just my initial gut feeling looking at the name.

> I'm thinking the type could also be:
```
type Optional[T any] struct {
	Value   *T
	Defined bool
}
```

might get a bit messy with mixed pointer and non-pointer T no? imo your current type implementation is less opinionated about the underlying type it handles which is good

oiime:
@Desuuuu It does always generate pointer types, but that assumes people only use modelgen models, in a lot of cases (mine included) this is not how it is used, types are created somewhere else and often represent null following `sql.NullString` style rather than `*string`.

oiime:
@Desuuuu 

> I'm not too familiar with SQL in Go, but if the intent is to use the type directly, is there a way to make the driver work correctly with Optional[sql.NullString] in any case? By that I mean making it skip fields that are undefined. I might be wrong but it doesn't seem possible, at least with sql.Valuer.

This is not really about SQL, NullType is a common way to represent null instead of *Type, I don't use it for SQL in my case but rather interacting with another legacy system that uses this approach. The purpose is not to pass onward Optional[NullString] but rather to be able to extract all types the same way, meaning both Optional[*string] and Optional[NullString] can be converted respectively into *string and NullString without caring for how the underlying mechanism works

> If that was possible and a use-case we wanted to support, there would be pretty much no reason to use sql.NullString instead of string since Optional could be made to skip undefined fields and to serialize nil as NULL.

Imo there are just too many assumptions there about the behavior and limitations down the line, I'd much rather fit gqlgen to the rest of the system, not vice versa. Up to you if you consider it a use-case you want to support.

> Besides that, you can already differentiate between null and undefined for custom types that have an unmarshaler!

How can that be done without using this Optional implementation? an umarshaler would simply not be called if it is not defined, meaning we'd need to create a custom optional wrapper for every type separately eg: 

```
type OptionalNullString struct {
	Value 	NullString
	Defined bool
}

func UnmarshalOptionalNullString(v interface{}) (OptionalNullString, error) {
	if v == nil {
		return OptionalNullString{Defined: true}
	}
	return OptionalNullString{Defined: true, NullString{Valid: true, String: "parse v"}}
}
```

carldunham:
> > But I would prefer nil meaning “not set” over having a bool field for that.

> It's not really the value that's nil in the *Optional case though, it's the wrapper itself since you can have a value that's nil and defined.

Right, so the nil *Optional can mean not provided, and a nil *T would mean a null value.

I'm not arguing in favor, necessarily, but it seems to clearly reflect the semantics of a "nullable" input field (or argument?) that can be provided or not, and where that is meaningful.

FWIW, I prefer avoiding terms like "optional" and "undefined", as they don't clearly reflect GraphQL semantics. Those are terms most often used to map gql concepts into language-specific ones, as I've seen them.

Which is why I'm a bit hesitant about this discussion at all, but I'm aware that in the real world there are implementations that distinguish between "set" and "null" for things like partial object updates. I've provided PRs for that use case myself. 🙂

Ned-el-ch:
this solves something i've been scratching my head at so would love to see it merged 😅

voslartomas:
I guess it is not possible to just leverage optional vs mandatory from graphql, with `String` vs `String!` right? This is great solution, but little bit verbose.

oiime:
@voslartomas 

> I guess it is not possible to just leverage optional vs mandatory from graphql, with String vs String! right? This is great solution, but little bit verbose.

Unless you intend to reinvent all the types you can't really do that, that would just get you in the same end result of having an Optiona[T] like implementation but with more extra steps. The issue is representing 3 distinct states, eg:
```
{value: null}
{value: "foobar"}
{}
```
This is simply not possible to express with `string` or `*string` so you need to add another type that can express those 3 states or convey that information through another source which is what all previous solutions did

carldunham:
Where would defaults fall here? I assume the executor is dropping them in if values aren't provided, but also smart enough to replace with `null` if it's explicitly given.

carldunham:
> Just remembered that input fields can have default values. The field's value will always be defined in this case and the value will either be the default value or the actual value. I'm not sure what would be expected in this case.

Makes sense. I've run into issues in the past over nullable arguments or input fields with defaults with a `null` explicitly passed, and the default being used instead.

oiime:
@carldunham I would think that is the intended behavior there, no? I guess supporting optional `database/sql` style scanner interface makes sense if somebody wants some unique behavior for defaults, there is already a standard library example on how to tackle a similar scenario

eg:
```
func (v *MyNullableDefaultType) Scan(value any) error {
   if any == nil {
      // do some v stuff representing a different default
   }
}
```

carldunham:
Well this is by the GraphQL spec, language independent. CarlOn Apr 1, 2023, at 09:10, Itamar Maltz ***@***.***> wrote:﻿
@carldunham I would think that is the intended behavior there, no? I guess supporting optional database/sql style scanner interface makes sense if somebody wants some unique behavior for defaults, there is already a standard library example on how to tackle a similar scenario
eg:
func (v *MyNullableDefaultType) Scan(value any) error {
   if any == nil {
      // do some v stuff representing a different default
   }
}

PR #2588, opened 2023-03-20
pin nancy because latest tag lookup is failing

StevenACoffman:
The `latest` lookup for nancy is very flaky, causing a number of failures that I need to retry manually.


StevenACoffman:
This magically started working again, so closing for now.

Issue #2559, opened 2023-02-20
Generated Federation code for fields with `@external` directive applied on a list which is used in a `@requires` field is wrong.


vamshiaruru-virgodesigns:
### What happened?
Hi everyone, thanks for the library!

I am trying to port one of my services from python to go. This service is a subgraph in a larger supergraph, and makes use of `@external` and `@requires` directives. As a simplified example, I have a type like this:

```
type MetadataItem {
  key: String! @external
  value: String! @external
}

type Checkout @key(fields: "id") {
  id: ID!
  metadata: [MetadataItem]! @external
  token: UUID! @external
  transformed: Transformation
    @requires(
      fields: "metadata { key value }  token"
    )
}
```
This seems to require implementation of a resolver called `findCheckoutByID` (I imagine this is the `resolve_reference` equivalent in gqlgen?). 

In `federation.go` I see this output:
```
			case "findCheckoutByID":
				id0, err := ec.unmarshalNID2string(ctx, rep["id"])
				if err != nil {
					return fmt.Errorf(`unmarshalling param 0 for findCheckoutByID(): %w`, err)
				}
				entity, err := ec.resolvers.Entity().FindCheckoutByID(ctx, id0)
				if err != nil {
					return fmt.Errorf(`resolving Entity "Checkout": %w`, err)
				}

				entity.Token, err = ec.unmarshalNUUID2string(ctx, rep["token"])
				if err != nil {
					return err
				}
				entity.Metadata.Key, err = ec.unmarshalNString2string(ctx, rep["metadata"].(map[string]interface{})["key"])
				if err != nil {
					return err
				}
				entity.Metadata.Value, err = ec.unmarshalNString2string(ctx, rep["metadata"].(map[string]interface{})["value"])
				if err != nil {
					return err
				}
				// More code here

```
and the error 
```
entity.Metadata.Value undefined (type []*model.MetadataItem has no field or method Value)
```
Which makes sense because, entity.Metadata is a list, not a struct. Here's the generated Checkout entity
```go
type Checkout struct {
	ID                  string              `json:"id"`
	Token               string              `json:"token"`
	Metadata            []*MetadataItem     `json:"metadata"`
	// rest of the fields
}
```
As you can see, there's a mismatch. I am not sure how to proceed here.

### What did you expect?
I expect that the metadata must be constructed as a list, not as a single struct.

### Minimal Reproduction Example
```
type Review {
  body: String
  author: User @provides(fields: "username")
  product: Product
}

[...]

PR #2594, opened 2023-03-27
EntityResolver input type fix

### LINKED ISSUE #2326: Federation plugin does not import autobind and generated models
### What happened?
When using the federation plugin with the `@entityResolver(multi: true)` directive, the entity's input struct is created in the generated models file.
If the generated models are located not in the same package as the federation package, we get an undeclared name error, as the generated federation file does not import the models.
### What did you expect?
When generating the federation file, it should include the generated models import and the autobind imports if needed.
Schema:
```gql
type MultiHello @key(fields: "name") @entityResolver(multi: true) {
    name: String!
}
```
Generated input struct in models_gen.go:
```go
type MultiHelloByNamesInput struct {
	Name string `json:"ID"`
}
```
### versions
 - `go run github.com/99designs/gqlgen version` v0.17.13
 - `go version` 1.17.9

erankor:
fixes #2326 - when using `@entityResolver(multi: true)` the InputType of the entity resolver may be in a different package (usually `model`). The fix is to pull the types.Type of the resolver input, and use templates.CurrentImports.LookupType in order to render it correctly (possibly adding another import)

I have:


StevenACoffman:
Thanks very much! This a nice improvement!

erankor:
@StevenACoffman, thank you very much for the super-fast merge!
Any chance you can create a new tag/release? we want to start using this feature :)

StevenACoffman:
I really wanted to look at the differences here and with the in progress #2593 and I was hoping that one of you would provide some test coverage before a new release.

erankor:
@StevenACoffman, I don't understand... from what I see, #2593 aims to resolve the same bug as this one, so I guess it's no longer relevant, right?
About test coverage, in this PR, I updated the gqlgen.yml of the existing test for entity resolver to use a dedicated package for the model. So, the bug solved by this PR is already covered - running go generate on 'entityresolver' fails without the fix (and since this PR uses the existing framework for type lookup, I don't think we need to test also the case of model sharing package with federation.go).
What do you think is missing?

StevenACoffman:
First, I try not to release a new version more than once a week unless it's entirely critical. Anyone who wants to can pin their gqlgen to a specific commit prior to that. `go get -d github.com/99designs/gqlgen@VERSION` where version is the commit sha1.

Second, In my volunteer OSS time in the mornings, I tend to only have enough time to see if new PRs fix legit bugs and pass the existing tests. Comparing competing solutions requires a deeper level of engagement. Drafting a new release also generally waits until I have more than the time to sip a single coffee. 

I'd like to see if one of the solutions better sets us up to fix some of the other federation issues people have reported such as #2559

erankor:
@StevenACoffman, no problem, we'll wait for the release, sorry if I was being pushy. I'm maintaining a couple of OSS repos myself, so I know how it's like... :)

RE #2593, as I wrote, I'm quite sure it's no longer relevant - it solves the exact same issue as this one and can be closed. @jclyons52, can you confirm?

RE #2559, it seems to be unrelated to either PR. If I'll have some free time, I'll try to give it a look.
Btw, on a similar topic to this issue, I also bumped into a problem with `@requires` - the entity resolver function (FindMany...ByIDs) doesn't get the 'required' fields, they are copied after the function returns. So, the function can't use these fields, which is, in my understanding, the main purpose of using `@requires`. 
The workaround I found for this is to pull required fields from the context -
```
	fc := graphql.GetFieldContext(ctx)
	_reps := fc.Args["representations"].([]map[string]interface{})
```
It's not the most elegant solution, but it works... :)

PR #2597, opened 2023-03-28
feat: removeDuplicateTags() validates tags and panic with meaningful …

gitxiongpan:
if schema defines a bad goTag, example: 
```
type AObject {
    id: ObjectID! @goTag(key: "bson", value: "_id, omitempty")
}
```
The value contains an empty space. This will break removeDuplicateTags function and panic which vague error message which makes it hard to debug.

My PR check goTag format and panic with  meaningful message which help debug a lot.

Before:
image
After:
image

Test result:
image

I have:
 - [✅] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [❌ ] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

### DIFF
--- a/plugin/modelgen/models.go
+++ b/plugin/modelgen/models.go
@@ -434,6 +434,11 @@ func removeDuplicateTags(t string) string {
 	// iterate backwards through tags so appended goTag directives are prioritized
 	for i := len(tt) - 1; i >= 0; i-- {
 		ti := tt[i]
+		// check if ti contains ":", and not contains any empty space. if not, tag is in wrong format
+		if !strings.Contains(ti, ":") || strings.Contains(ti, " ") {
+			panic(fmt.Errorf("wrong format of tags: %s. goTag directive should be in format: @goTag(key: \"something\", value:\"value1,value2,etc\"), no empty space is allowed", t))
+		}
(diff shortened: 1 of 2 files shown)


emanuel-gjini:
@gitxiongpan It looks like this update actually broke one of my running servers due to the usage of some tags that are used to generate `GORM` model tags. And some of those actually have spaces, like this one for example (from the gql schema):
```
type Proxy @private {
  id: Uint! @goTag(key: "gorm", value: "primaryKey")
  ip: String! @goTag(key: "gorm", value: "unique;not null")
}
```

gitxiongpan:
@emanuel-gjini I am sorry for what happened to you. In fact, it was broken before my PR, by other contributor.
have a look of that removeDuplicateTags()
`func removeDuplicateTags(t string) string {
	processed := make(map[string]bool)
	tt := strings.Split(t, " ")`

I totally understand your frustration. I have just submitted a PR to fix this issue. #2628

emanuel-gjini:
@gitxiongpan Many thanks for the fix ❤️

PR #2611, opened 2023-04-07
Transport for application/x-www-form-urlencoded content type

RatkoR:
This is one more (and the last) transport that we're missing in stock gqlgen (I'm done with transports after this PR).

It works with `application/x-www-form-urlencoded` transports that a big part of out clients use. It is similar to both multipart and graphql transports, but I still decided to do a separate file. Anyone interested in it can enable it in server config, so it's backwards compatible, it wont break anything.

Tests and doc changes are also part of this PR.

`transport_form.go` is renamed to `transport_form_multipart.go` as we now have two form transports (`transport_form_multipart.go` and `transport_form_urlencoded.go`).

Comments are welcome.

I have:


StevenACoffman:
Thanks!

PR #2621, opened 2023-04-21
remove unused `root_.gotpl` file

fiatjaf:
I was confused about the role of this file, but after a long time I realized it wasn't being used anywhere, so removing for clarity.

I have:


StevenACoffman:
Pretty sure it contains top-level definitions that should not be duplicated across the generated files for each schema file. https://github.com/99designs/gqlgen/blob/master/codegen/generate.go#L139

fiatjaf:
Hmm, that's true. Can you explain to me how that works? Because the `generated!.gotpl` file is almost exactly the same as `root_.gotpl` and that is what is being actually used in practice when generating schema files, no `root_.gotpl`. I've asked about this [on Discord](https://discord.com/channels/831012771810377778/831012771810377781/1098747517337870437).

In practice I can delete `root_.gotpl` and everything works fine. That function is never called by `gqlgen generate`.

StevenACoffman:
Sure, that is called by generatePerSchema [here](https://github.com/99designs/gqlgen/blob/a13eca12117720d9d6cccba8549c7e6934d09fbd/codegen/generate.go#L47), which is called by GenerateCode [here](https://github.com/99designs/gqlgen/blob/a13eca12117720d9d6cccba8549c7e6934d09fbd/codegen/generate.go#L29)

So `root_.gotpl` is needed when you have a "follow-schema" layout specified in your `gqlgen.yml` as in:
```
resolver:
  layout: follow-schema
```
But `root_.gotpl` is not needed when you have a `single-file` layout specified as in:
```
resolver:
  layout: single-file
```

StevenACoffman:
If you don't mind, I'm going to close this issue for now.

fiatjaf:
Yes, that makes sense, but I'm still confused. I did successfully run the `generate` flow using `follow-schema` and that codepath was never reached.

Also, this speaks to my stupidity, but nowhere in the code I could find a reference to the `generated!.gotpl` file, but clearly that is the file that is being used in practice during code generation (because if you modify it while deleting `root_.gotpl` you get the notifications).

Also `root_.gotpl` and `generated!.gotpl` are almost exactly the same.

I'm asking these questions because I feel there is something wrong that can be cleaned up even though everything seems to be working perfectly, but I can't figure it out for myself. Please ignore me if the questions are too annoying.

PR #2625, opened 2023-04-24
Bugfix: add missing return statements in GRAPHQL and UrlEncodedForm t…

RatkoR:
…ransports.

Two transports (GRAPHQL and UrlEncodedForm) did not have return statement at the end of `if err` block. Instead of returning a 'could not cleanup body' error, we continued processing.

User still got an error. But instead of early `could not cleanup` error, user got `Internal system error` which happened a few lines after the if block.

While at it, remove the not needed `log.Printf()` lines.

Tests are added.

I have:


StevenACoffman:
Thanks for fixing this!

PR #2617, opened 2023-04-13
feat: gqlgen ver in generated file notice and entire file notice optional

hexnaught:
## Describe your PR and link to any relevant issues.

The gqlgen version printed in the file notice adds a lot of noise to PRs in all of the services we use gqlgen at my current company when we regenerate files. There is an existing issue for removing the file notice entirely so I have added two new config options to.
- Omit the gqlgen version but keep the current file notice in tact
- Omit the file notice in its entirety.

Both are 'omit' options and defaulted to false to continue current behaviour.

```yaml
# Optional: turn on to not generate any file notice comments in generated files
omit_gqlgen_file_notice: false

# Optional: turn on to exclude the gqlgen version in the generated file notice. No effect if `omit_gqlgen_file_notice` is true.
omit_gqlgen_version_in_file_notice: false
```

### Both omit options `false`

```go
package graph

// This file will be automatically regenerated based on the schema, any resolver implementations
// will be copied through when generating and any unknown code will be moved to the end.
// Code generated by github.com/99designs/gqlgen version v0.17.29-dev

import (
	"context"
```

### Omit file version `true`

```go
package graph

// This file will be automatically regenerated based on the schema, any resolver implementations
// will be copied through when generating and any unknown code will be moved to the end.
// Code generated by github.com/99designs/gqlgen

import (
	"context"
```

### Omit file notice `true`

```go
package graph

import (
	"context"
```

### Both omit options `true`

```go
package graph

import (
	"context"
```

I have:
 - [ ] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [X] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

Resolves #2413

Issue #2413, opened 2022-10-27
Option to disable comment generation


arnemolland:
### What happened?
When upgrading from gqlgen v0.17.15 to v0.17.20, `gqlgen generate` started generating comments. In our case, these are useless and add nothing but noise, so I would they can be disabled by config. AFAIK this change was introduced in #2263.

### What did you expect?
I would expect the option to disable comment generation.

### Minimal graphql.schema and models to reproduce
```graphql
type Foo {
 bar: Int!
}
```

### versions
 - `go run github.com/99designs/gqlgen version`? v0.17.20
 - `go version` go1.19.2

StevenACoffman:
A PR would be welcome!

hexnaught:
@arnemolland cc @StevenACoffman 

I have opened a PR for this, hope this is acceptable!

PR #2627, opened 2023-04-29
New option to make comments on resolver optional

mokeko:
After the PR #2263 , the `gqlgen generate` command started generating comments such as `// Todos is the resolver for the todos field.` above  resolver implementations. However, these comments are not necessary for me. So I added an option to not generate such comments.

When `omit_template_comment: true` is set, the above comment will not be generated. However, pre-written comments are retained after the code re-generation. This feature is necessary to keep manually written comments.

Relevent issues and PRs: #2413 , #2617 

I have:


plugin/resolvergen/resolver.go
```diff
@@ -68,16 +68,17 @@ func (m *Plugin) generateSingleFile(data *codegen.Data) error {
 				continue
 			}
 
-			resolver := Resolver{o, f, nil, "// foo", `panic("not implemented")`}
```
mokeko, line 71: Currently, a fixed comment `// // foo` is generated for all resolvers in `single-file` mode, but it appears to me as if it was accidentally added to the code during development and was not intentional. Therefore, I have made a modification to generate the same comments as in `follow-schema` mode.

Issue #2469, opened 2022-12-14
Should be able to override enum capitalization


todor-mutafov:
Similar to: #1447

### What happened?
I'm using gqlgen with enums that look something like this:
```graphql
enum FilterType {
  CHILDREN
  IDENTITIES
}

enum PhoneType {
  IPHONE
  ANDROID
}
```

Unfortunately, the go 'enum' constants for these GQL enums are not... ideal...
```golang
type FilterType string

const (
	FilterTypeChildren   FilterType = "CHILDREN"
	FilterTypeIDEntities FilterType = "IDENTITIES"
)
```
```golang
type PhoneType string

const (
	PhoneTypeIPHone  PhoneType = "IPHONE"
	PhoneTypeAndroid PhoneType = "ANDROID"
)
```

### What did you expect?
When I say `IDENTITIES`, I mean the word 'identities', not 'ID Entities'. If I wanted to say 'ID Entities', the schema name would be `ID_ENTITIES`. The `IPHONE` example may be more contentious, because I could technically use `I_PHONE`. But I definitely would like a way to tell gqlgen how to name my enum constants, irrelevant of what abbreviations it thinks it's seeing in the names.

### versions
 - `go run github.com/99designs/gqlgen version`? `v0.17.21`
 - `go version`? `go version go1.19.2 linux/amd64`

nettrino:
+1 for this issue. Related https://github.com/99designs/gqlgen/discussions/1918

PR #2630, opened 2023-05-09
fix enum capitalization

dukhyungkim:
First of all, sorry for my bad english.

In my case, I want to use `IDLE` and it should be `Idle`. But the generated output shows me `IDLe`.
And I found similar issue #2469. So I try to satisfy both.

I have:

### DIFF
--- a/codegen/templates/templates.go
+++ b/codegen/templates/templates.go
@@ -519,6 +519,17 @@ func wordWalker(str string, f func(*wordInfo)) {
 		matchCommonInitial := false
 		upperWord := strings.ToUpper(word)
 		if commonInitialisms[upperWord] {
+			// If the uppercase word (string(runes[w:i]) is "ID" or "IP"
+			// AND
+			// the word is the first two characters of the str
+			// AND
+			// that is not the end of the word
+			// AND
+			// the length of the string is greater than 3
+			// AND
+			// the third rune is an uppercase one
+			// THEN
+			// do NOT count this as an initialism.
 			switch upperWord {
 			case "ID", "IP":
 				if word == str[:2] && !eow && len(str) > 3 && unicode.IsUpper(runes[3]) {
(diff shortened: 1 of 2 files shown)


codegen/templates/templates.go
```diff
@@ -517,7 +517,14 @@ func wordWalker(str string, f func(*wordInfo)) {
 		}
 
 		matchCommonInitial := false
-		if commonInitialisms[strings.ToUpper(word)] {
+		upperWord := strings.ToUpper(word)
+		if commonInitialisms[upperWord] {
+			switch upperWord {
```
StevenACoffman, line 522:
```suggestion
// If the uppercase word (string(runes[w:i]) is "ID" or "IP"
// AND
// the word is the first two characters of the str
// AND  
// that is not the end of the word
// AND 
// the length of the string is greater than 3
// AND
// the third rune is an uppercase one
// THEN
// do NOT count this as an initialism.
			switch upperWord {
```

StevenACoffman:
It took me a little while to understand what this code was doing. I suggested adding a comment (which will probably need to be indented differently), because in the future it may help someone to make this a more general fix than just for ID and IP.

codegen/templates/templates.go
dukhyungkim: Thank you for your suggestion!
I applied it and pushed.

StevenACoffman:
Thanks for making this improvement! Keep them coming!

dukhyungkim:
> Thanks! It took me a little while to understand what this code was doing. I suggested adding a comment (which will probably need to be indented differently), because in the future it may help someone to make this a more general fix than just for ID and IP.

Agree. I worried while I was writing this PR because it looks too tricky. 

> Thanks for making this improvement! Keep them coming! 

Thank you for accepting. I look forward to meet again on PR!

PR #2631, opened 2023-05-10
Fix uint32 unmarshal

mdvalv:
The string unmarshal for uint32 used ParseInt instead of ParseUint, which would parse the wrong range of valid numbers.

I have:


StevenACoffman:
Thanks for this!

PR #2634, opened 2023-05-11
Generate federation entity resolvers for interfaces with @key fields defined

nicolerenee:
When using interfaces and interfaceObjects as described by the Apollo spec you have to be able to resolve the interface types as described [here](https://www.apollographql.com/docs/federation/federated-types/interfaces/#interface-reference-resolver). 

This updates the generation to allow `@key` fields on interfaces and generates the resulting functions needed to handling resolving the entities. 

I left the existing test but changed it to be an unused interface, and then created a new test with an interface that is used to verify it works. I've also validated this works with the federated gateway to properly handle resolving interface return types coming from a different subgraph.

The only thing I'm not sure about is around line 223. I'm using an object that implements the interface to get the field types. This seemed easier to maintain long term as it seemed like I would need to fork to code and have one path for an interface and one path for an object since there isn't an object for the interface. Open to ideas to improve this though.

Not sure if any docs specifically for this are desired, the use of `@key` fields is documented already and it follows the same pattern.

I have:

### DIFF
--- a/plugin/federation/federation.go
+++ b/plugin/federation/federation.go
@@ -123,11 +123,18 @@ func (f *federation) InjectSourceLate(schema *ast.Schema) *ast.Source {
 	f.setEntities(schema)
 
 	var entities, resolvers, entityResolverInputDefinitions string
-	for i, e := range f.Entities {
-		if i != 0 {
-			entities += " | "
+	for _, e := range f.Entities {
+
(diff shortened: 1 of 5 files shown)


StevenACoffman:
I can't tell you how thrilled I am that you worked on this! Better federation support has been long desired. And you even thought about maintainability!

PR #2635, opened 2023-05-15
feat(federation): update Apollo Federation v2 definitions

dariuszkuc:
Fix Apollo Federation v2 directive definitions:
* `_FieldSet` was renamed to `FieldSet`
* new `@composeDirective` (fed v2.1 change)
* `@shareable` is now repeatable (fed v2.2 change)
* new `@interfaceObject` directive (fed v2.3 change)

NOTE: definitions are updated to the latest [Federation v2 spec](https://specs.apollo.dev/). Users might specify old version import (i.e. `@link` import v2.0) trying to use new federation features (e.g. `@interfaceObject`) which may lead to potential composition errors (i.e. server starts up fine but cannot be federated into a "supergraph").

---
I have:


plugin/federation/federation.go
```diff
@@ -86,28 +86,55 @@ func (f *federation) MutateConfig(cfg *config.Config) error {
...
+
 	// add version-specific changes on key directive, as well as adding the new directives for federation 2
 	if f.Version == 1 {
 		input += `
 	directive @key(fields: _FieldSet!) repeatable on OBJECT | INTERFACE
+	directive @requires(fields: _FieldSet!) on FIELD_DEFINITION
+	directive @provides(fields: _FieldSet!) on FIELD_DEFINITION
+	directive @extends on OBJECT | INTERFACE
 	directive @external on FIELD_DEFINITION
+	scalar _Any
```
StevenACoffman, line 99: This line is still shared by both v1 and v2, correct?

dariuszkuc: yeah but I thought it was cleaner to keep v1 and v2 definitions separate so they can be ordered (only common parts are `@extends` and `_Any`)

chadxzs:
@dariuszkuc Thanks for this contribution!

PR #2640, opened 2023-05-19
`@defer` support

StevenACoffman:
This implements `@defer` support according to https://github.com/graphql/graphql-wg/blob/66a80c7e7ff5d881a7de515eec991ca3150c829b/rfcs/DeferStream.md

There are no tests and there may be some bugs yet, but I'm just opening this to ask if this change is desired and/or if the code quality is too bad to be considered -- before I put more time polishing it and writing tests.

(Don't look at the first commit, please, just look at the final result.)

There is a demo app at https://github.com/fiatjaf/gqlgen-defer-demo, you can try it live on https://defer.fiatjaf.com/debug/. Only the fields `availability`, `here` and `there` are concurrent, so they're the only ones that can be deferred (the others will just return in the normal response regardless). Try the query

```graphql
query {
  fruits {
    name
    color
    availability {
      ... on Availability @defer {
        there
      }
      here
    }
  }
}
```

https://user-images.githubusercontent.com/1653275/233782726-97d3f129-e8d6-4190-b6fe-a419bc9809be.mp4

I have:


graphql/response.go
```diff
@@ -14,6 +15,9 @@ import (
 type Response struct {
 	Errors     gqlerror.List          `json:"errors,omitempty"`
 	Data       json.RawMessage        `json:"data"`
+	Label      string                 `json:"label,omitempty"`
+	Path       ast.Path               `json:"path,omitempty"`
+	HasNext    bool                   `json:"hasNext"`
```
StevenACoffman, line 20: What do you think about adding omitempty here? @fiatjaf @UnAfraid

StevenACoffman:
@UnAfraid @fiatjaf I put up a PR of your work, so I could look at it more easily. I am a little concerned that people who do not need/want `@defer` support will get the extra hasNext fields in their responses that they don't expect.

graphql/response.go
fiatjaf: Looks good to me.

StevenACoffman, line 20:
```suggestion
	HasNext    bool                   `json:"hasNext,omitempty"`
```

StevenACoffman:
Closed in favor of #2642

PR #2656, opened 2023-05-30
Addressing few issues in defer feature

UnAfraid:
Fixed hasNext being omitted when false, now it would only appear in the payload when there is deferred usage.
Applied [@fiatja](https://github.com/99designs/gqlgen/compare/master...fiatjaf:gqlgen:defer-prod) atomic patch

Here is modified version of GraphiQL playground that works with SSE https://gist.github.com/UnAfraid/c2f9ed63332c47a2d588213fdbcfd688

I did some testing in real world environment and it seems to work as intended.
I'm looking into writing some tests, but that would require implementation of SSE in the client package first correct?

### DIFF
--- a/_examples/chat/generated.go
+++ b/_examples/chat/generated.go
@@ -94,7 +94,7 @@ func (e *executableSchema) Schema() *ast.Schema {
 }
 
 func (e *executableSchema) Complexity(typeName, field string, childComplexity int, rawArgs map[string]interface{}) (int, bool) {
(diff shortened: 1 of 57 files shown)


StevenACoffman (approved):
LGTM.

StevenACoffman:
Thanks for this! I would love more tests, but the websocket testing story is not obvious. We are also using the unmaintained github.com/gorilla/websocket instead of the https://pkg.go.dev/nhooyr.io/websocket

UnAfraid:
I'll implement first set of tests using SSE if u don't mind

PR #2659, opened 2023-06-02
Safe cast for uint

monlu92:
Negative integers would convert into big uint's instead of failing.
I have added tests covering the case.


StevenACoffman:
Looks like some lint and test failures. Can you try this:
```
go generate ./...; cd _examples; go generate ./...; cd ..; go test ./...; cd _examples; go test ./...
```

chadxzs:
Nice small improvement @monlu92! Did you run into this as a bug in the wild? Or find it through some other means?

phughes-scwx:
Remaining aspects of this MR merged in #3430, I think it may be closed @StevenACoffman.

StevenACoffman:
Hey, thanks all. Let me know if there's anything else.

Issue #2632, opened 2023-05-10
gqlgen generates non compilable code when schema has uses @requires directive with array nested attribute dependency for federation  


mihirpmehta:
### What happened?

gqlgen generates non compilable code when schema has uses @requires directive with array nested attribute dependency for federation. in generated.go file i.e. see the sample schema that is mentioned below

### What did you expect?
It should generate compilable code 

### Minimal graphql.schema and models to reproduce

type Query {
  reviews(bookID: ID!): [Review]
}
type Book @key(fields: "id", resolvable: false ) {
  id: ID! @external
  publishers:[Publishers] @external
  reviews: [Review]  @requires(fields: "publishers {name}" )
}
type Publishers @key(fields: "id", resolvable: false) {
  id: ID! @external
  name: String @external
}
type Review @key(fields: "id") {
  id: ID! @external
  name: String
}

### versions
 - `go run github.com/99designs/gqlgen version`? v0.17.31
 - `go version`?  go1.20.3 darwin/amd64

if you try to run ---- 

`it will generate error validation failed: packages.Load: /graph/generated/federation.go:100 : 23: entity.Publishers.Name undefined (type []*model.Publishers has no field or method Name)`

on this line in generated.go file 

`entity.Publishers.Name, err = ec.unmarshalOString2ᚖstring(ctx, rep["publishers"].(map[string]interface{})["name"])`

This is because Publishers.Name is incorrect ... Publishers is array

dariuszkuc:
fixed in https://github.com/99designs/gqlgen/pull/2884

Issue #2357, opened 2022-09-07
non-nullable @external field that are not part of @key fieldset blows up federation


dariuszkuc:
### What happened?

If schema specifies non-nullable `@external` fields that are not part of the `@key` field set, entity resolution will blow up trying to coerce `nil` to a target type even if they are not part of the query. Those fields will only be correctly coerced if end user select a field that references those external fields using `@requires` field set. It appears that the current logic assumes that those fields will always be populated in entity representation.

Example use case from https://github.com/apollographql/apollo-federation-subgraph-compatibility. Given a type definition

```
extend type User @key(fields: "email") {
  averageProductsCreatedPerYear: Int @requires(fields: "totalProductsCreated yearsOfEmployment")
  email: ID! @external
  name: String @override(from: "users")
  totalProductsCreated: Int @external
  yearsOfEmployment: Int! @external
}
```

`@requires` directive allows to reference external fields that are necessary for calculating its value. Those `@external` fields will be provided in the entity representation ONLY if field referencing them is requested (i.e. router will only include `totalProductsCreated` and `yearsOfEmployment` in the user entity representation if `averageProductsCreatedPerYear` is in query selection set). Attempting to run a query without `@requires` field blows up the execution, e.g.

```graphql
# blows up
query userEntity ($representations: [_Any!]!) {
    _entities(representations: $representations) {
        ...on User { email name }
    }
}

# works
query userEntity ($representations: [_Any!]!) {
    _entities(representations: $representations) {
        ...on User { email name averageProductsCreatedPerYear }
    }
}
```

### What did you expect?

Query is executed successfully even if those fields are referenced by the selection set (i.e. they are not populated).

### Minimal graphql.schema and models to reproduce

Start example `reviews` server from https://github.com/99designs/gqlgen/tree/v0.17.16/_examples/federation 

Execute following query

```graphql
query userEntity ($representations: [_Any!]!) {
    _entities(representations: $representations) {
        ...on User { id }
    }
}
```

Variables

```json
{
    "representations": [
        {
            "__typename": "User",
            "id": "123"
        }
    ]
}
```

Execution blows up at https://github.com/99designs/gqlgen/blob/v0.17.16/_examples/federation/reviews/graph/generated/federation.go#L124 with `interface conversion: interface {} is nil, not map[string]interface {}`

### versions
 - `go run github.com/99designs/gqlgen version` v0.17.16
 - `go version` go1.19 darwin/arm64

[...]

Issue #1861, opened 2022-01-27
Federation: requiring a field with a list of objects generates invalid code


arnemolland:
### What happened?
When extending a federated entity with a field of the array type containing object (not primitives/scalars) and adding a field that requires the array field, `gqlgen generate` generates invalid code. It looks like the execution context does not have any unmarshaler for the specific type. I'm not sure if this kind of relation is valid in GraphQL or if it's missing in gqlgen.

### What did you expect?
When extending a federated entity with a field of the array type containing object (not primitives/scalars) and adding a field that requires the array field, `gqlgen generate` generates valid code.

### Minimal graphql.schema and models to reproduce
```graphql
type Bar {
  id: Int!
}

extend type Foo @key(fields: "id") {
  id: Int!
  arr: [Bar!]! @external
  xyz: Int! @requires(fields: "arr")
}

```

### Logs
```
gofmt failed on federation.go: /Users/arne/dev/orgs/gameflow/backend/graph-template-go/graph/generated/federation.go:105:33: expected ')', found ',' (and 1 more errors)
validation failed: packages.Load: /Users/arne/dev/orgs/gameflow/backend/graph-template-go/graph/generated/federation.go:105:33: expected ')', found ','
/Users/arne/dev/orgs/gameflow/backend/graph-template-go/graph/generated/federation.go:105:35: expected ';', found rep
/Users/arne/dev/orgs/gameflow/backend/graph-template-go/graph/generated/federation.go:105:26: invalid operation: ec (variable of type *executionContext) is not an interface
```

### Invalid code
```go
entity.Arr, err = ec.(ctx, rep["arr"])
```

### versions
 - `gqlgen version`?  v0.14.0-dev
 - `go version`? go version go1.17.6 darwin/arm64
 - dep or go modules ? modules

bpteam:
## Still actual for versions

- gqlgen version? v0.17.2
- go version? go version go1.17.6 darwin/arm64

andriihrachov:
can confirm it doesnt work on v0.17.13

```
type BoostedSKU {
    sku: String!
    position: Int
    advertising_id: String!
}

type ProductsResponse @key(fields: "global_entity_id vendor_id category_uuid customer_id session_id", resolvable: false) {
    boosted_items: [BoostedSKU] @external
    items(limit: Int = 20, page: Int = 0): ProductCategoryItems @requires(fields: "boosted_items")
}```

dariuszkuc:
[...]

Issue #1138, opened 2020-04-08
Support nested selection sets on _FieldSet (for @requires annotation)


bartcpostma:
### What happened?
We started using federation for our graphql solution. In our setup we have two graphql services, one providing items, another extending the item with breadcrumb information. In order to be able to resolve the breadcrumb information we need the id of a nested type on Item, i.e., the Variant of the Item.  

According to the apollo federation specifications on the Scalar _FieldSet, it supports nested selection sets as in "variant { id }" (https://www.apollographql.com/docs/apollo-server/federation/federation-spec/#scalar-_fieldset). This does not seem to be supported by gqlgen.

I observed that gqlgen is parsing the @requires annotation argument as being 4 fields, i.e., `variant`, `{`, `id` and `}` thereby ignoring that `variant` has a nested selection set defined. gqlgen execution crashes at some point with:
`federation: federation.gotpl: template: federation.gotpl:53:50: executing "federation.gotpl" at <.TypeReference.UnmarshalFunc>: error calling UnmarshalFunc: runtime error: invalid memory address or nil pointer dereference%`

### What did you expect?
That gqlgen generation works for a federation setup when having the schema setup as below.

### Minimal graphql.schema and models to reproduce
**service 1:**
```
type Item @key(fields: "id") {
     id: ID!
     variant: Variant
}
type Variant {
     id: ID!
}
```
**service 2:**
```
extend type Item @key(fields: "id") {
     id: ID! @external
     variant: Variant @external 
     breadcrumbs: [String!]! @requires(field: "variant { id }")
}
extend type Variant {
     id: ID! @external
}
```

### versions
 - `gqlgen version`: v0.11.3
 - `go version` 1.13.8

pepusz:
I have the same issue

We have a service where we store simple entities 

```
#Service 1
type BlogEntry  @key(fields: "id"){
   id: ID!
   date: Date!
   chapters: [Chapter]
}

type Chapter @key(fields: "id"){
    id: ID!
    category: Category!
    title: String!
    body: String
    score: Int
}


#Service 2 - extend blog entry with a recommendation. to be able to achieve this we need content from the entry 
extend type BlogEntry @key(fields: "id") {
    id: ID! @external
    chapters: [chapters] @external
    suggestedActivites: [Activity] @requires(fields: "chapters { id body }")
}

extend type Chapter @key(fields: "id"){
    id: ID! @external
    category: Category! @external
    body: String @external
}

```

the error is the following:

[...]

PR #2676, opened 2023-06-14
WIP: Refactor federation requires to populate on entity via a function

jesse-apollo:
The PR is WIP as I'd like to elicit feedback before regenerating the tests, etc.

This PR addresses several issues:

- https://github.com/99designs/gqlgen/issues/2559
- https://github.com/99designs/gqlgen/issues/2632
- https://github.com/99designs/gqlgen/issues/2357
- https://github.com/99designs/gqlgen/issues/1861
- https://github.com/99designs/gqlgen/issues/1138

The PR creates a new rendered template with functions that pass the requires representations to a generated function with the entity.  From here the user can apply the requires data as needed.

The output of the rendered templates looks like this:

```go
import (
	"context"
	"fmt"

	"github.com/mentat/graph/server/pkg/datasources"
)

// PopulateOrganizationRequires is the requires populator for the Organization entity.
func (ec *executionContext) PopulateOrganizationRequires(ctx context.Context, entity *datasources.Organization, reps map[string]interface{}) error {
	panic(fmt.Errorf("not implemented: PopulateOrganizationRequires"))
}
```

We (the Apollo team) think this approach is more flexible and future proof as it handles much more complex use cases like @requires nesting in repeated fields.  Our initial approach was to try and extend the existing requires functionality using recursive template rendering but this proved to be complex and difficult to test.

cc @brh55 @dariuszkuc 

I have:

### DIFF
--- a/plugin/federation/federation.go
+++ b/plugin/federation/federation.go
@@ -3,6 +3,9 @@ package federation
 import (
 	_ "embed"
 	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
 	"sort"
 	"strings"
 
@@ -11,6 +14,7 @@ import (
 	"github.com/99designs/gqlgen/codegen"
 	"github.com/99designs/gqlgen/codegen/config"
 	"github.com/99designs/gqlgen/codegen/templates"
+	"github.com/99designs/gqlgen/internal/rewrite"
 	"github.com/99designs/gqlgen/plugin"
 	"github.com/99designs/gqlgen/plugin/federation/fieldset"
 )
@@ -233,6 +237,13 @@ type Entity {
 }
 
 func (f *federation) GenerateCode(data *codegen.Data) error {
+	// requires imports
+	requiresImports := make(map[string]bool, 0)
+	requiresImports["context"] = true
+	requiresImports["fmt"] = true
+
+	requiresEntities := make(map[string]*Entity, 0)
+
 	if len(f.Entities) > 0 {
 		if data.Objects.ByName("Entity") != nil {
 			data.Objects.ByName("Entity").Root = true
@@ -272,9 +283,19 @@ func (f *federation) GenerateCode(data *codegen.Data) error {
 					fmt.Println("skipping @requires field " + reqField.Name + " in " + e.Def.Name)
 					continue
 				}
+				// keep track of which entities have requires
+				requiresEntities[e.Def.Name] = e
+				// make a proper import path
+				typeString := strings.Split(obj.Type.String(), ".")
+				requiresImports[strings.Join(typeString[:len(typeString)-1], ".")] = true
+
 				cgField := reqField.Field.TypeReference(obj, data.Objects)
 				reqField.Type = cgField.TypeReference
 			}
+
+			// add type info to entity
+			e.Type = obj.Type
+
 		}
 	}
 
@@ -292,6 +313,81 @@ func (f *federation) GenerateCode(data *codegen.Data) error {
 			}
 
 			resolver.InputType = obj.Type
+		}
+	}
+
+	if len(requiresEntities) > 0 {
+		// check for existing requires functions
+		type Populator struct {
+			FuncName       string
+			Exists         bool
+			Comment        string
+			Implementation string
+			Entity         *Entity
+		}
+		populators := make([]Populator, 0)
+
+		rewriter, err := rewrite.New(data.Config.Federation.Dir())
+		if err != nil {
+			return err
+		}
+
+		for name, entity := range requiresEntities {
+			populator := Populator{
+				FuncName: fmt.Sprintf("Populate%sRequires", name),
+				Entity:   entity,
+			}
+
+			populator.Comment = strings.TrimSpace(strings.TrimLeft(rewriter.GetMethodComment("executionContext", populator.FuncName), `\`))
+			populator.Implementation = strings.TrimSpace(rewriter.GetMethodBody("executionContext", populator.FuncName))
+
+			if populator.Implementation == "" {
+				populator.Exists = false
+				populator.Implementation = fmt.Sprintf("panic(fmt.Errorf(\"not implemented: %v\"))", populator.FuncName)
+			}
+			populators = append(populators, populator)
+		}
+
+		// find and read requires template
+		_, callerFile, _, _ := runtime.Caller(0)
+		currentDir := filepath.Dir(callerFile)
+		requiresTemplate, err := os.ReadFile(currentDir + "/requires.gotpl")
+
+		if err != nil {
+			return err
+		}
+
+		requiresFile := data.Config.Federation.Dir() + "/federation.requires.go"
+		existingImports := rewriter.ExistingImports(requiresFile)
+		for _, imp := range existingImports {
+			if imp.Alias == "" {
+				if _, ok := requiresImports[imp.ImportPath]; ok {
+					// import exists in both places, remove
+					delete(requiresImports, imp.ImportPath)
+				}
+			}
+		}
+
+		for k := range requiresImports {
+			existingImports = append(existingImports, rewrite.Import{ImportPath: k})
+		}
+
+		// render requires populators
+		err = templates.Render(templates.Options{
+			PackageName: data.Config.Federation.Package,
+			Filename:    requiresFile,
+			Data: struct {
+				federation
+				ExistingImports []rewrite.Import
+				Populators      []Populator
+				OriginalSource  string
+			}{*f, existingImports, populators, ""},
+			GeneratedHeader: false,
+			Packages:        data.Config.Packages,
+			Template:        string(requiresTemplate),
+		})
+		if err != nil {
+			return err
 		}
 	}
 
--- a/plugin/federation/federation.gotpl
+++ b/plugin/federation/federation.gotpl
@@ -103,10 +103,10 @@ func (ec *executionContext) __resolve_entities(ctx context.Context, representati
 						if err != nil {
 							return fmt.Errorf(`resolving Entity "{{$entity.Def.Name}}": %w`, err)
 						}
-						{{ range $entity.Requires }}
-							entity.{{.Field.JoinGo `.`}}, err = ec.{{.Type.UnmarshalFunc}}(ctx, rep["{{.Field.Join `"].(map[string]interface{})["`}}"])
+						{{ if $entity.Requires }}
+							err = ec.Populate{{$entity.Def.Name}}Requires(ctx, entity, rep)
 							if err != nil {
-								return err
+								return fmt.Errorf(`populating requires for Entity "{{$entity.Def.Name}}": %w`, err)
 							}
 						{{- end }}
 						list[idx[i]] = entity
--- a/plugin/federation/testdata/entityresolver/generated/federation.requires.go
+++ b/plugin/federation/testdata/entityresolver/generated/federation.requires.go
@@ -1,0 +1,38 @@
+package generated
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/99designs/gqlgen/plugin/federation/testdata/entityresolver/generated/model"
+)
+
+// PopulateMultiHelloMultipleRequiresRequires is the requires populator for the MultiHelloMultipleRequires entity.
+func (ec *executionContext) PopulateMultiHelloMultipleRequiresRequires(ctx context.Context, entity *model.MultiHelloMultipleRequires, reps map[string]interface{}) error {
+	panic(fmt.Errorf("not implemented: PopulateMultiHelloMultipleRequiresRequires"))
+}
+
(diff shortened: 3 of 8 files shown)


StevenACoffman:
@MiguelCastillo @dnerdy @a8m @giautm @vvakame I would appreciate your views on this PR. I'm leaning in favor, as this seems less messy than a lot of other approaches, and could potentially open up some more flexibility for other extensions (e.g. ent eager loading, caching, data loading).

StevenACoffman:
This PR is intended to address when extending a federated entity with a field of an array type containing object (not primitives/scalars) and then adding a field that requires the array field, gqlgen generate generates invalid code. 

@jesse-apollo @dariuszkuc  Questions from some other Khan engineers:
> The most common issue is why change the paradigm to require people to have a separate function to get the data when the current approach is to read it from the entity?

The prior PRs that @carldunham added did not require changing this paradigm, so why should supporting lists
https://github.com/99designs/gqlgen/pull/1669/files#diff-44eab2c3801fddd0439e1e18f5620dcc76950e205bd8a5d7b62dd9f7f4412bb8L249 
https://github.com/99designs/gqlgen/pull/1684
https://github.com/99designs/gqlgen/pull/1706
https://github.com/99designs/gqlgen/pull/1719
https://github.com/99designs/gqlgen/pull/1723
https://github.com/99designs/gqlgen/pull/1956

mihirpmehta:
Following the thread

jesse-apollo:
> > The most common issue is why change the paradigm to require people to have a separate function to get the data when the current approach is to read it from the entity?

The goal is to over the most versatility with the minimal amount of generated complexity.  I didn't see a way to cleanly utilize the current paradigm for all use cases that `@requires` supports.

> > The prior PRs that @carldunham added did not require changing this paradigm, so why should supporting lists now require such a paradigm change?

We initially took this route and built out some code to test it.  However, it didn't seem like a viable approach due to the fact that the requires could have multiple fields, some layered under array(s).  This would require generating code to do recursive array initialization and pointer to struct initialization to populate those leaf nodes at multiple levels.

> 
> See: https://github.com/99designs/gqlgen/pull/1669/files#diff-44eab2c3801fddd0439e1e18f5620dcc76950e205bd8a5d7b62dd9f7f4412bb8L249 #1684 #1706 #1719 #1723 #1956

MiguelCastillo:
@jesse-apollo thanks for this work!  the federation spec is fuzzy about whether or not fieldsets are supported in repeated fields.  https://www.apollographql.com/docs/federation/federated-types/federated-directives#requires.  But yeah i think it makes sense it's supposed to work.
Anyways - I wont be able to take a closer look at these changes until perhaps for a couple of days; Im out of town traveling.  But will do as soon as I can.

At a quick glance it seems like you are adding a function we call in the generated code to populate the `entity` field with the correct data?  Is that overall the idea here?

jesse-apollo:
> At a quick glance it seems like you are adding a function we call in the generated code to populate the entity field with the correct data? Is that overall the idea here?

Yes, that's correct.  The entity is first resolved and then passed to the generated function along with the representations `map[string]interface{}`.  From there the developer can apply the `@requires` data as they see fit.   This also cleanly supports using requires data for computed fields.

StevenACoffman:
I know it's WIP, but I regenerated and pushed it to your fork as I'm curious to see if this would be sufficient to pass the https://github.com/apollographql/apollo-federation-subgraph-compatibility suite.

Feel free to continue to work on it.

plugin/federation/federation.gotpl
```diff
@@ -103,10 +103,10 @@ func (ec *executionContext) __resolve_entities(ctx context.Context, representati
 						if err != nil {
 							return fmt.Errorf(`resolving Entity "{{$entity.Def.Name}}": %w`, err)
 						}
-						{{ range $entity.Requires }}
-							entity.{{.Field.JoinGo `.`}}, err = ec.{{.Type.UnmarshalFunc}}(ctx, rep["{{.Field.Join `"].(map[string]interface{})["`}}"])
+						{{ if $entity.Requires }}
+							err = ec.Populate{{$entity.Def.Name}}Requires(ctx, entity, rep)
```
StevenACoffman, line 107: This results in:
`PopulatePlanetRequiresRequires`

StevenACoffman, line 107:
```suggestion
							err = ec.Populate{{$entity.Def.Name}}(ctx, entity, rep)
```

plugin/federation/federation.go
```diff
@@ -295,6 +316,81 @@ func (f *federation) GenerateCode(data *codegen.Data) error {
...
+		}
+
+		requiresFile := data.Config.Federation.Dir() + "/federation.requires.go"
+		existingImports := rewriter.ExistingImports(requiresFile)
+		for _, imp := range existingImports {
+			if imp.Alias == "" {
+				if _, ok := requiresImports[imp.ImportPath]; ok {
+					// import exists in both places, remove
+					delete(requiresImports, imp.ImportPath)
+				}
```
StevenACoffman, line 367:
```suggestion
				delete(requiresImports, imp.ImportPath)
```
You don't need the conditional check, as delete will no-op if the key does not exist

plugin/federation/federation.gotpl
jesse-apollo: I don't think so.  You might be thinking of Popluator.FuncName.

plugin/federation/testdata/entityresolver/generated/federation.requires.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/99designs/gqlgen/plugin/federation/testdata/entityresolver/generated/model"
+)
+
+// PopulateMultiHelloMultipleRequiresRequires is the requires populator for the MultiHelloMultipleRequires entity.
+func (ec *executionContext) PopulateMultiHelloMultipleRequiresRequires(ctx context.Context, entity *model.MultiHelloMultipleRequires, reps map[string]interface{}) error {
+	panic(fmt.Errorf("not implemented: PopulateMultiHelloMultipleRequiresRequires"))
+}
+
+// PopulateMultiHelloRequiresRequires is the requires populator for the MultiHelloRequires entity.
+func (ec *executionContext) PopulateMultiHelloRequiresRequires(ctx context.Context, entity *model.MultiHelloRequires, reps map[string]interface{}) error {
```
StevenACoffman, line 16: `PopulateMultiHelloRequiresRequires` here is awkward and I'd like to avoid `RequiresRequires`

plugin/federation/testdata/entityresolver/generated/federation.requires.go
```diff
@@ -0,0 +1,38 @@
...
+	panic(fmt.Errorf("not implemented: PopulateMultiHelloRequiresRequires"))
+}
+
+// PopulateMultiPlanetRequiresNestedRequires is the requires populator for the MultiPlanetRequiresNested entity.
+func (ec *executionContext) PopulateMultiPlanetRequiresNestedRequires(ctx context.Context, entity *model.MultiPlanetRequiresNested, reps map[string]interface{}) error {
```
StevenACoffman, line 21: `RequiresNestedRequires` is not much better than `RequiresRequires`

plugin/federation/testdata/entityresolver/generated/federation.requires.go
jesse-apollo: Is MultiHelloRequires the entity name?

plugin/federation/testdata/entityresolver/generated/federation.requires.go
jesse-apollo: Normally the entity will not be named that.  I think this is just a side effect of a strangely named model.

plugin/federation/federation.gotpl
StevenACoffman: Pretty sure this template is what winds up causing https://github.com/99designs/gqlgen/pull/2676/files#diff-8574d02e76a5404ff6b01c08b88ce6e4a9f4e381e21de4238a418514e6db1443R31Z
```
func (ec *executionContext) PopulatePlanetRequiresRequires(ctx context.Context, entity *model.PlanetRequires, reps map[string]interface{}
```
And I would prefer to not have the `Requires` repeated as `RequiresRequires`.

jesse-apollo: The model is just named something strange for the test.  Requires is only added once. In the real world the model (the struct) wouldn't be named something like this.

plugin/federation/testdata/entityresolver/generated/federation.requires.go
StevenACoffman: Ah, gosh. I forgot. Thanks.

StevenACoffman:
For users who are happy with the current federation support (those with no need for slices/arrays), I would like for gqlgen to continue to work without configuration changes or any new coding. 

I think if people do not opt-in, you could still create a generated "PopulateUserRequires" but for them that it would not just be a stub but would continue to provide the current federation support. This would basically be a slight refactor.

If you want to add a configuration option so people can opt-in to creating a `PopulateUserRequires` function stub, it will facilitate more complicated use cases.

Also, it would be important to demonstrate an example of taking a stub `PopulateUserRequires` that would handle a more complicated use case like the current issues with federated GraphQL lists (slices/arrays).

dariuszkuc:
>For users who are happy with the current federation support (those with no need for slices/arrays), I would like for gqlgen to continue to work without configuration changes or any new coding.

While in general I agree that API breaking changes should be avoided. Hiding this improvement behind an opt-in mechanism is IMHO a bad idea as it leads to two different ways in how data is populated and will most likely end up with user errors. Current implementation is broken for list fields and nullable external fields. I guess you could highlight this in the doc but IMHO that will still lead to user confusion, i.e. how users would know whether to opt-in or use the defaults? in most cases they would try it first with default mechanism, encounter some failure (hopefully during testing and before they deploy to prod) and then try to figure out what went wrong. Unless exception message is self explanatory (right now it is not and I don't think it can be) then finding that their underlying issue is because they didn't opt-in to this mechanism may not be as simple as it sounds.

StevenACoffman:
This PR does not currently provide an example of how to code a PopulateUserRequires method to support list fields and nullable external fields, nor how to code a PopulateUserRequires for the currently working GraphQL federation use cases.

As a result, the current PR is not _just_ a breaking change requiring some minor work to update. This PR change requires a substantial engineering project for every user requiring a deep and nuanced view of both GraphQL and federation.  

I am going to be on vacation for the next 9 days. I believe my colleague @MiguelCastillo will be able to spend some time mulling this over and seeing what we can do here.

mihirpmehta:
In my opinion,  any real world complex use case that uses federation, will have use case where nested field @require would exist in schema. Ideally Gqlgen library should take care of it (Implicit parsing or explicit on developer). 

In my Opinion this PR fixes a bug, rather than implementing a corner use case. It changes underlying implementation a bit is debatable but there should not be opt-in/opt-out , this is must have functionality for federation to work for any real world use case.  Even if someone does not have nested @require fields dependency initially, they may introduce it at later stage (May be after moving to production) ... even if we mention this on doc people do not read Readme file line by line :)   

Also we are using this library for our project and due to this bug, we are not able to move forward with it. if this PR merge sooner it would be better for our team.

StevenACoffman:
Hey, I just merged https://github.com/99designs/gqlgen/pull/2720 which will make the resolver implementation configurable via a new template `resolver.gotpl`

I wonder if you could similarly facilitate more complicated federation use cases with some sort of optional template?

ldebruijn:
Two bugs I found with the changeset:
* Requires fields  that take arguments `foo(limit:4)` currently fail code gen because the argument is seen as part of the field name. This can be solved by cleaning field names by splitting from the argument opening character like `strings.Split(raw, "(")[0]`
* The argument key & value need to be typed without any spaces, otherwise they're seen as separate fields. `foo(limit:4)` works in codegen, `foo(limit: 4)` breaks.

I've forked this changeset in order to utilize this feature while waiting for this to be merged. If need by I can push my fixes for these to this PR.

StevenACoffman:
@ldebruijn  can you make a small repository that demonstrates federation using your gqlgen fork?

parkerroan:
Following the thread here, we at Shipt would love to see this fix get merged. I will try to help if I see an area where we can contribute.

StevenACoffman:
@parkerroan can you or your colleagues make a demo federation repository that uses this PR to demonstrate how this could possibly work?

parkerroan:
Yes, we have a demo repo for testing federation features that is pretty lightweight. I will take a look at this PR and test it out, I have not had a chance to really look into the details of this yet to be honest as I just found this PR today. Just reading the thread and description from @jesse-apollo it looks promising.

mihirpmehta:
Can someone expidite this merge ?

StevenACoffman:
This PR will break existing federation users, without any examples for how they can fix things.

StevenACoffman:
Please open a new PR that addresses one or more of the issues without breaking existing users.

Issue #2618, opened 2023-04-17
Subscription documentation is confusing


szgupta:
### What happened?
I was reading over the subscription example and in [the example resolver](https://gqlgen.com/recipes/subscriptions/#implementing-your-resolver), there's the following comment:

> // The channel may have gotten closed due to the client disconnecting.
> // To not have our Goroutine block or panic, we do the send in a select block.
> // This will jump to the default case if the channel is closed. 

From what I can tell, the subscription resolver returns the receiver-end of the channel, which implies that gqlgen cannot actually close the channel. 

Am I missing something? When would the send here actually fail (i.e. enter the `default` case)? Wouldn't sending over a closed channel actually cause a panic (if it were the case that the channel could be closed at this point)?

Was this a non-blocking send because the channel was unbounded? If so, I think the `default` case which describes the channel as "closed" is confusing.

Thanks!

szgupta:
cc @Unkn0wnCat since I believe you added these docs!

Unkn0wnCat:
Yes, I wrote that part of the docs. I don't quite know what caused the routine to block anymore, but i remember when i tested it when i wrote the docs and when not handling the send-failure my test program crashed a few times. This may not be the case anymore, but I have not tried. 

I'll try to find out more about it in the source code later, as I'm not at my computer currently

Hope this helps! 😊

szgupta:
Sounds good, thanks @Unkn0wnCat!

IIUC, the send can potentially block (if there isn't a ready receiver) which is when the `default` case would kick in. But I'm a little skeptical that the default case here actually indicates the channel was closed (if it was, attempting to send on it in the previous case would have panic'd) and also unsure how gqlgen can actually close the channel in library code if the resolver returns a receiver of the channel 🤔

Unkn0wnCat:
I've not had the time yet, and I don't think I'll have the time over the next few days to test removing the `default`-case, as I'm going away for an extended weekend. If you want to you can test how the code does without the `default` case. If you do please check if a) the program gracefully handles disconnected clients, and b) the count of goroutines stays the same (to check if sending loops/routines are exiting properly)

I have to honestly say, I don't know what exactly happens anymore. I'm sorry about that, it's been a little while

mokiat:
Hi, I was looking at the Subscription documentation and noticed a number of issues myself. Here are my findings:

1. Trying to send on a closed channel will panic (check [example](https://play.golang.com/p/nI4WHHbDiQM))

[...]

PR #2677, opened 2023-06-16
fix subscription example in documentation

leangaurav:

This  PR is to update the documentation on how to implement subscription resolvers. Doc is slightly ambiguous.
Referring to this doc: https://github.com/99designs/gqlgen/blob/master/docs/content/recipes/subscriptions.md

When writing to channel in a select case, 
`case ch <- t:`

the write can block if the consumer is not reading. Can happen during high loads or even randomly.
In such case, the go routine goes to default and exits for no reason.
It should exit when context is cancelled/subscription is closed.

Also documentation is a bit misleading about . 
```
			// The channel may have gotten closed due to the client disconnecting.
			// To not have our Goroutine block or panic, we do the send in a select block.
			// This will jump to the default case if the channel is closed
```
1. If writing to a closed channel always panics. Hence won't go to default block. If channel is really closed. Check this example https://go.dev/play/p/yRGO2uYh28M
2. I think ownership of the channel is with the producer (resolver in this case). So no one else closes channel and no panic happens.

I stumbled upon this when my PR got reviewed and I was asked to write the resolver like how it's written in the example. Writing with with `case <-ctx.Done()` works fine in our system.
Using `default` case really doesn't do much harm here, since usually clients consuming subscriptions have a retry. Hence the problem also goes unnoticed.

I have:

### DIFF
--- a/docs/content/recipes/subscriptions.md
+++ b/docs/content/recipes/subscriptions.md
@@ -133,16 +133,17 @@ func (r *subscriptionResolver) CurrentTime(ctx context.Context) (<-chan *model.T
 				TimeStamp: currentTime.Format(time.RFC3339),
 			}
 
-			// The channel may have gotten closed due to the client disconnecting.
-			// To not have our Goroutine block or panic, we do the send in a select block.
-			// This will jump to the default case if the channel is closed.
+			// The subscription may have got closed due to the client disconnecting.
+			// Hence we do send in a select block with a check for context cancellation.
+			// This avoids goroutine getting blocked forever or panicking,
 			select {
+			case <-ctx.Done(): // This runs when context gets cancelled. Subscription closes.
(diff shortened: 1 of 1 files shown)


leangaurav:
Also found an open issue regarding same
https://github.com/99designs/gqlgen/issues/2618

StevenACoffman:
Thanks very much! I really appreciate you clarifying the documentation, especially for subscriptions, which have needed some attention for a while. Even our current subscription tests tend to be slow and flaky, which is why I've got the CI set to retry them when they fail.

Thanks again, and I'm looking forward to your next PR!

leangaurav:
HI @StevenACoffman can you point me to the flaky subscription tests. I can take a look and try to fix them. 
Also the retry in CI is configured in github actions or somewhere else ?

Issue #1536, opened 2021-05-21
struct field is created even though `forceResolver: true`


quinn-freshly:
### What happened?
I added `@goField(forceResolver: true)` to a field. As expected, a new resolver was added, but a field was also added to the struct in `models_gen.go`.

### What did you expect?
I expected for only the resolver to be created. Since that field is not being resolved as part of the struct i expected no field to be added to the struct. 

### Minimal graphql.schema and models to reproduce
```
type User {
  friends: [User!]! @goField(forceResolver: true)
}
```

### versions
 - `gqlgen version`? 
 ```
github.com/99designs/gqlgen v0.13.0 h1:haLTcUp3Vwp80xMVEg5KRNwzfUrgFdRmtBY8fuB8scA=
github.com/99designs/gqlgen v0.13.0/go.mod h1:NV130r6f4tpRWuAI+zsrSdooO/eWUv+Gyyoi3rEfXIk=
```
 - `go version`?
 ```
go version go1.15.6 linux/amd64
```
 - dep or go modules?
 go mod

burdiyan:
Just ran into this, and was also confused. What is the expected behavior here?

cliedeman:
Saw this behaviour too. There are valid use cases though. The value can be populated in a parent resolver and you can use the existing value (sometimes).

I think another property to exclud the field would be reasonable to not cause mass breakages. 

E.g.

```
@goField(forceResolver: true, omitField: true)
```

mlaflamm:
We recently started to use the `name` argument of the `@goField` directive to unexport fields that we would like to omit. We change the field name to start with a lowercase letter. This is almost equivalent to omitting it.

```graphql
type User {
  id: ID!
  email: String!
  posts: [Post!]! @goField(forceResolver:true, name: "posts")
}
```

PR #2698, opened 2023-06-25
feat: add option `omit_force_resolver_fields`

abekoh:
This PR introduces a new option, `omit_force_resolver_fields` . When enabled, fields annotated with `@goField(forceResolver: true)` will not be generated by the modelgen plugin.

I frequently use the `forceResolver` option. However, fields with this option get generated in the same manner as those without it, leading to instances where I have mistakenly implemented mapping. This change is intended to prevent such mistakes.

I would greatly appreciate any suggestions or improvements.

Relevant issue: #1536

I have:


docs/content/config.md
```diff
@@ -71,6 +71,9 @@ resolver:
 # Optional: wrap nullable input fields with Omittable
 # nullable_input_omittable: true
 
+# Optional: omit generated model fields with `@goField(forceResolver: true)`
+# omit_force_resolver_fields: true
```
chadxzs, line 75: As someone relatively new to gqlgen, I don't think I would know what this option does without looking at the source code if I only had this documentation as a guide. Is there some way to make the comment here more illustrative, or some way you can provide additional context on when someone may want to do this?

abekoh: Thanks for your reviewing! I fixed at eb6ed62cf3a86229b7289f25a8456571364bd3b9

chadxzs:
@abekoh FYI I'm not a maintainer here, simply reviewing PRs as a way to help out some and become more familiar with the codebase. @StevenACoffman is primarily maintaining the codebase right now so you'll need to wait until he can review it.

UnAfraid:
I would prefer a bit more granular control over that feature within @goField(omitForceResolverField: true) directive also for cases where omitting is not an option.

Consider the following example: 
The Post wouldn't necessarily have all the user data, but only the user id, so i would initialise new User object only with the id, then use data loader to load the user by the id.

```graphql
type User {
  id: ID!
  email: String!
  posts: [Post!]! @goField(forceResolver: true) # This field can be omitted
}

type Post {
  id: ID!
  title: String!
  body: String!
  createdBy: User! @goField(forceResolver: true) # Here user id would be needed to know which user to load
  updatedBy: User  @goField(forceResolver: true) # ^
  deletedBy: User  @goField(forceResolver: true) # ^
  createdAt: DateTime!
  #...
}
```

abekoh:
@UnAfraid 
Thank you for your feedback! Indeed, in such cases, we would not want to omit it. I will consider a different option.

mlaflamm:
I know this PR is old and closed but for those ending here, you can rename the go field to start with a lowercase letter to unexport it. This is almost equivalent to omitting it.

You can also replace a struct field exported via a resolver by changing its name and its type.

Here is the adapted example previously provided by @UnAfraid:

```graphql
type User {
  id: ID!
  email: String!
  posts: [Post!]! @goField(forceResolver:true, name: "posts") # This field is unexported
}

type Post {
  id: ID!
  title: String!
  body: String!
  createdBy: User! @goField(forceResolver: true, name: "CreatedByUserID", type: "string") @goTag(key: "json", value: "-")
  updatedBy: User  @goField(forceResolver: true, name: "UpdatedByUserID", type: "*string") @goTag(key: "json", value: "-")
  deletedBy: User  @goField(forceResolver: true, name: "DeletedByUserID", type: "*string") @goTag(key: "json", value: "-")
  createdAt: DateTime!
  #...
}
```

PR #2723, opened 2023-07-20
Fixed Data Loader docs

UnAfraid:
Fixed compilation issues with the data loader docs
Also updated to v7

I have:


StevenACoffman (approved):
Thanks again!

WilfredoHQ:
The definition of the middleware the following
func Middleware(loaders *Loaders, next http.Handler) http.Handler

However, in use it is as follows
srv = loaders.Middleware(db, srv)

This ends up confusing (at least for me), what would the correct implementation look like?

PR #2725, opened 2023-07-24
fix: Don't set the package variable for the new Resolver Template

penard-cb:
My previous code #2720 introduced a bug when setting the package variable of `resolverTemplate`. I tested the code changes and noticed that subsequent runs would still use the template even if not set.

We need to make a copy of the template so that subsequent calls don't use the same template as the previous run. 😓

### DIFF
--- a/plugin/resolvergen/resolver.go
+++ b/plugin/resolvergen/resolver.go
@@ -81,8 +81,9 @@ func (m *Plugin) generateSingleFile(data *codegen.Data) error {
 		OmitTemplateComment: data.Config.Resolver.OmitTemplateComment,
 	}
 
+	newResolverTemplate := resolverTemplate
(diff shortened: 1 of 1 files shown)


StevenACoffman:
Thanks for catching this! Luckily, I hadn't pushed out a new release with the bug. I appreciate your working on this, and for diligently testing it. Do mind commenting on some of your motivating use cases here for posterity?

I'm also looking forward to your next contribution here! Thanks for your good work!

penard-cb:
Thanks @StevenACoffman! 

The motivation on this is to ensure I can apply some boilerplate logic to my resolvers. This will help implementation teams understand what needs doing a bit better and get help. Also in some schemas I can template out the full resolver implementation as it is fairly straight forward just extremely repetitive.

This makes it possible for me to do that per schema.

PR #2729, opened 2023-07-25
Make models configurable via template

penard-cb:
Similarly as in #2720 and #2725 this PR makes the models configurable as well via template. This opens up a lot of new options for configuration and using the [templating](https://pkg.go.dev/text/template) functionality

I have:

### DIFF
--- a/codegen/config/package.go
+++ b/codegen/config/package.go
@@ -10,9 +10,10 @@ import (
 )
 
 type PackageConfig struct {
-	Filename string `yaml:"filename,omitempty"`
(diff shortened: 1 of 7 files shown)


penard-cb:
I force pushed after rebasing. Should I not have done that? now some of the checks failed. Do I need to spin up a new PR? 
😩

cc: @StevenACoffman thoughts?

penard-cb:
I'm just going to abandon this one in favor of #2730 which i didn't rebase.

PR #2730, opened 2023-07-25
Make models configurable via template

penard-cb:
Similarly as in #2720 and #2725 this PR makes the models configurable as well via template. This opens up a lot of new options for configuration and using the [templating](https://pkg.go.dev/text/template) functionality

I have:

### DIFF
--- a/codegen/config/package.go
+++ b/codegen/config/package.go
@@ -10,9 +10,10 @@ import (
 )
 
 type PackageConfig struct {
-	Filename string `yaml:"filename,omitempty"`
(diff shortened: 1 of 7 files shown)


StevenACoffman:
This is pretty neat! I use sqlc + gqlgen, and quite frequently the models are duplicative, but you cannot easily translate from one struct type to the other. Using [mapstructure](https://github.com/mitchellh/mapstructure/) requires you to add a tedious amount of yaml config for gqlgen so that the structs will be annotated properly.
```
type User struct {
    Username string `mapstructure:"user"`
}
```

PR #2738, opened 2023-07-28
On a given interface store which models implement it

penard-cb:

On any given model we know of which interface it is. However there are cases where if you're writing some custom models, you would like to know what a type is implemented by:

For example:
```

type EntryCollection {
  total: Int!
  skip: Int!
  limit: Int!
  items: [Entry]!
}

interface Entry {
  sys: Sys!
  metadata: metadata!
} 

type Page implements Entry {
  sys: Sys!
  metadata: metadata!
  linkedFrom(allowedLocales: [String]): PageLinkingCollections
  name(locale: String): String
  ...
}
type PageArticle implements Entry {
  sys: Sys!
  metadata: metadata!
  linkedFrom(allowedLocales: [String]): PageArticleLinkingCollections
  name(locale: String): String
  ...
}
```

will allows for on a type like Page we'll know it implements an entry. But if we want to add additional model functionality for an interface, we don't know all the types that implement it. This PR makes it possible to add complex model definitions and helpers (marshalers/unmarshalers) for the Entry type.


StevenACoffman:
Thanks! I'm loving these improvements!

PR #2749, opened 2023-08-05
Introduce support for Universally Unique Identifiers

### LINKED ISSUE #2748: Built-in support for UUID type(s)
`gqlgen` should understand how to (un)marshal UUID types.

x80486:
A **Universally Unique Identifier** (UUID) is a 128-bit used for information in computer systems. The term **Globally Unique Identifier** (GUID) is also used, mostly in Microsoft systems.

Because Go (still) does not have built-in support for UUID types, module `gofrs/uuid` is imported. It's the most complete, correct (and starred) module nowadays.

This pull request introduces support for Universally Unique Identifiers (UUID) within `gqlgen`.

I have:

---

Closes #2748

### DIFF
--- a/graphql/uuid.go
+++ b/graphql/uuid.go
@@ -1,0 +1,24 @@
+package graphql
+
+import (
+	"errors"
+	"io"
+
+	"github.com/gofrs/uuid"
+)
+
+func MarshalUUID(t uuid.UUID) Marshaler {
+	if t.IsNil() {
+		return Null
+	}
+	return WriterFunc(func(w io.Writer) {
+		_, _ = io.WriteString(w, t.String())
+	})
+}
+
+func UnmarshalUUID(v interface{}) (uuid.UUID, error) {
+	if str, ok := v.(string); ok {
+		return uuid.FromString(str)
+	}
+	return uuid.Nil, errors.New("input must be an RFC-4122 formatted string")
+}
--- a/_examples/go.mod
+++ b/_examples/go.mod
@@ -19,6 +19,7 @@ require (
 require (
 	github.com/agnivade/levenshtein v1.1.1 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/gofrs/uuid v4.4.0+incompatible // indirect
 	github.com/hashicorp/golang-lru/v2 v2.0.3 // indirect
 	github.com/logrusorgru/aurora/v3 v3.0.0 // indirect
 	github.com/mattn/go-colorable v0.1.13 // indirect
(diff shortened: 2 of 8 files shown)


StevenACoffman:
Hi! So can you verify that with your PR someone can still override this type and use https://github.com/google/uuid ?

I would note that google/uuid is still more widely used than gofrs/uuid per this:
https://www.grank.io/pkg/github.com/google/uuid.html

x80486:
> Hi! So can you verify that with your PR someone can still override this type and use https://github.com/google/uuid ?

I will have to figure it out how to test that :grimacing:

> I would note that google/uuid is still more widely used than gofrs/uuid per this: https://www.grank.io/pkg/github.com/google/uuid.html
> 
> I don't mind picking a default implementation, but I would like to provide a working example that successfully overrides with an alternative since there are a number of popular choices that people are already using.

The Google implementation is the most popular indeed, but it looks like it's not actively maintained.

On the other hand, what are those failures in the pipeline? As far as I can tell it's complaining about `gofrs/uuid` not in `go.sum`, but it's there. I did try `go generate ./...` locally but it doesn't change anything.

x80486:
Nevermind. I was lurking in some other pull requests, and I saw the `go generate ./...` _trick_ inside `_examples`.

x80486:
If the relevant changes are done in `id.go`, and there is another `case` clause for `UUID` types...it is possible to use `UUID` transparently just by declaring it in the `Go` models, right?

StevenACoffman:
> If the relevant changes are done in id.go, and there is another case clause for UUID types...it is possible to use UUID transparently just by declaring it in the Go models, right?

I'm sorry, but I'm kind of confused by the wording of your question. Can you please rephrase that?

This PR is just like PR #2751 but here the choice is github.com/gofrs/uuid instead of there it is github.com/google/uuid

Many people will have already chosen one of those two (or something else) and so will have something like this in their config:
```yaml
models:
  UUID:
    model:
      - github.com/google/uuid
```

Please add to this PR a similar example that verifies that people can still override the default choice after your PR is applied.

x80486:
I didn't realize there was another pull request for the same work. I'm fine with either `UUID` type. I'll discard this one so you don't have to review the same work twice.

StevenACoffman:
Actually, yours was better quality, and it passed the tests! 😞

x80486:
I have another pull request to be able to use `ID` with `UUID`s as well. That's why I was asking if those changes should be done under `id.go`. I see a `switch` clause that's able to handle multiple types. I'm also planning to have a working example as well for that one.

graphql/uuid.go
```diff
@@ -0,0 +1,24 @@
...
+
+	"github.com/gofrs/uuid"
+)
+
+func MarshalUUID(t uuid.UUID) Marshaler {
+	if t.IsNil() {
+		return Null
+	}
+	return WriterFunc(func(w io.Writer) {
+		_, _ = io.WriteString(w, t.String())
```
it512, line 15: writeQuotedString(w, t.String()) ?
uuid type marsher is "00000000-0000-0000-0000-00    0000000000"

must Quoted

x80486: I don't fully understand why is that needed here. I think that might be the case for arbitrary string values that should be double-quoted because they could have escaped/control characters and non-printable ones, but the `uuid.UUID` will never have those.

StevenACoffman:
I really appreciate this contribution, as it had excellent documentation (and great tests!). I ended up applying your documentation and using #2751 as it used `github.com/google/uuid` which my company and most of the other maintainers use at work, despite the fact that `gofrs/uuid` is more complete and is actively maintained. 

It took a long time to discuss (argue) about which PR to merge, and we were pretty conflicted, so I'm very sorry for the delay (and that we did not pick merging this PR as is).

I hope that doesn't dissuade you from future contributions, as this was excellent quality.

PR #2751, opened 2023-08-07
add uuid type

it512:
A **Universally Unique Identifier** (UUID) is a 128-bit used for information in computer systems. The term **Globally Unique Identifier** (GUID) is also used, mostly in Microsoft systems.

Because Go (still) does not have built-in support for UUID types, module `google/uuid` is imported. It's one of the most complete, correct (and starred) modules nowadays (the other being `github.com/gofrs/uuid.UUID`).

This pull request introduces support for Universally Unique Identifiers (UUID) within `gqlgen`.

Many people may have already chosen a different UUID library:
```yaml
models:
  UUID:
    model:
      - github.com/gofrs/uuid.UUID
```


StevenACoffman:
Ok, so this is just like PR #2749 but here the choice is github.com/google/uuid instead of there it is github.com/gofrs/uuid

Many people will have already chosen one of those two (or something else) and so will have:
```yaml
models:
  UUID:
    model:
      - github.com/gofrs/uuid.UUID
```

Please add an example that uses this to verify that people can still override the default choice.

it512:
## UUID type

 add to gqlgen.yalm

 ```yaml
 models:
   UUID:
     model:
       - github.com/99designs/gqlgen/graphql.UUID
 ```

 add UUID scalar to schema.graphqls

 ```graphql
 scalar UUID
 ```
 See the _examples/uuid package for more examples.

it512:
add example in _examples/uuid  and modify scalar doc (docs/content/reference/scalar.md)

it512:
Does  need to add UUID to ID Model?
such as

models:                                                                                                             
  ID:                                                                                                               
    model:                                                                                                          
       - github.com/99designs/gqlgen/graphql.UUID

@StevenACoffman

it512:
ok， thank you！

StevenACoffman:
Thanks for your patience. I was quite ill for a while, and I'm only now getting caught up at my day job so I can return to gqlgen. I added a commit that used the documentation from #2749 here (thanks!) and I'm merging this one.

0x221A:
Is it better to return the `null` value instead of the zero value uuid?

PR #2770, opened 2023-08-26
vikstrous/dataloadgen replaces recommended dataloader package in example docs

### LINKED ISSUE #2755: replace recommended dataloader package
The docs at https://gqlgen.com/reference/dataloaders/ show an example of how to use the graph-gophers dataloader package.

There are some concerns I have with it:
* many places in the example where types have to be explicitly specified.
* error handling requires a custom "handleError" function in the docs
* it performs poorly in my benchmarks (not sure if this is important because they are very obviously synthetic and network latency is the real bottleneck)

I've been using this alternative package I built for a long time in production and it has a simpler API and implementation: https://github.com/vikstrous/dataloadgen

Should I make a pull request updating the docs? Is there any concern with merging such a change?

vikstrous:
This replaces the recommended dataloader package in example docs from [graph-gophers/dataloader](https://github.com/graph-gophers/dataloader) to [vikstrous/dataloadgen](https://github.com/vikstrous/dataloadgen)

This closes #2755
I updated the docs for the dataloader example in this PR.
I simplified the code so there would be less boilerplate.
I updated userIds to userIDs to follow the more common convention in Go.

~I haven't tested the example code.~
~I haven't made an example repo like this one https://github.com/zenyui/gqlgen-dataloader, but I'm thinking I can make one.~
There's an example repo at https://github.com/vikstrous/dataloadgen-example now.

~I recommend not merging this as is, but I want to get some feedback before I continue.~

### DIFF
--- a/docs/content/reference/dataloaders.md
+++ b/docs/content/reference/dataloaders.md
@@ -72,8 +72,9 @@ go get github.com/vikstrous/dataloadgen
 Next, we implement a data loader and a middleware for injecting the data loader on a request context.
 
 ```go
-package loader
+package loaders
 
+// import vikstrous/dataloadgen with your other imports
 import (
 	"context"
 	"database/sql"
@@ -84,46 +85,20 @@ import (
 	"github.com/vikstrous/dataloadgen"
 )
 
-// Get returns the Loaders bundle from the context. It must be used only in XXXXXXXXXXXX resolvers where Middleware has put the Loaders struct into the context already.
-func Get(ctx context.Context) *Loaders {
-	return ctx.Value(ctxKey{}).(*Loaders)
-}
+type ctxKey string
 
+const (
+	loadersKey = ctxKey("dataloaders")
+)
 
-// Loaders provide access for loading various objects from the underlying object's storage system while batching concurrent requests and caching responses.
-type Loaders struct {
-	User *dataloadgen.Loader[string, *model.User]
-}
-
-// Middleware injects data loaders into the context
-func Middleware(conn *sql.DB, next http.Handler) http.Handler {
-	// return a middleware that injects the loader to the request context
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		// Note that the loaders are being created per-request. This is important because they contain caching and batching logic that must be request-scoped.
-		loaders := newLoaders(conn)
-		r = r.WithContext(context.WithValue(r.Context(), ctxKey{}, loaders))
-		next.ServeHTTP(w, r)
-	})
-}
-
-type ctxKey struct{}
-
-// newLoaders creates the Loaders struct
-func newLoaders(conn *sql.DB) *Loaders {
-	ur := &userFetcher{db: conn}
-	return &Loaders{
-		User: dataloadgen.NewLoader(ur.getUsers, dataloadgen.WithWait(time.Millisecond)),
-	}
-}
-
-// userFetcher reads Users from a database
-type userFetcher struct {
+// userReader reads Users from a database
+type userReader struct {
 	db *sql.DB
 }
 
 // getUsers implements a batch function that can retrieve many users by ID,
 // for use in a dataloader
-func (u *userFetcher) getUsers(ctx context.Context, userIDs []string) ([]*model.User, []error) {
+func (u *userReader) getUsers(ctx context.Context, userIDs []string) ([]*model.User, []error) {
 	stmt, err := u.db.PrepareContext(ctx, `SELECT id, name FROM users WHERE id IN (?`+strings.Repeat(",?", len(userIDs)-1)+`)`)
 	if err != nil {
 		return nil, []error{err}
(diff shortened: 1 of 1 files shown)


vikstrous:
While working through the process of creating an example repo,, I realized that the current example (as well as the way my PR is written right now) creates a long lived loader, which is a mistake because that causes the caches to be reused across graphql requests. I'll update my example to show the correct pattern of creating a loader per request.

StevenACoffman:
@vikstrous Thanks! I have been sick, but the reason I delayed commenting was that I was trying to find time to demonstrate exactly that problem. If you could do a demo repo and update the docs like you are describing, I would love to merge it. I really appreciate this and your past contributions.

vikstrous:
I ran into the problem while I was building out my example repo https://github.com/vikstrous/dataloadgen-example. It's ready now, so you can take a look there.

In my example repo, I added a print statement and a `time.Sleep(time.Second)` to every storage call so that while I'm developing the example I can see exactly what requests are made and how they perform.

For the documentation issue that existed before, it seems to have been a relatively minor mistake. The current @zenyui 's example repo has the correct version. You can see in the current example that the middleware is instantiated with `loaders.Middleware(db, srv)`, but the `Middleware()` function takes `loaders *Loaders` instead of `db`. To fix the current example, you'd change `Middleware()` to take `db` instead and call `NewLoaders(db)` per request inside `HandlerFunc`.

My PR is not necessary to fix that issue.  Feel free to fix that separately while considering switching the example and recommended data loader implementation.

vikstrous:
I think you'll notice that the previous example encouraged top level functions like `loader.GetUser(ctx, obj.UserID)`. They do look pretty, but I felt like they added unnecessary indirection, so in my example, I encourage interacting with the loader object directly, like `loader.Get(ctx).User.Load(ctx, obj.UserID)`. I don't really want to recommend one over the other. I just wanted to keep things simple.

The top level functions have the benefit of providing a more implementation agnostic API and not exposing a mutable struct to the caller at the expense of more boilerplate. I don't feel strongly about any of this. Let me know if you'd prefer it if I don't change those aspects of the example.

zenyui:
@vikstrous cool -- so if i'm understanding correctly, you'd like to see the docs propose a short-lived loader? I can get behind that and make the change in the sample/docs if that helps.

zenyui:
@vikstrous @StevenACoffman how's this? https://github.com/99designs/gqlgen/pull/2778

Also, regarding your point above about the indirect with the `GetUser` method -- the reason I might advocate for this simple method is the type is more simple, so providing a DataLoader interface and mocking it in tests is way easier. The sample code doesn't demonstrate this, but it's what I do in my production app and it works well.

Of course, you can make a mocked `DataLoaders` implementation that returns the thunk, etc, but it makes testing much more verbose 🤷‍♂️

vikstrous:
Thanks for the docs fix PR!

Re: the API: Oh, interesting. I haven't tried mocking loaders. I normally mock the underlying storage. In any case, I think there are a lot of valid ways to set up the API for accessing the loaders. The way we do it in our proprietary code is completely different and way too complex for a one page example, so I wouldn't want to show the real way I've done it :)

Hopefully @StevenACoffman can make the call on what API should be shown in the docs. I'll happy to rewrite my example to show any pattern

StevenACoffman:
> In any case, I think there are a lot of valid ways to set up the API for accessing the loaders

So whatever seems simplest and most approachable seems like the thing to go with in the docs, and maybe just add a mention of alternatives that might be more complicated but desirable in other circumstances.

It looks like @zenyui mock loaders is currently a little more succinct, but it would also be nice to at least link to a mock storage example.

vikstrous2:
I reverted all non-essential changes and updated my example repo to follow the same pattern that's currently in the docs. https://github.com/vikstrous/dataloadgen-example/blob/master/graph/loader/loader.go

I think there are better patterns to show, but this works and I don't want to change too much in one PR and delay merging.

docs/content/reference/dataloaders.md
```diff
@@ -161,16 +153,17 @@ func For(ctx context.Context) *Loaders {
 	return ctx.Value(loadersKey).(*Loaders)
 }
 
+
```
vikstrous, line 156:
```suggestion
```

zenyui:
@vikstrous would love to play with your repo. I do think we should get some more community consensus / feedback before migrating to a new dataloader library...

vikstrous:
Yeah, no problem. Let me know how I can help

vikstrous2:
How can we make progress on this?

StevenACoffman:
@vikstrous2 I don't personally have a strong preference, so I was hoping to see evidence that community preferences were changing to your library before merging this PR, however, it's possible that the status quo documentation is discouraging that very thing!

Rather than replace the documentation on how to use [graph-gophers/dataloader](https://github.com/graph-gophers/dataloader) with [vikstrous/dataloadgen](https://github.com/vikstrous/dataloadgen), how documenting it twice and giving people the choice?

Currently, I'm not seeing a lot of community consensus that we should definitely use [vikstrous/dataloadgen](https://github.com/vikstrous/dataloadgen) instead of [graph-gophers/dataloader](https://github.com/graph-gophers/dataloader) for all cases:
+ https://pkg.go.dev/github.com/vikstrous/dataloadgen?tab=importedby
+ https://pkg.go.dev/github.com/graph-gophers/dataloader?tab=importedby

vikstrous2:
Yeah, I assume most people just follow the docs and have never searched for an alternative. Having both sounds okay. How do you recommend structuring things with both? Maybe it's better for you to take over from here?

zenyui:
My 2 cents - [graph-gophers/dataloader](https://github.com/graph-gophers/dataloader) works very well, especially with the updates to use generics, and it has significant adoption in production. I think Golang generics eliminate the need for codegen in a dataloader, as there isn't really anything dynamic about how a dataloader works given a type -- it's really just a generic problem of consolidating data fetching calls.

Now, super interested in your performance benchmarks -- do you think this is something you could introduce to [graph-gophers/dataloader](https://github.com/graph-gophers/dataloader)?

vikstrous:
I updated the benchmarks and wrote a blog post explaining some of the reasoning behind dataloadgen. https://viktorstanchev.com/posts/why-i-wrote-yet-another-data-loader-in-go/ 

The benchmarks cover all 4 packages I compared.

it512:
gqlgen need a plugin to generate data loader , if you can, I think that is cool !
@vikstrous

StevenACoffman:
Your article was great! At Khan, we always preferred your interface, but the benchmarks were very compelling. While [raph-gophers/dataloader](https://github.com/graph-gophers/dataloader) is still pretty popular, I'm going to go ahead and merge this PR as is.

If someone wants to copy-paste the original [raph-gophers/dataloader](https://github.com/graph-gophers/dataloader) 
 instructions into an alternative section for folks, I would merge that.

StevenACoffman:
thanks for your patience!

vikstrous2:
@it512 cool idea, but with dataloadgen, you don't need generated code as much and the code is very specific to how you structure your codebase. If you have a specific idea, feel free to email me more details or put it in an issue on this repo :)

PR #2779, opened 2023-09-05
fix flaky test TestSubscriptions

vikstrous:
The test failed in https://github.com/99designs/gqlgen/pull/2770

I noticed that the pattern of waiting for goroutines to match the number they were at the start of the test was missing from the second subtest.

I triggered the failure locally with
```
go test -race -test.run  'TestSubscriptions' -count 10000
```
This takes 3-4 minutes for me.

After the change, the test no longer failed after 10k iterations.

I noticed that there's a second copy of the same test and copied the fix to it too.


StevenACoffman:
OMG, I have wanted this for soooooo long. That's why I made the tests retry. Thanks!

PR #2793, opened 2023-09-14
Make it possible to pass UI headers to GraphiQL

marcusthelin:
#2740 made it possible to pass headers to the playground, but this doesn't cover all use cases. Instead of just being able to set the GraphiQL's fetcher headers, it would be nice to also be able to set default headers in the UI.

These changes very much mimics what was done in #2740.


marcusthelin:
@StevenACoffman I'm not sure why federation test failed 😅

StevenACoffman:
Just a transient error... I retried and it worked fine. Not sure why it is flaky though.

marcusthelin:
Great! Feel free to review

Issue #2801, opened 2023-09-22
Federation @entityResolver(multi: true) Breaks w/ omit_slice_element_pointers: true


parkerroan:
### What happened?
If you enable `omit_slice_element_pointers: true` and also use the federation plugin directive `@entityResolver(multi: true)` generated code will break. 

```
validation failed: packages.Load: /graph/generated/federation.go: cannot use _reps (variable of type []*model.HelloByNamesInput) as []model.HelloByNamesInput value in argument to ec.resolvers.Entity().FindManyHelloByNames
```

### What did you expect?
I would expect code generation to still work with these settings. 

### Minimal graphql.schema and models to reproduce

gqlgen.yml
```
federation:
  filename: graph/generated/federation.go
  package: generated
  version: 2
 
omit_slice_element_pointers: true
```

schema.graphql
```
type Hello @key(fields: "name") @entityResolver(multi: true) {
    name: String!
    secondary: String!
}
```

### versions
 - gqlgen version 0.17.38
 - go 1.18.1
 
 
 I am proposing we allow `omit_slice_element_pointers: true`  to work as expected but in this specific federated use case we just have pointers used the for FindMany entity resolvers in the array.
 
 If we change this [line](https://github.com/99designs/gqlgen/blob/af4d3943308b7a84464771e5ed6b578ba40ee0c3/plugin/federation/federation.go#L176) 
 
to this: 
 `resolvers += fmt.Sprintf("\t%s(reps: [%s]!): [%s]\n", r.ResolverName, r.InputTypeName, e.Name)`
 
 then generation will work and will consistently create `reps []*model.HelloByNamesInput` in the above case.

PR #2802, opened 2023-09-22
fix: CodeGen for omit_slice_element_pointers and GetMany Entity Resolvers

parkerroan:

Currently if you enable `omit_slice_element_pointers: true` and also use the federation plugin directive `@entityResolver(multi: true)` your generated code will break.  This is because `*` is hard coded in a few places in the federation plugin for this feature and when it is absent from the resolver signature the entity resolvers do not match. 

This PR simply makes the output of `@entityResolver(multi: true)` consistent despite the value of `omit_slice_element_pointers: true` allow users to continue to use both features without being forced to use 
`omit_slice_element_pointers: false` everywhere like it currently required if you enable multi. 

More details in the issue below
#2801 

I have:

### DIFF
--- a/codegen/testserver/followschema/models-gen.go
+++ b/codegen/testserver/followschema/models-gen.go
@@ -239,7 +239,7 @@ type ValidInput struct {
 	Underscore  string `json:"_"`
 }
 
-//  These things are all valid, but without care generate invalid go code
+// These things are all valid, but without care generate invalid go code
 type ValidType struct {
 	DifferentCase      string `json:"differentCase"`
 	DifferentCaseOld   string `json:"different_case"`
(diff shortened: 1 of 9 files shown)


parkerroan:
@StevenACoffman any hesitation with this change? Its a one char change, and does not change the current functionality. 

The generated code with the default setting `omit_slice_element_pointers: false` still will produce the same as it did prior. But it just allows for repo that set  `omit_slice_element_pointers: true` to also use the multi feature without breaking generation.

StevenACoffman:
Looks good, but I was waiting for you to sort out the lint failures.
Perhaps just try:
```
go generate ./...; cd _examples; go generate ./...
```

PR #2811, opened 2023-09-28
Store parsed Schema on ExecutableSchema rather than use global variable

gitxiongpan:
This PR moves from using the global variable `parsedSchema` to storing it on the `ExecutableSchema`.

---

In my use case, I have two GraphQL servers. One server's schema is trimmed down and it is publicly accessible. Another one  has a comprehensive schema and it is protected by a gateway 

I would like an easy way to implement these two GraphQL servers. At the moment `executableSchema.Schema()` returns a global variable `parsedSchema`. This behavior makes my use case quite challenging to solve. 
see code below
```
//go:embed "schema.graphql"
var sourcesFS embed.FS

func sourceData(filename string) string {
	data, err := sourcesFS.ReadFile(filename)
	if err != nil {
		panic(fmt.Sprintf("codegen problem: %s not available", filename))
	}
	return string(data)
}

var sources = []*ast.Source{
	{Name: "schema.graphql", Input: sourceData("schema.graphql"), BuiltIn: false},
}
var parsedSchema = gqlparser.MustLoadSchema(sources...)

......

func (e *executableSchema) Schema() *ast.Schema {
	return parsedSchema
}
```

I updated the `codegen/generated!.gotpl` so that executableSchema is configurable 

This update is backward compatible which means the change is almost transparent to those with no interests in hosting multiple GraphQL servers with different exposed schemas.

I have:


codegen/generated!.gotpl
```diff
@@ -268,14 +274,14 @@
...
-		return introspection.WrapSchema(parsedSchema), nil
+		return introspection.WrapSchema(ec.Schema()), nil
 	}
 
 	func (ec *executionContext) introspectType(name string) (*introspection.Type, error) {
 		if ec.DisableIntrospection {
 			return nil, errors.New("introspection disabled")
 		}
-		return introspection.WrapTypeFromDef(parsedSchema, parsedSchema.Types[name]), nil
+		return introspection.WrapTypeFromDef(ec.Schema(), ec.Schema().Types[name]), nil
```
gitxiongpan, line 284: I reckon line 277 and 284 have better coding style no matter we decide to make executableSchema configurable or not

gitxiongpan:
screenshots for vistual demo
image
image

StevenACoffman:
Hey it looks like you need to regenerate:
```
go generate ./...; cd _examples; go generate ./...
```

gitxiongpan:
> Hey it looks like you need to regenerate:
> 
> ```
> go generate ./...; cd _examples; go generate ./...
> ```

Thank you! It is my first time working on codegen feature. I didn't know I need to regenerate. 
Regenerated, please let me know if I am still in the wrong trrack

gitxiongpan:
I can see that Link/lint(1.19) is failed. I am not sure how I could fix it though.

codegen/generated!.gotpl
StevenACoffman: Avoiding global variables is good, but `ec.Schema().Types[name]` violates the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter).

If any of `ec`, `ec.Schema()`, or `ec.Schema().Types` are nil, this will have a fatal error (panic due to nil pointer dereference) that is difficult to figure out which of these happened.

A temporary local variable might be more verbose, but in the event of a nil pointer dereference, the line number would then tell you immediately which of those was problematic.

PR #2812, opened 2023-09-28
Consider go type name when autobinding

dany74q:
Currently, generated schema type names are normalized, for instance - SomeTYPE in the schema will be generated as SomeType in the model.

When autobinding, however, we only consider the schema type name when searching for it in the relevant package(s),
thus type names that differ post normalizations aren't auto-bound properly and are instead re-generated.

This commit suggests a fix where we'd try to autobind for both the schema type name (first, to maintain back compat), or the go type name if the former isn't found.

I have:

### DIFF
--- a/codegen/config/binder.go
+++ b/codegen/config/binder.go
@@ -5,12 +5,12 @@ import (
 	"fmt"
 	"go/token"
 	"go/types"
-	"strings"
 
+	"github.com/vektah/gqlparser/v2/ast"
 	"golang.org/x/tools/go/packages"
 
+	"github.com/99designs/gqlgen/codegen/templates"
 	"github.com/99designs/gqlgen/internal/code"
-	"github.com/vektah/gqlparser/v2/ast"
 )
 
 var ErrTypeNotFound = errors.New("unable to find type")
@@ -285,7 +285,7 @@ func (ref *TypeReference) UniquenessKey() string {
 		// Fix for #896
 		elemNullability = "ᚄ"
 	}
-	return nullability + ref.Definition.Name + "2" + TypeIdentifier(ref.GO) + elemNullability
+	return nullability + ref.Definition.Name + "2" + templates.TypeIdentifier(ref.GO) + elemNullability
 }
 
 func (ref *TypeReference) MarshalFunc() string {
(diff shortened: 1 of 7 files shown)


dany74q:
This could be solved w/ config modifications today by:
1. Binding to the normalized type names explicitly in the models section, or
2. Adding any unsupported suffixes to the list of go initialisms 

However I think that, although a potentially breaking change in a minutiae of cases, this could be beneficial to happen automatically.

Would love to hear your thoughts !

codegen/config/binder.go
```diff
@@ -285,7 +285,7 @@ func (ref *TypeReference) UniquenessKey() string {
 		// Fix for #896
 		elemNullability = "ᚄ"
 	}
-	return nullability + ref.Definition.Name + "2" + TypeIdentifier(ref.GO) + elemNullability
+	return nullability + ref.Definition.Name + "2" + templates.TypeIdentifier(ref.GO) + elemNullability
```
dany74q, line 288: Moved due to cyclic dep, only ref is in templates, so thought it fits there well

codegen/config/initialisms.go
```diff
@@ -69,7 +17,7 @@ type GoInitialismsConfig struct {
 // setInitialisms adjustes GetInitialisms based on its settings.
 func (i GoInitialismsConfig) setInitialisms() {
 	toUse := i.determineGoInitialisms()
-	GetInitialisms = func() map[string]bool {
+	templates.GetInitialisms = func() map[string]bool {
```
dany74q, line 20: Moved due to cyclic dep, only ref is in templates, so thought it fits there well

codegen/templates/templates.go
```diff
@@ -668,3 +667,97 @@ func write(filename string, b []byte, packages *code.Packages) error {
...
+	"VM":    true,
+	"XML":   true,
+	"XMPP":  true,
+	"XSRF":  true,
+	"XSS":   true,
+}
+
+// GetInitialisms returns the initialisms to capitalize in Go names. If unchanged, default initialisms will be returned
+var GetInitialisms = func() map[string]bool {
+	return CommonInitialisms
```
dany74q, line 762: Moved due to cyclic dep, only ref is in templates, so thought it fits there well

StevenACoffman:
Thanks! This is great. I am really looking forward to your next PR!

PR #2815, opened 2023-09-29
Add ability to not fail when pong is not received.

telemenar:
I also changed how the read deadline set works a little, the reason for this is that the protocol allows for pong to be sent without a ping.

So setting a read deadline on receiving pong isn't great. Instead we should always set the read deadline on sending ping. Though to do this we need to know whether we have received a pong or not. Because if we set the read deadline when the previous ping still hasn't received the pong. Then it will never hit the deadline.

I have:

### DIFF
--- a/graphql/handler/transport/websocket.go
+++ b/graphql/handler/transport/websocket.go
@@ -27,6 +27,15 @@ type (
 		CloseFunc             WebsocketCloseFunc
 		KeepAlivePingInterval time.Duration
 		PingPongInterval      time.Duration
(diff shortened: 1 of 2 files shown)


StevenACoffman:
Thanks! I appreciate you giving more attention to the websocket support in particular, as I'm not actively using it in production at the moment.

Do you mind looking over https://github.com/99designs/gqlgen/blob/master/docs/content/config.md and the other documentation that might be good to update for some of your recent changes to websockets?

PR #2699, opened 2023-06-28
fix: Skip federation2 directives generation

rlch:
This PR excludes the federation 2 directives added in #2635 from being added to `DirectiveRoot`. Without this exclusion, usage of `@interfaceObject` or `@composeDirective` fails at runtime with something like:

```json
{
  "data": null,
  "errors": [
    {
      "message": "directive interfaceObject is not implemented",
      "path": [
        "startTest",
        "currentQuestion"
      ]
    }
  ],
  "extensions": {
    "valueCompletion": [
      {
        "message": "Cannot return null for non-nullable field Mutation.startTest",
        "path": []
      }
    ]
  }
}
```

Please let me know if there's anything to do for this PR. I couldn't find documentation on how to regenerate `testdata/`, unsure if `go generate ./...` was sufficient. Thanks for reviewing + for this great package (and hello from Melbourne)!

I have:
 - [ ] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [ ] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

PR #2821, opened 2023-10-04
Add @interfaceObject and @composeDirective at Federation 2 directive lists.

wangmir:
This is just an update of PR #2699 
---
This PR excludes the federation 2 directives added in #2635 from being added to `DirectiveRoot`. Without this exclusion, usage of `@interfaceObject` or `@composeDirective` fails at runtime with something like:

```json
{
  "data": null,
  "errors": [
    {
      "message": "directive interfaceObject is not implemented",
      "path": [
        "startTest",
        "currentQuestion"
      ]
    }
  ],
  "extensions": {
    "valueCompletion": [
      {
        "message": "Cannot return null for non-nullable field Mutation.startTest",
        "path": []
      }
    ]
  }
}
```
---
Currently, when using `@interfaceObject`, the GraphQL schema generates the directive template in the auto-generated file because those directives are not included in the list of federation directives, and so are not excluded from the code generation.

So, it causes a "not implemented" error when using `@interfaceObject` even though this directive is a federation 2.3 directive.

This commit just adds those directives to the list.

I have:


StevenACoffman:
Thanks!

PR #2829, opened 2023-10-15
resolver: fix case-insensitive file name collision

erankor:
for compatibility with Windows/Mac OS X, Go does not allow files with identical case-insensitive names.
this commit changes the key of the 'files' map to use the lower case file name, keeping the original file name as a property on the File object.

I have:

### DIFF
--- a/plugin/resolvergen/resolver.go
+++ b/plugin/resolvergen/resolver.go
@@ -110,9 +110,12 @@ func (m *Plugin) generatePerSchema(data *codegen.Data) error {
 
 	for _, o := range objects {
 		if o.HasResolvers() {
-			fn := gqlToResolverName(data.Config.Resolver.Dir(), o.Position.Src.Name, data.Config.Resolver.FilenameTemplate)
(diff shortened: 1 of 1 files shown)


erankor:
To add a bit more color on the issue we've encountered - 
We have a type `Metric` in our schema, and we have a query `metric` for getting a single `Metric` object. The code of the `metric` resolver is generated to `metric.resolvers.go` (=based on the name of `metric` query). 
In addition, the `Metric` type has a field that accepts parameters. We are using `@goField(forceResolver: true)` on all parameterized fields, so that it will work as expected if such a field is requested multiple times with different parameters. The code of the field resolver is generated to `Metric.resolvers.go` (=based on the name of the `Metric` type)
So - we end up with two files with identical (case-insensitive) names, which Go does not allow.

StevenACoffman:
Thanks! This is the kind of issue that is easy to overlook, but can cause a lot of frustration, so I appreciate you fixing it. Looking forward to your next PR!

erankor:
Thanks @StevenACoffman !

PR #2830, opened 2023-10-18
Map based input types fields are now coerced to the right type

endSly:
Right now, input types mapped as `map[string]interface{}` were being returned as processed in `RawArgs`, that means that, for example nested input types were being returned as `map[string]interface{}` and scalars as `strings` without calling `UnmarshalGQL` contrary to how output types are treated.

I have:


StevenACoffman:
Oh, nice catch! Thanks very much for this. Looking forward to your next PR!

PR #2841, opened 2023-11-08
Small template fix to save space in the generated file

PatrLind:
This fix saves some space in the generated.go file.
It just removes some unused code.

### DIFF
--- a/_examples/config/generated.go
+++ b/_examples/config/generated.go
@@ -2832,8 +2832,6 @@ func (ec *executionContext) unmarshalInputNewTodo(ctx context.Context, obj inter
 		}
 		switch k {
 		case "text":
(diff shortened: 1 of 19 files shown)


StevenACoffman:
It looks like you would need to run `go generate ./...; cd _examples; go generate ./...`

PatrLind:
I did that, but it generates no files to check in.
It seems the test forbids anyone to change the templates, so I guess this PR is not allowed. I should remove it probably.

StevenACoffman:
`_examples` is a [separate module](https://github.com/99designs/gqlgen/blob/master/_examples/go.mod), so you need to re-generate from inside the `_examples` directory. I did that, and the linting works now.

Issue #2678, opened 2023-06-18
I have staticcheck's error during upgrading gqlgen from v0.17.31 to v0.17.33.


koheiyamayama:
# What happened
I have staticcheck's error during upgrading gqlgen from v0.17.31 to v0.17.33.
The method to reproduce this error is below.

```
$ cd /path/to/gqlgen-reporduce-error
$ go generate ./...
$ staticcheck --version
staticcheck 2023.1.3 (v0.4.3)
$ staticcheck ./...
graph/generated.go:3258:2: unreachable case clause: github.com/koheiyamayama/gqlgen-reproduce-error/graph/model.Post will always match before github.com/koheiyamayama/gqlgen-reproduce-error/graph/model.TweetPost (SA4020)
graph/generated.go:3260:2: unreachable case clause: github.com/koheiyamayama/gqlgen-reproduce-error/graph/model.Post will always match before *github.com/koheiyamayama/gqlgen-reproduce-error/graph/model.TweetPost (SA4020)
graph/generated.go:3265:2: unreachable case clause: github.com/koheiyamayama/gqlgen-reproduce-error/graph/model.Post will always match before github.com/koheiyamayama/gqlgen-reproduce-error/graph/model.FacebookPost (SA4020)
graph/generated.go:3267:2: unreachable case clause: github.com/koheiyamayama/gqlgen-reproduce-error/graph/model.Post will always match before *github.com/koheiyamayama/gqlgen-reproduce-error/graph/model.FacebookPost (SA4020)
graph/generated.go:3272:2: unreachable case clause: github.com/koheiyamayama/gqlgen-reproduce-error/graph/model.Post will always match before github.com/koheiyamayama/gqlgen-reproduce-error/graph/model.InstagramPost (SA4020)
graph/generated.go:3274:2: unreachable case clause: github.com/koheiyamayama/gqlgen-reproduce-error/graph/model.Post will always match before *github.com/koheiyamayama/gqlgen-reproduce-error/graph/model.InstagramPost (SA4020)
```

# minimal schema
- There are schema files in graph/schema directories.

# What I expected
No staticcheck error.
I read graph/generated.go. It is probably no problem.
But I'm glad if staticcheck error doesn't happen.

# versions
- golang
  - go1.20.4 darwin/arm64
- gqlgen
  - v0.17.33

# how to reproduce
https://github.com/koheiyamayama/reproduce-staticcheck-error-for-gqlgen

koheiyamayama:
[...]

PR #2846, opened 2023-11-23
Modify to prevent unreachable code from occurring

taako-502:
https://github.com/99designs/gqlgen/issues/2678

If they define the GraphQL files as follows, the case statements in generate.go will appear in the order shown below. However, since all types include Node, the second and subsequent case statements become unreachable code.

I believe that we should evaluate the case statements starting from the types with the most interface. That's why I created this pull request.

I'm looking forward to hearing about any issues or opinions.

```graphql
interface Post implements Node {
  id: ID!
  title: String!
  body: String!
}

type TweetPost implements Post & Node {
  id: ID!
  title: String!
  body: String!
  tweetId: ID!
}

type FacebookPost implements Post & Node {
  id: ID!
  title: String!
  body: String!
  facebookId: ID!
}

type InstagramPost implements Post & Node{
  id: ID!
  title: String!
  body: String!
  instagramId: ID!
}
```

I have:


codegen/testserver/singlefile/generated.go
```diff
@@ -15908,6 +16064,13 @@ func (ec *executionContext) _Animal(ctx context.Context, sel ast.SelectionSet, o
 	switch obj := (obj).(type) {
 	case nil:
 		return graphql.Null
+	case Horse:
+		return ec._Horse(ctx, sel, &obj)
+	case *Horse:
+		if obj == nil {
+			return graphql.Null
+		}
+		return ec._Horse(ctx, sel, obj)
```
taako-502, line 16073: I evaluate it before Dog.

StevenACoffman:
Thanks very much! I think this is generally a good idea, but it's hard for me to know if it breaks anyone's existing assumptions.

PR #2850, opened 2023-12-02
Improve ResolverImplementer.Implment 

roneli:
Changes the ResolverImplementor to be a function that is called in render time, rather then before, keeping compatibility with previous versions.

@StevenACoffman ,I will add tests if this change makes sense. 

The reason I changed it is sometimes the render requires some logic, such as calling the template.Render's ReserveImport , which impossible to call  if we are doing it before that call as it will be nil (templates.CurrentImports)

Moreover the original implementation had a flaw, that once an implementation existed, it was never called again, which doesn't make sense. I wanted to pass the PrevDecl of the function to the interface, but didn't want to break compatibility, so now it just re-renders each time. 

I have:


StevenACoffman:
@roneli I would like you to add some test coverage for this, as I expect others to build upon your work here.

mojtabacazi:
This breaks our entire project, replaces are existing resolver implementation with `panic(...`.

StevenACoffman:
@mojtabacazi Did you look at https://github.com/99designs/gqlgen/pull/2886 ? WDYT?

PR #2865, opened 2023-12-18
Return null if invalid enum is encountered

blmhemu:

In the current scenario, invalid enums are propagated to the client. This PR sends a null instead - while the graphql spec insists on returning a type error, I found this to be less breaking.

I have:

### DIFF
--- a/plugin/modelgen/models.gotpl
+++ b/plugin/modelgen/models.gotpl
@@ -98,7 +98,12 @@
 	}
 
 	func (e {{ goModelName .Name }}) MarshalGQL(w io.Writer) {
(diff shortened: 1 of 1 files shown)


blmhemu:
Hi @StevenACoffman (Sorry if I tagged the wrong person) and gqlgen team,
Wanted to gather information on this topic, before I do full on implementation.

StevenACoffman:
Hey, so breaking compatibility with the GraphQL spec is not something most people will want, so I would want to only enable this behavior with a config option and have it default to off.

That said, I'm not sure it is even a common enough desire to add this PR.

PR #2866, opened 2023-12-18
Remove archived dependency appdash

UnAfraid:
Removed appdash dependency since it has been archived back in 2021


StevenACoffman (approved):
Thanks!

PR #2871, opened 2023-12-28
fix: avoid panic from tracing on bad request

SteveRuble:
This fixes a panic which arises from the tracing components when a request has some defect which results in an error when creating the `OperationContext`. The transports consistently handle this by calling `DispatchError(graphql.WithOperationContext(ctx, rc), err)` where `rc` is the OperationContext which was not correctly constructed. This seems dangerous, because middleware may assume that if there in an `OperationContext` in the `context.Context` than they are being invoked on a normal codepath and can assume their other interceptors have been invoked in the normal order. Also, using a value returned by a function which also returned a non-nil error is very unusual. However, I have no idea what the impact of changing that dangerous behavior in the transports would be, so I opted to make the tracing component more resilient instead.

I have:


mihaimaruseac (approved):

StevenACoffman:
Thanks very much! This is much appreciated, and I look forward to future PRs from you!

PR #2886, opened 2024-01-20
Fix #2856: resolver receive previous implementation on render

### LINKED ISSUE #2856: ResolverImplementer has a break change that the implementor can't know a field status.
### What happened?

I implemented the `plugin.ResolverImplementer` to generate some CRUD method, but that breaks when bumping from v0.17.38 to v.0.17.41. 
Those implemented and edited code will be overridden by the original template initialization code.

I found #2850:

in v0.17.38 [/plugin/resolvergen/resolver.go](https://github.com/99designs/gqlgen/blob/d9077fac9c2320b28744fb43ab1d00c4d50f0af8/plugin/resolvergen/resolver.go#L13)
```go
                          if implementation == "" {
				// Check for Implementer Plugin
				var resolver_implementer plugin.ResolverImplementer
				var exists bool
				for _, p := range data.Plugins {
					if p_cast, ok := p.(plugin.ResolverImplementer); ok {
						resolver_implementer = p_cast
						exists = true
						break
					}
				}

				if exists {
					implementation = resolver_implementer.Implement(f)
				} else {
					implementation = fmt.Sprintf("panic(fmt.Errorf(\"not implemented: %v - %v\"))", f.GoFieldName, f.Name)
				}

			}
```

in v0.17.41 [/plugin/resolvergen/resolver.go](https://github.com/99designs/gqlgen/blob/fe60938c55308b1cd5562556cdb976771cfcc6cc/plugin/resolvergen/resolver.go#L134)

**diff:**

- 0.17.38  if the implementation is empty, it is passed to the implementer.  
- 0.17.41 the Field is passed to the ResolverImplementer regardless of whether it is already implemented

### What did you expect?

Let the ResolverImplementer can know the field status,  whether the field needs to be implemented, field is new or has been resolved, if not , can you keep the v0.17.38 behavior the way it was

### versions
 - `go run github.com/99designs/gqlgen version` 0.17.41 
 - `go version` 1.21

roneli:
Fix #2856

When implementing the `plugin.ResolverImplementer` (for example, to make a CRUD method), the original template initialization code will override that code.

This PR gives the resolver implementor the previous method body, so it will be empty if no previous body was created before.

It does break the interface, but I would assume its a minor inconvenience for anyone who uses this interface to update.

I have:

### DIFF
--- a/plugin/plugin.go
+++ b/plugin/plugin.go
@@ -33,5 +33,5 @@ type LateSourceInjector interface {
 
 // ResolverImplementer is used to generate code inside resolvers
 type ResolverImplementer interface {
-	Implement(field *codegen.Field) string
+	Implement(prevImplementation string, field *codegen.Field) string
 }
--- a/plugin/resolvergen/resolver.go
+++ b/plugin/resolvergen/resolver.go
@@ -257,12 +257,12 @@ type Resolver struct {
 	PrevDecl             *ast.FuncDecl
 	Comment              string
 	ImplementationStr    string
-	ImplementationRender func(r *codegen.Field) string
+	ImplementationRender func(prevImplementation string, r *codegen.Field) string
 }
 
 func (r *Resolver) Implementation() string {
 	if r.ImplementationRender != nil {
-		return r.ImplementationRender(r.Field)
+		return r.ImplementationRender(r.ImplementationStr, r.Field)
 	}
 	return r.ImplementationStr
 }
--- a/plugin/resolvergen/resolver.go
+++ b/plugin/resolvergen/resolver.go
@@ -127,14 +127,9 @@ func (m *Plugin) generatePerSchema(data *codegen.Data) error {
 			if !f.IsResolver {
 				continue
 			}
-
 			structName := templates.LcFirst(o.Name) + templates.UcFirst(data.Config.Resolver.Type)
 			comment := strings.TrimSpace(strings.TrimLeft(rewriter.GetMethodComment(structName, f.GoFieldName), `\`))
 			implementation := strings.TrimSpace(rewriter.GetMethodBody(structName, f.GoFieldName))
-			if implementation == "" {
-				// use default implementation, if no implementation was previously used
-				implementation = fmt.Sprintf("panic(fmt.Errorf(\"not implemented: %v - %v\"))", f.GoFieldName, f.Name)
-			}
 			resolver := Resolver{o, f, rewriter.GetPrevDecl(structName, f.GoFieldName), comment, implementation, nil}
 			var implExists bool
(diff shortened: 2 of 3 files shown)


roneli:
Hi @StevenACoffman, 

I would actually find this PR useful for my project if possible for us to merge this also?

I can create a config option that by default that keeps the old version that doesn't "regenerate" after first render which will keep the original intention, and when the flag is on then the resolver will always regenerate + get past generation 

WDYT?

StevenACoffman:
Sorry for not responding earlier.

I was sort of confused with some of the description of this PR and in the issue, so I've gone back and collected the original issues and changed the description.

Breaking an interface is not ideal, and breaking compatibility with someone's established usage is even worse.

If there was a config option that would be better.

tsingsun:
I added a rewriter to solve the interface changed in 17.41. Now I have updated to 17.45, the code generation is the same as before, thank you for your work.

bagnaram:
Is there an example of the rewriter? We are experiencing this issue in our codebase with schemeas being overwritten.

malcolmohare:
I'm also waiting for this to be fixed.  I think if this plugin is fundamentally broken because the lack of this parameter, we should ship the breaking change.  It's a quite minor update to anyone using the plugin today, and its clearly an issue for multiple users.

However, we CAN fix this in a way that isn't a breaking change. I achieved this by adding extending the plugin type for the resolver implementer.  The wrapping type has a method for determining whether to render or use the existing code. I've got a sample implementation here: https://github.com/99designs/gqlgen/pull/3079, and I've validated its working as intended on my own project.

StevenACoffman:
Would you recommend #2886 be merged as is?

malcolmohare:
I'd prefer this PR being merged.  My submission was just to offer an alternative if there was too much pushback to making the breaking change.

StevenACoffman:
I misunderstood @tsingsun  comment to mean this issue had a reasonable workaround, so I was waiting for an improvement that did not break backwards compatibility.

Thank you @malcolmohare for demonstrating both how it was basically broken for everyone, and how we could have avoided breaking backwards compatibility if we really needed to.

StevenACoffman:
Please try it out and verify it doesn't need any further improvements so I can cut a new release.

malcolmohare:
Confirmed this impl of the plugin works and returns the original behavior of the Resolver Codegen

```
func (m *Plugin) Implement(prevImplementation string, field *codegen.Field) string {
	if prevImplementation != "" {
		return prevImplementation
	}
	return fmt.Sprintf("panic(fmt.Errorf(\"not implemented: %v - %v\"))abcd", field.GoFieldName, field.Name)
}
```

PR #2891, opened 2024-01-30
Work with https://specs.apollo.dev/federation/v2.x

StevenACoffman:
The contributor intended to allow any version 2.x (2.0, 2.1, etc.) to be version `2` but the conditional precluded this from happening, so I fixed it.


trevor-scheer:
Good catch, lgtm. Thanks!

PR #2894, opened 2024-02-07
Update x/tools and add go v1.21,v1.22 in CI workflows

ryicoh:
In go v1.21, the test succeeds, but it fails in go v1.22. 
The reason was that x/tools was outdated, so I updated it to the latest version.

I have:


StevenACoffman:
So something in Go 1.22 changed and it got much more pedantic.
The `go generate ./...` would hit the comment in main.go:
```go
//go:generate go generate ./_examples/...
```
and give an error:
```
pattern ./_examples/...: directory prefix _examples does not contain main module or its selected dependencies
```
This was hard to see, as the old lint workflow did not echo commands and would run all of check-fmt, check-generate, check-coverage, golangci/golangci-lint-action.

The fix was to change the main.go comment to this:
```
//go:generate sh generate_examples.sh
```
and then in the script, I could do:
```shell
#!/bin/sh
cd _examples
go generate ./... || return 0
```
In github actions we don't really care if it works. The main.go comment is really only there to avoid having PRs from new contributors who would constantly stumble on forgetting to generate the _examples.

PR #2957, opened 2024-03-06
Add option to omit resolver fields from models

Desuuuu:
The `omit_resolver_fields` option, when enabled, makes modelgen skip fields that use the `@goField(forceResolver: true)` directive.

Since these fields use a resolver, they are never used directly by the generated exec code. This option makes it possible to simply not generate them.

I have:


StevenACoffman:
Thanks!

PR #2959, opened 2024-03-09
fix(docs): convert an unnecessarily capitalized word to lowercase

hxrxchang:
This is a minor correction.
I thought it would be more appropriate to use "keep" instead of "Keep" as it appears in the text, so I've  fixed it.


StevenACoffman:
Thanks!

PR #2965, opened 2024-03-11
Update explicit_requires to support models generated into same package

ericbock:
The `explicit_requires` feature added in https://github.com/99designs/gqlgen/pull/2884 failed to generate successfully when the models were generated into the same package as the new `federation.requires.go` file.

This PR [changes how the models](https://github.com/99designs/gqlgen/compare/master...ericbock:gqlgen:models_in_pkg?expand=1#diff-fa028f5dfc54acb372fd0d727ec7e799b93d8dbb34b2769918b7a55bbf051910R123) are named in `federation.requires.go`, with `LookupType` seeming to account for whether the package name should be included.

I updated the gqlgen config to [create the models ](https://github.com/99designs/gqlgen/compare/master...ericbock:gqlgen:models_in_pkg?expand=1#diff-a0db25d6541c9b382887a747214e1b2cde0ed1fc99c3142e5463894ca8440c30R10)in the same directory as the other generated files and then regenerated and verified the tests work with both the new configuration and the old.

I have:


StevenACoffman:
Thanks very much!

PR #2966, opened 2024-03-11
Add case for resolvers_always_return_pointers:false in explicit requires generation. 

jesse-apollo:
This PR addresses an edge case when resolvers_always_return_pointers:false.  The template code now adds an ampersand to return the pointer to the entity when the resolver returns a non-pointer.

I have:


StevenACoffman:
Thanks!

PR #2980, opened 2024-03-25
Add UintID type binding

nawivee:

I have:

I use Gorm to connect to database. By default it uses `uint` instead of `int`. I see that it type bindings for `int` to ID ( `string` ) works perfectly well. I'm lacking of automatic type bindings of `unit` to `ID`.

```graphql
Model {
  id: ID
}

```

```go

type Model struct {
  ID uint 
}
```

Now, default behaviour is to return `uint` id as `number`.
```json
{
  id: 123
}
```

I expect this model to return id as a `string`, to do so you need to extend type bindings

```yaml
models:
  ID: # The GraphQL type ID is backed by
    model:
      - github.com/99designs/gqlgen/graphql.IntID 
      - github.com/99designs/gqlgen/graphql.UintID # <------ add this little helper I wrote
```

in this case the result is supposed to be

```json
{
  id: "123"
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,7 @@ models:
     model:
       - github.com/99designs/gqlgen/graphql.IntID # a go integer
       - github.com/99designs/gqlgen/graphql.ID # or a go string
+      - github.com/99designs/gqlgen/graphql.UintID # or a go uint
 ```
 
 This means gqlgen will be able to automatically bind to strings or ints for models you have written yourself, but the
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -135,6 +135,7 @@ models:
     model:
       - github.com/99designs/gqlgen/graphql.IntID # a go integer
       - github.com/99designs/gqlgen/graphql.ID # or a go string
+			- github.com/99designs/gqlgen/graphql.UintID # or a go uint
```
StevenACoffman, line 138: Is this a tab character or spaces? it should match the lines above.

StevenACoffman: thanks!

nawivee: It seems to have been replaced by IDE. I've fixed it.

PR #2988, opened 2024-04-05
fix: stop loading package dependencies

### LINKED ISSUE #2981: High Memory Usage During Generation
### What happened?
Running gqlgen in our project with a few hundred types can lead to >12 GB of memory usage on my local MacBook M2, as well as consuming a large portion of the CPU. It runs for ~1-2 minutes locally, but can take 5 minutes or longer (and event time out) in our CI pipelines, likely due to memory constraints.

This seems to be due to binding medium-to-large Go packages, either through `autobind` or specifically binding particular models.

If I run gqlgen with a tiny dummy GQL schema and no Go bindings, the process uses ~500 MB and finishes within a few seconds. However if I declare a type binding to a single medium-to-large Go package, the process requires 3-5GB depending on the package, and takes >10s. This is true even if I bind to a tiny Go package that happens to import one of the larger packages. 

Our real project of course binds to a number of packages, so we see even greater memory usage.

### What did you expect?
Much lower memory usage, faster generation times on memory constrained systems.

```
type Example {
  id: ID!
}
```
Simply specifying an `autobind` to one of our larger packages with this tiny schema causes the problem.

### versions
 - `go run github.com/99designs/gqlgen version` v0.17.45
 - `go version` go version go1.21.8 darwin/arm64

agratta-silo:
fixes #2981 

Stop passing `NeedImports` and `NeedDeps` to `packages.Load()`.

Depending on the packages we are loading, these flags can cause gigabytes of extra data to be loaded, because they indicate we want type info for all transitive dependencies. This extra info is completely unused by gqlgen as far as I can tell.

Conceptually, if we are binding to types in a particular package, it doesn't make sense that we would need to know about the dependencies of that package, so I'm not sure why this data was ever being loaded. The `packages.Load` documentation is poor and the issue was likely never noticed because it only appears when binding to packages with many transitive dependencies.

Since we are no longer loading dependencies, I've also removed the logic around caching the dependencies.

### DIFF
--- a/internal/code/packages.go
+++ b/internal/code/packages.go
@@ -21,12 +21,10 @@ var (
 
 var mode = packages.NeedName |
 	packages.NeedFiles |
-	packages.NeedImports |
 	packages.NeedTypes |
 	packages.NeedSyntax |
 	packages.NeedTypesInfo |
-	packages.NeedModule |
-	packages.NeedDeps
+	packages.NeedModule
 
 type (
 	// Packages is a wrapper around x/tools/go/packages that maintains a (hopefully prewarmed) cache of packages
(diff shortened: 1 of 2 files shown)


StevenACoffman:
Thanks! If you get a chance, I would love a PR to improve the documentation.

PR #2998, opened 2024-04-08
Remove dedupExchange 

erka:
dedupExchange was deprecated in v4.0.0 and  is removed now.  That release note says

>- Deprecate the `dedupExchange`. The functionality of deduplicating queries and subscriptions has now been moved into and absorbed by the `Client`.
  Previously, the `Client` already started doing some work to share results between
  queries, and to avoid dispatching operations as needed. It now only dispatches operations
  strictly when the `dedupExchange` would allow so as well, moving its logic into the
  `Client`

In this release they just 
> Remove deprecated `dedupExchange`

### DIFF
--- a/integration/src/__test__/integration.spec.ts
+++ b/integration/src/__test__/integration.spec.ts
@@ -10,7 +10,7 @@ import {
     Observable,
     Operation,
 } from '@apollo/client/core';
-import {HttpLink} from '@apollo/client/link/http';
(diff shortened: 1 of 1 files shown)


UnAfraid (approved):

StevenACoffman:
@erka (and @UnAfraid too!) Thanks so much for looking at this while I was on my road trip!

PR #2999, opened 2024-04-10
Mark ctx as unused when no arguments for FieldContextFunc

halkeye:
gopls / vscode was spitting out hundreds of warnings about ctx being unused. I just updated the template to make sure its _ctx for when ctx isn't actually used

I don't think testing needs updating, nor there's any documentation change

PS, there's no `next` branch as mentioned by CONTRIBUTING.md

I have:
 - [ ] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [ ] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

PR #3001, opened 2024-04-11
Assign ctx, fs, sel to _ so no warnings about unused variables

halkeye:
alternative to https://github.com/99designs/gqlgen/pull/2999

Found a few more while doing development, gopls/vscode was warning about more variables that are undefined.
Instead of having even more complex logic for making variables _, just assigned them to nothing so they would get used.

That only leaves one left in generate.go for me

image

I have:

### DIFF
--- a/_examples/chat/generated.go
+++ b/_examples/chat/generated.go
@@ -540,6 +540,7 @@ func (ec *executionContext) _Chatroom_name(ctx context.Context, field graphql.Co
 }
 
 func (ec *executionContext) fieldContext_Chatroom_name(ctx context.Context, field graphql.CollectedField) (fc *graphql.FieldContext, err error) {
+	var _ = ctx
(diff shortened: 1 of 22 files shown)


StevenACoffman:
I would much rather we don't do this. The point of the linter warning you is to detect possible bugs where an argument or a variable declaration are unused, and it identified we were being a bit lazy in our codegen conditionals.

When the arguments are changed from `ctx` to `_` it is clear we only have the extraneous argument so the signature will match an interface.

When we assign `var _ = ctx` it obscures this intention.

StevenACoffman:
Closing in favor of #2999

halkeye:
Yea. That's totally cool. One of the arguments was only used inside a loop behind a conditional and I couldn't figure out how to nicely figure that out in the args, so this was idea b

I'll see if I can't find a way. Thanks

PR #3005, opened 2024-04-13
chore: update Automatic Persisted Queries Link

38elements:
Current url hash does not match `automatic persisted queries`. 
Therefore, this changed it.

* https://www.apollographql.com/docs/resources/graphql-glossary/#automatic-persisted-queries-apq

* https://www.apollographql.com/docs/resources/glossary/#automatic-persisted-queries

I have:


StevenACoffman:
Thanks!

PR #3029, opened 2024-04-29
Improve federation resolver selection

clayne11:
Just checking for existence of keys in the representations isn't enough.  If the values are null, we should skip the resolver.

Imagine the following case:

```graphql
type Location @key(fields: "id") @key(fields: "latitude longitude") {
  id: ID!
  latitude: Float!
  longitude: Float!
}
```

The Federation Router sends me the following query:
```graphql
query Location {
  _entities(representations: [{
        __typename: "GeoLocation",
        id: null,
        latitude: 37.7878792,
        longitude: -122.4027042
      }
    ]) {
    ... on GeoLocation {
      latitude
      longitude
      id
      fullAddress
      name
    }
  }
}
```

With the current logic, we'll select the `id` resolver, _even though the value is null_. With this change, we'll now see that the value for `id` is `null` and skip over it to select the `latitude longitude` resolver.

I have:

### DIFF
--- a/_examples/federation/accounts/graph/federation.go
+++ b/_examples/federation/accounts/graph/federation.go
@@ -197,8 +197,18 @@ func entityResolverNameForEmailHost(ctx context.Context, rep map[string]interfac
 			ok  bool
 		)
 		_ = val
+		// if all of the KeyFields values for this resolver are null,
+		// we shouldn't use use it
+		allNull := true
 		m = rep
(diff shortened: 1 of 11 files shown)


StevenACoffman:
Thanks! This is a performance improvement, and I can't think of any scenario where someone might have relied on the old behavior.

StevenACoffman:
I really appreciate you contributing this, and I am looking forward to future PRs from you!

PR #3077, opened 2024-05-16
Federation: Update docs to use IntrospectAndCompose

frankywahl:
`serviceList` now gets a deprecation warning to use IntrospectAndCompose instead. We update our docs to avoid referring to deprecated services


StevenACoffman:
Thanks!

PR #3088, opened 2024-05-21
Embed extra fields config

atzedus:
Added new configuration for models named "embedExtraFields", this extends "extraFields" with ability to "embed" other types into generated structure, for example:

```yaml
models:
    Person:
        embedExtraFields:
            - type: mycusomtype.PersonHelpers
            - description: some helpers for person
```

will generate something like:
```go
type Person {
    ID string
    // some helpers for person
    mycusomtype.PersonHelpers
}
```

@goExtraField directive also extended.
Field name now can be nullable, and if it not provided or filled as empty string, then this field become embedded. Example:

```graphql
type Person @goExtraField(type: "mycusomtype.PersonHelpers") {
    id: ID!
}
```

will generate the same result as described above.

I have:


StevenACoffman:
Thanks!

PR #3110, opened 2024-05-28
Correct dataloader example

mpldr:
Dataloader requires the value and error slice to be of equal length, in order to correctly return the values.

Link: https://github.com/vikstrous/dataloadgen/blob/7de6ebe3d882737607ce2ba646e8d6ec652b32e3/dataloadgen_test.go#L19-L20


StevenACoffman:
Thanks!

PR #3153, opened 2024-06-22
fix bug: POST transport Insufficient rigorous judgment leads to invalid SSE transport

o8x:
fix bug: POST transfer judgment was not strict enough, resulting in SSE transport never will be invoked

For example, the following POST based SSE request will be automatically taken over by POST transport and will not enter the SSE process

```
curl -N --request POST --url http://localhost:8080/query \
--data '{"query":"subscription { currentTime { unixTime timeStamp } }"}' \
-H "accept: text/event-stream" -H 'content-type: application/json' \
--verbose
```

Easy bug fixes, without updating documentation

I have:

### DIFF
--- a/graphql/handler/transport/http_post.go
+++ b/graphql/handler/transport/http_post.go
@@ -27,6 +27,10 @@ func (h POST) Supports(r *http.Request) bool {
 		return false
 	}
 
+	if strings.Contains(r.Header.Get("Accept"), "text/event-stream") {
+		return false
+	}
+
 	mediaType, _, err := mime.ParseMediaType(r.Header.Get("Content-Type"))
 	if err != nil {
 		return false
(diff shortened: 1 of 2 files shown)


StevenACoffman:
Hey, thank you very much! By the way, many users have asked me if there's a public repository that demonstrates using HTTP SSE with GraphQL. Are you able to make one public?

o8x:
> Hey, thank you very much! By the way, many users have asked me if there's a public repository that demonstrates using HTTP SSE with GraphQL. Are you able to make one public?

@StevenACoffman  
I would love to, but one issue prevented me from submitting the PR. I have submitted the details to the issue [https://github.com/99designs/gqlgen/issues/3154](https://github.com/99designs/gqlgen/issues/3154)

StevenACoffman:
Thanks, finding this issue is also helpful! 😅

StevenACoffman:
Hey, so it turns out this causes some problems with popular clients that advertise accepting `text/event-stream`, so I needed to revert this.

PR #3162, opened 2024-07-01
Nulls are now unmarshalled as zero values for primitive types

endSly:
## Description

The current implementation of Unmarshal to primitives in gqlgen does not handle null values consistently. Specifically, when input types are mapped to Go structs, optional fields defined as primitives fields in the Go struct can lead to errors or unexpected behaviors.

#### Example

Consider the following GraphQL input type:

```graphql
input ExampleInput @goModel(model: "example.ExampleInput") {
  name: String
  age: Int
}
```

This maps to the following Go struct:

```go
type ExampleInput struct {
  Name string `json:"name,omitempty"`
  Age  int    `json:"age,omitempty"`
}
```

When unmarshalling the following JSON input:

```json
{
  "name": null,
  "age": null
}
```

The resulting `Name` field in the Go struct is assigned the string value `"null"`, and the `Age` field causes an error. This behavior is inconsistent with the Go standard JSON library, which leaves the field untouched.

#### Proposed Solution

To align with the Go standard library's behavior, null values should be mapped to the zero-value of the corresponding Go type. This approach is more idiomatic to Go and reduces the potential for errors.

I have:

### DIFF
--- a/graphql/bool.go
+++ b/graphql/bool.go
@@ -20,6 +20,8 @@ func UnmarshalBoolean(v any) (bool, error) {
 		return v != 0, nil
 	case bool:
 		return v, nil
+	case nil:
+		return false, nil
 	default:
 		return false, fmt.Errorf("%T is not a bool", v)
 	}
--- a/graphql/executor/executor.go
+++ b/graphql/executor/executor.go
@@ -84,7 +84,6 @@ func (e *Executor) CreateOperationContext(
 
 	var err error
 	rc.Variables, err = validator.VariableValues(e.es.Schema(), rc.Operation, params.Variables)
-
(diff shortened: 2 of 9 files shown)


StevenACoffman:
Hey, thanks! I updated this PR to make the same changes for unsigned integers.  

While your PR is clearly in line with the Go standard JSON library, it's not necessarily the expectation of the JavaScript developers making GraphQL calls to gqlgen. 

As a result, this might be a controversial change. If people find that the GraphQL spec requires different behavior for null values, I would be open to reverting this change.

In the absence of that, I think this will make day-to-day life better for developers, so I very much appreciate you contributing it.

Issue #2991, opened 2024-04-07
Improve support for @requires Federation directives


clayne11:
### What happened?

#2884 landed recently to add support for nested selection sets in `@requires` directives. This implementation isn't great, however. There are two significant gaps:

1. The resolved fields are rendered as a `map[string]interface{}`. We know the types for this model, however, from the schema. We should be generating a typesafe model for these fields.
2. I would expect a separate resolver function to be generated for each `@requires` field since they should be treated like independently derived fields, at least in my opinion.
3. There's no way to access any errors that occurred when fetching the upstream fields. This is important because an upstream error can mean something very different from `null`.

For example, if I’m trying to compute `driverName` and do `@requires(fields: "driver { name }"` and `driver` comes back `null` due to an error the correct response might be to throw an error and return `null`, but if the `driver` is truly `null` I might return “No driver assigned”.

### Example
[From the tests](https://github.com/99designs/gqlgen/blob/master/plugin/federation/testdata/explicitrequires/schema.graphql#L41-L47):
```graphqls
type PlanetRequiresNested @key(fields: "name") {
    name: String! @external
    world: World! @external
    worlds: [World!] @external
    size: Int! @requires(fields: "world{ foo }")
    sizes: [Int!] @requires(fields: "worlds{ foo }")
}
```

Generated code:
```golang
// PopulatePlanetRequiresNestedRequires is the requires populator for the PlanetRequiresNested entity.
func (ec *executionContext) PopulatePlanetRequiresNestedRequires(ctx context.Context, entity *PlanetRequiresNested, reps map[string]interface{}) error {
	entity.Name = reps["name"].(string)
	entity.World = &World{
		Foo: reps["world"].(map[string]interface{})["foo"].(string),
	}
	return nil
}
```

### What did you expect?
In an ideal world, I would like to see something like the following be generated:
```golang
type PlanetRequiresNestedSizeRequires interface {
	Worlds() (PlanetRequiresNestedSizeWorldRequires, error)
}

type PlanetRequiresNestedSizeWorldRequires interface {
	Foo() (string, error)
}

// PopulatePlanetRequiresNestedRequires is the requires populator for the PlanetRequiresNested entity.
func (ec *executionContext) PopulatePlanetRequiresNestedRequiresSize(ctx context.Context, requires PlanetRequiresNestedSizeRequires) (int32, error) {
	worlds, err := requires.Worlds()
	if err != nil {
		return 0, err
	}

	foo, err := worlds.Foo()
[...]

PR #3183, opened 2024-07-14
chore: Refactor federation.go to make it easier to read

clayne11:
- Cut functions into smaller functions
- Remove mutation in several locations

This is setting up to make it easier to introduce improvements for #2991.

There are no behavior changes in this PR. There's one small change in how we output the Federation SDL into the generated code that is formatted more accurately than before.


StevenACoffman:
Thanks!

PR #3184, opened 2024-07-15
WIP — Implement improved API for Federation @requires

clayne11:
In order to do this correctly, you need to treat `gqlgen` as both a GraphQL client and a GraphQL server. The selection set inside `@requires(fields: "<selection set>")` is actually a fragment that gets resolved by the router and passed in the extensions object to use as part of the entity resolution.

The current `@requires` implementation (before this PR) simply isn't type safe to get around this requirement. The `@key` directive also includes a selection set and is type safe, but the implementation shortcuts the fragment requirement by flattening down the `@key.fields` into a flat struct — this is only possible because `@key.fields` cannot contain lists, interfaces, or unions.

`@requires` doesn't have this limitation, which leads to two issues:
1. Federation Entity resolvers are currently "faked" into the GraphQL schema. We can't use this pattern for `@requires` because the selection isn't guaranteed to fit into an input type (since inputs can't contain unions / interfaces). As a result, we can't model `@requires` as a GraphQL resolver.
2. We need to support generating types for fragments — currently the codegen is set up to generate models and resolver code for `ast.Schema` types, but doesn't support `ast.QueryDocument` types.

To solve these problems we introduce new abstractions for:
1. Optionally populating an `ast.QueryDocument` with  fragments and model generating types for them.
3. Hooking the resolvers up to the core engine / giving users a way to implement them that doesn't lean on faking GraphQL resolvers.

This PR makes a large improvement for #2991. It fixes the type safety issues, however the Federation spec still doesn't support passing errors downstream from the router for `@requires`.

I have:


clayne11:
Currently the first commit is directionally working. I had created the second two commits before I recognized the limitations I outlined above, so I need to rework the mechanism for injecting the resolvers.

clayne11:
I've thought more deeply about this, I think I have a pretty good plan for implementing this correctly. I can improve the overall Federation design in the process.

# Overview

We start with a schema that includes Federation directives like `@key` and `@requires`.

```graphql
"""
Original schema
"""
type User @key(fields: "id") @key(fields: "name") @entityResolver(multi: true) {
  id: ID!
  name: String!
  dateOfBirth: Int
  email(domain: String, requires: _Requires): String!
  accounts(filterByType: String): [Accounts]!
    @requires(fields: "email(domain: \"google.com\") dateOfBirth")
}

type Accounts {
  id: ID!
  name: String
}

type Product @key(fields: "id") {
  id: ID!
  name: String!
  price: Int!
}
```

We expand the schema to include the public-facing API required by the Federation router
to be able to resolve the required entities.

```graphql
"""
First order generated schema: exposed on the schema itself for the router to call.
"""
type Query {
  _entities(representations: [_Any!]!): [_Entity]!
}

scalar _Any

union _Entity = User | Product

type User @key(fields: "id") @key(fields: "name") @entityResolver(multi: true) {
  id: ID!
  name: String!
  dateOfBirth: Int
  email(domain: String, requires: _Requires): String!
  accounts: [Account]!
    @requires(fields: "email(domain: \"google.com\") dateOfBirth")
}

type Product @key(fields: "id") {
  id: ID!
  name: String!
  price: Int!
}
```

```graphql
"""
Second order generated schema: generated in the resolvers,
but not generated onto the external-facing schema.
"""
type Query {
  _entities(representations: [_Any!]!): [_Entity]!

  _entityKeyUserIDs(representations: [_EntityKeyUserID]): [User]!
    @goField(name: "findUserByIDs")
  _entityKeyUserNames(representations: [_EntityKeyUserName]): [User]!
    @goField(name: "findUserByNames")
  _entityKeyProductID(representation: _EntityKeyProductID): Product!
    @goField(name: "findProductByID")
}

scalar _Any

union _Entity = User | Product

scalar _EntityKeyUserID @entityDef(fields: "id", entity: "User")
scalar _EntityKeyUserName @entityDef(fields: "name", entity: "User")
scalar _EntityKeyProductID @entityDef(fields: "id", entity: "Product")

"""
This is a runtime directive used for multi entity resolvers.

It injects the current entity's position relative to the input representations list.
In order to inject the correct argument for _requires, we have to extract the
representations object from the input that corresponds to the current entity's position.

This is paired up with @extractFromRepresentations to read the position value injected
by this directive and populate the "_requires" argument with the correct input.
"""
directive @injectPosition on OBJECT

"""
This is a runtime directive used to implement @requires. It's automatically placed 
on the generated _requires argument, and the implementation of it extracts the
correct value from the input representations list. If the type is a multi entity resolver,
it will extract the correct value based on position injected by @injectPosition.
"""
directive @extractFromRepresentations on ARGUMENT_DEFINITION

type User
  @key(fields: "id", resolvable: false)
  @key(fields: "name")
  @injectPosition {
  id: ID!
  name: String!
  dateOfBirth: Int
  email(domain: String): String!
  accounts(
    """
    We autogenerate this argument based on the @requires directive. It's autopopulated
    using @extractFromRepresentations and @injectPosition.
    """
    _requires: _EntityRequiresUserAccounts @extractFromRepresentations
  ): [Account]!  # not generated in external-facing schema
    
    @requires(fields: "email(domain: \"google.com\") dateOfBirth")
    @goField(forceResolver: true)
}

scalar _EntityRequiresUserAccounts
  @entityDef(
    fields: "email(domain: \"google.com\") dateOfBirth"
    entity: "User"
  )
```

## Requests

What the request flow looks like:

```graphql
query Query($representations: [_Any]!) {
  _entities(respresentations: $representations) {
    ... on User {
      id
      name
      accounts(filterByType: "PAYMENT") {
        id
        name
      }
    }
    ... on Product {
      id
      price
    }
  }
}
```

```json
{
  "representations": [
    {
      "__typename": "User",
      "id": "1",
      "email": "one@foo.com",
      "dateOfBirth": 9000
    },
    {
      "__typename": "User",
      "id": "2",
      "email": "two@foo.com",
      "dateOfBirth": 9001
    },
    {
      "__typename": "User",
      "name": "David",
      "email": "two@foo.com",
      "dateOfBirth": 9002
    },
    {
      "__typename": "Product",
      "id": "4"
    }
  ]
}
```

We transform this into a new request that we send to the execution context root:

```graphql
query Query(
  $representation1: [_EntityKeyUserID]
  $representation2: [_EntityKeyUserName]
  $representation3: _EntityKeyProductName
) {
  representation1: _entityKeyUserIDs(representations: $representation1) {
    id
    name
    accounts {
      id
      name
    }
  }
  representation2: _entityKeyUserNames(representations: $representation2) {
    id
    name
    accounts {
      id
      name
    }
  }
  representation3: _entityKeyProductID(representation: $representation3) {
    id
    price
  }
}
```

```json
{
  "representation1": [
    {
      "__typename": "User",
      "id": "1",
      "email": "one@foo.com",
      "dateOfBirth": 9000
    },
    {
      "__typename": "User",
      "id": "2",
      "email": "two@foo.com",
      "dateOfBirth": 9001
    }
  ],
  "representation2": [
    {
      "__typename": "User",
      "name": "David",
      "email": "two@foo.com",
      "dateOfBirth": 9002
    }
  ],
  "representation3": {
    "__typename": "Product",
    "id": "4"
  }
}
```

Note: I may not actually send a string based request, as this would incur the cost
of parsing the string into an AST. Instead, I can send the AST directly to the
execution context root.

StevenACoffman:
This sounds great! Looking forward to next steps!

PR #3187, opened 2024-07-17
refactor: significantly clean up the federation.gotpl template

clayne11:
There were a number of inline structs and inline functions that made it extremely hard to reason about what the code is doing. Split these out into smaller functions with less closures and mutation.

Part of the setup to help with fixing #2991.

There should be no behavior changes in this PR, simply refactoring code.

I also added a new package `_examples/all` to make it easier to debug Federation subgraphs.

Tested by running this query in the `_example/federation` demo before and after my changes:
```graphql
query {
  topProducts {
    reviews {
      author {
        id
        reviews {
          body
        }
      }
    }
  }
}
```


_examples/federation/package.json
```diff
@@ -5,6 +5,7 @@
   "main": "gateway/index.js",
   "type": "module",
   "scripts": {
+    "start-gateway": "node gateway/index.js",
```
StevenACoffman, line 8: Thanks! Saves me time.

_examples/federation/readme.md
```diff
@@ -1,3 +1,14 @@
 ### Federation
 
 [Read the docs](https://gqlgen.com/recipes/federation/)
+
+## Testing
```
StevenACoffman, line 5: I am a huge fan of making it easier for people besides me to debug through this 😆

StevenACoffman (approved):
This looks good to me, but I would love to move things from `federation.gotpl` to `federation.go` wherever possible. Some of the static analysis tooling is defeated when logic is hidden in the federation.gotpl, and it's harder to read and maintain inside a template.

The only thing about this PR that gives me pause is that it doesn't improve things in that regard, but it maybe that is in your future plans.

PR #3188, opened 2024-07-18
feat: allow argument directives to be called even if the argument is null

clayne11:
The existing implementation assumes that if an input argument is null, you don't want to call the directive. This is a very constraining assumption — directives may want to not just mutate an argument but to actually outright set it.

**This is a breaking change** as argument directives now need to handle null input values. Added a new config switch to control this new behavior:

`call_argument_directives_with_nulls: bool`

I have:


StevenACoffman:
I can see your reasoning, but ouch. I'm worried that to support the upcoming federation changes this option will be required, and that is going to break a lot of people's expectations.

PR #3192, opened 2024-07-23
[resolvergen] enable rewrite features for single-file mode

hailyngx:
Enable rewrite in `single-file` mode, rewrite behaviors are as below:
* When there’s no change on the schema, re-run resolver generators should not move the old code that conforms the resolver interface to the comment.
* When there’s change on a single field, the rest of the resolver.go code that conforms to the resolver interface should not be moved to the comment.
* Only when the functions are no longer valid per the schema change, or functions that do not conform to the resolver.go should be moved to the comment.
* When checking which components to move to the comment, we fix `rewrite.go` to skip `token.VAR` and `token.TYPE` to avoid redeclare errors.

I have:

### DIFF
--- a/internal/rewrite/rewriter.go
+++ b/internal/rewrite/rewriter.go
@@ -188,7 +188,7 @@ func (r *Rewriter) RemainingSource(filename string) string {
 				continue
 			}
 
-			if d, isGen := d.(*ast.GenDecl); isGen && d.Tok == token.IMPORT {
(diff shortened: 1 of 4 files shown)


StevenACoffman:
Aww. Sorry I didn't provide feedback while I was on vacation.

hailyngx:
@StevenACoffman Hey Steve, appreciate your reply. I have put up a discussion here for a better solution. Just a minor improvement. Can you take a look and let me know your feedback? Thank you so much 😊
https://github.com/99designs/gqlgen/discussions/3205

Issue #3202, opened 2024-07-30
Add @concurrent directive for types


krupyansky:
### What happened?
For every element of list on marshaling graphql object we get one goroutine. It's bad for performance on load tests. We should use goroutine for network call or another system call. Or hard parallel calculations. By marshaling isn't hard calculation.

### What did you expect?
I propose to add "@concurrent" directive for control the behavior. That lists consists concurrent type will have generated code as is now. But for lists with not concurrent type (without @concurrent directive) will not have concurrent code.

PR #3203, opened 2024-07-30
[proposal] Add @concurrent directive for types

krupyansky:
We use one goroutine for every element of a list when marshaling the GraphQL object. This performs badly on load tests. We should use goroutines only for network calls, certain other system calls, or expensive parallel calculations. Marshaling isn't an expensive calculation.

In this PR, I propose adding a "@concurrent" directive that controls this behavior. Lists that have the concurrent type (with `@concurrent` directive) will have generated code as is now. But lists without the concurrent type (without @concurrent directive) will not have concurrent unmarshalling code.

Added pull request for issue https://github.com/99designs/gqlgen/issues/3202

### DIFF
--- a/_examples/chat/generated.go
+++ b/_examples/chat/generated.go
@@ -3579,7 +3579,7 @@ func (ec *executionContext) marshalNMessage2githubᚗcomᚋ99designsᚋgqlgenᚋ
 func (ec *executionContext) marshalNMessage2ᚕgithubᚗcomᚋ99designsᚋgqlgenᚋ_examplesᚋchatᚐMessageᚄ(ctx context.Context, sel ast.SelectionSet, v []Message) graphql.Marshaler {
 	ret := make(graphql.Array, len(v))
 	var wg sync.WaitGroup
-	isLen1 := len(v) == 1
+	isLen1 := true
 	if !isLen1 {
 		wg.Add(len(v))
 	}
@@ -3683,7 +3683,7 @@ func (ec *executionContext) marshalN__Directive2githubᚗcomᚋ99designsᚋgqlge
 func (ec *executionContext) marshalN__Directive2ᚕgithubᚗcomᚋ99designsᚋgqlgenᚋgraphqlᚋintrospectionᚐDirectiveᚄ(ctx context.Context, sel ast.SelectionSet, v []introspection.Directive) graphql.Marshaler {
 	ret := make(graphql.Array, len(v))
 	var wg sync.WaitGroup
-	isLen1 := len(v) == 1
+	isLen1 := true
 	if !isLen1 {
 		wg.Add(len(v))
 	}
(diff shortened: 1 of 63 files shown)


StevenACoffman:
I reworded the PR description from your initial issue. Please confirm it matches your intention.

krupyansky:
@StevenACoffman confirm

krupyansky:
@StevenACoffman сan you confirm that these changes are relevant for your library and are accepted?

StevenACoffman:
Thank you! I appreciate that you gave people the ability to opt-in to the current behavior that they have been relying on.

I've been a little busy at work lately, so I'm sorry I haven't been able to be as responsive as usual.

adomaskizogian:
Are there any benchmarks or data collected on what impact does this change have for services under high load?

razorness:
@krupyansky @StevenACoffman This merge broke dataloaders behind field resolvers as in [dataloadgen-example](https://github.com/vikstrous/dataloadgen-example) for so many users in `v0.17.50`. :-)

It took me some time to find out that field resolvers are not executed concurrently anymore and you have to set the newly introduced and undocumented `@concurrent` directive on `OBJECT` and cascade where needed. 

So, from now, if you want to use dataloader pattern behind field resolvers, you have these important places for preparation:

- add field resolver in `gqlgen.yml`
- add `directive @concurrent on OBJECT` definition to `schema.graphql`
- implement an empty `@concurrent` directive at `generated.Config.Directives.Concurrent`
- add `@concurrent` directive on object type with multiple field resolvers for dataloaders
- add `@concurrent` on every other object type which contains field resolvers at any depth and can be returned in an array

Honestly, this adds pretty much complexity. Field resolvers are meant for heavy work like fetching from database. Dataloaders are very common to get rid of N+1 problem. One missing `@concurrent` directive and your dataloaders shoot into the void. Furthermore does `@concurrent` change behavior in multiple directions:

1) underlying field resolvers at any depth that you want to work in parallel
2) arrays holding your object type

Here is an example with pagination and a field resolver for dataloader pattern which gets really complex right now:

`schema.graphql`:
```graphql

directive @concurrent on OBJECT

type Query {
  getBooks(first: Int!, after: String): BookConnection
}

type BookConnection {
  totalCount: Int!
  edges: [BookEdge!]!
  pageInfo: PageInfo!
}

# add @concurrent to make marshalling `BookEdge` conurrent 
# in `BookConnection.edges`, so that dataloder behind `Book.Writer` and `Book.Publisher`
# can work at all
type BookEdge @concurrent {
  cursor: String!
  node: Book!
}

# add @concurrent to make `writer` and `publisher` fetchable in parallel
type Book @concurrent {
  id: ID!
  name: String!
  isbn: String!
  writer: Writer!
  publisher: Publisher!
}

type Writer {
  id: ID!
  name: String!
}

type Publisher {
  id: ID!
  name: String!
}
```
`gqlgen.yml`:
```yml
models:
  Book:
    fields:
      # make Book.writer a field resolver
      writer:
        resolver: true
      # make Book.writer a field resolver
      publisher:
        resolver: true
```

Metamogul:
Thanks for the note, @razorness. I accidentally discovered this behavior today when debugging a failed test, eventually leading me here. I find the consequences in regard to dataloaders pretty devastating. I think this should be implemented in a non-breaking way, making the changed behavior opt-in, not the previous implementation.

I didn't find a ticket addressing the issue, so I created one: https://github.com/99designs/gqlgen/issues/3285

krupyansky:
@StevenACoffman @razorness @Metamogul i made fix

https://github.com/99designs/gqlgen/pull/3286

Please give me review

StevenACoffman:
This PR was reverted BTW

PR #3211, opened 2024-08-05
Fix marshaling interfaces and union types

krupyansky:
I added unit tests for query concrete types that I not insert in the query, but they return in a resolver.

Earlier, I can have same types
```
type Shape interface

type Square struct {
  radius float64
}

type Circle struct {
  width, lenght int
}

type Query {
  shapes: []Shape
}
```
and `query {shapes {on ...Square{width length}}}`

And resolver can return circles and I get result:
```
shapes{{}, {}, {width: 10, lenght: 50}}
```
I fixed it

### DIFF
--- a/codegen/testserver/followschema/generated_test.go
+++ b/codegen/testserver/followschema/generated_test.go
@@ -37,6 +37,12 @@ func TestUnionFragments(t *testing.T) {
 	resolvers.QueryResolver.ShapeUnion = func(ctx context.Context) (ShapeUnion, error) {
 		return &Circle{Radius: 32}, nil
 	}
+	resolvers.QueryResolver.Shapes = func(ctx context.Context) ([]Shape, error) {
+		return []Shape{
+			&Circle{Radius: 45},
+			&Circle{Radius: 54},
+		}, nil
+	}
 
 	srv := handler.NewDefaultServer(NewExecutableSchema(Config{Resolvers: resolvers}))
 	c := client.New(srv)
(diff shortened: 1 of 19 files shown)


krupyansky:
@StevenACoffman сan you confirm that these changes are relevant for your library and are accepted?

StevenACoffman:
Thanks! I appreciate your contributions, but I've been a little slammed at work lately.

StevenACoffman:
@rmennes Just to confirm, you get these problem using v0.17.52 correct?

rmennes:
@StevenACoffman  Yes. I got this issue starting from v0.17.50. But I'm experiencing the same issue with v0.17.52.

krupyansky:
@StevenACoffman @rmennes I create pull request https://github.com/99designs/gqlgen/pull/3293

@rmennes help please test out the new version

codegen/testserver/followschema/generated_test.go
```diff
@@ -78,4 +84,23 @@ func TestUnionFragments(t *testing.T) {
...
+		c.MustPost(`query {
+			shapes {
+				... on Rectangle {
+					length
+					width
+				}
+			}
+		}
+		`, &resp)
+		require.Empty(t, resp.Shapes)
```
werjo-gridfuse, line 104: Thats not compliant with the GraphQL spec. The resolver must return two empty objects (for the two circles) here. I also write the comment on the issue. Its not a but or inconvenience if the graphql server is returning empty objects in union or interface types. 

Thats part of the specification. https://spec.graphql.org/October2021/#sec-Unions 

and here is one quote: 

> interfaces must return the set of interfaces that an object implements (if none, interfaces must return the empty set).

werjo-gridfuse:
I already wrote it in the Pull Request but this issue is no issue. Interface types and Union types must return empty objects in this case.

https://spec.graphql.org/October2021/#sec-Unions

> interfaces must return the set of interfaces that an object implements (if none, interfaces must return the empty set).

in your example the inline fragment 

`...on Rectangle` is not telling the server to only return Rectangles. Its telling the server what field it should resolve if the type is a Rectangle. Thats why its absolutely correct if the resolver returns empty objects for all types that are not rectangles.

Issue #3222, opened 2024-08-14
Error on code generation


ankitpatial:
### What happened?
Moved go v1.22.6 to v1.23.0 and tried code generations

### What did you expect?
expecting to generate code successfully
but got following:-
panic: interface conversion: types.Type is *types.Alias, not *types.Named

goroutine 1 [running]:
github.com/99designs/gqlgen/codegen.(*builder).bindField(0x1402d611640, 0x140111ab830, 0x1401139b900)
         go/pkg/mod/github.com/99designs/gqlgen@v0.17.49/codegen/field.go:147 +0xdc0
github.com/99designs/gqlgen/codegen.(*builder).buildField(0x1402d611640, 0x140111ab830, 0x140003dea00)
         go/pkg/mod/github.com/99designs/gqlgen@v0.17.49/codegen/field.go:70 +0x314
github.com/99designs/gqlgen/codegen.(*builder).buildObject(0x1402d611640, 0x140003cdba0)
        go/pkg/mod/github.com/99designs/gqlgen@v0.17.49/codegen/object.go:78 +0x4fc
github.com/99designs/gqlgen/codegen.BuildData(0x140000f7688, {0x1402ae94600, 0x2, 0x2})
       go/pkg/mod/github.com/99designs/gqlgen@v0.17.49/codegen/data.go:116 +0x2cc
github.com/99designs/gqlgen/api.Generate(0x140000f7688, {0x0, 0x0, 0x102d8a120?})
        go/pkg/mod/github.com/99designs/gqlgen@v0.17.49/api/generate.go:97 +0x7bc
main.init.func2(0x140000bcd80?)
         go/pkg/mod/github.com/99designs/gqlgen@v0.17.49/main.go:190 +0xa8
github.com/urfave/cli/v2.(*Command).Run(0x102d96d60, 0x140000bcd80, {0x1400008f180, 0x1, 0x1})
         go/pkg/mod/github.com/urfave/cli/v2@v2.27.4/command.go:276 +0x600
github.com/urfave/cli/v2.(*Command).Run(0x14000134580, 0x140000bcc40, {0x14000092020, 0x2, 0x2})
         go/pkg/mod/github.com/urfave/cli/v2@v2.27.4/command.go:269 +0x83c
github.com/urfave/cli/v2.(*App).RunContext(0x140000f2400, {0x102a37340, 0x102dcb5a0}, {0x14000092020, 0x2, 0x2})
[...]

PR #3226, opened 2024-08-22
Go 1.23 support

UnAfraid:
Added support for *types.Alias conditionally only applies to go1.23
Also updated GitHub Actions to run the pipelines with go 1.23

Closes https://github.com/99designs/gqlgen/issues/3222

I have:

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -16,7 +16,7 @@ jobs:
   coverage:
     strategy:
       matrix:
-        go: ["1.22"]
+        go: ["1.22", "1.23"]
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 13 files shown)


StevenACoffman:
The coverage breaks if you run it for more than one version because coveralls will only accept the first one.

Also the matrix is already pretty slow, and current version + one previous seems sufficient for support.

indera:
This appears to be a breaking change.

Was this supposed to be included in a `minor` release instead of a `patch`?

UnAfraid:
This PR in particular should not break compatibility, it just adds support for types.Alias if compiled with go 1.23, the CI is running it with older versions of go and it was compiling fine, what kind of compatibility issues are you running into?

PR #3233, opened 2024-08-27
feat: allow argument directives to be called even if the argument is null

clayne11:
The existing implementation assumes that if an input argument is null, you don't want to call the directive. This is a very constraining assumption — directives may want to not just mutate an argument but to actually outright set it.

This is a breaking change as argument directives now need to handle null input values. Added a new config switch:

call_argument_directives_with_nulls: bool

to control this new behavior.

I have:

This is unreverting #3188 after fixing the issues with that PR.


StevenACoffman:
Thanks! A config option so that people can opt-in to this behavior is much appreciated.

PR #3243, opened 2024-09-04
[resolvergen] make rewrite default for single-file mode

hailyngx:
## BREAKING CHANGE
Resolvers generated using `layout: single-file` will now have their resolver files overwritten on generate.

## Why this change?
Make rewrite the default for single-file mode as no one is relying on the behavior of single file NOT supporting rewrite. Thus, it would be backwards compatible to use the existing `rewrite` instead of introducing the new `enable_rewrite_for_single_file` flag. 
Relevant Discussion: [https://github.com/99designs/gqlgen/discussions/3205]()

I have:


hailyngx:
Refactor unit tests -- Sorry I got mixed up some file changes in my last commit that failed 2 tests. Should passed all unit tests now with this new commit!

plugin/resolvergen/testdata/singlefile_enable_rewrite/gqlgen.yml
```diff
@@ -0,0 +1,17 @@
+schema:
+  - "testdata/schema.graphql"
+
+exec:
+  filename: testdata/singlefile_enable_rewrite/out/ignored.go
+model:
+  filename: testdata/singlefile_enable_rewrite/out/generated.go
+resolver:
+  filename: testdata/singlefile_enable_rewrite/out/resolver.go
+  type: CustomResolverType
```
clayne11, line 10: Does this work without this option?

plugin/resolvergen/testdata/singlefile_enable_rewrite/out/model.go
```diff
@@ -0,0 +1,13 @@
+package customresolver
+
+import "context"
+
+type Resolver struct{}
+
+type QueryResolver interface {
```
clayne11, line 7: Why are these interfaces being generated here? Aren't these normally inside the server executable?

plugin/resolvergen/testdata/singlefile_enable_rewrite/gqlgen.yml
hailyngx: @clayne11 This is the unit test for rewrite features with single-file, which follows exactly the available single-file feature. As I wrote a new function, it's good to have coverage. 
On another note, without rewrite, yes, it does work in our system with Bazel. However, it does not support any functional overwriting. The developer will generate the stubs, implement however they like after the stubs. If they happened to make large schema changes and want the new corresponding resolver stubs, they have to delete resolver.go in the directory. The main difference is that single-file only focus on generating the stubs, not keeping the implementation. 
Rewrite aims to support large schema changes while retaining the implementation of single-file mode.

plugin/resolvergen/testdata/singlefile_enable_rewrite/out/model.go
hailyngx: @clayne11 You are right! The test are supposed to include only one `model.go` with necessary interfaces (that are usually in the server.go). They supply it this way only for the purpose of a compact unit test I think.

StevenACoffman:
BTW, I very much appreciate you working on this, as it helps to narrow the behavioral gap between single file and multi-file (where rewrite is currently supported). 

I don't believe anyone is relying on the behavior of single file NOT supporting rewrite, so I think it would be backwards compatible to use the existing `rewrite` instead of introducing the new `enable_rewrite_for_single_file` but I am happy to merge this as is, or to make changes that better fit with the changes that @clayne11 is working on.

clayne11:
Makes sense, let's remove the flag then and just make it default behavior. Thanks for reviewing @StevenACoffman.

hailyngx:
Updated based on your suggestion @StevenACoffman 🙏 Thank you Steve!

plugin/resolvergen/testdata/singlefile/out/resolver.go
```diff
@@ -1,21 +1,22 @@
...
 	"context"
+	"fmt"
 )
 
 type CustomResolverType struct{}
 
 // Resolver is the resolver for the resolver field.
 func (r *queryCustomResolverType) Resolver(ctx context.Context) (*Resolver, error) {
-	panic("not implemented")
+	panic(fmt.Errorf("not implemented: Resolver - resolver"))
```
StevenACoffman, line 14: The string change is welcome, but why are you changing from:
```
panic("a problem")
```
to:
```
panic(fmt.Errorf("a problem"))
```

It seems unhelpful?

hailyngx: Looks like I have to go back to change the default stubs of rewrite single-file to `panic("not implemented")`. `panic(fmt.Errorf("not implemented: Resolver - resolver"))` is currently the template for the default rewrite follow-schema. That's why the change is necessary for a git diff check, but if I make this change, the lint check does not like it. Also, looks like deleting resolver config is not a solution. Let me know if you have any suggestions!

StevenACoffman: Not a big deal. We you would need to backtrack and change the template, but it's not actually important.

hailyngx: @StevenACoffman Made a new commit that works for all tests. Coverage and security should be good to go even though they fail right? Thanks Steve for being so responsive!

StevenACoffman: Yes! Thanks for sticking with it! Sorry this was such a long journey!

StevenACoffman:
Thanks for this work!

PR #3247, opened 2024-09-05
codegen: `Unalias` before lookup type

giautm:
This ensure the result always the same for single type.


giautm:
cc @StevenACoffman , @a8m

StevenACoffman:
Thanks very much! I really appreciate all of your contributions.

Issue #3160, opened 2024-06-25
Using @entityResolver(multi: true) causes "directive entityResolver is not implemented" to be thrown when the entity is used in another Query


RyanMcCabe-catapult:
### What happened?

I have a shared entity between two subgraphs, to keep things simple I'll call them the List service and the Item service. 
Items can be retrieved in 3 ways:

1. Items may appear as a collection of Items under a List (a federated entity between the List and Item services) 
2. They may be queried for against the Item service using some filters (Query definition: `items(filter: ItemsFilter!):ItemPage`) 
3. They may be queried for directly against the Item service if the ID is known (Query definition: `item(id:ID!): Item`)

I've set these up using the directive `@key(fields: "id")` and everything works as expected.
The requests to resolve large lists of items was a performance concern as each item was being resolved with a single request to the Item service, so in order to reduce this to a single call I introduced the `@entityResolver(multi: true)` directive to the entity definition in the Item service. This worked fine for the federated calls and only a single request was being made to resolve multiple Item IDs (scenario 1). 
However in scenario 2 and 3 I'm now getting a `Failed to query 'items': directive entityResolver is not implemented` and a `Failed to query 'item': directive entityResolver is not implemented` error message for the corresponding queries. 
From debugging the generated code, it seems that for scenario 2 the correct item filter resolver is called and returns the data, it then tries to resolve the fields and throws the error. For scenario 3, the item by id resolver is never called and the error is thrown.

### What did you expect?

I expected the queries in scenarios 2 and 3 above to continue working after the `@entityResolver(multi: true)` directive was introduced.

### Minimal graphql.schema and models to reproduce

If the resolvers are generated for the following, the single resolver query will work as expected, the multi resolver query will show the issue.

```
type RegularEntity @key(fields: "id") {
    id: ID!
    Name: String!
}

type RegularEntityWrapper {
    id: ID!
    Name: String!
    Entity: RegularEntity!
}

type MultiResolverEntity @key(fields: "id") @entityResolver(multi: true) {
    id: ID!
    Name: String!
}

type MultiResolverEntityWrapper {
    id: ID!
    Name: String!
    Entity: MultiResolverEntity!
}

type Query {
    singleResolver: RegularEntityWrapper
    multiResolver: MultiResolverEntityWrapper
}
```

### versions
 - gqlgen v0.17.45
 - go1.22.2 darwin/arm64

markkrijgsman:
I have the same issue, my minimum schema is as follows:
`
[...]

PR #3260, opened 2024-09-09
Adding entity resolver fix for entityResolver(multi: true) directive

RyanMcCabe-catapult:
A fix for the issue described in [issue 3160](https://github.com/99designs/gqlgen/issues/3160). If the multi entity is used in any other resolver the error would always be thrown, now it is just skipped. The multi entity resolver has the directive set and runs it, the other resolvers don't have it set so it gets skipped and works as expected.

I have:

### DIFF
--- a/codegen/directives.gotpl
+++ b/codegen/directives.gotpl
@@ -18,8 +18,11 @@
 					{{- end }}
 			{{- end }}
 			if ec.directives.{{$directive.Name|ucFirst}} == nil {
-				var zeroVal {{$zeroVal}}
-				return zeroVal, errors.New("directive {{$directive.Name}} is not implemented")
+				{{- if ne $directive.Name "entityResolver" }}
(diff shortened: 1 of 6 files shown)


clayne11:
@RyanMcCabe-catapult  thanks for trying to fix this. This isn't the best way to implement this.

Inside of [plugin/federation.go](https://github.com/99designs/gqlgen/blob/1855758d8642404df9bbd80386cd9030ae96efa4/plugin/federation/federation.go#L98) you should add `entityResolver` to the `skipRuntime` config.

clayne11:
This fix has [already landed](https://github.com/99designs/gqlgen/blob/master/plugin/federation/federation.go#L111). If you use the newest version this should be working for you now.

RyanMcCabe-catapult:
Thanks @clayne11, Just noticed that when I synced with latest. I'll close out the PR and the issue

PR #3278, opened 2024-09-15
SSE Transport must be inserted before POST, so add convenience for using the DefaultServer

StevenACoffman:
Replacement for #3153 to allow upgrade from POST request to SSE transport response

In order to enable the experimental Server-Sent Events (SSE) transport, you need to have the precedence before the POST transport, but many folks use the DefaultServer which only allows Adding (appending) transports.

This adds convenience methods to the server to both prepend (which is NOT appropriate for SSE) and a special one for inserting right before POST (which is ideal for SSE).

### DIFF
--- a/graphql/handler.go
+++ b/graphql/handler.go
@@ -93,6 +93,7 @@ type (
 	Transport interface {
 		Supports(r *http.Request) bool
 		Do(w http.ResponseWriter, r *http.Request, exec GraphExecutor)
+		String() string
 	}
 )
 
--- a/graphql/handler/server.go
+++ b/graphql/handler/server.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"slices"
 	"time"
 
 	"github.com/vektah/gqlparser/v2/ast"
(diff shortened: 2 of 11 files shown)


StevenACoffman:
@o8x  @jmic @UnAfraid People who were using the [NewDefaultServer](https://github.com/99designs/gqlgen/blob/master/graphql/handler/server.go#L34) were getting tripped up by an inability to add SSE support since it needs to be inserted in the middle of the transport precedence. 

Currently, the server just picks the first available transport, and the NewDefaultServer orders them:
+ Websocket
+ Options
+ GET
+ POST
+ MultipartForm

I did not just enable SSE for NewDefaultServer because when used over HTTP/1.1 (not HTTP/2), SSE suffers from a severe limitation to the maximum number of open connections of 6 per browser. See [here](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#sect1)

We might need to make our transport selection more sophisticated to enable default out-of-the-box support for SSE.

UnAfraid:
@StevenACoffman I got bit by NewDefaultServer before, because it initializes Websocket transport and adding new one doesn't have effect, we could also deprecate it in favor of 'New'.
Alternatively we could add 'Priority' method on the transport and sort on add to avoid the POST before SSE.
Also on add if the transport already exists we could just replace existing one with the provided one

phughes-scwx:
I'm just a rando, but imo NewDefaultServer is a hindrance to people who are using GQLGen for anything beyond toy projects. It should be deprecated and gqlgen init should add the above lines directly to the server code. Even then, maybe only add Options, Get and Post transports, and update the docs / quick start guide to point out that all additional transports have to be added by hand: so for uploads the first step is "add the multipart form" transport, etc etc.

PR #3289, opened 2024-09-17
Revert "Fix marshaling interfaces and union types (#3211)"

StevenACoffman:
This reverts commit 3556475adafe4b8b97a88fdc1dd7b4be21116522.

Got a report that since #3211 was merged, the generation seems to be broken.
The name of Go elements are no longer equal to the name of GQL types.
For instance, generating code with GQL type as: 
```
union CliTargetSettingsResult @goModel(model: "github.com/raito-io/appserver/lambda/appserver/services/cli.CliTargetSettingsResult") = CliTargetSettings | PermissionDeniedError

type CliTargetSettings @goModel(model: "github.com/raito io/appserver/lambda/appserver/services/cli.TargetSettings")
...
```

Will result in generated code like this:
```
if len(graphql.CollectFields(ec.OperationContext, sel, []string{"CliTargetSettingsResult", "TargetSettings"})) == 0 {
	return graphql.Empty{}
}
```
As `CliTargetSettingsResult` and `TargetSettings` are Go types and not a GQL types, this will always be 0 and return `graphql.Empty{}`. 

On top of that, this could lead to invalid JSON as a response. As Empty{} will be marshaled as an empty string I'm able to generate response data like
```
{"__typename":"xxxx","zzz":{"__typename":"yyy","id":"lR0QWfLtP7TsZcphexSvx","cliSettings":}}
```
Which panics on the JSON marshaling.

### DIFF
--- a/_examples/federation/accounts/graph/generated.go
+++ b/_examples/federation/accounts/graph/generated.go
@@ -3044,27 +3044,15 @@ func (ec *executionContext) __Entity(ctx context.Context, sel ast.SelectionSet, 
 	case nil:
 		return graphql.Null
 	case model.EmailHost:
-		if len(graphql.CollectFields(ec.OperationContext, sel, []string{"Entity", "EmailHost"})) == 0 {
-			return graphql.Empty{}
-		}
 		return ec._EmailHost(ctx, sel, &obj)
 	case *model.EmailHost:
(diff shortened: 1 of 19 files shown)


StevenACoffman:
@krupyansky If you can try this again while addressing the unforeseen problems that this caused, I would appreciate it.

PR #3292, opened 2024-09-18
feat: create @requires integration that functions like a normal resolver

clayne11:
The `explicit_requires` flag currently generates the `@requires` resolver inside the execution context. It also doesn't use the normal resolver resolution process and adds a custom implementation that is always called, even if the field isn't requested.

The new `computed_requires` flag looks for `@requires` directives and mutates the schema to add dynamic field arguments and directives that are used internally by `gqlgen`.

These additional directives (`@entityReference` and `@populateFromRepresentations`) are runtime directives that "computed_requires" adds implementations for that enable us to fetch the correct data from the representations array and populated it into the dynamic argument for the `@requires` field.

The new implementation still isn't typesafe and generates the required fields into a `map[string]any`. This is the same as the existing `explicit_requires` implementation.

We will follow up with type safety in a future PR. To support type safety we need to generate types and marshaling logic for GraphQL operations (vs just schemas like we do today) which is a large lift.

I have:


_examples/federation/products/gqlgen.yml
```diff
@@ -66,3 +67,5 @@ models:
       - github.com/99designs/gqlgen/graphql.Int
       - github.com/99designs/gqlgen/graphql.Int64
       - github.com/99designs/gqlgen/graphql.Int32
+
+omit_complexity: true
```
clayne11, line 71: Simply reduces generated code for unused features. Not a material change.

_examples/federation/reviews/graph/schema.resolvers.go
```diff
@@ -11,23 +11,62 @@ import (
 	"github.com/99designs/gqlgen/_examples/federation/reviews/graph/model"
 )
 
+// ManufacturerID is the resolver for the manufacturerID field.
+func (r *productResolver) ManufacturerID(ctx context.Context, obj *model.Product, federationRequires map[string]interface{}) (*string, error) {
```
clayne11, line 15: New resolver function automatically generated and includes the `@requires` fields in the `federationRequires` map.

codegen/data_test.go
```diff
@@ -41,9 +43,11 @@ func TestData_Directives(t *testing.T) {
 						},
 					},
 				},
-				Name:    "excludeDirective",
-				Args:    nil,
-				Builtin: false,
+				Name: "excludeDirective",
+				Args: nil,
+				DirectiveConfig: config.DirectiveConfig{
+					SkipRuntime: false,
```
clayne11, line 49: Used composition here in the underlying struct so I needed to update the instantiation.

codegen/directive.go
```diff
@@ -19,9 +20,10 @@ func (dl DirectiveList) LocationDirectives(location string) DirectiveList {
 
 type Directive struct {
 	*ast.DirectiveDefinition
-	Name    string
-	Args    []*FieldArgument
-	Builtin bool
+	Name string
+	Args []*FieldArgument
+
+	config.DirectiveConfig
```
clayne11, line 26: There can be multiple config properties, I thought it was easier to just pass this through. I can change this to an explicit map if you'd prefer.

codegen/directive.go
```diff
@@ -172,3 +178,23 @@ func (d *Directive) Declaration() string {
...
 	return res
 }
+
+func (d *Directive) IsBuiltIn() bool {
+	return d.Implementation != nil
+}
+
+func (d *Directive) CallPath() string {
+	if d.IsBuiltIn() {
+		return "builtInDirective" + d.CallName()
```
clayne11, line 188: We now have two types of directives: user provided and "built in" where the implementation is statically provided by the plug in.

For `builtIn` directives we call a static function vs an instance function.

plugin/resolvergen/testdata/filetemplate/out/schema.custom.go
```diff
@@ -36,9 +36,11 @@ type resolverCustomResolverType struct{ *CustomResolverType }
...
-//   - You have helper methods in this file. Move them out to keep these resolver files clean.
-func AUserHelperFunction() {
+//  - When renaming or deleting a resolver the old code will be put in here. You can safely delete
+//    it when you're done.
+//  - You have helper methods in this file. Move them out to keep these resolver files clean.
+/*
+	func AUserHelperFunction() {
 	// AUserHelperFunction implementation
 }
+*/
```
StevenACoffman, line 46: This change is causing one test to fail

plugin/resolvergen/testdata/followschema/out/schema.resolvers.go
```diff
@@ -36,9 +36,11 @@ type resolverCustomResolverType struct{ *CustomResolverType }
...
-//   - You have helper methods in this file. Move them out to keep these resolver files clean.
-func AUserHelperFunction() {
+//  - When renaming or deleting a resolver the old code will be put in here. You can safely delete
+//    it when you're done.
+//  - You have helper methods in this file. Move them out to keep these resolver files clean.
+/*
+	func AUserHelperFunction() {
 	// AUserHelperFunction implementation
 }
+*/
```
StevenACoffman, line 46: This change is causing one test to fail

plugin/resolvergen/testdata/filetemplate/out/schema.custom.go
clayne11: I reverted it but this change gets made automatically when I run `go test ./...`. I think we need to revert #3243.

StevenACoffman: Ah, crud. huh. Either that, or rethink how we do testing for that behavior.

clayne11: Yeah potentially. Either way, I would revert for now, we can fix it later. I don't want `master` to be in a broken state. We should also update CI to ensure the git state isn't dirty after running the tests.

cyberhck:
Hello @StevenACoffman 
First of all, thank you very much for adding this.

I don't have a lot of context surrounding gqlgen, so my question might not make sense, but I'm hoping someone can educate me on this.

When I use a `@requires` field that generates a separate resolver function, what's the rationale of adding the last parameter `federationResolver`?

I was wondering if it was possible to just update the initial object itself, like for example:
Let's say I have a schema where I return postal code, but I require 2 fields, let's say province and city name, when I generate code, this is what I get:

```
PostalCode(ctx context.Context, obj *model.IdentityDocument, federationRequires map[string]any) (*string, error)
```

My question is couldn't we have just hydrated the obj to include those two fields instead and kept the function looking like this?

```
PostalCode(ctx context.Context, obj *model.IdentityDocument) (*string, error)
```

Again, sorry if the question doesn't make sense, I'm trying to understand :slightly_smiling_face:

clayne11:
@cyberhck it's a good thought, but unfortunately it's not possible to do this for a few reasons.

The `@requires` directive allows you to input an arbitrary selection set starting at the current parent `Type` (effectively a fragment on the field's parent `Type`).

The biggest blocker is that in GraphQL, field arguments can only be `Input`, `Scalar`, and `Enum` types. But the `@requires` directive can specify any type, including `Interfaces` and `Unions`. There's no way to represent `Unions` or `Interfaces` in field arguments, thus we need a sidecar contract for including these.

Eventually we will want to codegen the types for these `@requires` fields (instead of using an untyped map), but that is another track of work that I haven't had time to prioritize. it will also require significant refactors to the library to support this.

cyberhck:
Thank you so much for the explanation :) that helps :)

PR #3302, opened 2024-09-26
Handle common initialisms as the non-first word

noamcohen97:
If a common initialisms was the prefix of the second word (or any word other the first one) of an identifier, it was treated as  initialism even though the next rune was uppercase. e.g. "USER_IDENTITY".

This PR should resolve this issue

I have:


StevenACoffman:
Thanks! I really appreciate the clarification in the comments and renaming of variables.

PR #3304, opened 2024-09-26
Call `Unalias()` in `IsNilable()` to support `gotypesalias=1`

### LINKED ISSUE #3301: Bug found when generating schema including arrays of type Any
### What happened?

We have a schema that uses an array of type Any as one of the field's types. Starting with `v0.17.48`, when we `go run github.com/99designs/gqlgen generate`, we encounter a breaking error that essentially adds an extra pointer when unmarshaling: Screenshot 2024-09-25 at 3 12 48 PM

### What did you expect?

Code generation to continue to work even when including a type of `[Any]` in the schema

```graphqls
scalar Any

type Broken {
   fakeField: [Any]!
}
```

### versions
 - `gqlgen >=v0.17.48`
 - `go v1.23.1`

noamcohen97:
Fixed false `IsNilable()` value for any (or any other aliased type) when `gotypesalias=1` (default since Go 1.23.0)
Fixes #3301

I have:

### DIFF
--- a/codegen/config/binder.go
+++ b/codegen/config/binder.go
@@ -520,7 +520,10 @@ func (b *Binder) CopyModifiersFromAst(t *ast.Type, base types.Type) types.Type {
 }
 
 func IsNilable(t types.Type) bool {
-	t = code.Unalias(t)
+	// Note that we use types.Unalias rather than code.Unalias here
+	// because we want to always check the underlying type.
+	// code.Unalias only unwraps aliases in Go 1.23
+	t = types.Unalias(t)
 	if namedType, isNamed := t.(*types.Named); isNamed {
 		return IsNilable(namedType.Underlying())
 	}
--- a/codegen/config/binder.go
+++ b/codegen/config/binder.go
@@ -520,6 +520,7 @@ func (b *Binder) CopyModifiersFromAst(t *ast.Type, base types.Type) types.Type {
 }
 
 func IsNilable(t types.Type) bool {
+	t = code.Unalias(t)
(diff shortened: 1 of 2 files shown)


codegen/config/binder.go
```diff
@@ -520,6 +520,7 @@ func (b *Binder) CopyModifiersFromAst(t *ast.Type, base types.Type) types.Type {
 }
 
 func IsNilable(t types.Type) bool {
+	t = code.Unalias(t)
```
noamcohen97, line 523: Maybe we should call `types.Unalias(t)` so we'll Unalias even in Go 1.22?
Currently in Go 1.22 calling this with a nilable alias type returns false.
@StevenACoffman, @a8m WDYT?

StevenACoffman: We really need to continue to support Go 1.22, and I don't think we can just assume people have gotypesalias=1

noamcohen97: So I'll call `types.Unalias()` here, that way we'll make sure `IsNilable()` is reliable no matter what.
Also - we don't need the pointer base type unaliasing from `code.Unalias()` here.

StevenACoffman:
Thanks! I really appreciate your fix here! Looking forward to your next contribution!

noamcohen97:
Thanks @StevenACoffman!
When can we expect a new release?

noamcohen97:
Thanks @StevenACoffman!
When can we expect a new release?

StevenACoffman:
Right now! Just released!

Issue #3168, opened 2024-07-02
Use embedded templates in root file generation


barr-code:
### What happened?
When a binary is shipped with gqlgen as one of its dependencies, with the follow-schema option set in the config, the following error is encountered:

`open <your file structure path>/github.com/99designs/gqlgen@v0.17.42/codegen/root_.gotpl: no such file or directory`

### What did you expect?
[The fix](https://github.com/99designs/gqlgen/pull/2262) for [issue 2260](https://github.com/99designs/gqlgen/issues/2260) offers the ability to pass in a `TemplateFS` option. However, that option is not being honored by the [generateRootFile](https://github.com/99designs/gqlgen/blob/a13eca12117720d9d6cccba8549c7e6934d09fbd/codegen/generate.go#L139) function. I would expect the same logic that was implemented [here](https://github.com/99designs/gqlgen/blob/a13eca12117720d9d6cccba8549c7e6934d09fbd/codegen/templates/templates.go#L171) to apply in `generateRootFile`.

### versions
 - `go run github.com/99designs/gqlgen version`?
 v0.17.42
 
- `go version`?
go 1.22.4

clayne11:
That makes sense, feel free to PR.

barr-code:
> That makes sense, feel free to PR.

Thanks! I opened a PR to resolve this issue: https://github.com/99designs/gqlgen/pull/3319

PR #3319, opened 2024-10-07
feat: Embed template for root file generation

barr-code:
This PR embeds the `root_.gotpl` file used for generating the rootFile rather than searching for that template at runtime. This change will allow codegen to be packaged as part of a binary without relying on root_.gotpl existing within the user's local file structure.

This PR will resolve this issue: https://github.com/99designs/gqlgen/issues/3168

I have:

### DIFF
--- a/codegen/generate.go
+++ b/codegen/generate.go
@@ -4,9 +4,7 @@ import (
 	"embed"
 	"errors"
 	"fmt"
-	"os"
 	"path/filepath"
-	"runtime"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


codegen/generate.go
```diff
@@ -129,24 +127,18 @@ func addBuild(filename string, p *ast.Position, data *Data, builds *map[string]*
 	}
 }
 
+// go:embed root_.gotpl
```
StevenACoffman, line 130: Looks like an extra space by accident here!

StevenACoffman:
Thanks!

PR #3329, opened 2024-10-14
fix: Ensure that nil http responses don't cause issues

### LINKED ISSUE #3328: Short Term Subscription Issue
### What happened?
I'm trying to create a short term subscription, that loops a few times and then closes. In the process, it sends out an extra response that there is an error (`internal system error`), prior to closing the connection. 

Screenshot 2024-10-14 at 17 32 31

In the process, I found a null pointer error here: 
https://github.com/99designs/gqlgen/blob/6ec416bac7b3d8a36c191e0d6fdf1fd0e5d54c35/graphql/handler/debug/tracer.go#L61-L66

For some reason (I couldn't fully track down why), `resp` is `null` after the `X` time I gave, and there is a panic when it tries calling `resp.Errors`. I can make a PR to guard against that, but I couldn't figure out why `resp` is null there. 

### What did you expect?
I expected that it would work, and run successfully. 

```
# graphql.schema
type Subscription {
  """
  `countFor` will return a stream of ints, for `count` times.
  """
  countFor(count: Int!): Int!
}
```

```
# schema.resolver.go
// CountFor is the resolver for the countFor field.
func (r *subscriptionResolver) CountFor(ctx context.Context, count int) (<-chan int, error) {
	ch := make(chan int)

	go func() {
		defer close(ch)

		for i := 0; i <= count; i++ {
			select {
			case <-ctx.Done():
				return
			case ch <- i:
				time.Sleep(100 * time.Millisecond)
			}
		}
	}()

	return ch, nil
}
```

### versions
 - `go run github.com/99designs/gqlgen version`? `v0.17.55`
 - `go version`? `go version go1.23.1 darwin/amd64`

df-wg:
In https://github.com/99designs/gqlgen/issues/3328, I investigated a debug error which had stopped local subscription development. This PR adds a simple `nil` guard around the problematic statement. 

I tested this PR  with our local setup and it solved our issues

I have:

Fixes #3328

### DIFF
--- a/graphql/handler/debug/tracer.go
+++ b/graphql/handler/debug/tracer.go
@@ -61,9 +61,12 @@ func (a Tracer) InterceptResponse(ctx context.Context, next graphql.ResponseHand
 	resp := next(ctx)
 
 	_, _ = fmt.Fprintln(a.out, "  resp:", aurora.Green(stringify(resp)))
-	for _, err := range resp.Errors {
-		_, _ = fmt.Fprintln(a.out, "  error:", aurora.Bold(err.Path.String()+":"), aurora.Red(err.Message))
+	if resp != nil {
+		for _, err := range resp.Errors {
+			_, _ = fmt.Fprintln(a.out, "  error:", aurora.Bold(err.Path.String()+":"), aurora.Red(err.Message))
+		}
 	}
+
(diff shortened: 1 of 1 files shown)


StevenACoffman:
Hey, thanks!

PR #3330, opened 2024-10-15
feat: add ability to unmarshal Upload from map[string]interface

golanglemonade:
## Summary:

When using the generated server client, the types come through as a `map[string]interface{}` rather than a `graphql.Upload`. This change allows the generic map to be unmarshalled into the `Upload` type:

example from `gen_server.go`, this currently fails because its coming through as a `map[string]interface{}` instead of a `graphql.Upload`

```
func (ec *executionContext) unmarshalNUpload2githubᚗcomᚋ99designsᚋgqlgenᚋgraphqlᚐUpload(ctx context.Context, v interface{}) (graphql.Upload, error) {
	res, err := graphql.UnmarshalUpload(v)
}
```

I have:

### DIFF
--- a/graphql/upload.go
+++ b/graphql/upload.go
@@ -1,15 +1,20 @@
 package graphql
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 )
 
+var (
+	invalidUpload = "%T is not an Upload"
+)
+
 type Upload struct {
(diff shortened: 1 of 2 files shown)


golanglemonade:
this actually looks like a problem with the gqlgenc client, rather than the parser here, going to close this out. sorry for the noise.

Issue #3331, opened 2024-10-15
Broken graphql.NoCache implementation


adomaskizogian:
`graphql.NoCache[T,T2]` struct implements `graphql.Cache[T]` interface but it is not assignable to any of the internally used fields. For example it is not possible to assign `NoCache` as apq cache. There's no way to instantiate `NoCache` with value type values. Only with pointer values.

Specifying key type on `graphql.NoCache[T,T2], // NoCache[T any, T2 *T]` forces the value of cache to be a pointer of key type. For example:
`graphql.NoCache[string, *string]`.

A code snippet like bellow is not possible:

```go

var cache graphql.Cache[string]
if config.UseAPQ {
  cache = &graphql.MapCache[string]{}
} else {
  cache = &graphql.NoCache[string]{}
}
```

Internally all caches use string type keys.

Type definitions were changed with https://github.com/99designs/gqlgen/pull/3179

adomaskizogian:
cc @sylr

PR #3332, opened 2024-10-15
refactor graphql.NoCache. use string type keys

adomaskizogian:
Fixes: https://github.com/99designs/gqlgen/issues/3331

All internal uses of cache use string type keys and leaves out the value type as generic parameter. The new generic `NoCache` implementation only allows specifying the key type and the value type is a pointer of the key type. Meaning if I specify the key type `string` the value type is `*string`. This implementation cannot be assigned in place of `graphql.Cache[any]`

I propose to fix this by updating the `NoCache` definition by dropping the key type and allowing to use only string as key, any as value

See #3179


StevenACoffman:
Thanks!

Issue #3342, opened 2024-10-23
@requires fields unable to deal with Union


rickbijkerk:
### What happened?
Code generation breaks when you apply an @requires to an Union field

if you run codegen for the following schema 
```
type Person @key(fields: "name"){
    name: String!
    gender: Gender!
    welcomeMessage: String @requires(fields:"gender { ... on Male {description} ... on Female {description}}")
}

union Gender = Male | Female

type Male {
    description: String!
}

type Female {
    description: String!
}
```

you'll get an exception:
```panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x30 pc=0x102a8df30]

goroutine 1 [running]:
github.com/99designs/gqlgen/plugin/federation/fieldset.fieldByName(0x102ce0a38?, {0x14000260681, 0x3})
        /Users/rickbijkerk/projects/gqlgen/plugin/federation/fieldset/fieldset.go:207 +0x20
github.com/99designs/gqlgen/plugin/federation/fieldset.Field.TypeReference({0x1400007f700?, 0x14000035990?, 0x14000029a70?}, 0x2a?, {0x140004f5100, 0x13, 0x6?})
        /Users/rickbijkerk/projects/gqlgen/plugin/federation/fieldset/fieldset.go:85 +0x100
github.com/99designs/gqlgen/plugin/federation.(*Federation).GenerateCode(0x1400025e150, 0x140000b8790)
        /Users/rickbijkerk/projects/gqlgen/plugin/federation/federation.go:297 +0xd48
github.com/99designs/gqlgen/api.Generate(0x140000e1688, {0x0, 0x0, 0x103029d20?})
        /Users/rickbijkerk/projects/gqlgen/api/generate.go:132 +0xb88
main.init.func2(0x1400008ab80?)
        /Users/rickbijkerk/projects/gqlgen/main.go:190 +0xa8
github.com/urfave/cli/v2.(*Command).Run(0x140000fc6e0, 0x1400008ab80, {0x14000020090, 0x1, 0x1})
        /Users/rickbijkerk/go/pkg/mod/pkg/mod/github.com/urfave/cli/v2@v2.27.5/command.go:276 +0x71c
github.com/urfave/cli/v2.(*App).RunContext(0x140000dc400, {0x102ce4530, 0x1030aa180}, {0x14000020090, 0x1, 0x1})
[...]

PR #3347, opened 2024-10-28
Fix union queries not working in @requires

rickbijkerk:
fixes: https://github.com/99designs/gqlgen/issues/3342

also tested in our project and its working ; )

I have:
 - [x ] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [x ] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

### DIFF
--- a/plugin/federation/federation.go
+++ b/plugin/federation/federation.go
@@ -279,8 +279,6 @@ func (f *Federation) GenerateCode(data *codegen.Data) error {
 				typeString := strings.Split(obj.Type.String(), ".")
 				requiresImports[strings.Join(typeString[:len(typeString)-1], ".")] = true
 
-				cgField := reqField.Field.TypeReference(obj, data.Objects)
-				reqField.Type = cgField.TypeReference
 			}
 
 			// add type info to entity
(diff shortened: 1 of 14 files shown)


plugin/federation/federation.go
```diff
@@ -4,6 +4,7 @@ import (
 	_ "embed"
 	"errors"
 	"fmt"
+	"github.com/99designs/gqlgen/internal/rewrite"
```
StevenACoffman, line 7: Oh thanks!

rickbijkerk: ah must have been auto formatted, let me change that!

clayne11:
`explicit_requires` is deprecated — did you test using `computed_requires` and does this issue still occur there?

rickbijkerk:
yes, also tried it with `computed_requires` same issue occurs there

clayne11:
Got it, thanks. I believe this code is not really used in the `computed_requires` flow but it is still executed until we fully remove the `explicit_requires` code path. Thanks for the fix.

StevenACoffman:
@rickbijkerk Hey, can you run:
```
go generate ./...
```
Thanks for this!

Issue #3321, opened 2024-10-12
Massive undocumented breaking change in v0.17.50


Vilsol:
I have hit a major issue due to this undocumented breaking change (https://github.com/99designs/gqlgen/pull/3243), which was introduced in a patch version bump.

This is how the `resolver.go` looked like before on version `v0.17.49`: https://github.com/satisfactorymodding/smr-api/blob/179d26156ff1bfc48119ab24ee8a4740d2a2a18b/gql/resolver.go

This is how it looks like after running the generator on version `v0.17.50`: https://gist.github.com/Vilsol/1d7ac5dabd75933604449e1d6faa3323

* It has not generated any imports for anything except for what it itself is using.
* It has blindly copied the code from within the resolver functions, disregarded the function argument names, therefore completely breaking code that used those variables.

Is there some config options I can change to make this work with newer versions? Or is my only possible path to not upgrade?

JonasDoe:
Gotta agree, I don't really understand what the intention behind this change was. My situation isn't as dramatic, I just added some fields to the `Resolver` which I probably have to pass by `ctx` right now. But what's the gain of this? It feels like it only introduces new incompatibilities which little benefit.

UnAfraid:
Yeah i feel you, behaviour altering changes should be behind configuration option usually opt-in

StevenACoffman:
This was bad situation is actually both my mistake and my miscommunication combined, and I'm quite sorry.

There was originally a config option `enable_rewrite_for_single_file` in that PR  to opt-in to rewriting resolvers or to preserve existing resolver code even when the schema changed to no longer match.

I suggested that instead the config option be `rewrite` apply equally to both `single-file` and `follow-schema`. My intent was to reduce the growing divergence in the code that places all resolvers in a single file and the code that places the resolvers in a directory layout that matches the schema.

The person who did that PR was an intern under the supervision of another maintainer, and the maintainer and I had side conversations so I assumed that it had been more carefully vetted and I merged it based on that assessment, rather that doing my own.

That was a big mistake. Again, I'm sorry.

I'd welcome a PR to add a config option to at least opt-out of this rewrite behavior.

JonasDoe:
Thanks for the response! If I find time, I can try to create a PR. But it might take a while, so if somebody is more into the topic and faster, I'll gladly yield it. ;)

StevenACoffman:
This appears to illustrate the PR change before the config option was removed:
[...]

PR #3359, opened 2024-11-08
Config option to preserve single file resolvers instead of always rewriting

StevenACoffman:
See issue #3321 - users of gqlgen who upgraded to v0.17.50 and had configured a layout of "single-file" for resolvers started to find their existing resolvers suddenly overwritten when re-generating models for schema changes. This change in behavior broke things for a number of people (sorry). It was intended that this behavior would be opt-in with a configuration option, but that was omitted.

This behavior change was introduced in #3243 to solve #3205 so that Bazel integration was possible (AFAICT Bazel needs a layout of "single-file" but needs to regenerate the resolver every time the schema updates).  This was necessary for a big rewrite of the Apollo Federation code contributed by an organizations that use Bazel.

This PR adds a configuration option `preserve_resolver` that if explicitly set to true will not avoid rewriting the resolver when gqlgen is run.

This means that people who want the v0.17.49 single-file resolver behavior will need to add preserve_resolver: true to their gqlgen.yaml config.

I'm actively soliciting feedback as to whether it should be `opt-out` with `preserve_resolver` vs `opt-in` with `rewrite_resolver`.

What has always been awkward (and is of special concern in this PR) is that the gqlgen resolver rewrite/preserve behavior for a configured layout of `follow-schema` is the opposite of the old v0.17.49 `single-file`.

If we made the config option `rewrite_resolver` so that the default was false, then that implies that everyone who uses `follow-schema` layout needs to add `rewrite_resolver: true` or silently fail to update their resolvers.

I've already been very disruptive to one set of gqlgen users, I would prefer not to "fix it" by inflicting similar disruption to others, but I'm open to feedback or alternatives.

---
Note, I also fixed the annoying test regeneration cycle in a few tests so those tests begin by resetting to a good state before running.


StevenACoffman:
FYI @Vilsol @UnAfraid @JonasDoe

JonasDoe:
Wow, thanks for the fast fix!

To answer your question: personally, the only reason why I'ld favor to default `preserve_resolver` to `true` is that it feels more like a guarantee this behavior isn't deprecated and will remain fully supported in the future. But from what I understand after reading your remarks, there's no reason to assume it might get dropped. So I'ld be fine with defaulting the setting to `false`, either. 🤷‍♂️ 

But if I understand correctly the big question you're asking is: how many ppl. have already updated their `gqlgen` since September, and how many haven't yet and might be startled by the changed behavior in the future? I'ld think most users don't update `gqlgen` avidly - you don't change your schemas that often, if even if you did, you wouldn't necessarily update this library each time. I might be wrong here, of course.

argoyle:
I run Renovate on all my repositories to keep all dependencies up to date at all time (and recommend everyone else to do the same). Just to show the other side of the argument. 😁

StevenACoffman:
We use `gopkg.in/yaml.v3` to unmarshal/deserialize/decode from the yaml config to the [resolver config struct](https://github.com/99designs/gqlgen/blob/75966dc09aacbcc0c1b9ee6f84ad21699dd97f58/codegen/config/resolver.go#L13) so we cannot easily have gqlgen yaml configs that are missing any value for `preserve_resolver` default to true, as the zero value for a boolean is false.

There are [4,513 public repositories](https://pkg.go.dev/github.com/99designs/gqlgen/graphql?tab=importedby) that import gqlgen and of the ones that appear actively maintained (commits in the last year) and use `single-file` layout, the majority have already updated to v0.17.50 or beyond.

I can't think of any way to get insight into private repository gqlgen version update information.

So it looks to me like if I flipped the single-file resolver rewrite back to the preserving behavior of v0.17.49  I would just make more trouble for most people who have already moved on.

BTW, there are 598 public repositories that have upgraded to v0.17.5x according to this: https://github.com/search?q=%22github.com%2F99designs%2Fgqlgen+v0.17.5%22&type=code

Based on that and the feedback above, I'm going to try merging this PR as is, cut a release, and update the v0.17.50 release notes. Hopefully we can advertise this workaround to the break better for anyone who is late to updating.

StevenACoffman:
Fixes #3321 as best it can be fixed

JonasDoe:
> I run Renovate on all my repositories to keep all dependencies up to date at all time (and recommend everyone else to do the same). Just to show the other side of the argument. 😁

Yeah, the three of us who were posting in this issue probably belong to this other side. ;)

Regarding default values: I think I was just a bit sloppy on responding. It would have been `rewrite_resolver: false` instead of `preserve_resolver: true` then ofc.

This quey is interesting! But if you go the versions backwards here, it looks like only a minority of the public repos was actually updated.

Anyway, every solution is a good one, thank you again!

Issue #3372, opened 2024-11-13
Improve: optimize transport.Post


lkeix:
### What happened?

The `transport.Post` implementation in the `Do` method has potential for optimization. Specifically, two areas could improve performance:
1. The `Do` method frequently converts `[]byte` to `string`, which incurs additional allocations.
2. `graphql.RawParams` could benefit from a `sync.Pool` to reduce allocations and reuse memory.

### What did you expect?

Optimizing memory allocation in the `Do` method should lead to better overall performance for the gqlgen server, especially for high-throughput applications.

### Minimal graphql.schema and models to reproduce

I tested using the gqlgen initial schema and an empty slice return for the `Todos` query, observing allocation patterns during benchmarking.

### Versions
- gqlgen version: `v0.17.55` (via `go run github.com/99designs/gqlgen version`)
- Go version: `go1.23.2 darwin/amd64`

### Pull Request

lkeix:
@StevenACoffman 
I took benchmark again, and check the profile.
It seems the gqlgen bottle-neck is json decoding below.

<img width="398" alt="スクリーンショット 2024-12-04 20 31 06" src="https://github.com/user-attachments/assets/667ee2c8-a613-48bf-9819-0818add5798d">

I took benchmark using goccy/go-json instead of encoding/json, total memory usage is 1.15% (832 B / 720 B) optimized.
<img width="642" alt="スクリーンショット 2024-12-04 20 30 56" src="https://github.com/user-attachments/assets/d6ce9e1c-e6c6-4209-bcbf-5f28995a8bec">

I created PR #3405

StevenACoffman:
@lkeix the benchmarking is very helpful! 

In #2842 it was proposed to allow JSON configuration to be configurable, and several people in the community have warned me against defaulting to using goccy/go-json [as in this comment](https://github.com/99designs/gqlgen/pull/3161#issuecomment-2350101842). In #2757 it was suggested to that there are more than one alternative JSON encoding library people might want to use:

Some of these libraries include:
- [goccy/go-json](https://github.com/goccy/go-json)
- [json-iterator/go](https://github.com/json-iterator/go)
- [bytedance/sonic](https://github.com/bytedance/sonic)
- [segmentio/encoding](https://github.com/segmentio/encoding)
- [mailru/easyjson](https://github.com/mailru/easyjson)
- [minio/simdjson-go](https://github.com/minio/simdjson-go)
[...]

PR #3373, opened 2024-11-13
Improve transport.Post Do method

lkeix:
### Description

This PR optimizes the `Do` method in `transport.Post` by addressing memory allocation inefficiencies. Specifically:

1. Removed unnecessary `[]byte` to `string` conversions in the `Do` method, which were causing additional memory allocations.
2. Introduced `sync.Pool` for `graphql.RawParams` to enable memory reuse, reducing overall allocations.

These changes aim to improve the performance of the gqlgen server, especially in high-throughput scenarios.

### Related Issue
Resolves [#3372](https://github.com/99designs/gqlgen/issues/I3372) (replace with actual issue link)

### Checklist

I have:
 - [ ] Added tests covering the optimization (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))

### DIFF
--- a/graphql/handler/transport/http_post.go
+++ b/graphql/handler/transport/http_post.go
@@ -1,11 +1,11 @@
 package transport
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"mime"
 	"net/http"
-	"strings"
 
 	"github.com/vektah/gqlparser/v2/gqlerror"
 
@@ -56,26 +56,27 @@ func (h POST) Do(w http.ResponseWriter, r *http.Request, exec graphql.GraphExecu
 		Start: start,
 		End:   graphql.Now(),
 	}
-
-	bodyString, err := getRequestBody(r)
+	
+	bodyBytes, err := io.ReadAll(r.Body)
 	if err != nil {
-		gqlErr := gqlerror.Errorf("could not get json request body: %+v", err)
-		resp := exec.DispatchError(ctx, gqlerror.List{gqlErr})
-		writeJson(w, resp)
-		return
+			gqlErr := gqlerror.Errorf("could not read request body: %+v", err)
+			resp := exec.DispatchError(ctx, gqlerror.List{gqlErr})
+			writeJson(w, resp)
+			return
 	}
 
-	bodyReader := io.NopCloser(strings.NewReader(bodyString))
-	if err = jsonDecode(bodyReader, &params); err != nil {
-		w.WriteHeader(http.StatusBadRequest)
-		gqlErr := gqlerror.Errorf(
-			"json request body could not be decoded: %+v body:%s",
-			err,
-			bodyString,
-		)
-		resp := exec.DispatchError(ctx, gqlerror.List{gqlErr})
-		writeJson(w, resp)
-		return
+
+	bodyReader := bytes.NewReader(bodyBytes)
+	if err := jsonDecode(bodyReader, &params); err != nil {
+			w.WriteHeader(http.StatusBadRequest)
+			gqlErr := gqlerror.Errorf(
+					"json request body could not be decoded: %+v body:%s",
+					err,
+					string(bodyBytes),
+			)
+			resp := exec.DispatchError(ctx, gqlerror.List{gqlErr})
+			writeJson(w, resp)
+			return
 	}
 
 	rc, opErr := exec.CreateOperationContext(ctx, params)
--- a/graphql/handler/transport/http_post.go
+++ b/graphql/handler/transport/http_post.go
@@ -74,26 +74,26 @@ func (h POST) Do(w http.ResponseWriter, r *http.Request, exec graphql.GraphExecu
 		Start: start,
 		End:   graphql.Now(),
 	}
-	
+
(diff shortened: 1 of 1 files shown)


lkeix:
I took benchmark simple code.
schema is generated project initizalization.

**resolver implementation**
```go
// CreateTodo is the resolver for the createTodo field.
func (r *mutationResolver) CreateTodo(ctx context.Context, input model.NewTodo) (*model.Todo, error) {
	return &model.Todo{}, nil
}

// Todos is the resolver for the todos field.
func (r *queryResolver) Todos(ctx context.Context) ([]*model.Todo, error) {
	return []*model.Todo{}, nil
}
```

benchmark implemantation
```go

func BenchmarkErrorReponse(b *testing.B) {
	srv := handler.NewDefaultServer(graph.NewExecutableSchema(graph.Config{Resolvers: &graph.Resolver{}}))

	query := `{
		"query": "query { todos { id text } }"
	}`

	testsrv := httptest.NewServer(srv)

	q := bytes.NewBufferString(query)

	client := &http.Client{}

	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		b.StopTimer()
		req, err := http.NewRequest(http.MethodPost, testsrv.URL, q)
		if err != nil {
			b.Fatalf("Failed to create request: %v", err)
		}
		req.Header.Set("Content-Type", "application/json")

		b.StartTimer()
		resp, err := client.Do(req)
		if err != nil {
			b.Fatalf("Request failed: %v", err)
		}
		b.StopTimer()
		resp.Body.Close() 
	}
	b.StopTimer()
}
```

graphql/handler/transport/http_post.go
```diff
@@ -46,32 +47,49 @@ func getRequestBody(r *http.Request) (string, error) {
...
-		gqlErr := gqlerror.Errorf("could not get json request body: %+v", err)
+		gqlErr := gqlerror.Errorf("could not read request body: %+v", err)
 		resp := exec.DispatchError(ctx, gqlerror.List{gqlErr})
 		writeJson(w, resp)
 		return
 	}
 
-	bodyReader := io.NopCloser(strings.NewReader(bodyString))
-	if err = jsonDecode(bodyReader, &params); err != nil {
+	bodyReader := bytes.NewReader(bodyBytes)
```
StevenACoffman, line 86: I'm a little nervous about removing the `io.NopCloser` here. We used to have quite a few issues with accidentally closing `io.Reader` more than once, and it was pretty hard to track down.

StevenACoffman:
I greatly appreciate you looking into performance improvements, but I want to ask how you have picked which places to optimize here and in your other PR.

It would be extremely valuable to the entire project to identify where the biggest source of allocations or CPU usage are coming from during a realistic load scenario. It may be that you have done that, but even if this is the most important place, I would like to hear what you found were second or third and why, as others in the community may be able to help on those.

Also, I'm not sure how far to take performance improvements in this area at the expense of readability and maintainability.

For instance, io.ReadAll is much more readable than using io.Copy, but it does produce more allocations. Make up a 10 MB text file:
```
func BenchmarkReadAll(b *testing.B) {
    for i := 0; i < b.N; i++ {
        in, _ := os.Open("sample.txt")
        io.ReadAll(in)
        in.Close()
    }
}

func BenchmarkCopy(b *testing.B) {
    for i := 0; i < b.N; i++ {
        buf := &bytes.Buffer{}
        in, _ := os.Open("sample.txt")
        io.Copy(buf, in)
        in.Close()
    }
}
```
Output:
```
goos: linux
goarch: amd64
pkg: example.org
cpu: Intel(R) Xeon(R) CPU E5-2620 v4 @ 2.10GHz
BenchmarkReadAll-32           26      45445764 ns/op    102635996 B/op        43 allocs/op
BenchmarkXxx-32               48      23985133 ns/op    67108544 B/op         21 allocs/op
```

Or take for example this article: https://klotzandrew.com/blog/speeding-up-json-processing-in-go/
This is very easily read and maintained:
```
  bodyBytes, _ := ioutil.ReadAll(body)
  box := BoxType{}
  _ = json.Unmarshal(bodyBytes, &box)
```
But it can definitely be made faster and allocate much less at the expense of complicating it as in the article.

It is worth it if the gains are worth it in comparison to everything else, but if it's a small part of the overall, then it's not worth it.

With this PR, I can't tell how much your performance improvement in this area relates to the overall CPU/memory performance. If you have already worked that out, I would love it if you could share that.

graphql/handler/transport/http_post.go
StevenACoffman: Ok, nevermind. It looks like that doesn't apply at all here (or at least it doesn't anymore). That was a different part of the codebase.

StevenACoffman:
I'm merging the PR here, but I would still appreciate someone sharing an analysis of the most performance critical areas of gqlgen runtime.

lkeix:
@StevenACoffman  
Apologies for the delayed response.  
Thank you for merging this PR.  

I will continue to share insights on gqlgen runtime bottlenecks in [this issue](https://github.com/99designs/gqlgen/issues/3372).

Issue #3371, opened 2024-11-12
proposal: Allow the custom template for `exec` code generation


OldBigBuddha:
### What happened?

The following template generates an unlimited number of goroutines at once, which sometimes requires a large amount of memory.

https://github.com/99designs/gqlgen/blob/aaf44f5b1ec9d391aa613d48a324074d8e2d0e3f/codegen/type.gotpl#L134

For example, the minimum size of a goroutine is 2KiB, so processing 10,000 objects that have some fields with child resolvers would require more than 20 MiB.

### What did you expect?

I hope that If I specify a template as follows, you want to use it for code generation in exec:

```yaml
exec:
  filename: graph/generated.go
  package: generated
  exec_template: graph/template/exec.gotpl
```

#### backgroud

My real goal is to restrict the number of generating goroutine somehow, like using semaphore.

The following is a part of the `generated.go` file generated by the latest version of gqlgen:

```go
ret := make(graphql.Array, len(v))
var wg sync.WaitGroup
isLen1 := len(v) == 1
if !isLen1 {
    wg.Add(len(v))
}
for i := range v {
    i := i
    fc := &graphql.FieldContext{
        Index:  &i,
        Result: &v[i],
    }
    ctx := graphql.WithFieldContext(ctx, fc)
    f := func(i int) {
        defer func() {
            if r := recover(); r != nil {
                ec.Error(ctx, ec.Recover(ctx, r))
                ret = nil
            }
        }()
        if !isLen1 {
            defer wg.Done()
        }
        ret[i] = ec.marshalNTodo2ᚖgithubᚗcomᚋOldBigBuddhaᚋgqlgenᚑgoroutineᚑrestrictionᚑworkaroundᚋgraphᚋmodelᚐTodo(ctx, sel, v[i])
    }
    if isLen1 {
        f(i)
    } else {
        go f(i)
    }

}
wg.Wait()

for _, e := range ret {
    if e == graphql.Null {
        return graphql.Null
    }
}

return ret
```

If I can rewrite it like this, I can limit the number of goroutines created.

```go
ret := make(graphql.Array, len(v))
var wg sync.WaitGroup
sm := semaphore.NewWeighted(1000)
isLen1 := len(v) == 1
if !isLen1 {
    wg.Add(len(v))
}
for i := range v {
    i := i
    fc := &graphql.FieldContext{
        Index:  &i,
        Result: &v[i],
    }
    ctx := graphql.WithFieldContext(ctx, fc)
[...]

PR #3375, opened 2024-11-14
proposal: allow the custom template for server code generation

OldBigBuddha:
# Overview

I have made `gqlgen` possible to set custom templates for the server code generation that I previously proposed at #3371.

If you set the file or directory path to the template in config as below, code generation will be performed by referring to those files rather than the templates managed by gqlgen:

```yaml
# Where are all the schema files located? globs are supported eg  src/**/*.graphqls
schema:
  - graph/schema/**/*.graphqls

# Where should the generated server code go?
exec:
  filename: graph/generated/autogen.go
  package: generated
  exec_template_dir: graph/template
  # you can specify a single template file
  # exec_template: graph/template/exec.gotpl
```

This change allows users who want to change the core logic of the server for exceptional use cases to continue to use the original `gqlgen` with only custom template preparation.

I have:

### DIFF
--- a/api/generate_test.go
+++ b/api/generate_test.go
@@ -16,7 +16,10 @@ func cleanup(workDir string) {
 	_ = os.Remove(filepath.Join(workDir, "graph", "resolver.go"))
 	_ = os.Remove(filepath.Join(workDir, "graph", "federation.go"))
 	_ = os.Remove(filepath.Join(workDir, "graph", "schema.resolvers.go"))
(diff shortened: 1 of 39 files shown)


StevenACoffman:
While this is excellent work, and is a very useful example for people who want to experiment with alternative executions, I'm going to close this PR given that we found a less invasive way to provide people with control over the trade off between memory and speed.

PR #3404, opened 2024-12-03
Deprecate NewDefaultServer

StevenACoffman:
This was only there for tests and as an example, but people have tried to use it in their production environments and found out the hard way it wasn't intended for that. Deprecating will hopefully make this more clear.

However, all the deprecation warnings for all the test usage cause the linter to fail. 😢 

I don't know of a great solution, so if you have a suggestion for how to fix this, I would appreciate it.

I'm not sure of a great way to make a test only equivalent function without using a special build constraint tag that would then cause more people to stumble on more regular `go test ./...` 

I could add a `//nolint:staticcheck // SA1019: handler.NewDefaultServer is deprecated but usage for test is ok` everywhere, but that ignores **_all_** staticcheck checks, not just the deprecation one.

I could copy-paste the NewDefaultserver function to be nested inside the various call sites, but that's a lot of duplication.

### DIFF
--- a/_examples/chat/chat_test.go
+++ b/_examples/chat/chat_test.go
@@ -5,7 +5,9 @@ import (
 	"runtime"
 	"sync"
 	"testing"
+	"time"
 
+	"github.com/99designs/gqlgen/graphql/handler/transport"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
@@ -14,7 +16,12 @@ import (
 )
 
 func TestChatSubscriptions(t *testing.T) {
-	c := client.New(handler.NewDefaultServer(NewExecutableSchema(New())))
+	srv := handler.New(NewExecutableSchema(New()))
+	srv.AddTransport(transport.Websocket{
+		KeepAlivePingInterval: time.Second,
+	})
+	srv.AddTransport(transport.POST{})
+	c := client.New(srv)
 
 	const batchSize = 128
 	var wg sync.WaitGroup
--- a/_examples/config/server/server.go
+++ b/_examples/config/server/server.go
@@ -6,13 +6,18 @@ import (
 
 	todo "github.com/99designs/gqlgen/_examples/config"
 	"github.com/99designs/gqlgen/graphql/handler"
+	"github.com/99designs/gqlgen/graphql/handler/transport"
 	"github.com/99designs/gqlgen/graphql/playground"
 )
 
 func main() {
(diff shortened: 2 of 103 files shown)


phughes-scwx:
What help do you want?

StevenACoffman:
All the deprecation warnings for all the test usage cause the linter to fail. 😢 

I don't know of a great solution, so if you have a suggestion, I would appreciate it.

I'm not sure of a great way to make a test only equivalent function without using a special build constraint tag that would then cause more people to stumble on more regular `go test ./...` 

I could add a `//nolint:staticcheck // SA1019: handler.NewDefaultServer is deprecated but usage for test is ok` everywhere, but that ignores **_all_** staticcheck checks, not just the deprecation one.

I could copy-paste the NewDefaultserver function to be nested inside the various call sites, but that's a lot of duplication.

phughes-scwx:
I can take a pass.

phughes-scwx:
https://github.com/99designs/gqlgen/pull/3406 has this branch as its base, as I can not push directly.

StevenACoffman:
Thanks again!

idc77:
How do you propose to replace it and what did "they find out the hard way"?

StevenACoffman:
@idc77 

In the code, the comment for it now tries to explain that you should make your own local version of it, customized to suit your own needs (only add the transports you want, and customize the cache and transports to match your workloads):

```
// NewDefaultServer is a demonstration only. Not for prod.
//
// Currently, the server just picks the first available transport,
// so this example NewDefaultServer orders them, but it is just
// for demonstration purposes.
// You will likely want to tune and better configure Websocket transport
// since adding a new one (To configure it) doesn't have effect.
//
// Also SSE support is not in here at all!
// SSE when used over HTTP/1.1 (but not HTTP/2 or HTTP/3),
// SSE suffers from a severe limitation to the maximum number
// of open connections of 6 per browser. See:
// https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#sect1
//
// Deprecated: This was and is just an example.
func NewDefaultServer(es graphql.ExecutableSchema) *Server {
	srv := New(es)

	})
	srv.AddTransport(transport.Options{})
	srv.AddTransport(transport.MultipartForm{})

	srv.SetQueryCache(lru.New[*ast.QueryDocument](1000))

	srv.Use(extension.Introspection{})
	srv.Use(extension.AutomaticPersistedQuery{
		Cache: lru.New[string](100),
	})

	return srv
}
```

Issue #2681, opened 2023-06-19
Large type system does not compile. executionContext too large


magiconair:
### What happened?

With very large type systems gqlgen can generate code that can no longer by compiled by the go compiler. Once more than 65.000 field marshaling methods have been added to the `executionContext` the Go compiler fails with:

```
<autogenerated>:1: internal compiler error: too many methods on *executionContext: <some number>
```

In our case we hit that limit at 78.000 methods. We realize that running such a large type system is probably not efficient and also notice that the gqlgen compiler hasn't been optimized for this. It takes quite some time to generate the code because the generation is single threaded. While we are working on refactoring our approach we are blocked by this problem. 

There is a simple fix since the majority of the methods on `executionContext` are field marshaling functions which are all private. Their number grows with the number of types **and** number of fields per type. 

They can safely be converted to regular functions where the `*executionContext` is passed as an argument. Their names are unique and most likely won't collide with any existing function. And if that would be the case the internal function can be renamed. 

In our case the number of methods went down from 78.000 to just 4.200. 

We are currently testing this with v0.17.4 and will bump to the latest version soon. 

The patch is rather small and can be found here: https://github.com/northvolt/gqlgen/commit/78b17868d0415cfe1c600d154d6f17ca2c8fb311

I will create a PR once we've moved to the latest released version.

PatrLind:
Perhaps a related question: I'm having problems with gopls being extremely slow and taking a long time to show code completions, when I'm writing code in my generated graphql package. Also the compile times are quite long. And I only have about 4000 methods in my executionContext.
Will this refactor help with this issue too you think?

magiconair:
I don't know. The easiest is to try it. This patch should work up to `v0.17.10` without change. I've rebased it internally to `v0.17.33` but we have to make some changes first. 

The change itself is very simple. You can clone this repo and use go workspace files to use it.

```sh
go work init
go use .
go use ../gqlgen # your cloned copy of the gqlgen repo
```

PatrLind:
Thanks, unfortunately I couldn't test it since there were some errors with compile errors with subscriptions.
` ec._Subscription_itemUpdates undefined (type *executionContext has no field or method _Subscription_itemUpdates)`
Perhaps the POC doesn't support subscriptions yet.

ericmeter:
[...]

PR #3407, opened 2024-12-05
added support for use_function_syntax_for_execution_context 

kanodia-parag:
**What happened?**
With very large type systems gqlgen can generate code that can no longer by compiled by the go compiler. Once more than 65.000 methods have been added to the executionContext the Go compiler fails with:

`<autogenerated>:1: internal compiler error: too many methods on *executionContext: <some number> 
`
**Proposed Solution**
Generate regular functions instead of receiver methods of `*executionContext` and pass `*executionContext` as function argument controlled by config - `use_function_syntax_for_execution_context`

Addressing this issue - https://github.com/99designs/gqlgen/issues/2681


StevenACoffman:
@kanodia-parag  Would you mind updating this a bit? There was some other PRs that landed recently that require some tweaks to this PR.

We deprecated the NewDefaultServer as it gives new people a false sense that they can just go to production with that config, and they instead really need to change things, especially if they want to use websockets or HTTP SSE. You can see how to update your tests here: https://github.com/99designs/gqlgen/pull/3404/commits/ca4954c3ecea0d5525721fe6b8cb10c7da11ade4

Also, there's a conflicting fix that landed for the annoying test coverage reporting situation by excluding generated files more aggressively.

kanodia-parag:
@StevenACoffman I have updated the PR by rebasing it and fixing the linting issues coming from the test files.

StevenACoffman:
Thanks for contributing this feature, especially with the tests! 

Organizations with smaller scales may work around their pain points, while these become blockers like this at the scale of Uber. I hope you and your organization can contribute awareness (and possibly solutions) for problems that would otherwise go unreported.

For instance, it may be more obvious to your organization where we can best improve our performance to the benefit of all.

Please let us know if you encounter any other pain points, as it's extremely valuable! Looking forward to your next PR!

PR #3411, opened 2024-12-08
[SECURITY] Implement a feature to disable the suggestion when a GraphQL query fails

tomoikey:
# Description
Hello! I recently contributed to gqlparser to disable the suggestion feature for security reasons. Since gqlgen also uses the updated version of gqlparser, I have made modifications to allow gqlgen users to disable the suggestion feature as well.

The suggestion feature can be convenient from the client's perspective, but it may pose security risks. Therefore, developers using gqlgen should have the option to enable or disable this feature as needed.

Thanks.

### The relevant pull request
https://github.com/vektah/gqlparser/pull/319

# Changes Made
* Added a configuration option to enable or disable the suggestion feature in gqlgen.

I have:
 - [ ] ~~Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))~~

### DIFF
--- a/graphql/executor/executor.go
+++ b/graphql/executor/executor.go
@@ -7,6 +7,7 @@ import (
 	"github.com/vektah/gqlparser/v2/gqlerror"
 	"github.com/vektah/gqlparser/v2/parser"
 	"github.com/vektah/gqlparser/v2/validator"
+	"github.com/vektah/gqlparser/v2/validator/rules"
 
 	"github.com/99designs/gqlgen/graphql"
 	"github.com/99designs/gqlgen/graphql/errcode"
@@ -24,7 +25,8 @@ type Executor struct {
 	recoverFunc    graphql.RecoverFunc
 	queryCache     graphql.Cache[*ast.QueryDocument]
 
-	parserTokenLimit int
+	parserTokenLimit  int
+	disableSuggestion bool
 }
 
 var _ graphql.GraphExecutor = &Executor{}
@@ -177,6 +179,10 @@ func (e *Executor) SetParserTokenLimit(limit int) {
 	e.parserTokenLimit = limit
 }
 
+func (e *Executor) SetDisableSuggestion(value bool) {
+	e.disableSuggestion = value
+}
+
 // parseQuery decodes the incoming query and validates it, pulling from cache if present.
 //
 // NOTE: This should NOT look at variables, they will change per request. It should only parse and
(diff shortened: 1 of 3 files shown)


graphql/executor/executor_test.go
```diff
@@ -169,6 +169,22 @@ func TestExecutor(t *testing.T) {
...
+		assert.Equal(t, "input:1: Cannot query field \"nam\" on type \"Query\". Did you mean \"name\"?\n", resp.Errors.Error())
+	})
+
+	t.Run("disable suggestion, the error message will not include suggestions", func(t *testing.T) {
+		exec.SetDisableSuggestion(true)
+		resp := query(exec, "", "{nam}")
+		assert.Equal(t, "", string(resp.Data))
+		assert.Equal(t, "input:1: Cannot query field \"nam\" on type \"Query\".\n", resp.Errors.Error())
+	})
+}
```
tomoikey, line 186: Since the settings of the Executor were being modified, we decided to use a different exec. This ensures that it does not affect anyone writing tests in the future

graphql/executor/executor.go
```diff
@@ -24,7 +25,8 @@ type Executor struct {
 	recoverFunc    graphql.RecoverFunc
 	queryCache     graphql.Cache[*ast.QueryDocument]
 
-	parserTokenLimit int
+	parserTokenLimit  int
+	disableSuggestion bool
```
tomoikey, line 29: Unless `SetDisableSuggestion(true)` is called, the behavior will remain exactly the same as before, so it does not affect backward compatibility. Only those who are aware of this feature will be able to turn off the suggestions.

StevenACoffman:
Thanks for this (and the gqlparser contribution)!

Just noticed you are in Yokohama, I miss the food there, so be sure to appreciate it for me! I'm always wistful when my Ramen here doesn't come with wood-ear mushrooms like in sanmamen.

tomoikey:
@StevenACoffman 
I love ramen too, so if you ever come to Yokohama, please let me show you around! 😎 Thanks again for all your help with gqlparser!

jedevc:
@StevenACoffman @tomoikey I *think* this might have accidentally been a breaking change. Maybe I'm wrong though!

Previously, `gqlgen` didn't import `github.com/vektah/gqlparser/v2/validator/rules`, so the rules that were added on `init` were never actually added (unless imported separately by the user). I think this means that now, anytime that `gqlgen` is imported, *all* validation is enabled, without an easy way to turn it off?

(obviously, yes, users like me should just make validation pass on schemas, but I'm not entirely sure if this change was intentional?)

PR #3428, opened 2024-12-10
Restore multiline resolver doc comments after #3243

StevenACoffman:


StevenACoffman:
@csilvers This should fix the issue.

plugin/resolvergen/resolver.go
```diff
@@ -159,6 +159,7 @@ func (m *Plugin) generatePerSchema(data *codegen.Data) error {
 				continue
 			}
 			structName := templates.LcFirst(o.Name) + templates.UcFirst(data.Config.Resolver.Type)
+			// TODO(steve): Why do we need to trimLeft "\" here? Some bazel thing?
```
StevenACoffman, line 162: @hailyngx @clayne11  Do you remember why this trimLeft is here?

Issue #2526, opened 2023-01-25
Why is GenerateCode called twice?


vikstrous2:
Reading the code in api/generate.go, I see `codegen.GenerateCode(data)` called twice. Once before plugins run and once after. Why is this? It seems very wasteful. I see that there was an attempt to remove it in #1079 but that had to be reverted. I see that some things don't work correctly if it's removed. Is anything more known about why? What exactly needs to be done before vs after plugins run?

vikstrous2:
@vektah can you provide some info about why #1079 was reverted? Do you see any way to avoid calling GenerateCode twice?

taako-502:
@vikstrous2 
https://github.com/99designs/gqlgen/pull/1100

Issue #3414, opened 2024-12-09
Inconsistencies in generated files


evellior:
### What happened?
Each time gqlgen runs it almost randomly picks whether it will use `interface{}` or `any` in some generated code.  I know they function the same, but is there a way to make the generated output more consistent?  

I don't think my use case is specifically important, I assume having a consistent output when generating code would be good regardless.  But the use case I have and how I'm reproducing the issue is with an automated test intended to catch when people change `.graphql` schema files and forget to regenerate the application code.  I have the following test workflow:
```
name: GQLGen Test

on:
  workflow_call: # Allows this to be executed by other workflows
  workflow_dispatch: # Allows this to be manually executed on any branch
  pull_request:
    branches: [ development ]

jobs:
  test:
    name: Generate graph
    runs-on: ubuntu-latest
    steps:

    - uses: actions/checkout@v4

    - name: Set up Go
      uses: actions/setup-go@v4
      with:
        go-version: 1.23.1

    - name: Run test gqlgen
      run: |
        go get -d github.com/99designs/gqlgen@v0.17.53
        cd app/api/graphql; go run github.com/99designs/gqlgen

    - name: Check file diff
      run: |
        if [ -n "$(git status --porcelain)" ]; then 
          echo "::error::The generated graphql resolvers do not match with those in the branch!"
          git status -v -v
          exit 1
        fi
```

The test failed as the generated files had the following diff:
```
Error: The generated graphql resolvers do not match with those in the branch!
On branch development
Your branch is up to date with 'origin/development'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   app/api/graphql/generated/graph/schema_gen.go

--------------------------------------------------
Changes not staged for commit:
diff --git i/app/api/graphql/generated/graph/schema_gen.go w/app/api/graphql/generated/graph/schema_gen.go
index f9ceeff..50af593 100644
--- i/app/api/graphql/generated/graph/schema_gen.go
+++ w/app/api/graphql/generated/graph/schema_gen.go
@@ -[23](https://github.com/orijinplus/backend/actions/runs/12228380133/job/34106552653#step:5:24)6836,12 +236836,12 @@ func (ec *executionContext) marshalNAnalyticsResult2ᚖorijinplusᚋappᚋapiᚋ
 	return ec._AnalyticsResult(ctx, sel, v)
 }
 
[...]

Issue #2308, opened 2022-07-29
Remove format at end of generation


clayne11:
The [`internal/imports.Prune`](https://github.com/99designs/gqlgen/blob/master/internal/imports/prune.go#L30-L46) method parses the entire generated file, removes unused imports, then reformats it. This can be very expensive in large files (we have some files that are 680k lines / 40MB+).

We should be able to determine the required imports as we build the file and only include the correct imports and avoid this entirely.

PR #3436, opened 2024-12-14
fix: only run generate code once

Brookke:
in issue #2526 it's noted that `codegen.GenerateCode` is ran twice theres a vague PR that reverted this duplication #1100, but I am not clear why.

In my testing using a particularly heavy schema file I found the following

```
# Build of the master branch
/Users/brooke/Developer/gqlgen/gqlgen  34.45s user 4.15s system 145% cpu 26.486 total

# Build of this PR
/Users/brooke/Developer/gqlgen/gqlgen  26.82s user 3.92s system 151% cpu 20.223 total
```

I have:

### DIFF
--- a/api/generate.go
+++ b/api/generate.go
@@ -136,10 +136,6 @@ func Generate(cfg *config.Config, option ...Option) error {
 		}
 	}
 
-	if err = codegen.GenerateCode(data); err != nil {
-		return fmt.Errorf("generating core failed: %w", err)
-	}
-
 	if !cfg.SkipValidation {
 		if err := validate(cfg); err != nil {
 			return fmt.Errorf("validation failed: %w", err)


Brookke:
Converted to draft while I investigate the CI errors

Brookke:
Hi @StevenACoffman I may need some pointers on this, the diff is all in the federated example. It's added `obj fedruntime.Entity` to a bunch of entity resolvers. Has this highlighted a previously unknown bug where the second run was overwriting things, or was the old behaviour correct?

Brookke:
> Hi @StevenACoffman I may need some pointers on this, the diff is all in the federated example. It's added `obj fedruntime.Entity` to a bunch of entity resolvers. Has this highlighted a previously unknown bug where the second run was overwriting things, or was the old behaviour correct?

I believe the issue was due to the order of running things. The plugins modify the data that gets passed to the main generateCode function. So by reordering things, the codegen now matches the examples

StevenACoffman:
Hey, sorry I've been slow to respond to this. I really appreciate you taking this on! This looks perfectly sound to me, but I would like to get some more people to try this out in case there's a reason I'm not seeing why something would need to run both before and after (more like a middleware). After all Hyrum's Law may apply:

> With a sufficient number of users of an API, it does not matter what you promise in the contract: all observable behaviors of your system will be depended on by somebody.

&mdash; [Hyrum Wright, 2012](https://www.hyrumslaw.com/)

+ @golanglemonade - Can you check if this PR [breaks any of your plugins](https://github.com/theopenlane/gqlgen-plugins) - which I just found by the way!
+ @vikstrous - Can you verify that this PR doesn't break [dataloadgen](https://github.com/vikstrous/dataloadgen)?
+ @Warashi - Can you verify that this PR doesn't break [compgen](https://github.com/Warashi/compgen)?
+ @ravilushqa - Can you verify that this PR doesn't break [otelgqlgen](https://github.com/ravilushqa/otelgqlgen)?

Brookke:
Thanks @StevenACoffman 🙏

vikstrous:
This won't cause an issue for dataloadgen. dataloadgen exists only at run time, not generation time. It's not a plugin

I got some data points for our biggest graph, though, to show that it helps us a lot too:

Version | Run 1 | Run 2
-- | -- | --
gqlgen v0.17.60 | 46.81 | 48.23
this PR | 35.52 | 30.46

StevenACoffman:
Ok, I'm going to merge this, but I'm going to hold off on a new release for a bit in case there's some late feedback and we need to alter things. BTW, this might help with #3414 

Unfortunately, very few people pay attention to changes until after a release and they try to put things into their own prod environments.

golanglemonade:
@StevenACoffman Look good on https://github.com/theopenlane/gqlgen-plugins against our main schema. Appreciate you checking!

Brookke:
@vikstrous is that schema available publicly? I'm looking into if there are any other obvious improvements and would like more schemas to test against. 

So far I've found that [this prune function](https://github.com/99designs/gqlgen/blob/3d0c92aca2af716903a726c777339e273e246020/internal/imports/prune.go#L29) takes a significant amount of time on our schema.

vikstrous:
No, sorry, it's a private codebase. I've noticed that autobind is one source of slowness because of the size of the repo. There are a lot of dependencies, so loading the packages can take time

StevenACoffman:
You might look at #3372 

@Brookke Sidenote: I'm not sure if it's helpful, but I shared with you a private repo with just Khan Academy's gqlgen configs, service schemas, and federation composed supergraph. We're a non-profit, so we don't have trade secrets, and we only accept the queries from our safelist, so I'm not super concerned with sharing. However, it would be better if it wasn't passed around too much, since that kind of thing invites attackers. 

The really huge gqlgen schemas I'm aware of like ones the reddit folks and uber folks use are not public. 

Honestly, one of the things that gqlgen is really missing is a set of realistic benchmarks and profiling to get a sense of where are the performance bottlenecks (during codegen but also in runtime), so this would be hugely valuable work. Performance optimizations can make the code harder to maintain, so we only want to do it where it makes a big difference, and right now we don't know where that would be.

Brookke:
Thanks for sharing @StevenACoffman. My initial findings centre around the Prune function, basically it calls `imports.Process` on generated.go which can get pretty large. Taking out the call to Prune significantly increases the speed. 

This suggestion might help https://github.com/99designs/gqlgen/issues/2308

Alternatively, I was wondering if a refactor could move the call to `imports.Process` to the end of the codegen and then run it in parallel across all the files it gets called on. Would potentially be really useful in conjunction with `layout: follow-schema`

StevenACoffman:
A PR for either/both would be very welcome!

If there's a breaking change necessary in the name of progress, we should put a config option and let people opt-in (or at least opt-out).

For instance, if the generated files were not formatted, people would be annoyed as their linters would complain (even though they can exclude generated files, or add a pre-lint format step after generation), even though it would be much faster.

StevenACoffman:
If you see any other userful performance improvement suggestions in issues and want to call them out to me here, I can tag them as performance for others.

I prioritize paying attention to PRs with my volunteer free time, but I often have stretches of time where I have to ignore issues (like camping) and then can never get enough time to circle back to look at them.

PR #3437, opened 2024-12-14
more actionable federation errors for nil key field queries

StevenACoffman:
In #3029 @clayne11 altered the federation behavior, so that null values for key Fields would now also skip using the resolver (since it would never work). This was both a security and performance enhancement, as expensive safelisted queries with nil KeyField values could have been sent for no reason.

However, the error messages for this change were not very specific and as a result, were not very clearly actionable.
I **think** I correctly understand the cases and have added more specific error messages, but I would definitely appreciate confirmation from @clayne11 or @kanodia-parag or anyone else that I'm not subtly misleading people.

I am collecting entity resolution errors as it is valid for entities with multiple key fields to fail with some resolvers and then succeed with a different resolver.

### DIFF
--- a/plugin/federation/federation.gotpl
+++ b/plugin/federation/federation.gotpl
@@ -294,6 +294,9 @@ func (ec *executionContext) resolveManyEntities(
 	{{- if .Resolvers }}
 
 		func entityResolverNameFor{{$entity.Name}}(ctx context.Context, rep EntityRepresentation) (string, error) {
+			// we collect errors because a later entity resolver may work fine
+			// when an entity has multiple keys
+			entityResolverErrs := []error{}
 			{{- range .Resolvers }}
 				for {
 					var (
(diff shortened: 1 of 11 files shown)


plugin/federation/federation.gotpl
```diff
@@ -310,11 +310,12 @@ func (ec *executionContext) resolveManyEntities(
...
 							val, ok = m["{{.}}"]
 							if !ok {
-								break
+								return "", fmt.Errorf("%w due to missing Key Field {{.}} for User", ErrTypeNotFound)
 							}
 							{{- if (ne $i $keyField.Field.LastIndex ) }}
 								if m, ok = val.(map[string]interface{}); !ok {
-									break
+									// nested field value is not a map[string]interface
+									return "", fmt.Errorf("%w for {{$entity.Name}} due to nested Keyfield not being map value", ErrTypeNotFound)
```
StevenACoffman, line 318: @clayne11 This is the one I am least confident if it is a good error message

plugin/federation/federation.gotpl
```diff
@@ -310,10 +313,15 @@ func (ec *executionContext) resolveManyEntities(
 						{{- range $i, $field := .Field }}
 							val, ok = m["{{.}}"]
 							if !ok {
+								entityResolverErrs = append(entityResolverErrs,
+									fmt.Errorf("%w due to missing Key Field \"{{.}}\" for {{$entity.Name}}", ErrTypeNotFound))
 								break
 							}
 							{{- if (ne $i $keyField.Field.LastIndex ) }}
```
StevenACoffman, line 320: I am not very sure I understand what this code is doing here.
We only want to possibly skip suggesting this resolver if the current field index does not match the last index of the key field?

PR #3438, opened 2024-12-15
enable testifylint.encoded-compare and fix lint issues (#3434)

StevenACoffman:


StevenACoffman:
@Brookke hey, after I rebased this with your PR, I'm getting this modelgen error. May be unrelated to your change, but if you get a chance to take a look, I would appreciate it.

StevenACoffman:
So I am a doofus. :) `any` is just a type alias for `interface{}`, so when you are making a conditional based on the type, you still need to look for `interface{}` rather than `any`.  This was a self-inflicted brain injury on my part.

PR #3456, opened 2024-12-25
[BUG] fix disable suggestion bug

tomoikey:
@StevenACoffman
First, please allow me to offer my apologies. I have realized that a minor bug exists in the [Pull Request I previously submitted](https://github.com/99designs/gqlgen/pull/3411).

The issue is that when disableSuggestion is set to true, and consecutive queries are made, the resulting error messages are duplicated.

Below, I’ve included a screenshot that demonstrates the bug, along with steps to reproduce it. Please feel free to try it out in your local playground environment:
1. Set disableSuggestion to true.
2. Execute a query that would trigger suggestions.
3. Observe that the errors are outputted multiple times, creating duplicates.

# Previous Pull Request
https://github.com/99designs/gqlgen/pull/3411

# Screenshot
image

I have:
 - [ ] ~~Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))~~

### DIFF
--- a/graphql/executor/executor.go
+++ b/graphql/executor/executor.go
@@ -227,6 +227,8 @@ func (e *Executor) parseQuery(
 		validator.RemoveRule("FieldsOnCorrectType")
 
 		rule := rules.FieldsOnCorrectTypeRuleWithoutSuggestions
+		// remove the rule added when it was last executed
+		validator.RemoveRule(rule.Name)
 		validator.AddRule(rule.Name, rule.RuleFunc)
 	}
 
--- a/graphql/executor/executor_test.go
+++ b/graphql/executor/executor_test.go
@@ -181,7 +181,14 @@ func TestExecutorDisableSuggestion(t *testing.T) {
 		exec.SetDisableSuggestion(true)
 		resp := query(exec, "", "{nam}")
 		assert.Equal(t, "", string(resp.Data))
+		assert.Equal(t, len(resp.Errors), 1)
 		assert.Equal(t, "input:1: Cannot query field \"nam\" on type \"Query\".\n", resp.Errors.Error())
+
+		// check if the error message is displayed correctly even if an error occurs multiple times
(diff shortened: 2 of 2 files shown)


graphql/executor/executor.go
```diff
@@ -227,6 +227,8 @@ func (e *Executor) parseQuery(
 		validator.RemoveRule("FieldsOnCorrectType")
 
 		rule := rules.FieldsOnCorrectTypeRuleWithoutSuggestions
+		// remove the rule added when it was last executed
+		validator.RemoveRule(rule.Name)
```
tomoikey, line 231: If gqlparser has a function like Contains, it might be better to use it and add a conditional `if` check to determine whether the rule is already duplicated.

However, in terms of computational cost, the current implementation should not cause any issues. Therefore, I think it would be sufficient to leave this as a future TODO for now.

StevenACoffman:
I added `ReplaceRule` as an alternative to `AddRule` since you are not the first to request it.
https://github.com/vektah/gqlparser/pull/338

StevenACoffman:
Ok, I cut a new release for gqlparser [v2.5.21](https://github.com/vektah/gqlparser/releases/tag/v2.5.21)

tomoikey:
@StevenACoffman 
Thank you for your prompt response! I apologize for the inconvenience I caused you...

StevenACoffman:
No worries. 
Honestly, the ruleset being in a global variable is poor implementation for gqlparser that leads to problems.
Having the `Init()` setting the RuleSet in gqlgen makes the situation worse, because it is hard to override later.

The current implementation could still be further improved by altering the ruleset once when the config option is first read, rather than doing it repeatedly in the executor.

I updated gqlparser and made the minor switch to use ReplaceRule
https://github.com/99designs/gqlgen/pull/3458

PR #3463, opened 2024-12-30
codegen: move templates comments from generated code

a8m:
Template comments should not be included in the generated assets (it repeats itself for every arg).

cc @giautm

### DIFF
--- a/_examples/chat/generated.go
+++ b/_examples/chat/generated.go
@@ -354,11 +354,7 @@ func (ec *executionContext) dir_user_argsUsername(
 	ctx context.Context,
 	rawArgs map[string]any,
 ) (string, error) {
-	// We won't call the directive if the argument is null.
-	// Set call_argument_directives_with_null to true to call directives
-	// even if the argument is null.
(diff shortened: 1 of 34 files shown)


a8m:
Also, something a bit worrying is that the latest upgrade significantly increases the size of the generated code. See this PR as an example (from 15K to 21K): https://github.com/ent/contrib/pull/607/files
image

StevenACoffman:
Ah, oops. Yeah, that's a good call. Not sure about the codegen size increase source (just more comments?).

a8m:
> Not sure about the codegen size increase source (just more comments?).

See the PR example I shared above.

StevenACoffman:
Sorry, I was unclear in my comment.
In https://github.com/ent/contrib/pull/607/files github.com/99designs/gqlgen was moved from v0.17.48 to v0.17.51

In that PR, I see that `entgql/internal/todo/generated.go` went from 15K to 21K lines, but it was not clear to me if the comments fully accounted for the difference.

PR #3466, opened 2025-01-02
Add `KeepAlivePingInterval` to SSE transport

jlaprise:
## Proposal
Add optional `KeepAlivePingInterval` parameter to the SSE transport.

```go
srv.AddTransport(transport.SSE{
})
```

SSE (Server-Sent Events) connections are designed to keep a connection open between the client and server to send real-time updates from the server to the client. While SSE connections are intended to be long-lived and persistent, they can be affected by idle timeout settings if there are long periods of inactivity.

### Idle Timeout :
Load balancers, proxies, or firewalls often have idle timeout settings that specify the maximum duration a connection can remain open without data being sent across it. If the idle timeout is exceeded without any data being transmitted, the connection may be closed.

### SSE Behavior :
In a properly functioning SSE setup, the server is typically expected to send periodic heartbeat messages or comments (using a colon : in SSE) to ensure that the connection stays alive and doesn't hit the idle timeout. This ensures that some traffic occurs to reset the idle timeout counter.

I have:


StevenACoffman:
@UnAfraid @oxisto @o8x @jmic @giulio-opal Do you have information or thoughts on this?

I'm unclear if Keep-Alive Ping should differ depending on HTTP/2 SSE vs HTTP/1.1 SSE? I don't know if Go's http2 automatically persists streaming connections within the data frames, or if we still need to do some sort of Ping.

I see a note that the Keep-Alive header is deprecated for HTTP2: 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Keep-Alive

oxisto:
> @UnAfraid @oxisto @o8x @jmic @giulio-opal Do you have information or thoughts on this?
> 
> I'm unclear if Keep-Alive Ping should differ depending on HTTP/2 SSE vs HTTP/1.1 SSE? I don't know if Go's http2 automatically persists streaming connections within the data frames, or if we still need to do some sort of Ping.
> 
> I see a note that the Keep-Alive header is deprecated for HTTP2: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Keep-Alive
> 
> AFAICT, this Keep-Alive Ping is a good idea for HTTP/1.1, but prohibited for HTTP/2, so this should at least have a comment to that effect (and pass the existing tests).
> 
> I also [read here](https://github.com/r3labs/sse/issues/101#issuecomment-962464451):
> 
> > In HTTP/2, the concept of persistent connections is built into the protocol itself.
> > HTTP/2 allows multiple streams of data to be multiplexed over a single connection, inherently supporting what "Keep-Alive" aims to achieve in HTTP/1.1.
> > Therefore, connection-specific headers like "Connection" and "Keep-Alive" are not only unnecessary in HTTP/2 but are also prohibited.
> > Browsers like Chrome and Firefox ignore these headers in HTTP/2 responses, and Safari may not load responses that contain them.
> > Given this, you do not need to turn off "Keep-Alive" in your Apache configuration when using HTTP/2, as it will be ignored by the protocol.

If I read the issues surrounding http2 correctly, there is now a `PingTimeout` in `http2.Server` that can be configured to send such pings: https://pkg.go.dev/golang.org/x/net/http2#Server. See also https://github.com/golang/go/issues/67812. This should be in x/net v0.33.0 and this will eventually end up in std. Not sure if this is already targeted for Go 1.24.

jlaprise:
Thanks for looking into this request. Since submitting this PR we found a solution, and a bit more info that hopefully will help anyone else having a similar issue.

SSE over an end-to-end HTTP/2 connection doesn't require a ping to keep the connection open.

The issue we had is related to SSE over HTTP/1. We didn't have any issues in our dev/test environments but once deployed to prod (AWS) we saw SSE connections get disconnected after 1 minute due to the default [ALB](https://aws.amazon.com/elasticloadbalancing/application-load-balancer/) idle connection timeout. We had ALB configured for HTTP/2 but the routed target group was configured for HTTP/1. ALB will accept HTTP 1 or 2 and translate the request to whichever protocol the load balancer's target group is configured to accept. 

We tried changing the target group protocol to HTTP/2, but then clients that sent HTTP/1 requests received response code [464 Incompatible protocol](https://http.dev/464).

The solution we ended at was to create two separate target groups for HTTP/1 and HTTP/2 pointing at the same backend, then configure a rule on the load balancer to forward to the target group matching the request's HTTP version.

Sending a ping like this PR suggested does keep the connection open when there's a network hop only compatible with HTTP/1. However, I'm unsure if there are caveats to doing that.

StevenACoffman:
Ok, this sounds like it's beneficial for HTTP/1.1 environments, and people can just opt-out if they don't need it. Can you fix the test to pass (and it would be nice if you rebased but not required)?

I appreciate you contributing it, as well as circling back to provide more context.

StevenACoffman:
Thanks for this!

PR #3468, opened 2025-01-03
Concurrent operation scheduler

jarreds:
In the spirit of optimized batching, I've been playing around with a coroutine implementation in Golang that allows cooperative multitasking to batch IO. I can share more details on that as the experiment progresses.

This is the code I used to shim in a new "scheduler" for gqlgen. Creating as draft for now until it proves out that there is actually reason to have this code.

### DIFF
--- a/_examples/chat/generated.go
+++ b/_examples/chat/generated.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 	"io"
 	"strconv"
-	"sync"
 	"sync/atomic"
 	"time"
 
@@ -291,7 +290,7 @@ func (ec *executionContext) processDeferredGroup(dg graphql.DeferredGroup) {
 	atomic.AddInt32(&ec.pendingDeferred, 1)
 	go func() {
 		ctx := graphql.WithFreshResponseContext(dg.Context)
-		dg.FieldSet.Dispatch(ctx)
+		dg.FieldSet.Dispatch(ctx, ec.OperationContext)
 		ds := graphql.DeferredResult{
 			Path:   dg.Path,
 			Label:  dg.Label,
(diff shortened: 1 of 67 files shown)


StevenACoffman:
Interesting! I'm very curious to learn what you find out!

However, I would suggest that in the Dispatch you _really_ need the scheduler bypass when you only have a single task. Having this case where it doesn't go through a scheduler or anything complicated is an important performance feature for a surprising number of use cases. Every time we've removed it, there's been a storm of complaints.

```
func (m *FieldSet) Dispatch(ctx context.Context) {
	if len(m.delayed) == 1 {
		// only one concurrent task, no need to spawn a goroutine or deal create waitgroups
```

jarreds:
Noted! FWIW I did put a special case in the default goroutine scheduler to run len one tasks (w/o wait groups) and the last `f` "on-thread:" https://github.com/99designs/gqlgen/pull/3468/files#diff-3a919dfc5028540a932a129e10b6e9cb749a1fd70be686dc7471ff440d823086R37-R40

jarreds:
First draft of libs to support IO batching and coroutines are here:

- https://github.com/webriots/corio
- https://github.com/webriots/coro

Note: they're still quite rough and I have more work to do on them.

With that said, I've confirmed gqlgen works with a GraphQL scheduler built on these primitives and it batches IO without the need for a wait-based data loader. I've been testing in a quite complicated GraphQL codebase that has a high degree of resolver concurrency without issue.

Additionally, because these libs utilize runtime supported coroutines, concurrent GraphQL resolvers bypass the go runtime scheduler. They use true cooperative multitasking instead.

Quick question: do you have a benchmark to measure performance that gqlgen developers typically use?

PR #3470, opened 2025-01-06
Adding additional documentation for int-based enums

oxisto:
Int-based enums need an additional marshal/unmarshal functionality in order to work with GraphQL client libraries.

### DIFF
--- a/_examples/enum/model/model.go
+++ b/_examples/enum/model/model.go
@@ -1,5 +1,10 @@
 package model
 
+import (
+	"encoding/json"
+	"fmt"
+)
+
 type IntTyped int
 
 const (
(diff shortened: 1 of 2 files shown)


_examples/enum/model/model.go
```diff
@@ -1,5 +1,10 @@
 package model
 
+import (
+	"encoding/json"
+	"fmt"
+)
+
```
StevenACoffman, line 7:
```suggestion

// this file is provided as an example for int-based enums
// but if you instead wanted to support arbitrary 
// english words for numbers to integers, consider
// github.com/will-lol/numberconverter/etoi or a similar library
```

_examples/enum/model/model.go
```diff
@@ -1,5 +1,14 @@
...
+// this file is provided as an example for int-based enums
+// but if you instead wanted to support arbitrary
```
StevenACoffman, line 9:
```suggestion
// but if you instead wanted to support arbitrary
```

Issue #3403, opened 2024-12-03
Error Info Missing From FTV1 Federated Trace Data


justindoherty:
### What happened?

We use Apollo GraphOS and have a federated supergraph with a number of subgraphs implemented with gqlgen. We are expecting to see some error details in GraphOS and after conferring with the Apollo team it seems like that information isn't being provided in the `ftv1` extension data which is needed for display in Apollo GraphOS.

From the [Apollo documentation](https://www.apollographql.com/docs/graphos/routing/graphos-reporting#errors) it states:

> Your subgraph libraries must support federated tracing (also known as FTV1 tracing) to provide errors to GraphOS.

Example from the Apollo guys, not related to the sample data below:
![image](https://github.com/user-attachments/assets/bdd58999-6ea1-445a-9552-9773043f0b23)

### What did you expect?

We'd expect that any errors reported as GraphQL errors during resolver execution would also make their way in to the `ftv1` extension data.

See the example data below. The `error` arrays are all empty.

### Minimal graphql.schema and models to reproduce

The subgraph response from the gqlgen subraph:

```json
{
  "timestamp": "2024-12-02T10:24:02.809833046Z",
  "level": "INFO",
  "trace_id": "...",
  "span_id": "...",
  "message": "Raw response body from subgraph \"asset-search\" received",
  "response.body": {
    "errors": [
      {
        "message": "",
        "path": ["assetSearch"],
        "extensions": {
          "code": 401
        }
      }
    ],
    "data": null,
    "extensions": {
      "ftv1": "GgwIwpW2ugYQ79TCgQMiDAjClba6BhDp0fXWAliGg80qcjBiLhoQQXNzZXRDb25uZWN0aW9uIUDjiAdI3owHagVRdWVyeQoLYXNzZXRTZWFyY2g="
    }
  },
  "apollo.subgraph.name": "asset-search",
  "target": "apollo_router::services::subgraph_service",
  "spans": [
    {
      "http.route": "/graphql",
      "http.request.method": "POST",
      "http.request.body.size": "131",
      "network.protocol.version": "HTTP/1.1",
      "network.type": "ipv4",
      "client.address": "...",
      "server.address": "...",
      "server.port": "4000",
      "url.path": "/graphql",
      "user_agent.original": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36",
[...]

PR #2132, opened 2022-04-27
Feature: Add FTV1 Support via Handler

lleadbet:
This adds the ability to properly respond to Apollo Federation v1 (aka `ftv1`) headers with traces to support field-level traces within Apollo Studio. This feature is entirely optional, and would be opted in as with the existing Apollo Tracing handler. 

Sample: https://github.com/lleadbet/apollo-gqlgen-ftv1/blob/master/server.go#L24

To support this, there are two major changes: 

- Added dependency on Protobuf, as Apollo uses it for marshaling the format
- Updates the `RawParams` to support passing request headers to Handlers

Beyond that, it adds the plugin itself alongside the source of the protobuff file to be able to use as reference. 

Not sure where docs should be updated, however. 

I have:
 - [x] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

PR #3506, opened 2025-01-27
feat: add errors to ftv1 traces

lleadbet:
This just adds the missing functionality that was missing from my initial `ftv1` implementation ( see #2132 ) for the `apollofederatedtracingv1` package. Addresses #3403.

This further adds the ability to define a transform function for the users that want to control what is sent in said traces; example:

```go
	srv.Use(&apollofederatedtracingv1.Tracer{
		ErrorOptions: &apollofederatedtracingv1.ErrorOptions{
			ErrorOption: apollofederatedtracingv1.ERROR_TRANSFORM,
			TransformFunction: func(err gqlerror.Error) gqlerror.Error {
				return gqlerror.Error{
					Message: "transformed",
				}
			},
		},
	})
```

The other valid options for `ErrorOption` are: 

* `ERROR_UNMODIFIED` - sends an unmodified error in the trace, which is potentially dangerous for sensitive information
* `ERROR_MASKED` - replaces the message in the error with `<masked>`; this is the default behavior

This matches with the [Apollo Server implementation](https://github.com/apollographql/apollo-server/blob/88a8db546bac6e18b167e55760a061daa9b119a1/packages/server/src/plugin/traceTreeBuilder.ts#L44-L47) to avoid sending sensitive information. 

If using `ERROR_TRANSFORM`, the `TransformFunction` is invoked and the user can customize the message/other data as needed, although modifying the message is the only expected thing in this case (although other data could be helpful). 

Visuals included in PR to help illustrate what it looks like in Apollo Studio. 
image
image
image

The `<redacted>` is coming from the Apollo Router, and is unimportant for this PR. `transformed` used the example above.  

I have:

### DIFF
--- a/graphql/handler/apollofederatedtracingv1/tracing.go
+++ b/graphql/handler/apollofederatedtracingv1/tracing.go
@@ -8,6 +8,13 @@ import (
 	"google.golang.org/protobuf/proto"
 
 	"github.com/99designs/gqlgen/graphql"
+	"github.com/vektah/gqlparser/v2/gqlerror"
+)
+
+const (
+	ERROR_MASKED     = "masked"
+	ERROR_UNMODIFIED = "all"
+	ERROR_TRANSFORM  = "transform"
 )
 
 type (
@@ -15,10 +22,20 @@ type (
 		ClientName string
 		Version    string
 		Hostname   string
+		Errors     *TraceErrors
 	}
 
 	treeBuilderKey string
 )
+
+type TraceErrors struct {
+	// ErrorOption is the option to handle errors in the trace, it can be one of the following:
+	// - "masked": masks all errors
+	// - "all": includes all errors
+	// - "transform": includes all errors but transforms them using TransformFunction, which can allow users to redact sensitive information
+	ErrorOption       string
+	TransformFunction func(gqlerror.Error) gqlerror.Error
+}
 
 const (
 	key = treeBuilderKey("treeBuilder")
@@ -62,7 +79,17 @@ func (t *Tracer) InterceptOperation(ctx context.Context, next graphql.OperationH
 	if !t.shouldTrace(ctx) {
 		return next(ctx)
 	}
-	return next(context.WithValue(ctx, key, NewTreeBuilder()))
+	if t.Errors == nil || t.Errors.ErrorOption == ERROR_MASKED {
+		t.Errors = &TraceErrors{
+			ErrorOption:       ERROR_MASKED,
+			TransformFunction: defaultErrorTransform,
+		}
+	} else if t.Errors.ErrorOption == ERROR_TRANSFORM && t.Errors.TransformFunction == nil {
+		t.Errors.TransformFunction = defaultErrorTransform
+	} else if t.Errors.ErrorOption == ERROR_UNMODIFIED {
+		t.Errors.TransformFunction = nil
+	}
+	return next(context.WithValue(ctx, key, NewTreeBuilder(*t.Errors)))
 }
 
 // InterceptField is called on each field's resolution, including information about the path and parent node.
@@ -96,8 +123,12 @@ func (t *Tracer) InterceptResponse(ctx context.Context, next graphql.ResponseHan
 
 	// now that fields have finished resolving, it stops the timer to calculate trace duration
 	defer func(val *string) {
-		tb.StopTimer(ctx)
+		errors := graphql.GetErrors(ctx)
+		if len(errors) > 0 {
+			tb.DidEncounterErrors(ctx, errors)
+		}
 
+		tb.StopTimer(ctx)
 		// marshal the protobuf ...
 		p, err := proto.Marshal(tb.Trace)
 		if err != nil {
@@ -109,4 +140,8 @@ func (t *Tracer) InterceptResponse(ctx context.Context, next graphql.ResponseHan
 	}(val)
 	resp := next(ctx)
 	return resp
+}
+
+func defaultErrorTransform(_ gqlerror.Error) gqlerror.Error {
+	return *gqlerror.Errorf("<masked>")
 }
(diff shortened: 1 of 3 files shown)


graphql/handler/apollofederatedtracingv1/tracing.go
```diff
@@ -8,18 +8,35 @@ import (
...
 	treeBuilderKey string
 )
 
+type TraceErrors struct {
+	// ErrorOption is the option to handle errors in the trace, it can be one of the following:
+	// - "masked": masks all errors
+	// - "all": includes all errors
+	// - "transform": includes all errors but transforms them using TransformFunction, which can allow users to redact sensitive information
+	ErrorOption       string
+	TransformFunction func(gqlerror.Error) gqlerror.Error
```
StevenACoffman, line 37: The `gqlerror.Error` struct you are passing around by values can be heavy, so I would prefer you use pointers where possible.
```suggestion
	TransformFunction func(*gqlerror.Error) *gqlerror.Error
```

graphql/handler/apollofederatedtracingv1/tree_builder.go
```diff
@@ -11,12 +12,14 @@ import (
...
 )
 
 type TreeBuilder struct {
-	Trace    *generated.Trace
-	rootNode generated.Trace_Node
-	nodes    map[string]NodeMap // nodes is used to store a pointer map using the node path (e.g. todo[0].id) to itself as well as it's parent
+	Trace        *generated.Trace
+	rootNode     generated.Trace_Node
+	nodes        map[string]NodeMap // nodes is used to store a pointer map using the node path (e.g. todo[0].id) to itself as well as it's parent
+	errorOptions TraceErrors
```
StevenACoffman, line 22:
```suggestion
	errorOptions *TraceErrors
```
It looks like this `TraceErrors` struct is passed by value, but when you modify, I think you always want to replace the subfields of the original version, so you should just change it to a pointer to avoid the possible bug.

graphql/handler/apollofederatedtracingv1/tracing.go
```diff
@@ -8,18 +8,35 @@ import (
...
 		ClientName string
 		Version    string
 		Hostname   string
+		Errors     TraceErrors
 	}
 
 	treeBuilderKey string
 )
 
+type TraceErrors struct {
```
StevenACoffman, line 31:
```suggestion
type TraceErrorOptions *struct {
```
`TraceErrors` implies that it contains one or more errors, but this is actually just the options. The word Trace might actually be redundant, so `ErrorOptions` would also be a good name.

graphql/handler/apollofederatedtracingv1/tracing.go
```diff
@@ -8,18 +8,35 @@ import (
...
+	ERROR_UNMODIFIED = "all"
+	ERROR_TRANSFORM  = "transform"
 )
 
 type (
 	Tracer struct {
 		ClientName string
 		Version    string
 		Hostname   string
+		Errors     TraceErrors
```
StevenACoffman, line 25:
```suggestion
		ErrorOptions     *TraceErrorOptions
```
This should be a pointer, and the name should not be easily confused with the actual errors.

lleadbet:
Great callouts @StevenACoffman - updated to pass around by reference rather than copying the value around to the various functions. Appreciate the fast feedback already.

StevenACoffman:
(btw, sorry I didn't batch my review comments)

The remaining `fmt.Println` statements are a problem for people who want to use their own logger (e.g. a structured logger like slog, zerolog, uber/zap, logrus, etc.).

I know these errors are sort of "meta" errors, errors from tracing the actual errors, and seem mostly unlikely to really occur.

I wonder if adding them to a list of errors and storing them in an exported list on the struct would avoid having to pass a loghook function?

lleadbet:
I guess if we do that, how do we expect folks to actually access them if needed? Agree structured logging would remain a problem here without providing a standard `Logger` attribute to set, but that feels heavy for, as you noted, issues unlikely to occur. 

Feels like we could remove them altogether if needed, since it's just cases where things _shouldn't_ be happening, but it's helpful for debugging.

StevenACoffman:
Yeah, ok, then I think we need a log hook like a stripped down version of :
https://github.com/hashicorp/go-retryablehttp/blob/main/client.go#L350 (but we don't need to provide levels etc.)

Even if the default logger is a No-op logger (something like https://github.com/rotationalio/x/blob/main/noplog/noplog.go ) but matches the signature of `fmt.Println(`. I've seen that done all the different ways, like a logrus one:

```
func NewDiscard() *Logger {
	l := logrus.New()
	l.Out = io.Discard
	return (*Logger)(l)
}
```

lleadbet:
@StevenACoffman - seeing you removing the prints; for the possible hook, did you want this in this PR or another?

StevenACoffman:
I don't like to hold contributors hostage to doing more work than they initially counted on, even though often they will stop at their more limited initial contribution and not come back to polish it up.

I was going to merge this PR without it, and then ask for a follow-up PR if I could impose upon you further.

I also hate to inflict on contributors asynchronous reviews where there's a chance of a protracted back-and-forth opinionated design process, rather than an unambiguous lint failure, and you've already been pretty patient with me in this one. If we were in the same timezone (US EST for me) I might try to do a call, but I think it's pretty late for you.

Another consideration is I updated gqlparser and gqlgen to support the apollo router's introspection queries that are using the draft graphql spec (`@deprecated` in input values, `@oneOf`) and until I cut a new release on gqlgen, no one can update gqlparser to latest without getting weird errors.

So if I can merge this as it is now, cut a new release, and then we can later (in another PR) add that log hook. That way it unblocks other folks, and you can more easily get feedback on this feature as it is in the real world.

lleadbet:
I'm actually in the US (NY)- and have a working version of the hook. I'll file another PR quickly for that and we can figure out timing. 

Appreciate the help here on this!

PR #1982, opened 2022-02-22
support to generate model for intermediate interface

Code-Hex:
Describe your PR and link to any relevant issues. 

Closes https://github.com/99designs/gqlgen/issues/1690
I have:
 - [x] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [ ] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

I hope "small change, small release". Thanks.

ref: https://github.com/vektah/gqlparser/pull/180

PR #455, opened 2018-11-28
Fix deprecated fields

vektah:
When answering introspection queries we were returning objects like this:
```
 {
          "kind": "OBJECT",
          "name": "Chatroom",
          "description": "",
          "fields": [
            {
              "name": "name",
              "description": "",
              "args": [
                
              ],
              "type": {
                "kind": "NON_NULL",
                "name": null,
                "ofType": {
                  "kind": "SCALAR",
                  "name": "String",
                  "ofType": null
                }
              },
              "isDeprecated": false,
              "deprecationReason": ""
            },
            {
              "name": "messages",
              "description": "",
              "args": [
                
              ],
              "type": {
                "kind": "NON_NULL",
                "name": null,
                "ofType": {
                  "kind": "LIST",
                  "name": null,
                  "ofType": {
                    "kind": "NON_NULL",
                    "name": null,
                    "ofType": {
                      "kind": "OBJECT",
                      "name": "Message",
                      "ofType": null
                    }
                  }
                }
              },
              "isDeprecated": false,
              "deprecationReason": ""
            }
```

**deprecationReason: ""**

This PR changes introspection to return null instead.

Fixes #436

I have:
 - [x] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))

PR #3507, opened 2025-01-27
Update gqlparser v2 to v2.5.22 to Support `@oneOf` and `@deprecated` on input values

StevenACoffman:
### Solved: Support `@oneOf` and `@deprecated` on input values

Some folks (e.g. @CodingContraption ) ran into a problem recently with [graphql-protect](https://github.com/ldebruijn/graphql-protect/) by @ldebruijn when used to protect an [apollo-router](https://github.com/apollographql/router) GraphQL API.

The latest apollo-router supports a draft version of the GraphQL spec and that draft adds support for deprecated input values: https://github.com/graphql/graphql-spec/pull/805

gqlparser v2.5.1 support was still at an earlier version of the GraphQL spec, resulting in errors when a client requests introspection fields like `isDeprecated` and `deprecationReason` on the `__InputValue`.

To fix this, in the gqlparser `prelude.graphql` was updated to match the draft introspection schema and the reference graphql-js implementation. This included a few updated validation rules. This was done in https://github.com/vektah/gqlparser/pull/342 and 
then github.com/vektah/gqlparser/v2 v2.5.2 was released, but now when updating gqlparser in gqlgen, I am finding a number of problems.

The last time there was a major GraphQL spec version support bump in gqlparser, gqlgen required some modifications #1982 so it may be that we need some modifications here in gqlgen, or it may be we need some tweaks to gqlparser or maybe both.

---

In this particular case, it looks like the gqlparser change is currently causing the generated ResolverRoot interface go from this:
```
type ResolverRoot interface {
	Mutation() MutationResolver
	Query() QueryResolver
	Subscription() SubscriptionResolver
}
```
To this:
```
type ResolverRoot interface {
	Mutation() MutationResolver
	Query() QueryResolver
	Subscription() SubscriptionResolver
	__InputValue() __InputValueResolver
	__Type() __TypeResolver
}
```
Which is causing this problem:
```
cannot use &resolver{…} (value of type *resolver) as ResolverRoot value in struct literal: *resolver does not implement ResolverRoot (missing method __InputValue)
```
This now looks like:
```
type Stub struct {
	QueryResolver struct {
		EchoIntToInt                func(ctx context.Context, n *int32) (int32, error)
		EchoInt64ToInt64            func(ctx context.Context, n *int) (int, error)
		EchoIntInputToIntObject     func(ctx context.Context, input Input) (*Result, error)
		EchoInt64InputToInt64Object func(ctx context.Context, input Input64) (*Result64, error)
	}
	__InputValueResolver struct {
		IsDeprecated      func(ctx context.Context, obj *introspection.InputValue) (bool, error)
		DeprecationReason func(ctx context.Context, obj *introspection.InputValue) (*string, error)
	}
	__TypeResolver struct {
		IsOneOf func(ctx context.Context, obj *introspection.Type) (*bool, error)
	}
}
```
Adding `IsDeprecated` was last touched in https://github.com/99designs/gqlgen/pull/455 so that might provide a model for how to support these changes.

---

**Updated:** Ok, it looks like it was pretty straightforward after all, following those two prior PR examples.

### DIFF
--- a/_examples/chat/generated.go
+++ b/_examples/chat/generated.go
@@ -526,6 +526,62 @@ func (ec *executionContext) field_Subscription_messageAdded_argsRoomName(
 	return zeroVal, nil
 }
 
+func (ec *executionContext) field___Directive_args_args(ctx context.Context, rawArgs map[string]any) (map[string]any, error) {
+	var err error
+	args := map[string]any{}
+	arg0, err := ec.field___Directive_args_argsIncludeDeprecated(ctx, rawArgs)
+	if err != nil {
+		return nil, err
+	}
+	args["includeDeprecated"] = arg0
+	return args, nil
+}
+func (ec *executionContext) field___Directive_args_argsIncludeDeprecated(
+	ctx context.Context,
+	rawArgs map[string]any,
+) (*bool, error) {
+	if _, ok := rawArgs["includeDeprecated"]; !ok {
+		var zeroVal *bool
+		return zeroVal, nil
+	}
+
+	ctx = graphql.WithPathContext(ctx, graphql.NewPathWithField("includeDeprecated"))
+	if tmp, ok := rawArgs["includeDeprecated"]; ok {
+		return ec.unmarshalOBoolean2ᚖbool(ctx, tmp)
+	}
+
+	var zeroVal *bool
+	return zeroVal, nil
+}
+
+func (ec *executionContext) field___Field_args_args(ctx context.Context, rawArgs map[string]any) (map[string]any, error) {
+	var err error
+	args := map[string]any{}
+	arg0, err := ec.field___Field_args_argsIncludeDeprecated(ctx, rawArgs)
+	if err != nil {
+		return nil, err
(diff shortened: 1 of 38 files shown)


peter-svensson:
Nice! Any idea when this will be released?
I cannot update my go.mod dependencies for vektah/gqlparser/v2 to v2.5.22 until then, not a big issue but it's always nice to get the latest stuff in (and keep Dependabot/Renovate happy 😊)

StevenACoffman:
@peter-svensson I released https://github.com/99designs/gqlgen/releases/tag/v0.17.64 a few minutes ago.

Sorry for the awkward period where upgrading gqlparser would break gqlgen. 😞

peter-svensson:
> @peter-svensson I released https://github.com/99designs/gqlgen/releases/tag/v0.17.64 a few minutes ago.
> 
> Sorry for the awkward period where upgrading gqlparser would break gqlgen. 😞

No worries what so ever! Thanks for the great work!

PR #3509, opened 2025-01-28
feat: add logger for use with tracing middleware

lleadbet:
In #3506, it was discussed to allow users to bring a custom logger to print out the various error logs that might happen during the tracing process to allow users to properly set the right structured logging consistently across the application. 

This adds a new `Logger` attribute on the `Tracer` struct that can take in a Logger that abides by the standard `log` library standard (that is it has at a minimum has the Print/ln/f functions). 

Defaults to a no-op logger. 

I have:

### DIFF
--- a/graphql/handler/apollofederatedtracingv1/logger/logger.go
+++ b/graphql/handler/apollofederatedtracingv1/logger/logger.go
@@ -1,0 +1,31 @@
+package tracing_logger
+
+import "log"
+
+// Logger is an interface that can be implemented to log errors that occur during the tracing process
+// This can use the default Go logger or a custom logger (e.g. logrus or zap)
+type Logger interface {
+	Print(args ...interface{})
+	Println(args ...interface{})
+	Printf(format string, args ...interface{})
+}
+
+func New() *NoopLogger {
+	return &NoopLogger{
+		log.New(NullWriter(1), "", log.LstdFlags),
+	}
+}
+
+type NullWriter int
+
+func (NullWriter) Write([]byte) (int, error) { return 0, nil }
+
+type NoopLogger struct {
+	*log.Logger
+}
+
+func (l *NoopLogger) Print(args ...interface{}) {}
+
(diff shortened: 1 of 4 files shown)


graphql/handler/apollofederatedtracingv1/logger/logger.go
```diff
@@ -0,0 +1,31 @@
...
+
+// Logger is an interface that can be implemented to log errors that occur during the tracing process
+// This can use the default Go logger or a custom logger (e.g. logrus or zap)
+type Logger interface {
+	Print(args any)
+	Println(args any)
+	Printf(format string, args any)
+}
+
+func New() *NoopLogger {
```
StevenACoffman, line 13:
```suggestion
// This will create a logger that discards all messages (i.e. does not output)
func NewNoopLogger() *NoopLogger {
```
I would rename this, otherwise folks who read the package name + `New()` will think they are going to get some output.

graphql/handler/apollofederatedtracingv1/logger/logger.go
```diff
@@ -0,0 +1,31 @@
+package tracing_logger
```
StevenACoffman, line 1: I would consider renaming this to just `logger` 
since your package is already inside the tracing parent, the tracing part is implied, and then it matches the directory name, and doesn't have an underscore in the name (even if people may still import it as an alias as `tracing_logger` to disambiguate between this and some other logger package they also import).
```suggestion
package logger
```

StevenACoffman (commented):
Hey, thanks for the quick turn around. I have a few nits, but this looks great.

lleadbet:
Agreed on both points; had considered the package name (and New function) but was semi-concerned about folks importing that on accident thinking it was related to something else, not the tracing package, but given the import path including the tracing piece it was probably less important.

lleadbet:
no rush on the new release, the actual needed feature was done and this is just a polishing piece :)

PR #3534, opened 2025-02-13
Update matrix to include Go 1.24

StevenACoffman:
Go 1.24 is out! Hooray!
In #3505 reports problems with gqlgen and Go 1.24 due to a dependency on an old version of x/tools

However, GCP Appengine still [doesn't support Go 1.23 as generally available](https://cloud.google.com/appengine/docs/standard/go/release-notes) so we cannot drop Go 1.22 support yet. This means that we cannot update x/tools to a newer version.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -16,7 +16,7 @@ jobs:
   coverage:
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


StevenACoffman:
Superceded by other PRs.

PR #3540, opened 2025-02-15
x/tools update go get golang.org/x/tools@v0.30.0

StevenACoffman:
The dependabot upgrade shows all kinds of failures, but I wonder if just also bumping it in the examples while doing a go mod tidy in both places is all that is necessary.


StevenACoffman:
OMG... that was all that was necessary.

PR #3552, opened 2025-02-21
codegen/type.gotpl: Avoid tautological nil check for slices

HaraldNordgren:


codegen/type.gotpl
```diff
@@ -18,9 +18,7 @@
 				return &res, graphql.ErrorOnPath(ctx, err)
 			{{- else if $type.IsSlice }}
 				var vSlice []any
-				if v != nil {
-					vSlice = graphql.CoerceList(v)
-				}
+				vSlice = graphql.CoerceList(v)
```
HaraldNordgren, line 21: Lines 9-11 already check if type is nullable, which a slices is.

`go/x/tools` nilness check reports the generated code as `tautological condition: non-nil != nil`.

StevenACoffman:
Thanks!

Issue #3553, opened 2025-02-22
Allow plugins to mutate the schema before models generation


islamaliev:
Currently plugins can implement `config.ConfigMutator` interface that defines only `MutateConfig` method. The problem is that there are built-in plugins (like modelgen) that are executed first, hence there is no way to change it before.
 
In certain cases it's very handy to be able to change the schema before models are generated. This can be done by adding another interface that plugins can implements and that are executed before `config.ConfigMutator`.

PR #3554, opened 2025-02-22
feat: Add SchemaMutator method to be used by plugins

islamaliev:
Resolves https://github.com/99designs/gqlgen/issues/3553

This PR adds another interface `SchemaMutator` that allows plugins to mutate the schema before models are generated.

### Background:

Currently, plugins can implement `config.ConfigMutator` interface that defines only `MutateConfig` method. The problem is that there are built-in plugins (like modelgen) that are executed first, hence there is no way to change it before.
 
In certain cases, it's convenient to be able to change the schema before models are generated. This can be done by adding another interface that plugins can implement and that is executed before `config.ConfigMutator`.

I have:

### DIFF
--- a/api/generate_test.go
+++ b/api/generate_test.go
@@ -1,13 +1,12 @@
 package api
 
 import (
+	"errors"
 	"os"
 	"path/filepath"
 	"testing"
 
 	"github.com/stretchr/testify/require"
-
-	"fmt"
 
 	"github.com/99designs/gqlgen/codegen/config"
 	"github.com/vektah/gqlparser/v2/ast"
@@ -69,7 +68,7 @@ func (t *testSchemaMutator) Name() string {
 
 func (t *testSchemaMutator) MutateSchema(schema *ast.Schema) error {
 	if t.shouldError {
-		return fmt.Errorf("deliberate schema mutation error")
+		return errors.New("deliberate schema mutation error")
 	}
 	schema.Types["TestType"] = &ast.Definition{
 		Kind: ast.Object,


StevenACoffman:
Thank you for contributing this! (It looks like there is a minor lint nit by the way)

This looks quite useful, but it would be good to give people a clear example when they would want to use this.

If you have a public plugin that you are thinking about contributing (either here or in another repository), or something in the `_examples` directory, or just a clear description of a how you plan to use this yourself, that would be very helpful.

StevenACoffman:
@islamaliev I cannot push to your PR, so [8bc81158ba79b31ce8cd2865152c9826268f3af3](https://github.com/99designs/gqlgen/commit/8bc81158ba79b31ce8cd2865152c9826268f3af3) looks like the lint fix

islamaliev:
@StevenACoffman thanks for your feedback. I applied your fix lint suggestion.

Regarding an example: I could share my plugin, but I'm not sure if it's worth it, because it's a bit specific for our needs. Basically what it does is for every field with `@perm` directive it adds to the resulting union possible errors:
```graphql
getUsers(input: GetUsersInput!): GetUsersResult! @perm(resource: "user", actions: ["user_read"])
```
In this example `GetUsersResult` among other types will get extended with two more error types.

Which file do you think is good to add some information about what the new plugin hook can do?

StevenACoffman:
Honestly, that would be a very broadly useful example! If you don't mind putting some form of that up in a public repo somewhere, I would gladly link to it in the documentation. People often wish to enrich their errors with custom information that comes from schema directives. If you do, you could just reply in this PR to notify me and I would take care of it.

For instance, we have popular documentation on using schema directives to implement Role-Based Authentication Controls (RBAC) here: https://gqlgen.com/reference/directives/ (and a little bit more here https://gqlgen.com/recipes/authentication/ ) but the resulting developer experience could be much improved with a plugin like you are describing. 

--- 

If you'll indulge me in reading just a bit more, you have accidentally come close to providing something that has been very frequently requested from the community. I know it's a digression as you're just trying to get something straightforward done, but you've made me very excited!

People frequently report they struggle to implement Attribute-Based Authentication Controls (ABAC) with schema directives. ABAC is more complicated and bespoke than RBAC, and so it's not as easy to make a simple, generic, public gqlgen example with ABAC. 

However, gqlgen currently requires a certain amount of developer friction and boilerplate to accomplish this, and your schema mutator would help to make it easier.

In the [Apollo Router authorization](https://www.apollographql.com/docs/graphos/routing/security/authorization) you have:
The @requiresScopes directive allows granular access control through the scopes you define.
The @authenticated directive allows access to the annotated field or type for authenticated requests only.

islamaliev:
@StevenACoffman sure. I'd be glad to help the community.

First of all, for our ABAC to work the plugin is not required. So let's leave it out for now.

We have the following directives defined:
```graphql
"""
Represents a permission directive that can be used to enforce permissions on fields.

Field marked with this directive will only be accessible if the user has the required permissions.
"""
directive @perm(resource: String!, actions: [String!]!) on FIELD_DEFINITION

"""
Represents a directive that can be used to enforce self-user permissions on fields.

Field marked with this directive will only be accessible if the user requests their own data.
The entity under the resource will be checked to have the current user id in the idField.

If specified without parameters, the directive will just check the user's authentication.
"""
directive @ownData(resource: String, idField: String) on FIELD_DEFINITION
```
And as an example for where it's used we take users:
```graphql
union UpdateOwnUserResult = User | UserWithEmailAlreadyExistsError | LoginRequiredError 
union UpdateUserResult = User | NoUserWithIDError | UserWithEmailAlreadyExistsError | LoginRequiredError | NoPermissionError

extend type Mutation {
  """
  Updates user's own data.
  """
  updateOwnUser(input: UserDataUpdateInput!): UpdateOwnUserResult!  @ownData

  """
  Updates a user's information.
  """
  updateUser(input: UserUpdateInput!): UpdateUserResult!  @perm(resource: "user", actions: ["user_update"])
}
```
Here is the handler for the `@perm` directive:
```go
// handlePermDirective creates a directive handler function for permission checking in GraphQL resolvers.
// It takes an auth.Service as input and returns a function that validates user permissions for specific resources and actions.
//
// The returned handler function:
// - Extracts user ID from the context
// - Checks if user is logged in, returns LoginRequiredError if not
// - Validates if user has required permissions using authService
// - Returns NoPermissionError if user lacks permission
// - Proceeds with resolution if user has permission
//
// Usage:
//
// Apply as @perm directive in GraphQL schema to protect resolvers
//
// Example:
//
//	type Query {
//	  updateUser(input: UserInput!): UpdateUserResult! @perm(resource: "user", actions: ["user_update"])
//	}
func handlePermDirective(authService auth.Service) func(
	context.Context,
	any,
	graphql.Resolver,
	string,
	[]string,
) (any, error) {
	return func(
		ctx context.Context,
		obj any,
		next graphql.Resolver,
		resource string,
		actions []string,
	) (any, error) {
		userID := user.FromContext(ctx)
		if userID == 0 {
			return model.LoginRequiredError{Field: graphql.GetPath(ctx).String()}, nil
		}

		success, e := authService.HasUserPermission(
			ctx,
			authModel.UserID(userID),
			authModel.Permission{Resource: resource, Actions: actions},
		)

		if e != nil {
			return e, nil
		}

		if !success {
			return model.NoPermissionError{Field: graphql.GetPath(ctx).String()}, nil
		}
		return next(ctx)
	}
}
```
Here the handler requests the auth service to provide information about user's permissions.

The `@ownData` directive also does authentication, but (parameters provided) also checks if the authenticated user own the resource it's requesting. For this the corresponding service needs to implement a certain interface. I won't give the implementation as it might not be that relevant to the community and it's not yet very well tested.

Now, coming back to the plugin. It basically allow you not to specify `LoginRequiredError` and `NoPermissionError` for every restricted field. It will automatically extend the unions. Here is the code:
```go
package perm

import (
	"strings"

	"github.com/99designs/gqlgen/codegen/config"
	"github.com/99designs/gqlgen/plugin"
	"github.com/vektah/gqlparser/v2/ast"
)

type Plugin struct {
	// extensions is a map from union name to a list of types that should be added to the union.
	// This map is populated during schema mutation and used during config mutation to add an
	// additional source to the list of raw graphql sources.
	// It is necessary so that the `parsedSchema` from `generated` package contains the union extensions.
	extensions map[string][]string
}

var (
	_ plugin.SchemaMutator = (*Plugin)(nil)
)

func New() plugin.Plugin {
	return &Plugin{
		extensions: make(map[string][]string),
	}
}

func (p *Plugin) Name() string {
	return "perm"
}

func (p *Plugin) MutateSchema(schema *ast.Schema) error {
	for _, schemaType := range schema.Types {
		if schemaType.Kind == ast.Object || schemaType.Kind == ast.Interface {
			for _, field := range schemaType.Fields {
				if hasPermDirective(field) {
					p.addTypesToUnion(schema, field, []string{"LoginRequiredError", "NoPermissionError"})
				} else if hasOwnDataDirective(field) {
					p.addTypesToUnion(schema, field, []string{"LoginRequiredError"})
				}
			}
		}
	}
	return nil
}

func (p *Plugin) MutateConfig(cfg *config.Config) error {
	sb := strings.Builder{}
	for unionName, types := range p.extensions {
		sb.WriteString("extend union ")
		sb.WriteString(unionName)
		sb.WriteString(" = ")
		for i, typeName := range types {
			if i > 0 {
				sb.WriteString(" | ")
			}
			sb.WriteString(typeName)
		}
		sb.WriteString("\n")
	}

	cfg.Sources = append(cfg.Sources, &ast.Source{
		Name:  "extend_union_with_perm",
		Input: sb.String(),
	})
	return nil
}

func hasPermDirective(field *ast.FieldDefinition) bool {
	for _, dir := range field.Directives {
		if dir.Name == "perm" {
			return true
		}
	}
	return false
}

func hasOwnDataDirective(field *ast.FieldDefinition) bool {
	for _, dir := range field.Directives {
		if dir.Name == "ownData" {
			return true
		}
	}
	return false
}

func (p *Plugin) addTypesToUnion(schema *ast.Schema, field *ast.FieldDefinition, typesToAdd []string) {
	typ := field.Type
	for typ.Elem != nil {
		typ = typ.Elem
	}

	// If the return type is not a union, skip
	unionDef := schema.Types[typ.Name()]
	if unionDef == nil || unionDef.Kind != ast.Union {
		return
	}

	for _, typeToAdd := range typesToAdd {
		hasUnitType := false
		for _, existingType := range unionDef.Types {
			if existingType == typeToAdd {
				hasUnitType = true
				break
			}
		}
		if !hasUnitType {
			addTypeToUnion(schema, unionDef, typeToAdd)
			p.extensions[unionDef.Name] = append(p.extensions[unionDef.Name], typeToAdd)
		}
	}
}

func addTypeToUnion(schema *ast.Schema, unionDef *ast.Definition, newType string) {
	// Add the new type to the union and the union to the new type's implements list
	// These 2 step below are necessary for models (model/models_gen.go) to be generated correctly.
	unionDef.Types = append(unionDef.Types, newType)
	schema.Implements[newType] = append(schema.Implements[newType], unionDef)

	// Add the new type to possible types of the union
	// This step is necessary for the rest of code generation (generated.go) so that it can
	// generate code for strong type checking.
	// For example for CreateCourseProductResult union within the function _CreateCourseProductResult
	// it will generate code like this:
	// case model.LoginRequiredError:
	//	   return ec._LoginRequiredError(ctx, sel, &obj)
	// case *model.LoginRequiredError:
	//	   if obj == nil {
	//		   return graphql.Null
	//	   }
	//	   return ec._LoginRequiredError(ctx, sel, obj)
	schema.AddPossibleType(unionDef.Name, schema.Types[newType])
}
```
What we are doing now with the `@perm` directive is making it work for nested response fields. For example, here is the part of `Product` definition:
```graphql
type Product {
  "Unique identifier for the product"
  id: ProductID!
  "Category of the product"
  category: ProductCategory!
  "Merchant's share of the product revenue"
  merchantShare: Rate @perm(resource: "product", actions: ["details_read"])
  # ...
}
```
The field `merchantShare` should be accessible only to authorized users. But this requires a bit more intricate handling or gql errors which is slightly out of scope for this discussion.

Let me know if my examples are clear, if it's helpful at all and if you have any other questions.

Issue #3573, opened 2025-03-06
Expose GraphQL extensions via the OperationContext


NavidMitchell:
Currently if a client passes extensions in their request body during a GraphQL request there is no way to access those extensions from our resolvers. I would like to have the extensions available when our service receives the request.

PR #3574, opened 2025-03-06
feat: Expose GQL Extensions to OperationContext

NavidMitchell:

https://github.com/99designs/gqlgen/issues/3573

I have:
 Added the GQL Extensions to the OperationContext

### DIFF
--- a/graphql/context_operation.go
+++ b/graphql/context_operation.go
@@ -17,6 +17,7 @@ type OperationContext struct {
 	Variables     map[string]any
 	OperationName string
 	Doc           *ast.QueryDocument
+	Extensions    map[string]any
 	Headers       http.Header
 
 	Operation              *ast.OperationDefinition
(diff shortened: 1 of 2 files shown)


StevenACoffman:
Thanks for this contribution. I can see this being useful (although unit tests would be appreciated in the future). Can you say more about which specific GraphQL extensions you are hoping to expose to the operation context? Right now, by getting them from the RawParams you don't have a lot of type safety and there might be a better way to get the extension you want.

NavidMitchell:
Hello @StevenACoffman thank you for the super fast response. :)

I will add tests in the future. Sorry about that.

In our case we are looking to pull a list of strings out of a specific extension key. I added the code as I did since GQL extensions are not strictly defined. https://graphql.org/learn/serving-over-http/#body . 

As for type safety I generally prefer to have it as well. I hadn't thought about it in this case. 

I guess we could add something like below somewhere?  I'm somewhat of an experienced golang programmer so I am not sure if this adequate or not as well. 

```
func GetExtensionValue[T any](c *OperationContext, key string) (T, error) {
	val, ok := c.Extensions[key]
	if !ok {
		var zero T
		return zero, errors.New("key not found")
	}
	tVal, ok := val.(T)
	if !ok {
		var zero T
		return zero, errors.New("type mismatch")
	}
	return tVal, nil
}

```

PR #3587, opened 2025-03-12
feat: add StoragePrefix to graphiql playground

eiixy:
When multiple endpoints exist under the same domain, localStorage might overwrite each other, leading to data loss or inconsistency.

Example:
```go
playground.Handler("Graphiql", "/query", playground.WithStoragePrefix("serviceName_"))
```

### DIFF
--- a/graphql/playground/playground.go
+++ b/graphql/playground/playground.go
@@ -10,7 +10,7 @@ var page = template.Must(template.New("graphiql").Parse(`<!DOCTYPE html>
 <html>
   <head>
   	<meta charset="utf-8">
-  	<title>{{.title}}</title>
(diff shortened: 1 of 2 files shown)


StevenACoffman:
Thanks! This is a nice improvement all by itself, and I know you came just for that, but it makes me notice there's a maintenance problem here that this change (only slightly) makes worse (because it changes the line numbers).

Currently, when there's a new version of Graphiql playground, these obscure values are do not have a link to any documentation that tells people how to discover and update the correct newer values. People have to do a git blame, find the old PRs, look up the descriptions, follow the breadcrumbs in some of them.

If you are interested in updating Graphiql further, can you please document the procedure you follow to find the new values in a comment or something?

PR #3596, opened 2025-03-15
Being extra careful in the sort makes me sleep better

StevenACoffman:
This is a follow-up from #3591 with a suggestion from @csilvers so that we not be as dependent on input order for sorting, but instead explicitly use Name as a sorting secondary sort criteria.

Also, he noticed we can be tighter with the explicitness of sorting our important file rule as well. 

For both, there is still a possibility of a genuine tie (identical in every regard) so we still need to use SortStable so that the sort is at least deterministic based on input order rather than randomly changing every time and causing test failures.

Note: There is some other completely unrelated minor stuff I lazily threw in this same PR while I was here.

The linter warned me that the `template` variable shadowed the package name which is just a bad practice, so I renamed it.

The generated `exec.go` files in Khan Academy's code has a bunch of useless if conditional wrapping:
```
	if v != nil {
		vSlice = graphql.CoerceList(v)
	}
```
Which I find very odd, because the `type.gotpl` here in gqlgen doesn't have that. While I was looking into that, I made a minor refactor of the `graphql.CoerceList` here to make it more obvious that having all those redundant checks is not necessary.

### DIFF
--- a/_examples/federation/accounts/graph/generated.go
+++ b/_examples/federation/accounts/graph/generated.go
@@ -3297,13 +3297,6 @@ func (ec *executionContext) __Entity(ctx context.Context, sel ast.SelectionSet, 
 	switch obj := (obj).(type) {
 	case nil:
 		return graphql.Null
-	case model.EmailHost:
-		return ec._EmailHost(ctx, sel, &obj)
-	case *model.EmailHost:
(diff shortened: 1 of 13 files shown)


csilvers (commented):
Thanks @StevenACoffman !

PR #3617, opened 2025-03-25
New 'type' argument for goField directive

atzedus:
Sometimes it needed to customize go-type of some field in generated model.
But i don't want to replace it with fully user-defined model with goModel directive.
This PR adds ability to do this.

Example:

```graphql
type User {
    id: Int!
    name: String! @goField(type: "my.module/my-package.SuperString")
    created: Time!
    updated: Time!
}
```

```go
type User struct {
  id int
  name SuperString
  created time.Time
  updated time.Time
}
```

I have:


StevenACoffman:
Thanks for this!

PR #3630, opened 2025-04-01
breaking feat: add GraphQL models and update complexity functions to use context

### LINKED ISSUE #3629: Regarding the use of context.TODO() in the code for complexity in codegen, which results in the resolver not retrieving the context correctly
Below are the changes I have introduced for this issue

Image

z9905080:
### Note: Breaking Change as this requires re-generating to match this new Complexity function signature

Without context, all complexity weights had to be static constants baked into the generated code, which was a significant constraint for any application where "how expensive is this field?" depends on who's asking.

Custom complexity functions previously lacked access to the request context, making it impossible to vary complexity based on runtime information. With `context.Context` now threaded through the entire complexity calculation — from Calculate down through selectionSetComplexity, interfaceFieldComplexity, and fieldComplexity to the Complexity interface method itself — a user-defined complexity function can now inspect anything attached to the context when deciding how expensive a field is.

###  Practical examples of what this enables:

  - Per-user or per-tenant limits: a complexity function can read the authenticated user from the context and assign different
  weights depending on their tier or role
  - Feature flags: weighting an expensive field differently when a feature is in a rollout phase
  - Tracing/observability: attaching spans or logging inside complexity calculations using a tracer from the context

This PR updates the Complexity functions to use the context correctly.

Fixes #3629


z9905080:
> It looks like you need to do this:
> 
> ```
> go generate ./...
> ```
> 
> and then commit your changes

Thank you, @StevenACoffman! 
Yes, this is my first time committing to this project.

StevenACoffman:
Thanks! I appreciate you working on this. It looks like you are still missing a change in at least two other places:
- https://github.com/z9905080/gqlgen/blob/master/graphql/executor/testexecutor/testexecutor.go#L120
- https://github.com/z9905080/gqlgen/blob/master/graphql/executor/testexecutor/testexecutor.go#L173

z9905080:
> [覆蓋狀態](https://coveralls.io/builds/73027441)
> 
> 覆蓋率：73.349％。**將 z9905080:master 上的****[b1eda1f](https://github.com/99designs/gqlgen/commit/b1eda1fa3395e61213266a4902dd444da79227af)**拉 入**99designs:master 上的****[492bf92](https://github.com/99designs/gqlgen/commit/492bf92b6dd38fcb90584fb92f13bcc261656fa3)** 時保持不變。**[](https://github.com/99designs/gqlgen/commit/b1eda1fa3395e61213266a4902dd444da79227af)****[](https://github.com/99designs/gqlgen/commit/492bf92b6dd38fcb90584fb92f13bcc261656fa3)**

Do I need to take additional steps to address the test coverage issues?

z9905080:
image

Perhaps I could add some inconsequential test cases, but I'm not sure if that's a good approach. 
In general, we don't specifically pursue such test cases.

StevenACoffman:
Thank you for your persistence in continuing to work on this! 

I'm not overly concerned with the specific code coverage number.

However, your original issue did not describe the symptom that would have been seen without your fix, or how to reproduce the circumstances when it would occur.

Now, we do not have a test that prevents someone from accidentally re-introducing the same bug.

Before your change, when resolver did NOT retrieve the context correctly, what would the error have been?

Would there be any possible test case that could have simulated the old error before your change?

Could you think of a test case that would provide some regression safety here?

PR #3631, opened 2025-04-01
Example of separating schema and resolvers for a large project structure with for multi-team setup

afreeland:
This MR includes an additional example of how it is possible to separate schema and resolvers into separate packages, even external packages, allowing support for larger projects. Hoping this kind of dovetails in the this MR: https://github.com/99designs/gqlgen/pull/3595

The project has the following layout:
```
tree 
.
├── integration
│   ├── go.mod
│   ├── go.sum
│   ├── integration.go
│   └── schema.graphqls
├── main
│   ├── go.mod
│   ├── go.sum
│   ├── gqlgen.yml
│   ├── graph
│   │   ├── generated.go
│   │   ├── model
│   │   │   └── models_gen.go
│   │   ├── resolver.go
│   │   ├── schema.graphqls
│   │   └── schema.resolvers.go
│   ├── server.go
│   └── tools.go
└── shared
    ├── go.mod
    ├── go.sum
    ├── schema.graphqls
    └── shared.go

6 directories, 18 files
```

The main package is the centralized graph and will consume the schema files from both the `shared` and the `integration` packages.  

The main package, after initial `generate` command, has commented out the `resolver:` section in the `gqlgen.yml` file.  This prevents the main package from generating *all* resolvers, which we would like other teams to be able to manage. When `generate` is ran, it will still generate all model information in the main package, which the external packages can manage.

Additionally, the main package in the `resolver.go` file tries to lift the generated interface for external resolvers into its own interface called `ExternalQueryResolver`.  This interface is then embedded in in the `Resolver` like so:
```
type Resolver struct {
	ExternalQueryResolver
}
```
This allows the `RootResolver` to still function as it should but lifts some of the `MutationResolver` and `QueryResolver` interfaces that were generated up in a way that other teams can implement the resolver.

Then upon initialization, you can define the `ExternalQueryResolver` from another package like so:
```
// Create a new executable schema with the composed resolver
	srv := handler.NewDefaultServer(graph.NewExecutableSchema(graph.Config{
		Resolvers: &graph.Resolver{
			ExternalQueryResolver: &integration.Resolver{},
			// Add other team resolvers here
		},
	}))
```

It was challenging to find a good way to provide a balance of gqlgen built-in functionality in a way that allowed for ease of use, while also allowing still providing some autonomy to other teams.

There are very few discussions on how this is doable and I couldn't really find any examples of doing this in practice.  I wanted to put this up in hopes it would help others...or at the very least call out any major flaws that may exist with this.

I have:


_examples/large-project-structure/integration/go.mod
```diff
@@ -0,0 +1,17 @@
+module github.com/corelight/integration
```
StevenACoffman, line 1: Do you mind if we rename this something that matches the directory like:
```suggestion
module github.com/99designs/gqlgen/_examples/large-project-structure/integration
```

StevenACoffman:
So, some larger installations also use `use_function_syntax_for_execution_context`, `omit_root_models` and a few other config options to further customize behavior.

StevenACoffman:
I wonder if some members of the community who have experience using gqlgen in larger orgs or teams might offer guidance here? 
@a8m @giautm @kanodia-parag @duckbrain 

For a single gqlgen project, what gqlgen config works best really changes at a few key points during the course of that project's normal growth and evolution. For instance:

1. quick proof of concept - single file layout
2. medium-sized team(s) - follow schema layout
3. many teams - separate schema and resolvers into separate packages as in this example
4. very large type systems (more than 65,000 methods) - `use_function_syntax_for_execution_context`

However, some will instead choose to adopt GraphQL Federation and split into multiple gqlgen instances before one of these growth points is even reached.

PR #3632, opened 2025-04-02
feat: bind from basic type to named type

markya0616:
According this [pull request](https://github.com/99designs/gqlgen/pull/3617), we can set a custom type for a field.

Example:
```
type User {
    id: Int!
    number: Int! @goField(type: "my.module/my-package.SuperNumber")
}

type User struct {
  id int
  number SuperNumber
}
```

However, it will generate a new resolver like below:
```
type ResolverRoot interface {
	User()UserResolver
}
type UserResolver interface {
	Number(ctx context.Context, obj *User, number int) error
}
```

This pull request provides a convenient way without the Resolver interface via graphql.ContextMarshaler/graphql.ContextUnmarshaler/graphql.Marshaler/graphql.Unmarshaler interface

```
type SuperNumber uint64

func (e *SuperNumber) UnmarshalGQLContext(ctx context.Context, v interface{}) error {
	id, ok := v.(int)
	if !ok {
		return errors.New("not a int")
	}

	*e = SuperNumber(id)
	return nil
}

func (e SuperNumber) MarshalGQLContext(_ context.Context, w io.Writer) error {
	fmt.Fprint(w, e)
	return nil
}
```

I have:
 - [V] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [V] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

### DIFF
--- a/codegen/config/binder.go
+++ b/codegen/config/binder.go
@@ -476,7 +476,18 @@ func (b *Binder) TypeReference(schemaType *ast.Type, bindTarget types.Type) (ret
 		ref.GO = b.CopyModifiersFromAst(schemaType, ref.GO)
 
 		if bindTarget != nil {
-			if err = code.CompatibleTypes(ref.GO, bindTarget); err != nil {
+			// if the bind type implements the graphql.ContextMarshaler/graphql.ContextUnmarshaler/graphql.Marshaler/graphql.Unmarshaler interface, we can use it
+			if hasMethod(bindTarget, "MarshalGQLContext") && hasMethod(bindTarget, "UnmarshalGQLContext") {
+				ref.IsContext = true
+				ref.IsMarshaler = true
+				ref.Marshaler = nil
+				ref.Unmarshaler = nil
+			} else if hasMethod(bindTarget, "MarshalGQL") && hasMethod(bindTarget, "UnmarshalGQL") {
+				ref.IsContext = false
+				ref.IsMarshaler = true
+				ref.Marshaler = nil
+				ref.Unmarshaler = nil
+			} else if err = code.CompatibleTypes(ref.GO, bindTarget); err != nil {
 				continue
 			}
 			ref.GO = bindTarget


codegen/config/binder.go
```diff
@@ -476,7 +476,18 @@ func (b *Binder) TypeReference(schemaType *ast.Type, bindTarget types.Type) (ret
 		ref.GO = b.CopyModifiersFromAst(schemaType, ref.GO)
 
 		if bindTarget != nil {
-			if err = code.CompatibleTypes(ref.GO, bindTarget); err != nil {
+			// if the bind type implements the graphql.ContextMarshaler/graphql.ContextUnmarshaler/graphql.Marshaler/graphql.Unmarshaler interface, we can use it
+			if hasMethod(bindTarget, "MarshalGQLContext") && hasMethod(bindTarget, "UnmarshalGQLContext") {
+				ref.IsContext = true
+				ref.IsMarshaler = true
+				ref.Marshaler = nil
+				ref.Unmarshaler = nil
```
StevenACoffman, line 484: I'm not sure about these two lines.

codegen/config/binder.go
```diff
@@ -476,7 +476,18 @@ func (b *Binder) TypeReference(schemaType *ast.Type, bindTarget types.Type) (ret
...
+			} else if hasMethod(bindTarget, "MarshalGQL") && hasMethod(bindTarget, "UnmarshalGQL") {
+				ref.IsContext = false
+				ref.IsMarshaler = true
+				ref.Marshaler = nil
```
StevenACoffman, line 488: Again, not sure about these two lines. I would need think through a number of possible scenarios.

codegen/config/binder.go
markya0616: It's because we may set them on https://github.com/99designs/gqlgen/pull/3632/files#diff-ab508c289540d69c88d9f1146004bbb03b2fbefa9b9ce932e044927fea85aeccL449-L450.

For the `bindTarget` implementing these interface, it's Marshaler and Unmarshaler should be nil.

atzedus:
So, when i worked on PR (https://github.com/99designs/gqlgen/pull/3617) i used test data like follow:

```graphql
type User {
    id: ID! @goField(type: "github.com/google/uuid.UUID")
}
```

BUT also i have gqlgen.yaml:

```yaml
  ID:
    model:
      - github.com/99designs/gqlgen/graphql.UUID
```

My goal was to give the ability to specify an alternative type for generated field, rather than the first one from the list (github.com/99designs/gqlgen/graphql.ID in this case).
With this configuration, no additional resolver is required and not generated.

As a result, to solve the problem described above this is neccessary to configure `gqlgen.yaml` like:

```yaml
  Int:
    model:
      - my.module/my-package.SuperNumber
```

And in package implement marshal/unmarshal methods like:

```go
func MarshalSuperNumber(id SuperNumber) graphql.Marshaler {
	// implement me
}

func UnmarshalSuperNumber(v any) (SuperNumber, error) {
	// implement me
}
```

My opinion is that it works more clearly this way.

Unfortunately, I do not have expertise in the binder code and it difficult to predict what problems the proposed changes may cause.

But this feature looks cool.

codegen/config/binder.go
atzedus: According on comment at https://github.com/99designs/gqlgen/blob/master/codegen/config/binder.go#L151-L152
and then calling `FindObject` at https://github.com/99designs/gqlgen/blob/master/codegen/config/binder.go#L436 settings this properties to nil looks like as breaking changes.

markya0616: https://github.com/99designs/gqlgen/pull/3632#discussion_r2024829492
Here is the main reason why I set to nil.

Why is this a breaking change?
For those implementing the graphql.ContextUnmarshaler/graphql.Marshaler interface, it's `Marshaler` and `Unmarshaler` should be nil.
https://github.com/99designs/gqlgen/pull/3632/files#diff-ab508c289540d69c88d9f1146004bbb03b2fbefa9b9ce932e044927fea85aeccR451-R458

In this PR, we only check if the `bindTarget` meets the interface. 
If they implement the interface, we follow the same behavior in [here](https://github.com/99designs/gqlgen/pull/3632/files#diff-ab508c289540d69c88d9f1146004bbb03b2fbefa9b9ce932e044927fea85aeccR451-R458
). In current behavior, if they implement the interface, `Marshaler` and `Unmarshaler` should be nil.

atzedus: Ok, i will try to explain.

First, implement marshalers

```go
type SuperNumber int

func MarshalSuperNumber(id SuperNumber) graphql.Marshaler {
	return graphql.MarshalAny(`"Hello from MarshalSuperNumber"`)
}

func UnmarshalSuperNumber(v any) (SuperNumber, error) {
	return 100, nil
}

func (b SuperNumber) MarshalGQL(w io.Writer) {
	w.Write([]byte(`"Hello from MarshalGQL"`))
}

func (b *SuperNumber) UnmarshalGQL(v any) error {
	n := SuperNumber(200)
	b = &n
	return nil
}
```

Second, use schema from example:

```graphql
type User {
    id: Int!
    number: Int! @goField(type: "my.module/my-package.SuperNumber")
}

type Query {
    user: User
}
```

And finally configure gqlgen.yaml like:

```yaml
models:
    Int:
        model:
            - github.com/99designs/gqlgen/graphql.Int
            - my.module/my-package.SuperNumber
```

After generating server code, you can invoke `user` method in playground and you can see, that it use `func (b SuperNumber) MarshalGQL(w io.Writer)`, not `func MarshalSuperNumber(id SuperNumber) graphql.Marshaler`, but according to source code and again at https://github.com/99designs/gqlgen/blob/master/codegen/config/binder.go#L151-L152 it `func MarshalSuperNumber` must be used here.

OR NOT? :) @StevenACoffman @duckbrain  help :)

StevenACoffman:
I'm sick and feverish at the moment, and the binder.go code is really complicated and delicate.

I think what you are attempting to do is a neat idea, but right now I can't really think through the implications of `ref.Unmarshaler = nil` and `ref.Marshaler = nil` in those spots and whether or not that is correct. 

There is no test to prove one way or the other in this PR (if I remove those lines, the tests all still pass, right?).

Do you mind if I delay deciding until I feel better or someone can think of a test that proves what the implications are?

markya0616:
Please take good care of yourself. We can talk about how to improve this when you're feeling better.

markya0616:
The reason to set Unmarshaler/Marshaler to nil is because we check them before IsMarshaler.
https://github.com/99designs/gqlgen/blob/master/codegen/type.gotpl#L51-L91
We also can switch the order of evaluation and check IsMarshaler before Marshaler.

However, I think we set ref.Unmarshaler = nil and ref.Marshaler = nil here are ok.
In current behavior, we create a new ref in each loop and stop here because of incompatible type. 
https://github.com/99designs/gqlgen/blob/master/codegen/config/binder.go#L430
https://github.com/99designs/gqlgen/blob/master/codegen/config/binder.go#L480
I prefer to set ref.Unmarshaler and ref.Marshaler to nil.

StevenACoffman:
@sonatard Can you please take a look at this and see what you think?

sonatard:
This idea is very interesting, and like everyone else, I didn’t understand whyit need to set ref.Unmarshaler = nil and ref.Marshaler = nil after reading it just once.

It would help my understanding to see what kind of code would be generated in the following three versions:

1. Code generated in the current master branch.
2. Code generated in this PR
3. Code generated in this PR if the following settings were not applied:
```
ref.Unmarshaler = nil
ref.Marshaler = nil
ref.IsContext = true
ref.IsMarshaler = true
ref.IsContext = false
ref.IsMarshaler = false
```

markya0616:
> I didn’t understand why it need to set ref.Unmarshaler = nil and ref.Marshaler = nil after reading it just once.

The Unmarshaler/Marshaler may be not nil because it may be set [here](https://github.com/99designs/gqlgen/pull/3632/files#diff-ab508c289540d69c88d9f1146004bbb03b2fbefa9b9ce932e044927fea85aeccR447-R451).
In [codegen](https://github.com/99designs/gqlgen/blob/master/codegen/type.gotpl#L51-L91), it check Unmarshaler before IsMarshaler.

[case 1: Code generated in the current master branch.](https://github.com/markya0616/gqlgen/commit/18ac05cc4c6e692a6566f0dcd8f6146057b6372a#diff-e25937fb3596da32de14a5e55d9ff00861d59d6c29062ae36d3ce1587bebb911R43)
It generate a new [interface](https://github.com/markya0616/gqlgen/commit/18ac05cc4c6e692a6566f0dcd8f6146057b6372a#diff-e25937fb3596da32de14a5e55d9ff00861d59d6c29062ae36d3ce1587bebb911R43-R78)

[case 2: Code generated in this PR](https://github.com/99designs/gqlgen/pull/3632/commits/9368a23da473a255388b964a2cbf036b484d8777)
It use the [UnmarshalGQLContext](https://github.com/99designs/gqlgen/pull/3632/commits/9368a23da473a255388b964a2cbf036b484d8777#diff-e25937fb3596da32de14a5e55d9ff00861d59d6c29062ae36d3ce1587bebb911R3712).

[case 3: Code generated in this PR, if the following settings were not applied](https://github.com/markya0616/gqlgen/tree/support-bind-types-from-basic-types-to-named-type-case-3)
In this case, we try not set [Unmarshaler/Marshaler to nil](https://github.com/markya0616/gqlgen/commit/8a5de4f510bdf8f74a907c0569dfe4fc69d16a15).
It use [UnmarshalInt](https://github.com/markya0616/gqlgen/commit/b1e99d4fce6619b001b3f1cdc2219a3a0892e8b3#diff-e25937fb3596da32de14a5e55d9ff00861d59d6c29062ae36d3ce1587bebb911R3711) and this case is failed to build.

sonatard:
I really appreciate that! The provided code has made the scope of the changes clear.
Now, I’d like to take some time to read and understand the code to ensure there are no issues. If you have any comments that could help with understanding the code, I’d appreciate it if you could add them to the relevant code sections in the Pull Request comments.

sonatard:
- Case1: Code generated in the current master branch.
The issue with case1 is that `unmarshalNInt2githubᚗcomᚋ99designsᚋgqlgenᚋ_examplesᚋtodoᚐNumber` is not being generated.

スクリーンショット 2025-04-16 11 22 07

スクリーンショット 2025-04-16 11 18 51

- Case2: Code generated in this PR
Case2 has no issues.`unmarshalNInt2githubᚗcomᚋ99designsᚋgqlgenᚋ_examplesᚋtodoᚐNumber` is being generated, and `UnmarshalGQLContext `is executed within `unmarshalNInt2githubᚗcomᚋ99designsᚋgqlgenᚋ_examplesᚋtodoᚐNumber`.

スクリーンショット 2025-04-16 11 22 57
スクリーンショット 2025-04-16 11 19 44

- Case3: Code generated in this PR if the following settings were not applied:
In case3, `unmarshalNInt2githubᚗcomᚋ99designsᚋgqlgenᚋ_examplesᚋtodoᚐNumber` is being generated, but `graphql.UnmarshalInt(ctx, v)` is executed.

スクリーンショット 2025-04-16 11 23 52
スクリーンショット 2025-04-16 11 20 50

StevenACoffman:
@sonatard If I understand you last comment, it seems like you have some issues you want to be addressed in Case 1 and Case 2, is that still to be done by @markya0616 ?

sonatard:
My last comment summarized markya0616's verification results, so there’s nothing specific for him to do.

StevenACoffman:
@sonatard  Thank you! Please let me know when you have finished your review. I'm interested in your opinion on this.

codegen/config/binder.go
```diff
@@ -481,7 +481,20 @@ func (b *Binder) TypeReference(schemaType *ast.Type, bindTarget types.Type) (ret
...
+					ref.Marshaler = nil
+					ref.Unmarshaler = nil
+				} else if hasMethod(bindTarget, "MarshalGQL") && hasMethod(bindTarget, "UnmarshalGQL") {
+					ref.IsContext = false
+					ref.IsMarshaler = true
+					ref.Marshaler = nil
+					ref.Unmarshaler = nil
+				} else {
+					continue
+				}
```
sonatard, line 497: @markya0616 Why is this change necessary only when code.CompatibleTypes(ref.GO, bindTarget) results in an error?

markya0616: This is the smallest change I can think of for now.
The original issue was that an incompatible type would generate an interface that needs to be implemented. (Case 1)
Now, I just want to provide an alternative way to handle those incompatible types.

However, I also think it doesn't necessarily have to be limited to only incompatible types.
It could also be rewritten like this:

```
if hasMethod(bindTarget, "MarshalGQLContext") && hasMethod(bindTarget, "UnmarshalGQLContext") {
	ref.IsContext = true
	ref.IsMarshaler = true
	ref.Marshaler = nil
	ref.Unmarshaler = nil
} else if hasMethod(bindTarget, "MarshalGQL") && hasMethod(bindTarget, "UnmarshalGQL") {
	ref.IsContext = false
	ref.IsMarshaler = true
	ref.Marshaler = nil
	ref.Unmarshaler = nil
} else if err = code.CompatibleTypes(ref.GO, bindTarget); err != nil {
	continue
}
```

sonatard: This is not a request for changes, but a pure question: Is it possible to simplify the rules and handle it by modifying the following conditional branching?



```go
		if values := b.enumValues(def); len(values) > 0 {
			err = b.enumReference(ref, obj, values)
			if err != nil {
				return nil, err
			}
		} else if fun, isFunc := obj.(*types.Func); isFunc {
			ref.GO = code.Unalias(t.(*types.Signature).Params().At(0).Type())
			ref.IsContext = code.Unalias(t.(*types.Signature).Results().At(0).Type()).String() == "github.com/99designs/gqlgen/graphql.ContextMarshaler"
			ref.Marshaler = fun
			ref.Unmarshaler = types.NewFunc(0, fun.Pkg(), "Unmarshal"+typeName, nil)
		} else if hasMethod(t, "MarshalGQLContext") && hasMethod(t, "UnmarshalGQLContext") {
			ref.GO = t
			ref.IsContext = true
			ref.IsMarshaler = true
		} else if hasMethod(t, "MarshalGQL") && hasMethod(t, "UnmarshalGQL") {
			ref.GO = t
			ref.IsMarshaler = true
		} else if hasMethod(t, "MarshalJSON") && hasMethod(t, "UnmarshalJSON") {
			ref.GO = t
			ref.IsJSONMarshaler = true
		} else if underlying := basicUnderlying(t); def.IsLeafType() && underlying != nil && underlying.Kind() == types.String {
			// TODO delete before v1. Backwards compatibility case for named types wrapping strings (see #595)

			ref.GO = t
			ref.CastType = underlying

			underlyingRef, err := b.TypeReference(&ast.Type{NamedType: "String"}, nil)
			if err != nil {
				return nil, err
			}

			ref.Marshaler = underlyingRef.Marshaler
			ref.Unmarshaler = underlyingRef.Unmarshaler
		} else {
			ref.GO = t
		}
```

sonatard: Thank you. I understood that the change is simple.

Since the TypeReference function is very difficult to understand, I thought it would be good to add unit tests to help with that. Please add tests with reference to TestEnumBinding in binder_test.go.

markya0616: Add some tests in [75643dd](https://github.com/99designs/gqlgen/pull/3632/commits/75643dd1ac10b8f2505b1333d067cfe7d1e7b78b)

StevenACoffman:
@sonatard I think we are good to merge this as is now, do you agree, or do you think the test should be changed?

sonatard:
I was thinking about adding missing unit tests for TypeReference to confirm that the test results don’t change with this PR’s fixes. However, I can only add those tests over the weekend, so I think it’s fine to merge this PR first.

PR #3638, opened 2025-04-04
Gracefully close websocket connections

### LINKED ISSUE #3633: Graceful Websocket Connection Shutdown
Hi! How do we currently gracefully shutdown websocket connections and is it actually graceful? 

This has been brought up before but does not seem like the discussion is complete: https://github.com/99designs/gqlgen/discussions/2847. The solution there implies canceling outstanding request contexts when the server context shuts down in `InitFunc` is enough and `gqlgen` should just handle the Websocket shutdown rigamarole for us. Seems sensible.

Using this approach (or not -- this solution seems to have no effect which suggests to me something is up in gqlgen) I see tons of these errors in my `ErrorFunc` on deploy:

```
websocket write: websocket: close 1006 (abnormal closure): unexpected EOF
websocket read: websocket connection closed
websocket write: websocket: close sent
```

My `InitFunc` (see below for full code) looks something like:
```go
InitFunc: func(reqCtx context.Context, _ transport.InitPayload) (context.Context, *transport.InitPayload, error) {
	liveWebsocketConnections.Add(1)

	newReqCtx, initFuncCancel := context.WithCancel(reqCtx)

	go func() {
		select {
		case <-serverCtx.Done():
			fmt.Printf("Server shutting down. Gracefully shutting down WS based request. Outstanding WS connections %d", liveWebsocketConnections.Load())
			initFuncCancel()
		case <-newReqCtx.Done():
			return
		}
	}()

	return newReqCtx, nil, nil
},
```
I see those error messages immediately after my `Server shutting down` log.

Looking at gqlgen's code, I should hit `closeOnCancel` then `c.close(...)` on server shutdown
```go
func (c *wsConnection) closeOnCancel(ctx context.Context) {
	<-ctx.Done()

	if r := closeReasonForContext(ctx); r != "" {
		c.sendConnectionError("%s", r)
	}
	c.close(websocket.CloseNormalClosure, "terminated")
}

...

func (c *wsConnection) close(closeCode int, message string) {
	c.mu.Lock()
	if c.closed {
		c.mu.Unlock()
		return
	}
	_ = c.conn.WriteMessage(websocket.CloseMessage, websocket.FormatCloseMessage(closeCode, message))
	for _, closer := range c.active {
		closer()
	}
	c.closed = true
	c.mu.Unlock()
	_ = c.conn.Close()

	if c.CloseFunc != nil {
		c.CloseFunc(c.ctx, closeCode)
	}
}
```
 
1. **Why don't we wait for a closing frame from the client before closing `c.conn`?** Isn't that part of the [Websocket RFC](https://datatracker.ietf.org/doc/html/rfc6455#page-9)?
2. **Where do we prevent in-flight writes and reads (besides in the close func which is where we should read the client's response) from happening after we've initiated shutdown but before shutdown is complete? Should we?** It doesn't seem like we do, which I'm guessing is partially what explains the 1006's. 

These 1006 errors ultimately show up as `Protocol(ResetWithoutClosingHandshake)` in my downstream Rust proxy. I don't think these errors should always be retried (see https://github.com/snapview/tokio-tungstenite/issues/101), so it makes it hard to know what to do even further downstream on the client. These errors are nearly 100% correlated with my deployments and pods being naturally cycled out and not reproducible in any other way.

It would be nice if `gorilla/websocket` provided an API for this like `net/http` does for shutting down non-Websocket connections, but it seems like it does not and will not (https://github.com/gorilla/websocket/issues/448). `gqlgen` seems like the place we should handle gracefully closing websocket connections, no? It seems like others in the community have started to work around this issue (https://github.com/ethereum/go-ethereum/issues/30482#issuecomment-2389142357), which is really not ideal.

My full code:
```go
transport.Websocket{
	Upgrader: websocket.Upgrader{
		CheckOrigin: func(r *http.Request) bool {
			return true
		},
		Error: func(w http.ResponseWriter, r *http.Request, status int, reason error) {
			_, _ = fmt.Fprintf(w, "Websocket upgrade failed: %s with status %v", reason, status)
			http.Error(w, http.StatusText(status), status)
		},
	},
	InitFunc: func(reqCtx context.Context, _ transport.InitPayload) (context.Context, *transport.InitPayload, error) {
		liveWebsocketConnections.Add(1)

		newReqCtx, initFuncCancel := context.WithCancel(reqCtx)

		go func() {
			select {
			case <-serverCtx.Done():
				fmt.Printf("Server shutting down. Gracefully shutting down WS based request. Outstanding WS connections %d", liveWebsocketConnections.Load())
				initFuncCancel()
			case <-newReqCtx.Done():
				return
			}
		}()

		return newReqCtx, nil, nil
	},
	PingPongInterval: 1500 * time.Millisecond,
	ErrorFunc: func(ctx context.Context, err error) {
		_, _ = fmt.Println("Websocket error: ", err)
	},
	CloseFunc: func(ctx context.Context, closeCode int) {
		liveWebsocketConnections.Add(-1)
		fmt.Printf("Gracefully shut down WS based request. Remaining WS connections %d", liveWebsocketConnections.Load())

		if closeCode == websocket.CloseNormalClosure {
			return
		}
		_, _ = fmt.Println("Websocket closed with close code: ", closeCode)
	},
},
```

john-markham:
Fixes #3633 as follow-up to discussion #2847

This is intended to gracefully close WebSocket connections, because the Gorilla/websocket doesn't provide us a nice way to do that.

I have:

### DIFF
--- a/graphql/handler/transport/websocket.go
+++ b/graphql/handler/transport/websocket.go
@@ -238,13 +238,12 @@ func (c *wsConnection) init() bool {
 
 func (c *wsConnection) write(msg *message) {
 	c.mu.Lock()
+	defer c.mu.Unlock()
 	// don't write anything to a closed / closing connection
 	if c.serverClosed {
-		c.mu.Unlock()
 		return
 	}
 	c.handlePossibleError(c.me.Send(msg), false)
-	c.mu.Unlock()
 }
 
 func (c *wsConnection) run() {
--- a/graphql/handler/transport/websocket.go
+++ b/graphql/handler/transport/websocket.go
@@ -563,6 +563,8 @@ func (c *wsConnection) close(closeCode int, message string) {
 			// either we get net.ErrClosed or some other error
 			// either way, bail on the graceful shutdown as it's either
 			// impossible or very likely not happening
+			// TODO: optimize this to bypass the select statement to avoid
+			// waiting the entire closeTimeout
 			if err != nil {
 				return
 			}
--- a/graphql/handler/transport/websocket.go
+++ b/graphql/handler/transport/websocket.go
@@ -332,9 +332,12 @@ func (c *wsConnection) run() {
 		case connectionCloseMessageType:
 			c.mu.Lock()
 			c.clientClosed = true
-			// normal termination
+			// server already initiated the graceful shutdown
+			// we don't need to send another close message
 			if c.serverClosed {
 				c.clientCloseReceiver <- struct{}{}
+				c.mu.Unlock()
+				return
 			}
 			c.mu.Unlock()
 			c.close(websocket.CloseNormalClosure, "terminated")
--- a/graphql/handler/transport/websocket.go
+++ b/graphql/handler/transport/websocket.go
@@ -53,8 +53,14 @@ type (
 		pingPongTicker  *time.Ticker
 		receivedPong    bool
 		exec            graphql.GraphExecutor
-		closed          bool
 		headers         http.Header
+
+		closeTimeout time.Duration
+
+		serverClosed bool
+		clientClosed bool
+
+		clientCloseReceiver chan struct{}
 
 		initPayload InitPayload
 	}
@@ -115,13 +121,14 @@ func (t Websocket) Do(w http.ResponseWriter, r *http.Request, exec graphql.Graph
 	}
 
 	conn := wsConnection{
-		active:    map[string]context.CancelFunc{},
-		conn:      ws,
-		ctx:       r.Context(),
-		exec:      exec,
-		me:        me,
-		headers:   r.Header,
-		Websocket: t,
+		active:              map[string]context.CancelFunc{},
+		conn:                ws,
+		ctx:                 r.Context(),
+		exec:                exec,
+		me:                  me,
+		headers:             r.Header,
+		Websocket:           t,
+		clientCloseReceiver: make(chan struct{}),
 	}
 
 	if !conn.init() {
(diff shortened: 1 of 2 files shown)


graphql/handler/transport/websocket.go
```diff
@@ -283,6 +295,16 @@ func (c *wsConnection) run() {
 	go c.closeOnCancel(ctx)
 
 	for {
+		c.mu.Lock()
```
UnAfraid, line 298: This will have performance implications

graphql/handler/transport/websocket.go
```diff
@@ -231,6 +238,11 @@ func (c *wsConnection) init() bool {
 
 func (c *wsConnection) write(msg *message) {
 	c.mu.Lock()
+	// don't write anything to a closed / closing connection
+	if c.serverClosed {
+		c.mu.Unlock()
```
UnAfraid, line 243: I would move that as defer c.mu.Unlock() right after the lock

graphql/handler/transport/websocket.go
john-markham: Yeah totally. I'm open to suggestions as to how we should measure how much this impacts performance as this is a concern for me and my team as well. 

Overall though - if the thesis of this PR is correct (that we are not gracefully handling shutdown), I'm inclined to trade perf off to get us closer to WS spec compliance.

john-markham: Ideally we have a solution that does both: doesn't impact perf and gracefully shuts down the connection. Again open to suggestions on how to refactor the implementation to do both.

UnAfraid: There are multiple ways to do that that would not require mutext in a loop, also i would check for server closed after c.me.NextMessage() call, because it would block on that until next message comes, so you probably wanna handle it as soon as you get any message and not on the next iteration

john-markham: Like what?

PR #3639, opened 2025-04-04
Fix: Apollo FTv1 tracing doesn't wait for field resolver completion

kolia-kaploniuk:
This PR fixes an issue in Apollo Federation Tracing (FTv1) where the `EndTime` of a field span was recorded too early — before the field resolver had actually completed execution. This resulted in inaccurate tracing data, as the span duration didn't reflect the true time spent resolving the field.

### DIFF
--- a/graphql/handler/apollofederatedtracingv1/tracing.go
+++ b/graphql/handler/apollofederatedtracingv1/tracing.go
@@ -94,11 +94,18 @@ func (t *Tracer) InterceptField(ctx context.Context, next graphql.Resolver) (any
 	if !t.shouldTrace(ctx) {
 		return next(ctx)
 	}
+
+	var stop func()
 	if tb := t.getTreeBuilder(ctx); tb != nil {
-		tb.WillResolveField(ctx)
+		stop = tb.WillResolveField(ctx)
 	}
 
-	return next(ctx)
+	res, err := next(ctx)
+	if stop != nil {
+		stop()
+	}
+
+	return res, err
 }
 
 // InterceptResponse is called before the overall response is sent, but before each field resolves; as a result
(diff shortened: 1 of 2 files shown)


graphql/handler/apollofederatedtracingv1/tree_builder.go
```diff
@@ -114,25 +114,25 @@ func (tb *TreeBuilder) StopTimer(ctx context.Context) {
 
 // On each field, it calculates the time started at as now - tree.StartTime, as well as a deferred function upon full resolution of the
 // field as now - tree.StartTime; these are used by Apollo to calculate how fields are being resolved in the AST
```
StevenACoffman, line 116:
```suggestion
// field as (now - tree.StartTime); these are used by Apollo to calculate how fields are being resolved in the AST
```

graphql/handler/apollofederatedtracingv1/tracing.go
```diff
@@ -94,11 +94,18 @@ func (t *Tracer) InterceptField(ctx context.Context, next graphql.Resolver) (any
 	if !t.shouldTrace(ctx) {
 		return next(ctx)
 	}
+
+	var stop func()
 	if tb := t.getTreeBuilder(ctx); tb != nil {
-		tb.WillResolveField(ctx)
+		stop = tb.WillResolveField(ctx)
```
StevenACoffman, line 100: I think `stop()` needs to be called in a defer if it is non-nil here.

StevenACoffman, line 100:
```suggestion
		stop = tb.WillResolveField(ctx)
		if stop != nil {
		  defer stop()
		}
```

graphql/handler/apollofederatedtracingv1/tracing.go
```diff
@@ -94,11 +94,18 @@ func (t *Tracer) InterceptField(ctx context.Context, next graphql.Resolver) (any
...
+	}
+
+	res, err := next(ctx)
+	if stop != nil {
+		stop()
 	}
 
-	return next(ctx)
+	return res, err
```
StevenACoffman, line 108: This should just be `return next(ctx)` because of the `defer stop()` above

daviddavid533:
can we merge this asap? @StevenACoffman

StevenACoffman:
Thanks for the ping! I hadn't noticed you had corrected updated it since the defer comment.

Issue #3586, opened 2025-03-11
Generate groups and reorders imports incorrectly


mdakovac:
gqlgen version: v0.17.66
go version: 1.23.3

full repro: https://github.com/mdakovac/gqlgen-import-sorting/tree/main

Go lets you group imports by separating them by a newline, however when you run the `generate` command, this grouping is broken.

StevenACoffman:
Which file is getting altered in this way?

mdakovac:
`graph/schema.resolvers.go` is the file, added to description

StevenACoffman:
Thanks! not sure why that is happening. If you care to make a PR to remove the extra newline, I would appreciate it.

PR #3646, opened 2025-04-07
add config local_prefix for imports group split

light-white:
#3586 

I think this question is abort "golang.org/x/tools/imports" LocalPrefix

add config local_prefix and use it in api.Generate before all generate plugins

### DIFF
--- a/api/generate.go
+++ b/api/generate.go
@@ -5,6 +5,8 @@ import (
 	"regexp"
 	"syscall"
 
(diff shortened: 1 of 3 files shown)


StevenACoffman:
This is great! Can you add this to the docs here https://github.com/99designs/gqlgen/blob/master/docs/content/config.md too?

StevenACoffman:
Awesome! This is great! Thanks alot!

PR #3653, opened 2025-04-07
Add websocket shutdown grace period

john-markham:
Simpler approach to graceful shut down than https://github.com/99designs/gqlgen/pull/3638. Got this idea from following the discussion around https://github.com/gorilla/websocket/issues/248

Relying explicitly on the handshake complicates the implementation [past what I think is worth it](https://github.com/99designs/gqlgen/pull/3638/files#r2029408714) if alternatives like this could get us 80% of the way there. 

I'm still validating that hypothesis. Work to address https://github.com/99designs/gqlgen/issues/3633

I have:

### DIFF
--- a/graphql/handler/transport/websocket.go
+++ b/graphql/handler/transport/websocket.go
@@ -56,6 +56,8 @@ type (
 		closed          bool
 		headers         http.Header
 
+		shutdownGracePeriod time.Duration
+
 		initPayload InitPayload
 	}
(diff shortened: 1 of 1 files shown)


StevenACoffman:
@john-markham Hey, I'm just curious what the status of this is. There was a flurry of other websocket activity, and I got confused where this PR fits with all that.

Issue #3227, opened 2024-08-23
Support for Custom Omittable Type


Dan6erbond:
### What happened?

I saw that #2585 addresses the problem with `undefined` vs `null` in input types, and it's already a great solution for smaller APIs that need to check if a field was explicitly set as `nil` or not defined at all.

However, in my project we have a lot of DTOs and use `github.com/moznion/go-optional.Option` to represent our optional types internally. This allows us to decouple our service methods from the GQL layer.

For smaller projects it would be fine to manually map `graphql.Omittable` to `Option` with a generic function like this:

```go
func MapToOption[T any](o graphql.Omittable[T]) optional.Option[T] {
	if o.IsSet() {
		return optional.Some(o.Value())
	}
	return optional.None[T]()
}
```

However, for larger projects it would be nice to use the Modelgen hooks to replace all the `Omittable` with `Option` and a custom unmarshaler/marshaler for that type, but as GQLGen doesn't support this because `nil` values never hit the unmarshaler, I was wondering if an extension to #2585 would be possible with a custom "Omittable" type, in this case `Option`?

The simplest way to allow this would be via a mapper, that users can provide in the config for each `Omittable` type that we want to convert:

```yml
omittableTypes:
  *string: mypkg.UnmarshalStringOption
```

### What did you expect?

My proposal above or some other way to convert `Omittable` to `Option`. Entirely replacing `Omittable` with `Option` would be the most elegant.

### Minimal graphql.schema and models to reproduce

```gql
input UpdateTodoInput {
  text: String @goField(omittable: true, omittableType: "github.com/moznion/go-optional.Option") # alternative to providing omittable types in the config
}
```

### versions
 - `go run github.com/99designs/gqlgen version`?
 - `go version`?

StevenACoffman:
By the way, I would welcome a PR for this!

deitrix:
I propose foregoing the use of `graphql.Omittable` in unmarshallers. This potentially removes a library dependency where it may not be needed, and makes the use of generics optional too. For concrete types that are omittable, the function signature could be
```go
// play.ground/types
package types

import "play.ground/pkg/null"

func UnmarshalOptionalString(s *string, isSet bool) null.String {
   ...
}
```

And then for the generic case
```go
// play.ground/types
package types

import "play.ground/pkg/optional"

func UnmarshalOptional[T any](v *T, isSet bool) optional.Of[T] {
   ...
}
```

[...]

PR #2839, opened 2023-11-07
Omittable can now be serialized as json

endSly:
The `Omittable` struct lacks the necessary JSON unmarshaller and marshaller implementations. As a result, instances of this struct are serialized as empty JSON objects (`{}`) even when they contain data.

For instance, consider the following Go struct:
```go
struct {
	ID     graphql.Omittable[*string]     `json:"id,omitempty"`
	Bool   graphql.Omittable[*bool]       `json:"bool,omitempty"`
	Str    graphql.Omittable[*string]     `json:"str,omitempty"`
}
```

This struct is being serialized as:
```json
{
  "id": {},
  "bool": {},
  "str": {},
}
```

I have:
 - [x] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [x] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

PR #3660, opened 2025-04-13
Omittable supports IsZero method for omitzero

sonatard:
This PR adds IsZero to Omittable. It also adds support for MarshalGQL and MarshalGQLContext, which are necessary for correctly parsing GraphQL enums.

## Background
API may behave differently when JSON is undefined versus when it is null.

- undefined
  - no effect to name
  - json: `{}`

```go
input := UserUpdateInput{
		Name:         graphql.Omittable[string]{},
}
```

- null
  - set empty string to  name
  - json: `{"name":null}`

```go
input := UserUpdateInput{
		Name:         graphql.OmittableOf[string](null),
}
```

However, the current Omittable does not omit fields when json.Marshal is executed with set=false.
Additionally, omitempty could not specify whether the field’s type itself should be omitted. However, with the introduction of omitzero and the IsZero method in Go 1.24, it is now possible to express whether a field should be omitted.

## Reference
- https://github.com/99designs/gqlgen/issues/1416
- https://github.com/99designs/gqlgen/pull/2585
- https://github.com/99designs/gqlgen/pull/2839
- https://github.com/99designs/gqlgen/issues/3227

I have:

### DIFF
--- a/_examples/go.mod
+++ b/_examples/go.mod
@@ -1,8 +1,8 @@
 module github.com/99designs/gqlgen/_examples
 
-go 1.23.0
+go 1.23.8
 
-toolchain go1.23.7
+toolchain go1.24.2
 
 replace github.com/99designs/gqlgen => ../
 
--- a/_examples/websocket-initfunc/server/go.mod
+++ b/_examples/websocket-initfunc/server/go.mod
@@ -1,6 +1,8 @@
 module github.com/gqlgen/_examples/websocket-initfunc/server
 
-go 1.23.6
+go 1.23.8
+
+toolchain go1.24.2
 
 require (
 	github.com/99designs/gqlgen v0.17.70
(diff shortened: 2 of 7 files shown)


StevenACoffman:
This is cool, and I always love to see your contributions!

Unfortunately, we cannot drop support for Go 1.23 yet. Our policy is the last 2 versions of Go (current release and the last one). This is the same thing we do in [ent](https://github.com/ent/ent), and we do it because we found that it takes some time for some companies to upgrade the Go version.

This also matches Go's own [security policy](https://go.dev/doc/security/) where each major Go release is supported until there are two newer major releases.

Selfishly, at work, we use Google AppEngine, and we use gqlgen and they currently only support Go 1.23.
https://cloud.google.com/appengine/docs/standard/go/release-notes

sonatard:
@StevenACoffman Thank you for the review!
I’ve made changes to set the build tag so it works with Go 1.23 as well.

PR #3663, opened 2025-04-14
Add json.Marshaler

### LINKED ISSUE #3661: Proposal: Generate MarshalJSON Method
Currently, [gqlgenc](https://github.com/Yamashou/gqlgenc) implements a [custom Marshal function](https://github.com/Yamashou/gqlgenc/blob/12e712934d85d6b660d7f9fe0a7b635b5bf6beb9/clientv2/client.go#L490) instead of using [json.Marshal](https://pkg.go.dev/encoding/json#Marshal). This is because, to correctly send enum values as JSON, gqlgenc needs to call MarshalGQL.
Therefore, I propose that gqlgen generate [MarshalJSON](https://pkg.go.dev/encoding/json#Marshaler)

- `generate_json_marshaler: true`
```go
func (e Object) MarshalGQL(w io.Writer) {
    fmt.Fprint(w, strconv.Quote(e.String()))
}

func (e *Object) UnmarshalJSON(b []byte) error {
       s, err := strconv.Unquote(string(b))
       if err != nil {
               return err
       }
       return e.UnmarshalGQL(s)
}

func (e Object) MarshalJSON() ([]byte, error) {
       var buf bytes.Buffer
       e.MarshalGQL(&buf)
       return buf.Bytes(), nil
}
```

With the upcoming JSON v2 package, maintaining a custom Marshaler to keep up with v2 introduces the risk of implementation differences.
I would appreciate your advice on this. If it seems like a good idea, I will implement it and create a Pull Request.

sonatard:
Fix #3661

I have:


skaji:
I think this change will likely cause an error if the enum already implements MarshalJSON/UnmarshalJSON.
https://github.com/skaji/gqlgen-v0.17.71-issue

sonatard:
@skaji Thank you for reporting the bug.

@StevenACoffman I fixed it in this PR, so please review it.
https://github.com/99designs/gqlgen/pull/3675

ysdanielkim:
Hi, I'm not seeing how https://github.com/99designs/gqlgen/pull/3675 addresses "only provide this behavior only if the enum does not already implement MarshalJSON/UnmarshalJSON". Thankfully, upgrading to this version revealed a bug we had thanks to the GQL unmarshaler checking for valid enum values. However, I need to able to override the default JSON unmarshaller to fix the bug.

jarreds:
> Can either of you make a PR (other than just a revert) that would be able to only provide this behavior only if the enum does not already implement MarshalJSON/UnmarshalJSON?

@StevenACoffman also noting here that https://github.com/99designs/gqlgen/pull/3675 did not fix this functionality. We're trying to upgrade our gqlgen version and have a bunch of custom JSON marshallers that are now throwing `already declared` errors.

StevenACoffman:
@jarreds @ysdanielkim
Does #3940 and the new configuration option `omit_enum_json_marshalers: true` fully disable this behavior and meet your needs?

Issue #3662, opened 2025-04-13
Unaliasing type behavior breaks usage of type alias to export internal


ChrisRx:
I have a federated supergraph made up of subgraphs that exist in different Go modules. Some of these subgraphs define models in packages like `internal/store/example` so to provide them to other Go modules I was attempting to use a type aliases in non-internal packages:

```go
// subgraph A
package model

import (
	"...internal/store/example"
)

type ExampleItem  = example.Item
```

which then gets used in another subgraph's schema like:

```graphql
# subgraph B
extend type ExampleItem @key(fields: "id") @goModel(model: "graph/model.ExampleItem") {
  id: ID!
  updatedAt: Time!
}
```

The issue I have run into seems to revolve around a workaround made in #3471 to help address another issue in #3414.  Since the aliased type gets unaliased `packages.Load` attempts to load the type from the internal package and fails with:

```shell
validation failed: packages.Load: graph/entity.resolvers.go:11:2: use of internal package ... not allowed
```

The first thought I had to address it was to conditionally not unalias when the resulting type will be an internal package, which felt a little dirty to do a workaround on top of a workaround at first, but I also couldn't think of a situation outside of dealing with internal packages where it would make a difference. I'd be more than happy to make a PR for this if that sounds like an acceptable solution (I really love/appreciate this project!).

StevenACoffman:
Thanks for looking at this. Workarounds for workarounds does seem like making a tippy stack of band-aids, so if you have a more penetrating insight, please share it! That said, I am ok with PRs fix things while having good test coverage.

PR #3666, opened 2025-04-14
fix: dont unalias when used to expose internal types

ChrisRx:
This is an attempt at fixing #3662, where the unaliasing behavior prevents using type aliases to expose unexported types. The check attempts to narrow down the types that will skip unaliasing as much as possible to target only that particular use case.

I have:

### DIFF
--- a/internal/code/alias.go
+++ b/internal/code/alias.go
@@ -2,10 +2,23 @@ package code
 
 import (
 	"go/types"
+	"strings"
 )
 
 // Unalias unwraps an alias type
 func Unalias(t types.Type) types.Type {
+	if alias, ok := t.(*types.Alias); ok {
+		// If the type is an alias type, it must first check if the non-alias
+		// type is in an internal package. Only the last type in the alias
+		// chain is provided as the RHS.
+		if isAliasInternal(t.String(), unalias(t).String()) {
+			return types.NewNamed(alias.Obj(), alias.Underlying(), nil)
+		}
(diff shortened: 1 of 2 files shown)


internal/code/alias.go
```diff
@@ -15,3 +28,29 @@ func Unalias(t types.Type) types.Type {
 	}
 	return types.Unalias(t)
 }
+
+// isAliasInternal checks if an alias type path is declared for a type within
+// an internal package. A best-effort attempt is made to mirror the Go internal
```
StevenACoffman, line 33: Can you site your source for the Go internal visibility rules here?

ChrisRx: I'm still trying to see if there is a better source for this, but I translated the logic of the function used by the compiler: [disallowInternal](https://github.com/golang/go/blob/adf21a9b56f3498385152d701ddd0e8679fe57ed/src/cmd/go/internal/load/pkg.go#L1466) (but removing aspects that don't pertain to just the import path, since it considers more context about the packages and file system).

It seems silly but I couldn't find anything in the language spec or Go module reference, and I haven't found anything yet in the official [references docs](https://go.dev/doc/#references). I know it is deliberate behavior, because it couldn't be all us Go developers just hallucinating the existence of `internal` :joy:.

StevenACoffman: That seems probably the best you are going to get!

PR #3679, opened 2025-04-22
fix: lower Go version requirement to 1.23.0

zerok:
This should prevent compatibility issues with tools that are on 1.23 but not on the latest patch release there.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/99designs/gqlgen
 
-go 1.23.8
+go 1.23.0
 
 require (
 	github.com/PuerkitoBio/goquery v1.10.3


StevenACoffman:
I'm not really entirely sure I understand how this works. I thought that the minor version of the Go directive would not have caused an issue.

zerok:
@StevenACoffman This also surprised me but perhaps it somehow different when explicitly setting a toolchain. Unfortunately, I didn't have time to set up a complete test-suite around this 🙁 Thanks for merging this! Hopefully this will get us unstuck soonish 😄

StevenACoffman:
Can you circle back if you get more information? I worry I am downgrading people to Go 1.23.0 from later minor versions.
I have read over https://go.dev/doc/toolchain and am not sure why the problem you report would have happened.

zerok:
Absolutely!

Just for the record, that's the setup that we have with Dagger right now:

- dagger/dagger uses `go 1.23.0` and `toolchain go1.23.6`
- gqlgen used `go 1.23.8`

My interpretation of the spec is that setting `go 1.23.0` forces users to have a toolchain >= `1.23.0` when compiling the code. Setting the toolchain to `go1.23.6` enforce a minimum version of the compiler. Setting `go ...` should never downgrade a toolchain from what I've understood.

One thing that I've noticed which might explain the whole cascade-of-conflicts here:

When I add v0.17.72 as a dependency in a go.mod file and run `go mod tidy`, it will update the `go ...` line in *my* go.mod file to `go 1.23.8` which would conflict with the toolchain defined by Dagger.

https://github.com/dagger/dagger/issues/6706#issue-2147149550

StevenACoffman:
I really need a test case that detects Go version directive changes and enforces the security policy with an actionable resolution to prevent this from just happening again and again.

Apparently this is what's necessary to fix the `go.mod` up:
```
go get go@1.23.0
go get toolchain@none
go mod tidy
```
So then a script that runs that and checks whether git is dirty or something.

Also, probably good to add `export GOTOOLCHAIN=go1.23.0` in all my CI stuff.

lionelfleury:
Go directive reference
https://go.dev/ref/mod#go-mod-file-go
It's really about minimal required version

PR #3684, opened 2025-04-24
Assign unused parameters to placeholder to avoid too many warning in IDE

### LINKED ISSUE #3348: Unused Variable in generated.go
### What happened?
Afer running `gqlgen generate`

In VSCode using gopls, I got this warning that display
```md
unused parameter: sel
```
Screenshot 2024-10-29 at 08 58 42

That is caused by `sel` param not used
```go
// generated.go

...
func (ec *executionContext) marshalNBoolean2bool(
	ctx context.Context,
	sel ast.SelectionSet, // <= this part
	v bool,
) graphql.Marshaler {
	res := graphql.MarshalBoolean(v)
	if res == graphql.Null {
		if !graphql.HasFieldError(ctx, graphql.GetFieldContext(ctx)) {
			ec.Errorf(ctx, "the requested element is null which the schema does not allow")
		}
	}
	return res
}
...
```

### What did you expect?
I need the VSCode warning to dissappear, it's annoying and messup development experience.
The variable should be used or removed, but I can't change that because it is an auto generated file.
But I can't exclude that from `gopls` either. I have tried. To elaborate, I opened a question about it on StackOverflow:
https://stackoverflow.com/questions/79135399/exclude-directory-or-file-in-gopls

```graphqls
# GraphQL schema example
#
# https://gqlgen.com/getting-started/

type Query {
  character(id: ID!): Character
  pogues: [Character]!
  kooks: [Character]!
}

type Mutation {
  upsertCharacter(input: CharacterInput!): Character!
}

type Character {
  id: ID!
  name: String!
}

input CharacterInput {
  name: String!
  id: String
}
```

### versions
 - gqlgen version: v0.17.55
 - go version go1.23.2 darwin/arm64

git-hulk:
This closes #3348.

I have:

### DIFF
--- a/_examples/chat/generated.go
+++ b/_examples/chat/generated.go
@@ -3912,6 +3912,7 @@ func (ec *executionContext) unmarshalNBoolean2bool(ctx context.Context, v any) (
 }
 
 func (ec *executionContext) marshalNBoolean2bool(ctx context.Context, sel ast.SelectionSet, v bool) graphql.Marshaler {
+	_ = sel
 	res := graphql.MarshalBoolean(v)
 	if res == graphql.Null {
 		if !graphql.HasFieldError(ctx, graphql.GetFieldContext(ctx)) {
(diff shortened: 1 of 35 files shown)


git-hulk:
The first commit https://github.com/99designs/gqlgen/pull/3684/commits/d02b502882ee843dabd7ab1ea13f1de39df0697d changed `type.gotpl` template and the second commit is generated by `go generate`.

StevenACoffman:
Nice! Thanks for calling out the real change!

Issue #567, opened 2019-03-01
Question: Setting cookies


DouglasHoang:
I was looking at #98 and was wondering if I should set cookies through a resolver or a handler.

mathewbyrne:
If you want to access cookie information from a resolver, then follow the advice given in #98 and use a mmiddleware to inject a value into context.

saschajullmann:
I think this question was misunderstood. From my perspective it's not about accessing cookie information, it's about where to **set** a cookie. I'm currently struggeling with that myself and don't know how to set the cookie. Is it possible to extend the authentication example on the website with "how to set a cookie"? For example when you have a login mutation and after confirming that the passwords match you want to set a cookie that should be attached to the http response.

LIYINGZHEN:
@saschajullmann Did you find the answer? I think maybe we should set the cookie from the cline. GraphQL only return the token string.

saschajullmann:
@LIYINGZHEN unfortunately no. That's an idea. Although tbh I gave up on this and didn't look into this any further.

DouglasHoang:
@LIYINGZHEN I couldn't find a way to set cookies via a graphql resolver so I just made a separate login route that handled authentication.

LIYINGZHEN:
@DouglasHoang That's probably a way to go. I also want to upload data to GraphQL but look's like gqlgen doesn't support `multipart/form-data`, so I have to create a rest API to do that.

ravilution:
This sucks. There should be a way to do this. I cannot create a separate app to handle login.

mathewbyrne:
Ah right, my mistake I had misread the original request.  That said, I don't think the advice it too different here.  You would probably just be injecting some shared state into context, that a HTTP middleware could read after the query has been executed and set as a cookie.

ravilution:
@mathewbyrne can you elaborate with a sample code? Because inside a resolver, as far as my understanding goes, context is read-only. It cannot be set to a request, because there is no request object present to set.

mathewbyrne:
I don't have time to put an example together right now.  You wont be writing to context, but writing to a struct pointed to by something inside context.  Your HTTP middleware would probably create and keep a pointer to this so that it can read it on the way out.

ravilution:
I understand now. I will try and share the code.

ravilution:
works as you explained. I added the userId pointer from the request into the context in the middleware.
```
userID := "12345"
[...]

PR #3706, opened 2025-05-13
Allows to rewrite `writeJson` to customize headers according to graphql.Response

### LINKED ISSUE #3694: Is it possible to hijack the `writeJson` to add customized HTTP headers
Currently, some of the transport allows adding fixed customized HTTP response headers.
But in our case, we would like to add customized headers based on GraphQL's response.
And as far as I know, it has no way to add dynamic headers(for example: operation/operation name)
unless we could hijack the `writeJson` method.

A possible solution to support this is to export a way to rewrite `writeJson` method, but I'm not sure
if this sounds good:

```Go

type JsonWriter interface {
	WriteJson(*graphql.Response)
}

func writeJson(w http.ResponseWriter, response *graphql.Response) {
	if jsonWriter, ok := w.(JsonWriter); ok {
		jsonWriter.WriteJson(response)
		return
	}

	b, err := json.Marshal(response)
	if err != nil {
		panic(fmt.Errorf("unable to marshal %s: %w", string(response.Data), err))
	}
	w.Write(b)
}
```

I will be happy to submit a PR to support this if the direction is good.

git-hulk:
This closes #3694

Currently, some of the transport allows adding fixed customized HTTP response headers.
But in our case, we would like to add dynamic headers based on GraphQL's response.
And as far as I know, it has no way to add dynamic headers(for example: operation/operation name)
unless we could hijack the `writeJson` method.

This PR exports a way to rewrite the `writeJson` method:

```Go

type JsonWriter interface {
	WriteJson(*graphql.Response)
}

func writeJson(w http.ResponseWriter, response *graphql.Response) {
	if jsonWriter, ok := w.(JsonWriter); ok {
		jsonWriter.WriteJson(response)
		return
	}

	b, err := json.Marshal(response)
	if err != nil {
		panic(fmt.Errorf("unable to marshal %s: %w", string(response.Data), err))
	}
	w.Write(b)
}
```

I have:

### DIFF
--- a/graphql/handler/transport/util.go
+++ b/graphql/handler/transport/util.go
@@ -3,29 +3,38 @@ package transport
 import (
 	"encoding/json"
 	"fmt"
-	"io"
+	"net/http"
 
 	"github.com/vektah/gqlparser/v2/gqlerror"
 
 	"github.com/99designs/gqlgen/graphql"
 )
 
-func writeJson(w io.Writer, response *graphql.Response) {
+type JsonWriter interface {
+	WriteJson(*graphql.Response)
+}
+
+func writeJson(w http.ResponseWriter, response *graphql.Response) {
+	if jsonWriter, ok := w.(JsonWriter); ok {
+		jsonWriter.WriteJson(response)
+		return
+	}
+
 	b, err := json.Marshal(response)
 	if err != nil {
 		panic(fmt.Errorf("unable to marshal %s: %w", string(response.Data), err))
 	}
-	w.Write(b)
+	_, _ = w.Write(b)
 }
 
-func writeJsonError(w io.Writer, msg string) {
+func writeJsonError(w http.ResponseWriter, msg string) {
 	writeJson(w, &graphql.Response{Errors: gqlerror.List{{Message: msg}}})
 }
 
-func writeJsonErrorf(w io.Writer, format string, args ...any) {
+func writeJsonErrorf(w http.ResponseWriter, format string, args ...any) {
 	writeJson(w, &graphql.Response{Errors: gqlerror.List{{Message: fmt.Sprintf(format, args...)}}})
 }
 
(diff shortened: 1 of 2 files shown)


StevenACoffman:
@git-hulk I'm very sorry I've been consumed at work and in my personal life and not able to provide more timely guidance on your original issue and this PR. I'm just not able to immediately able to devote time to providing quality design advice, but I do feel like this is an important area to think deeply about (and to improve). I just can't work out all the implications of this proposal right now, and to reconcile it with articles like https://anto.pt/articles/go-http-responsewriter

@sonatard Do you have any advice?

git-hulk:
@StevenACoffman Thanks for your reply and efforts.

 > I just can't work out all the implications of this proposal right now, and to reconcile it with articles like https://anto.pt/articles/go-http-responsewriter

Yes, Go supports intercepting the HTTP writer with another wrapper. But in our scenarios, we would like to add some customized HTTP headers according to the GraphQL response, and this cannot be achieved by simply rewriting the HTTP writer.

StevenACoffman:
@CodiumAI-Agent /review

StevenACoffman:
Ok, well, that experiment was not especially useful. 😆

StevenACoffman:
Discussion in #567 is relevant to this. The supported method for bridging between GraphQL and HTTP is through context and HTTP middleware. This is because GraphQL is intended to be transport agnostic, so we do our best to keep HTTP handlers and resolvers segregated.

git-hulk:
> Discussion in #567 is relevant to this. The supported method for bridging between GraphQL and HTTP is through context and HTTP middleware. This is because GraphQL is intended to be transport agnostic, so we do our best to keep HTTP handlers and resolvers segregated.

@StevenACoffman Thanks for your reply. I guessed the original intention when it used `io.Writer` instead of `http.ResponseWriter`. I will close this PR and the related issue.

PR #3723, opened 2025-06-05
Add an option to allow using `appliation/json` on wildcard accept header

### LINKED ISSUE #3712: Should use `application/graphql-response+json` as the default content-type while ACCEPT is wildcard
As mentioned in the comment https://github.com/99designs/gqlgen/issues/3602#issuecomment-2894775785, using `application/graphql-response+json` will introduce the compatibility issue if users are using a wildcard in the ACCEPT header like `application/*`/`*/*`.

From the spec, it doesn't explicitly state how to handle this situation. But it recommends choosing the `application/graphql-response+json` over `application/json` for the sake of the first one adds improved of HTTP status. See: https://github.com/99designs/gqlgen/issues/3602#issuecomment-2896415854.

I believe there are advantages and disadvantages to both approaches, to see if maintainers have any comments on this.

git-hulk:
This closes #3712.

I have:

### DIFF
--- a/graphql/handler/transport/headers.go
+++ b/graphql/handler/transport/headers.go
@@ -11,7 +11,7 @@ const (
 	acceptApplicationGraphqlResponseJson = "application/graphql-response+json"
 )
 
-func determineResponseContentType(explicitHeaders map[string][]string, r *http.Request) string {
+func determineResponseContentType(explicitHeaders map[string][]string, r *http.Request, useGrapQLResponseJsonByDefault bool) string {
 	for k, v := range explicitHeaders {
 		if strings.EqualFold(k, "Content-Type") {
 			return v[0]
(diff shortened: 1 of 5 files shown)


graphql/handler/transport/headers.go
```diff
@@ -11,17 +11,18 @@ const (
 	acceptApplicationGraphqlResponseJson = "application/graphql-response+json"
 )
 
-func determineResponseContentType(explicitHeaders map[string][]string, r *http.Request) string {
+func determineResponseContentType(explicitHeaders map[string][]string, r *http.Request, useGrapQLResponseJsonByDefault bool) string {
```
StevenACoffman, line 14: It seems a little odd to have the parameter order be `r *http.Request, useGrapQLResponseJsonByDefault bool` instead of `useGrapQLResponseJsonByDefault bool, r *http.Request` but I suppose it doesn't really matter.

StevenACoffman:
Thanks!

graphql/handler/transport/headers.go
git-hulk: @StevenACoffman Thanks for your review and suggestions. Should I adjust the order accordingly?

PR #3744, opened 2025-06-20
Fix/explicit requires with mutli entity resolver

parkerroan:

I have:

### DIFF
--- a/plugin/federation/federation.gotpl
+++ b/plugin/federation/federation.gotpl
@@ -268,10 +268,13 @@ func (ec *executionContext) resolveManyEntities(
 								{{- if $options.ComputedRequires }}
 									{{/* We don't do anything in this case, computed requires are handled by standard resolvers */}}
 								{{- else if and $options.ExplicitRequires $entity.Requires }}
-									err = ec.Populate{{$entity.Def.Name}}Requires(ctx, {{- if (not $usePointers) -}}&{{- end -}}entity, rep)
-									if err != nil {
-										return fmt.Errorf(`populating requires for Entity "{{$entity.Def.Name}}": %w`, err)
-									}
+									{{- if not (index $emitted $entity.Def.Name) }}
+										err = ec.Populate{{$entity.Def.Name}}Requires(ctx, {{- if (not $usePointers) -}}&{{- end -}}entity, rep)
+										if err != nil {
+											return fmt.Errorf(`populating requires for Entity "{{$entity.Def.Name}}": %w`, err)
+										}
+										{{- $_ := set $emitted $entity.Def.Name true }}
+									{{- end }}
 								{{- else }}
 									{{ if $useFunctionSyntaxForExecutionContext -}}
 										entity.{{.Field.JoinGo `.`}}, err = {{.Type.UnmarshalFunc}}(ctx, ec, reps[i].entity["{{.Field.Join `"].(map[string]any)["`}}"])
--- a/plugin/federation/federation.gotpl
+++ b/plugin/federation/federation.gotpl
@@ -265,14 +265,23 @@ func (ec *executionContext) resolveManyEntities(
 
 						for i, entity := range entities {
 							{{- range $entity.Requires }}
-								{{ if $useFunctionSyntaxForExecutionContext -}}
-								entity.{{.Field.JoinGo `.`}}, err = {{.Type.UnmarshalFunc}}(ctx, ec, reps[i].entity["{{.Field.Join `"].(map[string]any)["`}}"])
-								{{- else -}}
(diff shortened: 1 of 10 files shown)


plugin/federation/federation.gotpl
```diff
@@ -264,15 +264,27 @@ func (ec *executionContext) resolveManyEntities(
...
+									return fmt.Errorf(`populating requires for Entity "{{$entity.Def.Name}}": %w`, err)
+								}
+							{{- end }}
 							{{- range $entity.Requires }}
-								{{ if $useFunctionSyntaxForExecutionContext -}}
-								entity.{{.Field.JoinGo `.`}}, err = {{.Type.UnmarshalFunc}}(ctx, ec, reps[i].entity["{{.Field.Join `"].(map[string]any)["`}}"])
-								{{- else -}}
-								entity.{{.Field.JoinGo `.`}}, err = ec.{{.Type.UnmarshalFunc}}(ctx, reps[i].entity["{{.Field.Join `"].(map[string]any)["`}}"])
-								{{- end }}
+								{{- if $options.ComputedRequires }}
```
parkerroan, line 274: I was not sure about this but saw it in the other `ResolveEntity` section.

@StevenACoffman 
@alexandear

parkerroan: Looks like test cases addresses this concern. I will need to update computed requires test for this I think

StevenACoffman:
Thanks for working on this!

parkerroan:
> Thanks for working on this!

I decided to limit these changes to explicit requires.  Computed requires will need a similar fix from what I can tell, but I am less familiar with it. 

@clayne11 - tagging you as you seem to have introduced the computed_requires changes. From what I can tell, those changes will also fail generation when `@entityResolver(multi: true)` is set.  I was not aware the portion I was working on was deprecated.

plugin/federation/testdata/explicitrequires/schema.graphql
```diff
@@ -65,7 +65,9 @@ type PlanetRequiresNested @key(fields: "name") {
 type MultiPlanetRequiresNested @key(fields: "name") @entityResolver(multi: true) {
     name: String! @external
     world: World! @external
+    worlds: [World!] @external
     size: Int! @requires(fields: "world{ foo }")
+    sizes: [Int!] @requires(fields: "worlds{ foo }")
```
parkerroan, line 70: This would have failed generation in previous versions.

clayne11:
As far as I know, `@entityResolver(multi: true)` does work with `computed_requires` — that was one of the issues that I addressed with the new design.

My opinion is that `explicit_requires` is fundamentally broken from a design perspective and is deprecated. I would encourage you to move away from it if you can.

parkerroan:
> As far as I know, `@entityResolver(multi: true)` does work with `computed_requires` — that was one of the issues that I addressed with the new design.
> 
> My opinion is that `explicit_requires` is fundamentally broken from a design perspective and is deprecated. I would encourage you to move away from it if you can.

@clayne11 If you take a look at the old code it was still generating explicit mapping unlike resolveEntity which was skipping it.     Also generation would fail on the bug I have listed in the PR using computed_requires also. I just pushed up a fix I would love for you to take a look?

plugin/federation/federation_computedrequires_test.go
```diff
@@ -162,22 +162,23 @@ func TestMultiComputedRequires(t *testing.T) {
...
-				"MultiHelloRequires {name, key1}",
+				"MultiHelloRequires {name, key2}",
 			}),
 			&resp,
 			client.Var("representations", representations),
 		)
 
 		require.NoError(t, err)
 		require.Equal(t, "first name - 1", resp.Entities[0].Name)
-		require.Equal(t, "key1 - 1", resp.Entities[0].Key1)
```
parkerroan, line 178: Computed requires _should not_ populate entity like this as far as I know. This would fail if not for Multi true.

plugin/federation/testdata/explicitrequires/schema.graphql
parkerroan: @clayne11 this is the bug I was referencing. It will fail generation for both options, explicit and computed when mutli is enabled. 

I think I addressed both issues in this PR, the computed fix was incredibly simple.

StevenACoffman:
I really appreciate your engaging with this. There haven't been enough folks carefully going over federation recently and the more eyes on this the better. I'm going to merge this as is for now, but I would love for further follow-up PRs.

StevenACoffman:
@parkerroan I tried to connect with you on https://www.linkedin.com/in/steve-coffman-79322175/ because it might be easier to talk over a few things in a quick chat, if you are up for that.

Issue #3758, opened 2025-07-04
Docs examples all use deprecated NewDefaultServer


brackendawson:
The examples in the docs such as [this one](https://gqlgen.com/getting-started/) use a main package which calls [NewDefaultServer](https://pkg.go.dev/github.com/99designs/gqlgen@v0.17.76/graphql/handler#NewDefaultServer). This function is deprecated but the deprecation note does not say what to use instead, this makes the need to show how to properly configure the server more important.

I also raised https://github.com/issues/created?issue=vektah%7Cgqlgen-tutorials%7C11 in the repo for the examples' code.

## Expectation
I expected to be able to find out how to use gqlgen for a production server.

## Actuality
The documentation in the getting started guide and the [handler package](https://pkg.go.dev/github.com/99designs/gqlgen@v0.17.76/graphql/handler) only show how set up an experimental server.

StevenACoffman:
@brackendawson Yes! This is on purpose! If you read the docs for the NewDefaultServer, it is intended as an example. You can copy it into your own code base and customize it based on your needs. You should not include websockets if you don't want/need websockets, for example. You should not include SSE if you don't want/need SEE. And so on.

brackendawson:
Thanks for the reply. Should the deprecation notice for NewDefaultServer be updated to include "copy it into your own code base and customize it based on your needs"? I'm happy to raise the PR.

StevenACoffman:
@brackendawson Sure! That is what the [current comment](https://github.com/99designs/gqlgen/blob/fadac6f93abf5b7516c0d16e815aeafbc8d63c93/graphql/handler/server.go#L34-L48) was intended to convey, so maybe your phrasing would be more clear.


```
// NewDefaultServer is a demonstration only. Not for prod.
//
// Currently, the server just picks the first available transport,
// so this example NewDefaultServer orders them, but it is just
// for demonstration purposes.
// You will likely want to tune and better configure Websocket transport
// since adding a new one (To configure it) doesn't have effect.
//
// Also SSE support is not in here at all!
// SSE when used over HTTP/1.1 (but not HTTP/2 or HTTP/3),
// SSE suffers from a severe limitation to the maximum number
// of open connections of 6 per browser. See:
// https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#sect1
//
// Deprecated: This was and is just an example.
```

PR #3764, opened 2025-07-14
Better document how to construct a Server

brackendawson:
Expecting a functional constructor; I had a hard time working out how to make a server appropriate for production. This was compounded by all examples using NewDefaultServer. What was not written down is that you should use New and provide properly configured transports, so write that in the deprecation notice.

I have:
 - [x] Changed the NewDefaultServer deprecation notice to say what to use in place of NewDefaultServer.
 - [x] Added documentation to Server.AddTransport.
 - [x] Added documentation to Server.Use.
 - [x] Added documentation to New.
 
 Here's how it renders with pkgsite:
bilde

Addresses #3758

### DIFF
--- a/graphql/handler/server.go
+++ b/graphql/handler/server.go
@@ -36,12 +36,13 @@ func New(es graphql.ExecutableSchema) *Server {
 }
 
 // NewDefaultServer returns a Server for the given executable schema which is
-// suitable as an example.
+// only suitable for use in examples.
 //
 // Deprecated:
-// NewDefaultServer is only suitable as an example. Use [New] and add transports
-// configured for your use case. See the implementation of this constructor for
-// an example.
+// The Server returned by NewDefaultServer is not suitable for production use.
+// Use [New] instead and add transports configured for your use case,
+// appropriate caches, and introspection if required. See the implementation of
+// NewDefaultServer for an example of starting point to construct a Server.
 //
 // SSE is not supported using this example. SSE when used over HTTP/1.1 (but not
 // HTTP/2 or HTTP/3) suffers from a severe limitation to the maximum number of


brackendawson:
IME documentation PRs have more rounds of review than code so I'm of course happy for as much back-and-forward as is needed here.

graphql/handler/server.go
```diff
@@ -25,27 +25,29 @@ type (
...
+// example.
 func New(es graphql.ExecutableSchema) *Server {
 	return &Server{
 		exec: executor.New(es),
 	}
 }
 
-// NewDefaultServer is a demonstration only. Not for prod.
+// NewDefaultServer returns a Server for the given executable schema which is
+// suitable as an example.
```
StevenACoffman, line 39:
```suggestion
// suitable **only** as an example of a starting point.
// 
// Please copy this (or something similar) into **your own project** 
// and then customize it to match your specific project needs:
// 
// + Add only the Transports your project needs, in precedence order. 
// + Set up appropriate caches (Memcached? Redis? What size?) 
// + Enable Introspection if required
//
```

brackendawson: Using the standard library as an example, I aimed for the following properties:
* The first line of the doc string says what the function does.
* "Deprecated:" preceeds the reason for deprecation and then says what to use instead.
* The documentation concisely states facts and consequences. No emphasis, repetition, or verbosity, which only dilutes the message. The standard library never uses bold type and never says please.

> suitable **only** as an example of a starting point.

The returned Server isn't a starting point, that suggests it can be modified to become production ready, but it can't. The source Code of NewDefaultServer is the starting point, I'll add a commit to make that clearer and include your other recommended configurations.

StevenACoffman: > The returned Server isn't a starting point, my first attempt got this wrong too. It suggests it can be modified to become production ready, but it can't.

Oooooh! Now I see how that could be confusing.

PR #3773, opened 2025-07-25
Reuse argument parsing function

UnAfraid:
Added new function processArgField that can be reused for every field argument parsing instead of generating one for each field
(Whenever possible, in the case where directives are involved - falls back the old way)

I've tested this on pretty big schema (over 400 types/inputs) and it it results in -20k lines of code

I have:

### DIFF
--- a/_examples/chat/generated.go
+++ b/_examples/chat/generated.go
@@ -279,6 +279,26 @@ func (e *executableSchema) Exec(ctx context.Context) graphql.ResponseHandler {
 	}
 }
 
+func processArgField[T any](
+	ctx context.Context,
+	rawArgs map[string]any,
+	fieldName string,
+	valueMapperFn func(ctx context.Context, value any) (T, error),
+) (T, error) {
+	if _, ok := rawArgs[fieldName]; !ok {
+		var zeroVal T
+		return zeroVal, nil
+	}
+
+	ctx = graphql.WithPathContext(ctx, graphql.NewPathWithField(fieldName))
+	if tmp, ok := rawArgs[fieldName]; ok {
+		return valueMapperFn(ctx, tmp)
+	}
+
+	var zeroVal T
+	return zeroVal, nil
+}
+
 type executionContext struct {
 	*graphql.OperationContext
 	*executableSchema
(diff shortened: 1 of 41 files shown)


StevenACoffman:
@UnAfraid This is a massive improvement!  Thank you so much! 

I noticed that the only conditional template difference between the two variations is due two is whether or not it is using $useFunctionSyntaxForExecutionContext : 
[codegen/generated!.gotpl](https://github.com/99designs/gqlgen/pull/3773/files#diff-116ba37e32ebcd0b716dc9d8ef840329a5173a17e247a87385356e9b17358c1c) 

What do you think about changing the name of the ExecutionContext variant and exporting both variants so they don't need to be part of the template code anymore at all?

Screenshot 2025-07-26 at 12 43 08 PM

StevenACoffman:
@UnAfraid  I'm going to merge this PR as it is, since it's already a huge improvement, but I would like to move as much logic out of templates as possible. If you are willing or interested in another PR to do this I would be hugely appreciative!

The code in gqlgen that is in a template receives significantly fewer contributions than the traditional library code. I believe it's primarily due to the difficulty for most developers (and tools) to understand and manipulate it.

If these functions were just exported functions in a library, we could unit test them, and they wouldn't be intimidating for any developer to help us improve. What do you think?

UnAfraid:
> @UnAfraid This is a massive improvement! Thank you so much!
> 
> I noticed that the only conditional template difference between the two variations is due two is whether or not it is using $useFunctionSyntaxForExecutionContext : [codegen/generated!.gotpl](https://github.com/99designs/gqlgen/pull/3773/files#diff-116ba37e32ebcd0b716dc9d8ef840329a5173a17e247a87385356e9b17358c1c)
> 
> What do you think about changing the name of the ExecutionContext variant and exporting both variants so they don't need to be part of the template code anymore at all?
> 
> Screenshot 2025-07-26 at 12 43 08 PM

Yeah i am interested in moving more of the generated code to reusable functions, where would be a good place/package to place those in?

StevenACoffman:
@UnAfraid Thank you for volunteering! The easiest place to put these particular reusable functions is in "github.com/99designs/gqlgen/graphql" as all of the code that gets generated using those functions already **at least** imports:
```
	"github.com/99designs/gqlgen/graphql/introspection"
	gqlparser "github.com/vektah/gqlparser/v2"
	"github.com/vektah/gqlparser/v2/ast"
```

Unless there's an import cycle, or something seems like it _needs_ to belong elsewhere, I haven't been getting any complaints concerned with there being too much in that package, and reorganizing it is easier than moving things out of templates.

PR #3813, opened 2025-09-01
Documentation of the Omittable type

evellior:
I couldn't find any information about how to use [the Omittable type](https://github.com/99designs/gqlgen/pull/2585) in the documentation other than a single mention of it existing in the config options.  The type itself is fairly self explanatory, but I didn't actually realize it existed and had been using maps for changesets until now.  I thought I'd point it out in the changeset documentation, seeing as in most situations it's a strictly better (type safe) way of doing the same thing.

I have:
 - [x] `n/a` Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))

### DIFF
--- a/docs/content/reference/changesets.md
+++ b/docs/content/reference/changesets.md
@@ -5,8 +5,9 @@ description: Falling back to map[string]interface{} to allow for presence checks
 menu: { main: { parent: 'reference', weight: 10 } }
 ---
 
-Occasionally you need to distinguish presence from nil (undefined vs null). In gqlgen we do this using maps:
+Occasionally you need to distinguish presence from nil (undefined vs null). In gqlgen this can be done using either maps or the Omittable type.
 
+## Maps
 
 ```graphql
 type Mutation {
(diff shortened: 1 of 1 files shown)


StevenACoffman:
Thanks!

StevenACoffman:
@evellior What do you think about #3227 ?

evellior:
> @evellior What do you think about #3227 ?

Our codebase has basically the same issue, we have a custom type for optional fields and our service layer doesn't import any of the API layer code.  If it were handled in a similar way to how you can map to custom scalars, if we could just write an unmarshal method for a custom omittable model and add it to the config, that would be ideal.

The only thing I don't like about their suggestion is the idea of implementing multiple omittable types (one per scalar?).  Using [custom scalars](https://gqlgen.com/reference/scalars/#custom-scalars-with-third-party-types) as a reference and the `go-optional` type mentioned in that issue (which isn't actually what I use so excuse if I declare it incorrectly), I would suggest the custom omittable handler should look something like this:

(I'm using the word unmarshaller even though I'm not sure that's really what this is, but I'm not sure what else to call it..)
```
package mypkg

import (
	"https://github.com/moznion/go-optional"
)

func UnmarshalMyCustomOmittable[T any](v *T) (optional.Option[*T], error) {
	// This method being called means the field was included in the changeset
	return optional.Some[*T](v)
}
```
with the config either following the model pattern:
```
omittable:
	model: github.com/me/mypkg.MyCustomOmittable
```
or the option pattern:
```
nullable_input_omittable_type: github.com/me/mypkg.MyCustomOmittable
```

Overall I'd place the issue low priority, someone like them or myself who's working on the issue can add that functionality if/when the current way becomes too annoying.  It isn't a major or blocking issue or anything. 
I hadn't seen that issue but our solution was the same as theirs, a generic function to convert from Omittable to our type.  Having to map every optional field manually does make writing the resolvers a little tedious but not a big deal.

PR #3832, opened 2025-09-09
Move field generated code to graphql package 

dloukadakis:
Reduced code generation size by reusing common field resolution logic.

I have:


StevenACoffman:
Thanks for working on this! This is a welcome improvement, as moving code out of templates makes it much easier for the community to maintain.

Looks like gotestsum now requires Go 1.24: https://github.com/gotestyourself/gotestsum/blob/main/go.mod#L3
```
go: gotest.tools/gotestsum@latest: gotest.tools/gotestsum@v1.13.0 requires go >= 1.24.0 (running go 1.23.12; GOTOOLCHAIN=local)
```

Looks like some of the other lint failures are still legit though:
```
  Error: graphql/resolve_field_test.go:201:12: context-keys-type: should not use basic type untyped string as key in context.WithValue (revive)
  					ctx = context.WithValue(ctx, "resolver", "middleware")
  					      ^
  Error: graphql/resolve_field_test.go:258:4: require-error: for error assertions use require (testifylint)
  			assert.EqualError(t, GetErrors(ctx), test.expectedErr)
```

dloukadakis:
Both of them are now fixed, thank you!

StevenACoffman:
Thanks! 

@UnAfraid I'm merging this PR, despite the test failures on Go 1.23 due to gotestsum latest no longer supporting Go 1.23, as I was also planning on dropping support for Go 1.23. I'll make a follow-up PR for that.

PR #3833, opened 2025-09-11
Drop support for Go 1.23 and Add support for Go 1.25

StevenACoffman:
The time has come to finally say good bye to Go 1.23 and hello to Go 1.25!

Our official support policy (which I should definitely put into a document somewhere) is the same as the Go one: Each major Go release is supported until there are two newer major releases.

We delay that by a bit sometimes because my current employer (Khan Academy) uses Google AppEngine, and AppEngine is often a bit slow to release support for the last version of Go: https://cloud.google.com/appengine/docs/standard/go/release-notes

### DIFF
--- a/codegen/config/config_test.go
+++ b/codegen/config/config_test.go
@@ -70,7 +70,7 @@ func TestReadConfig(t *testing.T) {
 		if runtime.GOOS == "windows" {
 			require.ErrorContains(t, err, "failed to walk schema at root not_walkable/: ")
 			// Go 1.24 and below report "CreateFile" but 1.25 and above report "GetFileAttributesEx" in error
-			// See https://tip.golang.org/doc/go1.25#ospkgos
+			// See https://go.dev/doc/go1.25#ospkgos
 			require.ErrorContains(t, err, " not_walkable/: The system cannot find the file specified.")
 		} else {
 			require.EqualError(t, err, "failed to walk schema at root not_walkable/: lstat not_walkable/: no such file or directory")


StevenACoffman:
Hmmm... on Windows for Go 1.25, it looks like we are surprisingly getting test failures(?!?!):
```
=== Failed
=== FAIL: codegen/config TestReadConfig/unwalkable_path (0.00s)
    config_test.go:70: 
        	Error Trace:	github.com/99designs/gqlgen/codegen/config/config_test.go:70
        	Error:      	Error message not equal:
        	            	expected: "failed to walk schema at root not_walkable/: CreateFile not_walkable/: The system cannot find the file specified."
        	            	actual  : "failed to walk schema at root not_walkable/: GetFileAttributesEx not_walkable/: The system cannot find the file specified."
        	Test:       	TestReadConfig/unwalkable_path

=== FAIL: codegen/config TestReadConfig (0.00s)

DONE 1225 tests, 2 failures in 117.358s
```

Not sure what in Go 1.25 release notes would have caused this: https://tip.golang.org/doc/go1.25

StevenACoffman:
I couldn't find an easy way to reliably futureproof this test based on the Go version.

I started to think about adding https://github.com/caarlos0/go-version/blob/main/version.go

This would give you the GoVersion as a string like `1.25.0` vs `1.24.2` but then parsing that out into major/minor and ensuring that the minor ... I just decided it was all not worth it for this one test.

Issue #2373, opened 2022-09-14
Proposal: deterministic batch loading


n-e:
### Dataloaders

As [described in the documention](https://gqlgen.com/reference/dataloaders/), calls to databases or other external services need to be batched to achieve decent performance. This is generally achieved by using the [dataloader pattern](https://github.com/graphql/dataloader#dataloader).

In golang, the dataloader implementations ([dataloaden](https://github.com/vektah/dataloaden), [graph-gophers/dataloader](https://github.com/graph-gophers/dataloader)) work roughly this way:

- dataloader.Load(id) is called
- dataloader.Load blocks for a set amount of time (for example 5ms)
- once the amount of time has elapsed, the user-provided Fetch function is called with the ids passed to the calls of dataloader.Load made during the blocking time, returns the values and the dataloader.Load calls return

This approach has several drawbacks:
- each call to dataloader.Load that starts a new batch blocks N ms, which means nested queries take at least depth * N ms
- dataloader.Load calls need to be thread-safe, which results in significant overhead
- Most importantly, under load (including large graphql queries), unless the wait time is excessively large, the dataloader.Load calls will be made too slowly, resulting in poor batching, reduced performance and increased database load.

### gqlgen

gqlgen builds the query results depth-first, top to bottom, and spawns a goroutine for each field in a fieldset, and each element in an array (with a few optimizations). For example, the query ` query { users { stars; likes }}` will roughly result in this `Users(); go marshallUser(user1); go Stars(user1); go Likes(user1); go marshallUser(user2)...`

This compounds the dataloaders drawbacks since:
- the goroutines aren't spawned in the right order to optimize batching (ideally all the calls to the Stars resolver would be made first, then all the calls to the Likes resolver)
- the approach can result in spawning a large number of goroutines (for example 3000 goroutines for 1000 users), further adding overhead

### Previous discussion

https://github.com/99designs/gqlgen/issues/518

## Proposal

Add built-in batch-loading to gqlgen. Instead of having the following resolver signature:

```go
func (r *itemResolver) Children(ctx context.Context, obj *model.Item, ...) ([]*model.Item, error)
```

have this one:

```go
func (r *itemResolver) Children(ctx context.Context, obj []*model.Item, ...) ([][]*model.Item, []error)
```

[Proof of concept](https://github.com/n-e/gqlgen/tree/array-resolvers) (simple schemas work. nulls, error handling, introspection, middlewares and likely many other things don't)

[...]

PR #3853, opened 2025-09-25
feat(dataloader): handle thunk resolvers

lbarthon:
With how dataloadgen works, we don't need to create as many goroutines, and we can let it handle all the batching for us, and use their "thunk" function, that will resolve when the batch of data resolved.

This commit adds a new resolver configuration, "thunk_resolvers", that changes the return type of resolvers from `(T, error)` to `(func() (T, error), error)`. This allows us to implement resolvers using `LoadThunk` instead of `Load`, thus using the whole potential of dataloadgen.

This will end up creating a lot less goroutines in heavy setups that rely on dataloading, and should greatly improve performance.

This setting could also be enabled for anyone that wants it, without the need to rely on dataloadgen. Effectively, it gives control to the engineer writing the implementation of how they handle their goroutines.

I have:

### DIFF
--- a/graphql/resolve_field.go
+++ b/graphql/resolve_field.go
@@ -28,8 +28,17 @@ func ResolveField[T any](
 		recoverFromPanic,
 		nonNull,
 		Null,
-		func(ctx context.Context, res T) Marshaler {
-			return marshal(ctx, field.Selections, res)
+		func(ctx context.Context, fc *FieldContext, res func() (T, error)) Marshaler {
+			return WriterFunc(func(w io.Writer) {
+				ret, err := res()
+				fc.Result = ret
+				if err != nil {
+					oc.Error(ctx, err)
+					Null.MarshalGQL(w)
+					return
+				}
+				marshal(ctx, field.Selections, ret).MarshalGQL(w)
+			})
 		},
 	)
 }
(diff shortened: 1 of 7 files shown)


StevenACoffman:
@coady Do you have thoughts on this?

graphql/resolve_field.go
```diff
@@ -28,8 +28,15 @@ func ResolveField[T any](
 		recoverFromPanic,
 		nonNull,
 		Null,
-		func(ctx context.Context, res T) Marshaler {
-			return marshal(ctx, field.Selections, res)
+		func(ctx context.Context, res func() (T, error)) Marshaler {
+			return WriterFunc(func(w io.Writer) {
+				ret, err := res()
```
dloukadakis, line 33: This circumvents the non null check that errors with "must not be null" in `resolveField` because this function could return nil

dloukadakis (changes requested):

lbarthon:
Being fully honest, this isn't perfect, but as I said in the commit message, our goal is to drastically reduce the # of goroutines, and this should help a ton

StevenACoffman:
This has a lot of potential, but I think it needs a bit more polish before it's fair to unleash it on others' production environments, don't you think?

I would like to encourage you to keep going on it though.

StevenACoffman:
@vikstrous Do you have thoughts on how this could be further improved?

vikstrous:
I don't know the details of the generated code well enough to comment on the PR content.

An API change like this would be hard to take back if it's added.

I've been hoping to see progress on a different API change: https://github.com/99designs/gqlgen/issues/2373
Allowing to write batch loaders directly in gqlgen means there's no need for loaders or thunks in the first place. I'm concerned that loaders don't know when it's time to run, so they sleep for a fixed amount of time.

So, this PR might be an incremental improvement, but the new interface forces you to use loaders without solving the loaders sleep issue. I would prefer to see the problem solved by gqlgen directly without third party loaders.

lbarthon:
A few realizations I've had since working on this:
- We had a goroutine leak in our app, which caused a pressure due to the large amount of goroutines. Once we fixed it, everything came back to normal, and although the quantities of goroutines were high, they were within reasonable bounds.
- This PR is bad, as it removes any potential batching logic. Because the results are only retrieved once we start calling `MarshalGQL`, this causes the child resolvers to be called one by one, instead of in a batch, which defeats the whole purpose of GraphQL / dataloading.

I'll close it for now, if someone want to look into it later, they can start with this I guess!

Issue #3863, opened 2025-10-06
Feature Request: Add response context support for interceptors (esp. Subscriptions)


michelepra:
## Problem

Interceptors in gqlgen only receive the **request context** as their first parameter.  
For queries and mutations, this works fine because execution is linear:

1. Receive request  
2. Execute resolver  
3. Return response (context == request context)  

For **subscriptions**, however:

- gqlgen starts a goroutine that publishes values to a channel  
- gqlgen reads from this channel and encodes responses  
- At this point, the response has **no associated context** that interceptors can access  

As a result, interceptors cannot access or enrich the "response context" (e.g., metadata, tracing, request-scoped values) for subscription payloads.
They only see the request context from when the subscription was created.

---

### 1. Extend `Response` struct with a `Context` field

Current:  
```go
type Response struct {
	Errors     gqlerror.List   `json:"errors,omitempty"`
	Data       json.RawMessage `json:"data"`
	Label      string          `json:"label,omitempty"`
	Path       ast.Path        `json:"path,omitempty"`
	HasNext    *bool           `json:"hasNext,omitempty"`
	Extensions map[string]any  `json:"extensions,omitempty"`
}
```
Proposed:
```go
type Response struct {
	Context context.Context `json:"-"`

	Errors     gqlerror.List   `json:"errors,omitempty"`
	Data       json.RawMessage `json:"data"`
	Label      string          `json:"label,omitempty"`
	Path       ast.Path        `json:"path,omitempty"`
	HasNext    *bool           `json:"hasNext,omitempty"`
	Extensions map[string]any  `json:"extensions,omitempty"`
}
```
This allows interceptors to access response-level context.

### 2. Introduce a SubscriptionField interface for subscription result payloads
Instead of subscription resolvers returning `(<-chan *T, error)`, have them return something like:
```go
type SubscriptionField[T any] interface {
	GetContext() context.Context
	GetField() T
}
```
So subscription resolvers could return:
```go
(<-chan SubscriptionField[*T], error)
```
This lets each published item carry its own context, which flows naturally into the modified Response.Context.

---

## Some benefit 

- Consistent interceptor behavior across queries, mutations, and subscriptions withouth break current interceptor code
- Enables request/response tracing, per-event metadata, and richer observability for subscriptions
- Clean separation of concerns: business payload stays in GetField(), while per-response metadata lives in context.Context

michelepra:
[...]

PR #3864, opened 2025-10-06
response context support for interceptors

michelepra:
This is PR for https://github.com/99designs/gqlgen/issues/3863

This allows interceptors to access response-level context so that interceptors can work on subscriptions similar to how they work on queries and mutations.  The breaking part is changing the subscription resolver’s return type. We could avoid that by making the feature opt-in with a disabled-by-default parameter.

### 1. Extend `Response` struct with a `Context` field

Current:  
```go
type Response struct {
	Errors     gqlerror.List   `json:"errors,omitempty"`
	Data       json.RawMessage `json:"data"`
	Label      string          `json:"label,omitempty"`
	Path       ast.Path        `json:"path,omitempty"`
	HasNext    *bool           `json:"hasNext,omitempty"`
	Extensions map[string]any  `json:"extensions,omitempty"`
}
```
Proposed:
```go
type Response struct {
	Context context.Context `json:"-"`

	Errors     gqlerror.List   `json:"errors,omitempty"`
	Data       json.RawMessage `json:"data"`
	Label      string          `json:"label,omitempty"`
	Path       ast.Path        `json:"path,omitempty"`
	HasNext    *bool           `json:"hasNext,omitempty"`
	Extensions map[string]any  `json:"extensions,omitempty"`
}
```
This allows interceptors to access response-level context.

### 2. Introduce a SubscriptionField interface for subscription result payloads
Instead of subscription resolvers returning `(<-chan *T, error)`, have them return something like:
```go
type SubscriptionField[T any] interface {
	GetContext() context.Context
	GetField() T
}
```
So subscription resolvers could return:
```go
(<-chan SubscriptionField[*T], error)
```
This lets each published item carry its own context, which flows naturally into the modified Response.Context.

---

## Benefit 

- Consistent interceptor behavior across queries, mutations, and subscriptions without breaking current interceptor code
- Enables request/response tracing, per-event metadata, and richer observability for subscriptions
- Clean separation of concerns: business payload stays in GetField(), while per-response metadata lives in context.Context

I have:

### DIFF
--- a/_examples/chat/generated.go
+++ b/_examples/chat/generated.go
@@ -83,7 +83,7 @@ type QueryResolver interface {
 	Room(ctx context.Context, name string) (*Chatroom, error)
 }
 type SubscriptionResolver interface {
-	MessageAdded(ctx context.Context, roomName string) (<-chan *Message, error)
+	MessageAdded(ctx context.Context, roomName string) (<-chan graphql.SubscriptionField[*Message], error)
 }
 
 type executableSchema struct {
@@ -222,6 +222,8 @@ func (e *executableSchema) Exec(ctx context.Context) graphql.ResponseHandler {
 					return nil
 				}
 			}
+			response.Context = ctx
+
 			var buf bytes.Buffer
 			data.MarshalGQL(&buf)
 			response.Data = buf.Bytes()
@@ -244,7 +246,8 @@ func (e *executableSchema) Exec(ctx context.Context) graphql.ResponseHandler {
 			data.MarshalGQL(&buf)
 
 			return &graphql.Response{
-				Data: buf.Bytes(),
+				Context: ctx,
+				Data:    buf.Bytes(),
 			}
 		}
 	case ast.Subscription:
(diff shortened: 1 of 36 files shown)


UnAfraid:
That is very interesting idea, but i am wondering, couldn't we wrap the internals only so we don't introduce breaking changes?

michelepra:
> That is very interesting idea, but i am wondering, couldn't we wrap the internals only so we don't introduce breaking changes?

Adding Context to Response isn’t breaking.
The breaking part is changing the subscription resolver’s return type. We could avoid that by making the feature opt-in with a disabled-by-default parameter.

Unfortunately, I don’t have other ideas on how to propagate the subscription response context up to the interceptors (and beyond) without introducing a breaking change on the resolver.

StevenACoffman:
@michelepra There are _a lot_ of test failures and lint errors. I think adding a disabled by default config option would avoid fix these, and avoid breaking backwards compatibility for the community.

@john-markham @nabishec @giulio-opal @szgupta @zhixinwen @telemenar @giulio-opal @jmic @alexvelea Aside from the broken tests, what do you think about the general concept of passing context in a channel here for subscriptions for interceptors?

StevenACoffman:
Closing due to lack of response from @michelepra so he appears to not be serious about wanting to contribute.

PR #3865, opened 2025-10-06
fix(modelgen): use ToGoModelName instead of ToGo for model fields     

lbarthon:
When there are conflicting model names in the codegen, this could create issues
where the wrong model would be referenced in a struct field. This is now fixed,
and models will refer to other types properly!

I have:


lbarthon:
@StevenACoffman when you have a moment :p

StevenACoffman:
Thanks!

lbarthon:
Anytime! Mind bumping the version of gqlgen, so we can start using this?

lbarthon:
cc @StevenACoffman 👀

PR #3885, opened 2025-10-30
Embedded base types for interfaces: support external types embedding, transitional embedding of repeated fields

PaulVasilenko:
…ding

I have:

### DIFF
--- a/plugin/modelgen/interface_graph.go
+++ b/plugin/modelgen/interface_graph.go
@@ -1,7 +1,7 @@
 package modelgen
 
 import (
-	"fmt"
+	"errors"
 
 	"github.com/vektah/gqlparser/v2/ast"
 )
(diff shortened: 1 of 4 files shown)


PaulVasilenko:
@StevenACoffman @adrianlungu 

Original pull request https://github.com/99designs/gqlgen/pull/3242 introduced this logic, but misses a lot of edge cases (including our internal use-case) 

So this pull request handles all different variations, namely

1. Interfaces implementing interfaces -> chained embedding
2. Diamond problem -> attempt to infer most fitting struct to embed
3. Interface covariance: when interface has interface field, GraphQL allows implementation to replace this fields with implementing types, which doesn't allow to embed base type

This feature appeared much more complicated than originally anticipated, but i guess somewhat useful (at least in our organisation and we would be glad to avoid using forks)

PR #3896, opened 2025-11-04
fix: allow error presenter to silence errors

lbarthon:
The `ErrorPresenterFunc` might return `nil` in some cases, if we want to silence the errors. This isn't handled properly, and causes panics when we go over the errors later in this file. We need to keep this ability, as the default error presenter can return nil too (though unlikely, as it only does it when err is nil).

I have:


lbarthon:
cc @StevenACoffman 😄

StevenACoffman:
Thanks! I really appreciate all the fixes and improvements!

Issue #2367, opened 2022-09-13
ComplexityLimit


alichherawalla:
Hi, 

Thank you for this amazing library, and all the hard work put into making this. It definitely makes my day-to-day life easier

###  What happened?
- I used extenstion.FixedLimitComplexity(5)
### What did you expect?
- I expected the following behavior
```
query {
  shop {                  # Object  - 1 point
    id                    # ID      - 0 points
    name                  # String  - 0 points
    timezoneOffsetMinutes # Int     - 0 points
    customerAccounts      # Enum    - 0 points
  }
}
```
https://shopify.engineering/rate-limiting-graphql-apis-calculating-query-complexity

However, I noticed that the complexity is incremented by 1 even for scalar types. [This](https://github.com/99designs/gqlgen/blob/master/complexity/complexity.go#L45) is the line that is responsible for it. 

If this is not the expected behavior, I'm happy to make the change to fix the issue. Please let me know.

Thanks

alichherawalla:
@vektah possible to please review this question?

jakezhu9:
agree with you. I don't think the default complexity calculation is satisfying, and the only way to change it now is to use the custom complexity function, which is inconvenient. Maybe we should add some different complexity calculation strategies for users to choose, or add useful functions to make customization easier for users (such as setting complexity by type)

ryicoh:
This value must be set to 0, but it is an breaking change.
https://github.com/99designs/gqlgen/blob/3087cf3a9830f80d01f508cf49d55e2370f60a19/complexity/complexity.go#L78

ryicoh:
It's related to https://github.com/99designs/gqlgen/issues/508.

As you say, current behavior may be not good.
However, this method only solves the problem of cost calculation by maximum depth.

It should be configurable for production use, and gqlgen is the only one that can do so.

* [graphql-go/graphql](https://github.com/graphql-go/graphql) stops at [this](https://github.com/graphql-go/graphql/issues/231) discussion.
* [graph-gophers/graphql-go](https://github.com/graph-gophers/graphql-go) can specify only [MaxDepth](https://github.com/graph-gophers/graphql-go/blob/0135d515ff3f9a4524720cac0727b48273354b78/graphql.go#L113).

gqlgen is more flexible than others.
https://github.com/99designs/gqlgen/blob/3087cf3a9830f80d01f508cf49d55e2370f60a19/_examples/todo/generated.go#L48

[...]

Issue #508, opened 2019-01-24
Implement complexity as a plugin


mia0x75:
schema:
```
type Query {
	posts(count: Int = 10): [Post!]!
}

type Post {
	title: String!
	text: String!
	related(count: Int = 10): [Post!]!
}
```
when depth limit equals to 2:
valid query:
```
{
	posts(count: 100) { # depth = 1
		related(count: 100) { # depth = 2
		}
	}
}
```

invalid query:
```
{
	posts(count: 100) { # depth = 1
		related(count: 100) { # depth = 2
			related(count: 100) { # depth = 3
				related(count: 100) { # depth = 4
					title
				}
			}
		}
	}
}
```

mathewbyrne:
We have a [query complexity](https://gqlgen.com/reference/complexity/) feature that you can use to limit which queries are allowed to run.  The `handler.ComplexityLimit` option allows you to control the limit.

qhenkart:
I strongly second this. It is the field depth that causes issues, not the field length.

Kay-Zee:
Also interested in a depth based complexity limit as an option.

vektah:
We want to pull query complexity out into a plugin which will at the very least means you can provide a custom implementation without forking. 

I don't think it's as simple as what you describe though in practice. While 
```
posts -> related posts -> related posts
```
Is too deep
```
Viewer -> user > name
```
Isn't

Complexity is meant to capture what is hard and make it multiplicitivly harder the deeper you go, in this case related posts should have a high complexity, so you can't go very deep, but user, viewer and name have low complexity so going deep is ok

qhenkart:
@vektah that would be great! But in terms of your examples, if I were designing my own plugin, I would design it so all depth is considered equal. Which means both of your examples would be equally complex (and I would set the complexity number and model my subresolvers appropriately), I think the issue that folks are touching on is that 

```
posts -> related posts -> related posts
```
has the same complexity as 
```
Viewer -> user > name
```

and

```
Viewer -> user > name
``` 
has the same complexity as 
```
Viewer -> user > name
                             email
                             id
```

qhenkart:
keeping this alive because it's an important feature despite it's marking as low priority

PR #3909, opened 2025-11-11
fix(complexity): fieldComplexity calculation override allows reducing score

mahboubii:
There is a legitimate use case to consciously choose to manually _reduce_ the complexity cost associated with some known queries instead of the alternative of having to increase the global query complexity limit, since that would allow a larger attack surface.

Also see somewhat related #508 and #2367

### DIFF
--- a/complexity/complexity_test.go
+++ b/complexity/complexity_test.go
@@ -163,19 +163,6 @@ func TestCalculate(t *testing.T) {
 		requireComplexity(t, query, 2)
 	})
 
-	t.Run("custom complexity must be >= child complexity", func(t *testing.T) {
-		const query = `
-		{
-			customObject {
-				list(size: 100) {
-					scalar
-				}
-			}
-		}
-		`
-		requireComplexity(t, query, 101)
-	})
-
 	t.Run("interfaces take max concrete cost", func(t *testing.T) {
 		const query = `
 		{


StevenACoffman:
Can you say more about why you would want that?

mahboubii:
For example a single query or mutation has a complexity of 100(tons of attributes for example instead of returning serialized json) but every other op has a query cost lower than 20, you can either override the cost of this single op or increase the complexity limit for all ops to 100.

You can also argue that if an override function is provided what it returns should be respected.

StevenACoffman:
This was originally decided way back in #315 by @edsrzf and I wonder if someone should pass another explicit option or directive to make it explicit that they are not only overriding the current complexity but also exceeding any children.

That may be helping to find unintentional problems.

edsrzf:
I no longer work with gqlgen and don't have anything extra to share about my reasoning, so feel free to do whatever makes sense. 😄

StevenACoffman:
Well, thanks for your past work, and just appreciate living in New Zealand for me! It's one of the best places on Earth.

mahboubii:
Behvaiour of cost override for children remains the same here or am i missing something?

Previously it was possible to increase certain children's costs but not decreasing, which is a legitimate use case as described.  

Not sure if we can even call this a breaking change? depending on how you are approaching it can be a new feature(cost decreasing) or a bug fix(override function values not being respected without any documentation)

mahboubii:
Thanks for looking into this one @StevenACoffman, please let me know if i could provide more context or change anything in this PR.

StevenACoffman:
Related: 
+ #508 suggested making Compexity a plugin so folks could provide their own implementations
+ #2367 ComplexityLimit increments one point for Scalars as well

StevenACoffman:
Thanks! I am going to try this out in a release, but if someone reports that spec compliance precludes this, I may need to revert it.

PR #3911, opened 2025-11-13
change interface embedding from config to directive

PaulVasilenko:
Changed implementation of interface embedding from using config parameter to GraphQL directive. 

I have:


PaulVasilenko:
@adrianlungu Hey there! I've changed a plugin to follow different configuration approach, however, it would break your code, so if you want to upgrade - you would need to get rid of config variable and put directive everywhere you want to.

adrianlungu:
That's actually an interesting approach, since we don't need embedding on everything, and it reduces the surface of issues that could arise from having the plugin go through everything. 

Sounds good on my end 👍

PR #3913, opened 2025-11-13
Migrate from archived go-yaml to goccy/go-yaml

### LINKED ISSUE #3667: go-yaml is deprecated
migrate to https://github.com/goccy/go-yaml from https://github.com/go-yaml/yaml

irubey:
Replaces gopkg.in/yaml.v3 with github.com/goccy/go-yaml to address the archived status of the original library and ensure continued maintenance and security updates.

## Changes:
- Updated go.mod dependency to github.com/goccy/go-yaml v1.18.0
- Updated import and decoder API in codegen/config/config.go
- Added defensive panic recovery for malformed YAML tags
- Updated test error message assertions for new library format
- Added comprehensive fuzz testing (FuzzReadConfig)

## Fuzz Testing:
As requested in #3667, comprehensive fuzz testing was performed to assess library stability:
- **497,967 iterations** across various input patterns
- **Discovered edge case:** Malformed YAML tags (e.g., `!0000`) can trigger a nil pointer panic in the decoder
- **Impact:** Only affects invalid/corrupted input; real gqlgen configs never use YAML tags
- **Solution:** Added defensive panic recovery in `ReadConfig()` to convert panics into proper errors
- **Corpus saved:** Failing input preserved in `testdata/fuzz/` for regression testing

## Testing:
- All existing tests pass (60+ packages)
- Integration tests verified (server builds, code generation works)
- Example projects compile and run

The new error messages from goccy/go-yaml are more user-friendly, including visual pointers to show exactly where errors occur.

Fixes #3667

I have:

### DIFF
--- a/codegen/config/config_test.go
+++ b/codegen/config/config_test.go
@@ -35,11 +35,8 @@ func TestReadConfig(t *testing.T) {
 		t.Cleanup(func() { _ = cfgFile.Close() })
 		_, err = ReadConfig(cfgFile)
 
-		if runtime.GOOS == "windows" {
-			require.EqualError(t, err, "unable to parse config: [1:1] string was used where mapping is expected\n>  1 | asdf\r\n       ^\n")
-		} else {
-			require.EqualError(t, err, "unable to parse config: [1:1] string was used where mapping is expected\n>  1 | asdf\n       ^\n")
-		}
+		actualErr := strings.ReplaceAll(err.Error(), "\r\n", "\n")
+		require.Equal(t, "unable to parse config: [1:1] string was used where mapping is expected\n>  1 | asdf\n       ^\n", actualErr)
 	})
 
 	t.Run("unknown keys", func(t *testing.T) {
@@ -48,11 +45,8 @@ func TestReadConfig(t *testing.T) {
 		t.Cleanup(func() { _ = cfgFile.Close() })
 		_, err = ReadConfig(cfgFile)
 
-		if runtime.GOOS == "windows" {
-			require.EqualError(t, err, "unable to parse config: [2:1] unknown field \"unknown\"\n   1 | schema: outer\r\n>  2 | unknown: foo\n       ^\n")
-		} else {
-			require.EqualError(t, err, "unable to parse config: [2:1] unknown field \"unknown\"\n   1 | schema: outer\n>  2 | unknown: foo\n       ^\n")
-		}
+		actualErr := strings.ReplaceAll(err.Error(), "\r\n", "\n")
+		require.Equal(t, "unable to parse config: [2:1] unknown field \"unknown\"\n   1 | schema: outer\n>  2 | unknown: foo\n       ^\n", actualErr)
 	})
 
 	t.Run("globbed filenames", func(t *testing.T) {


codegen/config/testdata/fuzz/FuzzReadConfig/d76a9a281aa1168d
StevenACoffman: Neat! I'm not as familiar with fuzzing practices. Is it common for the corpus files to be checked into the repository?

StevenACoffman:
I think you are going to want to fix the `_examples` directory dependencies by doing this:
```
cd _examples
go mod tidy
cd ..
```

StevenACoffman:
Thanks for doing this! I'm shocked it was such a minor change. There's some windows test failures because of the `\r\n` line ending differences.

codegen/config/config_test.go
```diff
@@ -34,15 +34,15 @@ func TestReadConfig(t *testing.T) {
...
+		require.EqualError(t, err, "unable to parse config: [1:1] string was used where mapping is expected\n>  1 | asdf\n       ^\n")
 	})
 
 	t.Run("unknown keys", func(t *testing.T) {
 		cfgFile, err := os.Open("testdata/cfg/unknownkeys.yml")
 		require.NoError(t, err)
 		t.Cleanup(func() { _ = cfgFile.Close() })
 		_, err = ReadConfig(cfgFile)
-		require.EqualError(t, err, "unable to parse config: yaml: unmarshal errors:\n  line 2: field unknown not found in type config.Config")
+		require.EqualError(t, err, "unable to parse config: [2:1] unknown field \"unknown\"\n   1 | schema: outer\n>  2 | unknown: foo\n       ^\n")
```
StevenACoffman:
```suggestion
		require.EqualError(t, err, "unable to parse config: [2:1] unknown field \"unknown\"\n   1 | schema: outer"+string(os.PathSeparator)+">  2 | unknown: foo\n       ^\n")
```

codegen/config/config_test.go
```diff
@@ -310,3 +310,33 @@ func TestLoadSchema(t *testing.T) {
...
+	f.Add([]byte(`schema: schema.graphql
+model:
+  filename: models.go
+  package: models
+exec:
+  filename: exec.go`))
+	f.Add([]byte(``))
+	f.Add([]byte(`asdf`))
+	f.Add([]byte(`schema: outer
+unknown: foo`))
```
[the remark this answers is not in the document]
StevenACoffman: Ah, that makes what you were doing make much more sense. Thanks for unpacking that for me!

irubey:
Okay after some investigation into the CI failures a few things came up:

1. Circular deps issue - goccy is a transitive dependency of the `_examples` module, so when examples run `go mod tidy` it would remove the dep, then the next example would fail because the dep was missing. 
- Solved this by adding `goccy/go-yaml` to `_examples/tools.go` as a blank import so it's now a direct dependency and doesn't get cleared by tidy 


2. `_examples/enum/gqlgen.yaml` file failed to parse - This file had unquoted bools which passed in old package, caught in new package
- quoted the bools properly


3. Windows line endings issues - the tests I updated assumed that the error codes were platform specific but looking at the error code in go-yaml, it normalizes all error messages to use `\n` on all platforms.
- updated test to expect `\n` on all platforms


I ran local tests and the check-generate workflow and all tests pass and workflow passes, so I think this should go green

irubey:
Looking at the windows failure apparently goccy/go-yaml is normalizing to \n, but it's using the original YAML errors which include windows line endings. so the error message template is \n but the embedded yaml message has the \r\n. 🤦‍♂️

irubey:
Now the error is exactly opposite, expecting `\r\n` but getting `\n` ????

StevenACoffman:
I actually had found this same error super maddening the last time I had to deal with it, as I had a heckuva time tracking down where this was coming from. The pattern of just making the test have the platform specific string was a hack since I punted.

irubey:
Yeah, a tricky bug for sure! Not sure what caused the inconsistency, but normalization should handle both LF and CRLF. Thanks for reviewing!

PR #3928, opened 2025-11-19
feat(resolveField): allow field middlewares to return a marshaler directly

lbarthon:
In some specific cases, we'd like the ability from a field resolver to return a marshaler directly. This is needed for specific performance-sensitive codepaths we have in our application, where part of the request might already be resolved and we don't want to resolve it again, as the performance hit from doing so would be too high.

This shouldn't impact any already existing usecase, and only make those field middlewares more capable.

I have:

### DIFF
--- a/graphql/resolve_field.go
+++ b/graphql/resolve_field.go
@@ -126,17 +126,20 @@ func resolveField[T, R any](
 		}
 		return defaultResult
 	}
-	res, ok := resTmp.(T)
-	if !ok {
-		var t T
-		oc.Errorf(
-			ctx,
-			`unexpected type %T from middleware/directive chain, should be %T`,
-			resTmp,
-			t,
-		)
-		return defaultResult
+	if res, ok := resTmp.(T); ok {
+		fc.Result = res
+		return result(ctx, res)
 	}
-	fc.Result = res
-	return result(ctx, res)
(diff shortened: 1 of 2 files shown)


lbarthon:
@StevenACoffman is that something you'd consider landing? will fork otherwise, as this is much needed. happy to add test in followup diffs

StevenACoffman:
I would be ok with merging this, but would you be ok with adding some test coverage and/or documentation for this new behavior?

StevenACoffman:
@lbarthon I was checking in on this, as I would like to merge this before the next release, but I'm holding off merging this PR until you get a chance to add some test coverage and/or documentation for this new behavior. Do you anticipate being able to spend some time on that soon?

lbarthon:
Yes @StevenACoffman, on it now

lbarthon:
I added tests - is that enough? Got another PR I want to make, working on this now

StevenACoffman:
Thanks! BTW, you refer to "we", and I'm curious who "we" is? Are you working with a group of developers? Is there a public project I can peek at somewhere? If you can't speak about it publically, maybe reach out to me directly on linkedin at https://www.linkedin.com/in/steve-coffman-79322175/ or via my work email steve@khanacademy.org

PR #3930, opened 2025-11-23
add decimal scalar

it512:
add decimal scalar
I have:
 - [ x] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [ x] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))


StevenACoffman:
This adds a new dependency to generated code, without an option to customize which decimal library to use, so I'm not sure if that's ideal, but it looks like this might be the most popular choice?

At the same time, the generated code and format tests are currently still failing, so that's what's holding me back from merging it.

PR #3939, opened 2025-11-25
Add functional options to control complexity calculation

vibridi:
No AI was ever involved in the creation of this PR.

The PR adds functional options to `complexity.Calculate` and `extension.FixedComplexityLimit`. Clients can control what complexity value scalars and enums should have (in some literature that's expected to be 0, despite gqlgen defaulting to 1). Clients can also control which fields to outright ignore (this helps with graphql clients that automatically add fields to each request, such as Apollo adding `__typename` fields).

Changes:
- Add `Option` type to `complexity` package
- Add `WithFixedScalarValue` option to control complexity value of scalars and enums
- Add `WithIgnoreFields` option to control which fields should not contribute to the complexity calculation
- Add vararg `...Option` to `complexity.Calculate`. Usage of varargs allows to preserve the behavior of existing call sites.
- Add check for fields to ignore in `selectionSetComplexity`
- Add scalar and enum default complexity evaluation in `complexityWalker.fieldComplexity`. When no option is supplied, it defaults to `1` to maintain backward compatibility
- Add vararg `...complexity.Option` to `extension.FixedComplexityLimit`. The options are passed to `complexity.Calculate`

I have also:

### DIFF
--- a/complexity/complexity.go
+++ b/complexity/complexity.go
@@ -13,11 +13,17 @@ func Calculate(
 	es graphql.ExecutableSchema,
 	op *ast.OperationDefinition,
 	vars map[string]any,
+	opts ...Option,
 ) int {
+	os := defaultOptions
+	for _, o := range opts {
+		o(&os)
+	}
 	walker := complexityWalker{
 		es:     es,
 		schema: es.Schema(),
 		vars:   vars,
+		opts:   os,
 	}
 	return walker.selectionSetComplexity(ctx, op.SelectionSet)
 }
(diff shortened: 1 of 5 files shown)


vibridi:
Hi @StevenACoffman I fixed a bug yesterday after the first workflow run, please can you run the workflows again? Thanks

vibridi:
formatted according to `gofumpt`

StevenACoffman:
Thanks! I think this gives people a great deal of flexibility.

PR #3940, opened 2025-11-25
Optional json marshallers generation

PaulVasilenko:
https://github.com/99designs/gqlgen/pull/3663 modified the way how json library work with the type across the whole codebase, not only in GQL related code, which may lead to very unexpected issues for users.

It is better to make generating those marshalers optional, however it is already a default lib behaviour, so making opposite - giving a fallback for users out of this situation.

I have:


StevenACoffman:
Thanks! I really appreciate your continuing to contribute and improve things!

edeetee:
Thank you. This was an unexpected breaking change but its very good to have a solution already here for us.

PR #3945, opened 2025-12-01
feat(codegen): allow users to manually extend GraphQL unions

lbarthon:
In some scenarios, we need caching or prefetching of parts of the GraphQL query tree. This is useful if we have a very complex tree and common fragments. This can allow us to cache the marshalers for some fields that can appear more than once in the schema.

We use it in a real-world scenario for lookahead fetching. A query returns IDs of an entity, let's say, an User. For those entities, we have an extra layer to handle visibility filtering, i.e. am I allowed to see said users (are they blocked, hidden, restricted, etc.).
The query looks like that:
```graphql
query {
  listUsers {
    userResult {
      ... on User {
        fullname
        age
      }
      ... on UserUnavailable {
        error
      }
    }
  }
}
```

A plain GraphQL implementation, would do, in order:
- Resolve listUsers
- Resolve userResult
- Resolve the fields on the user fragment

We implement lookahead, which will do:
- Resolve listUsers
- Resolve userResult **AND** the fields on the user fragment
- Return the user fragment only if userResult is an User

Lookahead makes requests faster, at the expense of more resolver calls (thus potentially more RPCs, DB calls, or whatever your resolvers do). We implement it by having a different type, `LookaheadUserResult`, which is a new member of the `UserResult` union (which can be either `User` or `UserUnavailable` in the query above). This ensures we satisfy type requirements, however this change by itself fails without the template changes of this commit.

This commit extends the generated code for unions: if generated union types are returned, then we proceed with standard resolving. However, if we receive a variant that satisfies type constraints (otherwise it won't build eh), but that the codegen doesn't know about, check if it is a `graphql.Marshaler`, and just return it, for the parent to call `MarshalGQL` on, as we do for all `Marshaler`!

This is the cleanest way I found to do this, to add a tiny bit of flexibility without breaking the type system altogether.

I have:


lbarthon:
@StevenACoffman would love that in the next release too!

lbarthon:
cc @StevenACoffman 👀

StevenACoffman:
@lbarthon Sorry, I've been sitting on this for a bit!

I've found this PR actually quite fascinating! I wanted time to play around with it a bit, as it has some very interesting implications for many uses that you _didn't_ mention.

However, I'm not as sure I understand how you are successfully using this for lookahead or caching from your example. How would fields be resolved for your LookaheadUserResult type? You mentioned that you have an extra layer to handle visibility filtering, so perhaps this fits in somehow?

BTW, you refer to "we", and I'm curious who "we" is? Are you working with a group of developers? Is there a public project I can peek at somewhere? If you can't speak about it publicly, maybe reach out to me directly on linkedin at https://www.linkedin.com/in/steve-coffman-79322175/ or via my work email [steve@khanacademy.org](mailto:steve@khanacademy.org)

lbarthon:
Thanks @StevenACoffman for landing this!

polastre:
@StevenACoffman here's a fix for the bug noted above ^^

https://github.com/99designs/gqlgen/pull/3971

PR #3949, opened 2025-12-02
fix(introspection): support @deprecated directive on field arguments

AlexanderArvidsson:
Field arguments are currently not introspected as deprecated due to a typo in the field arguments definition in the introspection code. It uses the directives from the field instead of the directives on the field argument:
```diff
		for _, arg := range f.Arguments {
			args = append(args, InputValue{
				Type:         WrapTypeFromType(t.schema, arg.Type),
				Name:         arg.Name,
				description:  arg.Description,
				DefaultValue: defaultValue(arg.DefaultValue),
-				deprecation:  f.Directives.ForName("deprecated"),
+				deprecation:  arg.Directives.ForName("deprecated"),
			})
```

The **[GraphQL spec was updated in September](https://spec.graphql.org/September2025/#sec--deprecated)** to allow deprecation on field arguments, and gqlgen seems to have followed, though not working due to the above typo.

I have also added tests, although I must admit I am not entirely sure what the difference is between `testserver/singlefile` and `testserver/followschema`. However, I mimicked what I saw from other tests.

I also added introspection tests in the `integration/` folder, but running the GraphQL codegen generated code that was unrelated to my changes (perhaps these have not been re-generated for a while, and GraphQL codegen changed?), so I have not included that. I'm open to add those changes to this PR if you're OK with extra stuff being added by codegen.

I have:

### DIFF
--- a/codegen/testserver/followschema/introspection_test.go
+++ b/codegen/testserver/followschema/introspection_test.go
@@ -67,6 +67,45 @@ func TestIntrospection(t *testing.T) {
 			require.Equal(t, "id", resp.Type.Fields[0].Name)
 			require.Nil(t, resp.Type.Fields[0].DeprecationReason)
 		})
+
+		t.Run("deprecated directive on field arguments", func(t *testing.T) {
+			var resp struct {
+				Type struct {
+					Fields []struct {
+						Name string
+						Args []struct {
+							Name              string
+							DeprecationReason *string
+						}
+					}
+				} `json:"__type"`
+			}
+
+			err := c.Post(
+				`{ __type(name:"Query") { fields { name args { name deprecationReason }}}}`,
+				&resp,
+			)
+			require.NoError(t, err)
+
+			var args []struct {
+				Name              string
+				DeprecationReason *string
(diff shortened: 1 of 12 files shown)


AlexanderArvidsson:
Fixed the linting, I did not see it because I had v1 of golangci-lint instead of v2 which caused it to not run in Neovim 😓

StevenACoffman:
Thanks for this! I could also use help updating https://github.com/vektah/gqlparser (which gqlgen relies on) to the latest spec update.

AlexanderArvidsson:
> Thanks for this! I could also use help updating https://github.com/vektah/gqlparser (which gqlgen relies on) to the latest spec update.

I might have a look if I find some time, am happy to help!
This one case I only found when deprecating a field and wondering why my introspected schema didn't include it.

StevenACoffman:
@AlexanderArvidsson If you get a chance, https://github.com/vektah/gqlparser/pull/401 is where I started to update it, and then there are just various test failures to work through. Anything you can contribute would be a huge help!

PR #3954, opened 2025-12-06
fix(codegen): correctly unmarshal slice of input object mapped to map…

### LINKED ISSUE #3953: Panic: Unmarshalling recursive input slice mapped to map[string]interface{}
When an `input` object is mapped to `map[string]interface{}` in `gqlgen.yml` and is used as a list field (e.g., `[TodoWhereInput!]`), the generated unmarshalling code is incorrect.

The generated function for the slice field attempts to pass the entire slice (as `[]interface{}`) directly to the single-item unmarshaller function, instead of iterating over the elements. This causes a runtime panic:
`panic: interface conversion: interface {} is []interface {}, not map[string]interface {}`

### Steps to reproduce

1.  Define an input type in your schema that contains a list field of itself (recursive) or just a list of that type.
```graphql
input TodoWhereInput {
  textContains: String
  done: Boolean
  AND: [TodoWhereInput!]
}

type Query {
  todos(where: TodoWhereInput): [Todo!]!
}
```

2.  Map this input type to `map[string]interface{}` in `gqlgen.yml`.
```yaml
models:
  TodoWhereInput:
    model:
      - map[string]interface{}
```

3.  Run the server and execute a query that uses the list field.
```graphql
{
  todos(where: {AND: [{textContains: "x"}, {done: true}]}) {
    id
  }
}
```
## Expected Behavior
The generated code should iterate over the input slice, unmarshal each item, and return a slice of maps.

### Actual Behavior
The application panics.

**Generated Code:**
```go
// graph/generated.go

// Incorrect signature: returns map[string]any instead of []map[string]any
func (ec *executionContext) unmarshalOTodoWhereInput2mapᚄ(ctx context.Context, v any) (map[string]any, error) {
	if v == nil {
		return nil, nil
	}
    // v is []interface{}, but it is passed to a function that expects map[string]interface{}
	res, err := ec.unmarshalInputTodoWhereInput(ctx, v)
}

func (ec *executionContext) unmarshalInputTodoWhereInput(ctx context.Context, obj any) (map[string]any, error) {
    // Panic occurs here because obj is []interface{}
	it := make(map[string]any, len(obj.(map[string]any)))
    // ...
}
```

### Versions
- `gqlgen`: v0.17.84
- `go`: 1.25.1

**Note:** this issue is similar to #3657 but there the input is an array, it is not the same case here.

rodcorsi:
This fix #3953

I have:

### DIFF
--- a/codegen/testserver/singlefile/maps.graphql
+++ b/codegen/testserver/singlefile/maps.graphql
@@ -30,3 +30,12 @@ scalar CustomScalar @goModel(model: "singlefile.CustomScalar")
 input NestedMapInput {
     map: MapStringInterfaceInput
 }
+
+input MapNestedMapSliceInput @goModel(model: "map[string]interface{}") {
+    name: String
+    recurse: [MapNestedMapSliceInput!]
+}
+
+extend type Query {
+    mapNestedMapSlice(input: MapNestedMapSliceInput): Boolean
+}
--- a/codegen/config/binder.go
+++ b/codegen/config/binder.go
@@ -421,7 +421,7 @@ func (b *Binder) TypeReference(
 			return &TypeReference{
 				Definition: def,
 				GQL:        schemaType,
-				GO:         MapType,
+				GO:         b.CopyModifiersFromAst(schemaType, MapType),
 				IsRoot:     b.cfg.IsRoot(def),
 			}, nil
 		}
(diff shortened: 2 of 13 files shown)


codegen/testserver/singlefile/maps.graphql
```diff
@@ -30,3 +30,12 @@ scalar CustomScalar @goModel(model: "singlefile.CustomScalar")
 input NestedMapInput {
     map: MapStringInterfaceInput
 }
+
```
StevenACoffman: Thanks for adding a specific example here so we can better avoid regressions and demonstrate correctness!

StevenACoffman:
Hey, I'm not able to look into this more deeply right now, but I appreciate you working on this! It looks like there's some test failures currently.

rodcorsi:
Hey, I have fixed the tests, thanks
cc @StevenACoffman

StevenACoffman:
Thanks very much for this!

PR #3971, opened 2025-12-17
fix: do not use if-else scoped var in type assertion

polastre:
The variable `obj` is redeclared in the `if` block, so the `else` statement is using the locally scoped `obj` instead of one the one that was being evaluated in the switch statement for a type assertion. Change the variable name so that it doesn't shadow the original `obj` definition.

I have:

### DIFF
--- a/codegen/interface.gotpl
+++ b/codegen/interface.gotpl
@@ -24,8 +24,8 @@ func (ec *executionContext) _{{$interface.Name}}(ctx context.Context, sel ast.Se
 			{{- end }}
 	{{- end }}
 	default:
-		if obj, ok := obj.(graphql.Marshaler); ok {
-			return obj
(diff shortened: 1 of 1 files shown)


StevenACoffman:
Thank you! There's a ton of this variable shadowing still lurking from the old days.  Always glad to merge PRs that clean some of it up. Can you re-generate and commit the result?
```
go generate ./...
```

StevenACoffman:
I closed this and re-opened in #3972 so I could apply the go generate.

Thanks for contributing the fix!

polastre:
Thanks for the quick turn @StevenACoffman, sorry I didn't get to `go generate` sooner!

StevenACoffman:
No worries! There's a bunch of other similar variable shadowing in other places, so any more help with that would be much appreciated!

PR #3972, opened 2025-12-19
fix: do not use if-else scoped var in type assertion

StevenACoffman:
From @polastre

The variable `obj` is redeclared in the `if` block, so the `else` statement is using the locally scoped `obj` instead of one the one that was being evaluated in the switch statement for a type assertion. Change the variable name so that it doesn't shadow the original `obj` definition.

I have:


polastre:
@StevenACoffman any chance we can get a release cut with this included? thanks!

PR #3978, opened 2025-12-24
Caching for packages loaded by a relative path

### LINKED ISSUE #3961: Relative package path references always miss the cache
When using relative package paths for model binding, the `packages.Load` cache is being missed every time, resulting in really slow performance.

For example, this is using the relative (much shorter) package path. This is the one that's slow. Each binding results in a separate `packages.Load` call.
```go
input Match_Time @goModel(model: "../../pkg/match.Time") {
  ...
}
```

And this works as expected, with `packages.Load` calls being cached and reused across multiple binding lookups.
```go
input Match_Time @goModel(model: "github.com/{org}/{repo}/apps/{app}/pkg/match.Time") {
  ...
}
```

For a relatively small GraphQL schema, with ~40 goModel/goEnum mappings, this takes generation performance from less than 2 seconds to over 45 seconds. The obvious fix is to not use relative package paths, though I feel gqlgen should handle this a bit better.

deitrix:
Fixes #3961

### DIFF
--- a/internal/code/packages_test.go
+++ b/internal/code/packages_test.go
@@ -57,7 +57,7 @@ func TestPackages(t *testing.T) {
 		require.Equal(t, 1, p.numLoadCalls)
 	})
 
-	t.Run("albe to load relative package again after evict", func(t *testing.T) {
+	t.Run("able to load relative package again after evict", func(t *testing.T) {
 		p := initialState(t)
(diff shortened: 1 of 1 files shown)


StevenACoffman:
Thank you! I really appreciate your improving things! Looking forward to your next PR!

PR #3985, opened 2026-01-06
Add gqlgen.yml schema configuration file

Namyts:
I would like my IDE (VSCode) to understand the expected schema for the `gqlgen.yml` file. I generated, proofread, and quickly tested this schema file.

This removes the slightly frustrating choice of commenting out most properties and their descriptions in my gqlgen.yml vs cleaning up the file and having to refer back to the documentation whenever I make changes.

There may be difficulties keeping this in sync with the source `./init-templates/gqlgen.yml.gotmpl`, but overall this standard is probably more useful long-term

To use the following to the top of the `gqlgen.yml`:
```yaml
# yaml-language-server: $schema=[LINK TO THIS FILE]
```

I have:

### DIFF
--- a/gqlgen.schema.json
+++ b/gqlgen.schema.json
@@ -1,0 +1,217 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "gqlgen Configuration",
+  "description": "Configuration file for gqlgen code generation",
+  "type": "object",
+  "properties": {
+    "schema": {
+      "description": "Where are all the schema files located? globs are supported",
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
+    },
+    "exec": {
+      "description": "Where should the generated server code go?",
+      "type": "object",
+      "properties": {
+        "filename": {
+          "type": "string",
+          "description": "Path to the generated file (required for single-file layout)"
+        },
+        "package": {
+          "type": "string",
+          "description": "The go package name"
+        },
+        "layout": {
+          "type": "string",
+          "enum": ["single-file", "follow-schema"],
+          "default": "single-file"
+        },
+        "dir": {
+          "type": "string",
+          "description": "Directory for generated files (required for follow-schema layout)"
+        },
+        "filename_template": {
+          "type": "string",
+          "default": "{name}.generated.go",
+          "description": "Template for filenames (required for follow-schema layout)"
+        },
+        "worker_limit": {
+          "type": "integer",
+          "description": "Maximum number of goroutines in concurrency to use per child resolvers",
+          "default": 1000
+        }
+      }
+    },
+    "federation": {
+      "description": "Federation configuration",
+      "type": "object",
+      "properties": {
+        "filename": { "type": "string" },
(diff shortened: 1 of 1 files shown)


Namyts:
Looks like this.

image

StevenACoffman:
Thanks! This is very useful! Sorry for the delay in getting back to you. We should at least document that this is even an option for folks. It looks like this same config would also work for other IDE's like JetBrains (GoLand)? 

What should the default schema location be? "https://raw.githubusercontent.com/99designs/gqlgen/refs/heads/master/schemas/gqlgen.schema.json"

Is there some authoritative place we should consider registering our config schema? It looks like maybe [schemastore](https://www.schemastore.org/) maybe? Looking at 
https://github.com/SchemaStore/schemastore/pull/5273/changes it seems like a PR like this:

```
    {
      "name": "gqlgen GraphQL server go generate configuration",
      "description": "gqlgen configuration to go generate a GraphQL server. See https://gqlgen.com/config/",
      "fileMatch": [
        "**/gqlgen.yaml", 
        "**/gqlgen.yml"
      ],
      "url": "https://raw.githubusercontent.com/99designs/gqlgen/refs/heads/master/schemas/gqlgen.schema.json"
    },
```

We could add a configuration option so that they could add (and override the default) the schema location when we init a new gqlgen config. 

golangci-lint uses their own configuration schema to enable configuration verification but chose **_not_** to embed their configuration schema into their binary in https://github.com/golangci/golangci-lint/issues/4646 . However I get rate limit flakiness in my CI as a result, so I'm kind of personally annoyed by this decision on their part.

We could alternatively try moving this file into the `docs/static/` directory, like we do for `main.css`, so people can access it using `https://gqlgen.com/gqlgen.schema.json` by default. (At least, I think so, but the website's Hugo render.com setup was made by 99designs folk before my time and is unclear to me.) That *might* allow more generous rate limits? Or less? I don't have any idea.

Namyts:
Thanks for merging!

SchemaStore could be a good idea 👀 

It seems to support [versioning](https://github.com/SchemaStore/schemastore/pull/5270/changes#diff-1b8f038f5afb1158263a1fc83b9c0ca5a7438cceb90adad99d87168f70edc815R9175-R9177), hosting, and even [storing the schema itself](https://github.com/SchemaStore/schemastore/pull/5270/changes#diff-a026488cc1bc9323141cf6c3fce30b99cfedb3bdd73b37ac74a02a4f23eaa54eR1) - all of which would solve the rate limiting issues. But probably best (like you suggested) to point to this repo, keep default as `master`, and per-version use the tags `https://raw.githubusercontent.com/99designs/gqlgen/refs/tags/v0.17.86/gqlgen.schema.json`. Might be a good idea to create a github action to create a PR whenever the schema file is modified. SchemaStore seems to have considered [code ownership](https://github.com/OSS-Docs-Tools/code-owner-self-merge) to automate PR approvals!

For the file location within this repo - completely up to you (I am just happy it exists 😅 ). It's a meta file so a schema folder or keeping in the root make sense to me.

PR #4017, opened 2026-02-07
refactor(codegen): use graphql.Config type

tomoikey:
### Description
Some of the code currently living in the templates doesn't need to be part of the generated output. Moving these pieces into static .go files helps keep generated.go simpler and easier to maintain.

There are several places where this applies, so I'll be sending a series of smaller PRs rather than one large change. As a first step, this PR extracts the Config struct into the graphql package.

I have:

### DIFF
--- a/codegen/testserver/generated_test.go
+++ b/codegen/testserver/generated_test.go
@@ -1,16 +1,17 @@
 package testserver
 
 import (
-	"fmt"
+	"bytes"
 	"go/ast"
 	"go/parser"
+	"go/printer"
 	"go/token"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 	"testing"
 
-	eqgo "github.com/kevinmbeaulieu/eq-go/eq-go"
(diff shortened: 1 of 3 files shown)


codegen/testserver/generated_test.go
```diff
@@ -21,19 +22,15 @@ func TestLayouts(t *testing.T) {
...
-		singlefileFSet,
-		followschemaPkg,
-		followschemaFSet,
-		nil,
-	)
-	if !eq {
-		// When msg is too long, require.True(...) omits it entirely.
-		// Therefore use fmt.Fprintln to print it manually instead.
-		fmt.Fprintln(os.Stderr, msg)
-		require.Fail(t, "Packages not equivalent")
```
tomoikey: eq-go does not support Go 1.18 generics (ast.IndexListExpr) and panics when it encounters them.
Since the library is no longer maintained, this PR replaces it with comparison using the standard library's go/printer.

PR #4021, opened 2026-02-10
federation: apply OBJECT-level directives to entity resolvers

### LINKED ISSUE #3910: Directives not applied to entity resolvers in Federation mode
Hi, gqlgen team!

We came across an issue with federation generation. Gqlgen does not generate directives on entity resolvers, which creates a security gap when using Apollo Federation.

### Problem Description
When using Apollo Federation, directives (like `@guard`) are correctly applied to query resolvers but are completely bypassed for entity resolvers (`_entities` query).
This means that authorization checks work for direct queries but being skipped when Apollo Router fetches the same entity through federation's entity resolution mechanism.

### Environment
gqlgen version: v0.17.63
Federation version: v2
Go version: 1.24

```
# schema
directive @guard(name: String, parent: String, skip: String) on OBJECT

type Person @key(fields: "id") @guard(name: "PersonGuard") {
    id: String!
    email: String!
    phone: String!
}

extend type Query {
    getPerson(id: String!): Person
}
```

gqlgen configuration
```
# gqlgen.yml
federation:
  version: 2
```

generated code for query resolver
```
// generated/schema.generated.go
func (ec *executionContext) _Query_getPerson(ctx context.Context, field graphql.CollectedField) graphql.Marshaler {
    resTmp, err := ec.ResolverMiddleware(ctx, func(rctx context.Context) (any, error) {
        directive0 := func(rctx context.Context) (any, error) {
            return ec.resolvers.Query().GetPerson(rctx, personID)
        }

        directive1 := func(ctx context.Context) (any, error) {
            name, err := ec.unmarshalOString2ᚖstring(ctx, "PersonGuard")
            if err != nil {
                return nil, err
            }
            // ✅ Guard IS applied here
            return ec.directives.Guard(ctx, nil, directive0, name, nil, nil)
        }

        return directive1(rctx)
    })
    // ...
}
```

generated code for entity resolver (directivies aren't being applied)
```
// generated/federation.go
func (ec *executionContext) resolveEntity(
    ctx context.Context,
    typeName string,
    rep EntityRepresentation,
) (e fedruntime.Entity, err error) {
    switch typeName {
    case "Person":
        id, err := ec.unmarshalNString2string(ctx, rep["id"])
        if err != nil {
            return nil, err
        }
        // ❌ Guard is NOT applied - direct resolver call
        entity, err := ec.resolvers.Entity().FindPersonByID(ctx, id)
        if err != nil {
            return nil, err
        }
        return entity, nil
    }
}
```

ravisastryk:
Fixes #3910: Federation entity resolvers now correctly execute OBJECT-level directives.

When using Apollo Federation, OBJECT-level directives (example:- `@auth`, `@guard`) applied to federated entities were not being executed on entity resolvers. Regular queries worked fine, but `_entities` queries bypassed these directives entirely.

Previously, directives declared on a type such as:
```
type Person @key(fields: "id") @guard(name: "PersonGuard")
```

were enforced for Query.getPerson but silently skipped when the Apollo Router resolved the entity through federation's _entities query.

**High Level Changes**

1. Added `ImplDirectives` field to `Entity` struct
2. Populated directives during code generation
* Filters out federation-internal directives (@key, @requires, etc.)
* Preserves user-defined OBJECT-level directives
3. Applied directive chains in entity resolver template
* Wraps entity resolver calls with directive middleware
* Maintains same behavior as regular query resolvers

**Tests**

* Unit tests for directive detection and filtering
* Integration test with schema including multiple directive types
* Added `//go:generate` directives for automatic testdata regeneration
* Existing federation tests pass

**Note:**
* Improved testability by moving template logic into library code
* minor code refactors

### DIFF
--- a/_examples/embedding/subdir/federation_gen.go
+++ b/_examples/embedding/subdir/federation_gen.go
@@ -10,6 +10,11 @@ import (
 	"github.com/99designs/gqlgen/plugin/federation/fedruntime"
 )
 
+var (
+	ErrUnknownType  = errors.New("unknown type")
+	ErrTypeNotFound = errors.New("type not found")
+)
+
 func (ec *executionContext) __resolve__service(ctx context.Context) (fedruntime.Service, error) {
 	if ec.DisableIntrospection {
 		return fedruntime.Service{}, errors.New("federated introspection disabled")
--- a/_examples/embedding/subdir/gendir/federation_gen.go
+++ b/_examples/embedding/subdir/gendir/federation_gen.go
@@ -10,6 +10,11 @@ import (
 	"github.com/99designs/gqlgen/plugin/federation/fedruntime"
 )
 
+var (
+	ErrUnknownType  = errors.New("unknown type")
+	ErrTypeNotFound = errors.New("type not found")
+)
+
 func (ec *executionContext) __resolve__service(ctx context.Context) (fedruntime.Service, error) {
 	if ec.DisableIntrospection {
 		return fedruntime.Service{}, errors.New("federated introspection disabled")
--- a/plugin/federation/testdata/entitydirectives/generated/exec.go
+++ b/plugin/federation/testdata/entitydirectives/generated/exec.go
@@ -22,12 +22,7 @@ import (
 
 // NewExecutableSchema creates an ExecutableSchema from the ResolverRoot interface.
 func NewExecutableSchema(cfg Config) graphql.ExecutableSchema {
-	return &executableSchema{
-		schema:     cfg.Schema,
-		resolvers:  cfg.Resolvers,
-		directives: cfg.Directives,
-		complexity: cfg.Complexity,
-	}
+	return &executableSchema{SchemaData: cfg.Schema, Resolvers: cfg.Resolvers, Directives: cfg.Directives, ComplexityRoot: cfg.Complexity}
 }
 
 type Config = graphql.Config[ResolverRoot, DirectiveRoot, ComplexityRoot]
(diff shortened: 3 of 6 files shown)


StevenACoffman:
Thanks @ravisastryk ! The more code that can be moved _out_ of codegen templates, and _into_ shared library code in packages like `graphql` or `federation` the better.

Codegen templates are much more difficult to statically analyze, test, and pose a substantial barrier to most contributions.

The interactions between the results of rendered code from two _different_ codegen templates with all the possible permutations makes this exponentially more difficult.

I really appreciate you not only finding this problem, but also working on fixing it and being willing to improve gqlgen at the same time!

ravisastryk:
Thank you @StevenACoffman for the comment. Can you please re-review?

StevenACoffman (approved):
Ok, thanks! Unfortunately, we had some big PRs land recently that caused significant conflicts. @tomoikey  FYI

@ravisastryk  Can you resolve the conflicts?

ravisastryk:
> Ok, thanks! Unfortunately, we had some big PRs land recently that caused significant conflicts. @tomoikey FYI
> 
> @ravisastryk Can you resolve the conflicts?
Thank you @StevenACoffman I have rebased, so we should be good.

ravisastryk:
> Ok, thanks! Unfortunately, we had some big PRs land recently that caused significant conflicts. @tomoikey FYI
> 
> @ravisastryk Can you resolve the conflicts?

@StevenACoffman All the conflicts have been resolved, awaiting final approval and merge. Thank you.

PR #4026, opened 2026-02-17
fix: run `go generate ./...` & `golangci-lint run --fix`

tomoikey:
### Description
It looks like there are some differences in the go generate output.
This seems to have been caused by a merge that went through without rebasing after #4019 was merged.
I ran `go generate ./...` to fix this up.

Also, while creating this PR, I noticed that the lint was failing, so I fixed it.

image

### Suggestion
As a countermeasure, it might be a good idea to change the repository settings to enforce running "Update branch" when the base branch is outdated.

I have:

### DIFF
--- a/codegen/directive.go
+++ b/codegen/directive.go
@@ -186,12 +186,11 @@ func (d *Directive) Declaration() string {
 
 	var resSb173 strings.Builder
 	for _, arg := range d.Args {
-		resSb173.WriteString(
-			fmt.Sprintf(
-				", %s %s",
-				templates.ToGoPrivate(arg.Name),
-				templates.CurrentImports.LookupType(arg.TypeReference.GO),
-			),
+		fmt.Fprintf(
+			&resSb173,
+			", %s %s",
+			templates.ToGoPrivate(arg.Name),
+			templates.CurrentImports.LookupType(arg.TypeReference.GO),
 		)
 	}
 	res += resSb173.String()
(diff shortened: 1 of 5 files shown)


StevenACoffman:
Screenshot 2026-02-17 at 9 22 23 AM
I flipped this option on, but I'm not sure if you meant to suggest to change another setting as well @tomoikey

StevenACoffman:
@tomoikey 

Just a side note. I had noticed that the PR was not rebased, and I knew it would require a follow-up `go generate ./...`, but I merged it anyway. I knew that this would then "bubble up" to cause the same problem for your PR, but merged yours expecting to just fix both at the same time.

For minor problems that I feel confident I can quickly manually sort out afterwards, I often will ask but _not_ require first-time casual contributors to rebase or `go generate ./...` or `golangci-lint fmt --config=./.golangci.yml .` prior to my merging their PRs.

I've found that if the contribution process is tedious, many contributors will never complete it. If their first-time contribution experience is quick, smooth, and welcoming, then contributors are far more likely to come back and repeat their experience. They are then more invested in the project and more willing to learn the etiquette to make maintenance easier for everyone.

tomoikey:
@StevenACoffman 
really great! it honestly hadn't occurred to me to think about it that way. Prioritizing a smooth and welcoming first-time experience over strict process adherence makes so much sense. Thanks for sharing that perspective!

PR #4038, opened 2026-02-25
feat: move codec implementations to shards for split-package support

codelite7:
## Summary

This branch adds support for split-package (sharded) code generation in gqlgen, where generated code is split into per-entity packages for faster parallel compilation.

- **Move codec implementations to shards**: codec marshal/unmarshal logic moved from root package into per-entity shard packages
- **Add codec registration to shardruntime**: shardruntime registers codec implementations from all shards
- **Fix field resolution templates**: restore field resolution templates to root to prevent infinite recursion when shards reference root types
- **Fix shard codec template**: add fmt import and type assertions

These changes enable faster parallel compilation by splitting generated GraphQL code across multiple packages while maintaining correct codec and field resolution behavior.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/codegen/split_codecs_.gotpl
+++ b/codegen/split_codecs_.gotpl
@@ -1,6 +1,7 @@
 {{ define "split_codecs_.gotpl" }}
 {{ reserveImport "reflect" }}
 {{ reserveImport "sync" }}
+{{ reserveImport "fmt" }}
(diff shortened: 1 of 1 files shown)


codelite7:
Opened against wrong repo, will be handled in the fork.

PR #4040, opened 2026-02-28
feat: support @goField(batch: true) directive

dpulpeiro:
## Summary

- Add `batch` argument to the `@goField` directive, allowing batch resolver configuration directly in the schema instead of only via `gqlgen.yml`
- Update the `batchresolver` example with directive-configured fields alongside the existing YAML-configured ones
- Add `TestBatchDirectiveConfig` to verify directive parsing

As suggested in https://github.com/99designs/gqlgen/pull/4008#discussion_r2747348032.

## Test plan

- [x] `go generate` produces correct batch resolver signatures for directive-annotated fields
- [x] `TestBatchDirectiveConfig` verifies `cfg.Init()` parses `batch: true` from directives
- [x] All existing `TestBatchResolver_*` tests continue to pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,7 @@ directive @goField(
 	type: String
 	autoBindGetterHaser: Boolean
 	forceGenerate: Boolean
+	batch: Boolean
 ) on INPUT_FIELD_DEFINITION | FIELD_DEFINITION
 
 type User @goModel(model: "github.com/you/pkg/model.User") {
--- a/_examples/batchresolver/batchresolver_test.go
+++ b/_examples/batchresolver/batchresolver_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/99designs/gqlgen/client"
+	"github.com/99designs/gqlgen/codegen/config"
 	"github.com/99designs/gqlgen/graphql/handler"
 	"github.com/99designs/gqlgen/graphql/handler/transport"
 )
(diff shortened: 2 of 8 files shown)


dpulpeiro:
One thing to consider: if we ever want to add batch-specific options like a _limit_, having batch config under `@goField` could get crowded. A separate `@batch` directive might give us more room to grow. Not sure what the long-term vision is for the library here

StevenACoffman:
@tomoikey @atzedus Do you have thoughts about this particular directive?

atzedus:
Regarding your comment about a separate `@batch` directive: creating a separate directive for this purpose is most likely not necessary.

Indeed, there is a high probability that batch size limits/workers per operation will be needed in certain scenarios in the future. But I'm confident that such settings are unlikely to be needed for one specific field or a few. Most likely, it would be more convenient to have a separate configuration in `gqlgen.yaml` that applies to all batch resolvers at once.

With the introduction of this feature and directive, I'd like to point out that it would be worth adding a global setting in `gqlgen.yaml` like `resolvers_batch` (`true`/`false`) that enables the new behavior by default, while the directive/per-field setting could override it to disable it.

It might even be worth building in future functionality right away by making it an object:

```yaml
resolvers_batch:
  enabled: true/false
```

What do you think about this?

dpulpeiro:
I think that deserves its own discussion in a separate issue or at least its own PR. Changing the default resolver signature for an entire schema is a pretty significant change IMO. I like the opt-in approach per field that we currently have. 

I would like to keep this PR small, as the main reason for this change is that we are using @goModel and @goField to configure other options. Right now `batch` is only configurable via YAML. That said, I'm happy to keep the YAML option added by @tomoikey  for people who prefer that approach.

atzedus:
This change would only take effect if explicitly opted in — fully analogous to:

```yaml
federation:
  ...
  options:
    entity_resolver_multi: true
```

I believe this is a very flexible approach to cover different user needs through a priority chain: **global config → field config → directive**.

But I agree with you, I think this is a topic for a separate PR.

However, I still believe there is no need for a separate @batch directive. The batch option for @goField is quite sufficient and fits well here. After all, if the need arises, I see nothing wrong with adding an additional @goBatchConfig or simply @goBatch directive.

atzedus:
My experience shows that directive-based settings like `@entityResolver(multi: true)` or the new `@goField(batch: true)` are temporary in nature — they are used during the refactoring period while adopting new capabilities. Once the refactoring is complete, they are removed and `gqlgen.yaml` takes over.

dpulpeiro:
Opened #4044 to track the global batch config discussion separately.

Regarding the directive naming, I'm happy to keep `@goField(batch: true)` and not introduce a separate `@goBatch` directive.

dpulpeiro:
Assuming we're good with `@goField(batch: true)`,  I'm not planning to make any more changes here.

StevenACoffman:
Thanks! I really appreciate you both! It's great that we're sprucing up the documentation and easing people's migration path!

PR #4043, opened 2026-03-01
Add nested batch resolver tests and documentation to batchresolver example

dpulpeiro:
## Summary
- Add nested batch resolver example (`User → Profile → Image`) with both batch and non-batch paths to document current batching behavior
- Add `TestBatchResolver_Nested_CallCount` that verifies batch call counts and documents that nested batch parent context is not yet propagated (coverBatch is called N times instead of 1)
- Add benchmarks comparing batch vs non-batch for single-level and nested resolution
- Add readme.md documenting the batchresolver example

## Test plan
- `go test ./_examples/batchresolver/...` passes
- Benchmarks run: `go test -bench=. ./_examples/batchresolver/`

### DIFF
--- a/_examples/batchresolver/batchresolver_test.go
+++ b/_examples/batchresolver/batchresolver_test.go
@@ -554,6 +554,125 @@ func TestBatchResolver_Nested_CallCount(t *testing.T) {
 	)
 }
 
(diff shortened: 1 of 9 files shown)


StevenACoffman:
@tomoikey FYI
@dpulpeiro Thank you so much!

PR #4059, opened 2026-03-07
Allow gqlgen to run in sandbox/bazel mode with injected type information

angl:
Gqlgen uses go/packages API to load source code from the entire git source tree to find out the type information for goModel binding. This doesn't play well with bazel as by default bazel doesn't expose source tree and it will take a lot of effort to force bazel to do so.

This is an approach suggested by Cursor (Opus 4.6-high) to inject type information directly into gqlgen and bypass the use of go/packages. It is quite elegant: the type information can be obtained directly from go archive files that are built by go_rule already. Only a minimal manifest JSON file is required to map import path to archive file is required.

I asked Cursor to build an end-to-end PoC and here it is: https://gerrit.observeinc.com/c/observe/+/83903. AFAICT, it works.

So making this PR to add the type injection capability to our forked gqlgen, and once this is in, we can actually start bazelify gqlgen.

### DIFF
--- a/codegen/config/config.go
+++ b/codegen/config/config.go
@@ -887,7 +887,7 @@ func (c *Config) injectBuiltins() {
 }
 
 func (c *Config) LoadSchema() error {
-	if c.Packages != nil {
+	if c.Packages != nil && !c.Packages.HasInjected() {
 		c.Packages = code.NewPackages(
 			code.WithBuildTags(c.GoBuildTags...),
 			code.PackagePrefixToCache("github.com/99designs/gqlgen/graphql"),
--- a/codegen/config/typecache.go
+++ b/codegen/config/typecache.go
@@ -1,0 +1,98 @@
+package config
+
+import (
+	"bufio"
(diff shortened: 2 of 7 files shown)


StevenACoffman:
@angl Hi! Thanks for contributing! 

> AFAICT, it works.

I am not sure what you mean by this. Have you verified that with this change, everything is working in your production systems? 

I am very suspicious of AI-generated code solutions, as they often appear plausible but can have subtle problems. For me to merge this PR, I would like you to first update the code to match our latest and resolve all the code conflicts, fix any lint errors, pass all tests, and run the result to verify it actually works.

Otherwise, you are asking me to force a large number of other engineers at other companies to do your testing for you, and to clean up any possible mess.

angl:
Hi @StevenACoffman ,

This was a mistake :) I didn't intend to push a PR for the official branch but rather to our fork of it, as I totally agree this is not ready for the official thing. Will delete this PR. Sorry for the noise.

Best,
-Ang

PR #4070, opened 2026-03-09
Feature/optimize packages load validation

jhonDoe15:
**Summary:** Adds 4 new opt-in config options to dramatically speed up code generation for large projects:

| Option | Benefit |
|--------|---------|
| `fast_validation` | ~2x faster validation (cold cache) |
| `skip_import_grouping` | ~10x faster formatting |
| `use_light_mode_prefetch` | ~200x faster package loading |
| `use_buffer_pooling` | ~25% less GC pressure |

**Key points:**
- All options default to `false` (no breaking changes)
- Users explicitly opt in for performance gains
- Targets warm-cache generation of ~10s for 1.1M LOC codebases
- Includes tests, docs, and schema updates

I have:
 - [x] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [x] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

relates to #4025

PR #4080, opened 2026-03-16
Remove use_light_mode_prefetch option

jhonDoe15:
### Follow-up from #4070 : Feature/optimize packages load validation

This option was intended to speed up package loading by using light mode `(NeedName|NeedFiles|NeedModule)` instead of full mode `(NeedTypes|NeedSyntax|...)`.

However, the option doesn't work well with:
- Autobind packages (need full type info to match Go types)
- `@goModel` mappings (need full type info to resolve types)

For projects that use these features (most real-world projects), the option either causes incorrect code generation or provides no benefit, since the types need to be loaded anyway.

Keeping the 3 other performance options that do work:
- `fast_validation`: Uses `-gcflags="-N -l"` for faster validation
- `skip_import_grouping`: Uses format.Source instead of imports.Process
- `use_buffer_pooling`: Reuses byte buffers via sync.Pool

Also adds tests for the remaining performance options.

I have:

### DIFF
--- a/api/generate_test.go
+++ b/api/generate_test.go
@@ -128,3 +128,143 @@ func TestGenerateWithSchemaMutator(t *testing.T) {
 		})
 	}
 }
+
+// TestPerformanceOptionsWithAutobind tests that the three working performance
+// optimization options (fast_validation, skip_import_grouping, use_buffer_pooling)
+// work correctly with autobind and @goModel type mappings.
+//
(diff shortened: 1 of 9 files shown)


jhonDoe15:
@StevenACoffman I'm removing one of the options that doesnt seem to make sense in a real project.
caused import loops localy when used

StevenACoffman:
Thanks!

PR #4087, opened 2026-03-21
fix: batch resolver being called multiple times for interface implementations

### LINKED ISSUE #4081: Bug: batch resolver don't work as expected on interface parent types
**Batch resolver invoked multiple times when parent type of resolving field is interface**

The response data is correct, but the batching optimization is not working as expected, resulting in N resolver calls instead of 1.

### 1. Schema Definition
```graphql
interface Animal {
    id: ID!
    batchProp: String! @goField(batch: true)
}

type Cat implements Animal {
    id: ID!
    batchProp: String! @goField(batch: true)
}

extend type Query {
    animals: [Animal!]!
}
```

### 2. Generate Code
Run `gqlgen` generate as usual.

### 3. Resolver Implementation
```go
// BatchProp is the batch resolver for the batchProp field.
func (r *catResolver) BatchProp(ctx context.Context, objs []*entity.Cat) ([]string, error) {
    result := make([]string, 0, len(objs))
    fmt.Println("BatchProp for Cats called with the len: ", len(objs))
    for _, obj := range objs {
        fmt.Printf("BatchProp for Cat ID: %s\n", obj.ID)
        result = append(result, "catBatchProp: "+obj.ID)
    }
    return result, nil
}

// Animals is the resolver for the animals field.
func (r *queryResolver) Animals(ctx context.Context) ([]entity.Animal, error) {
    return []entity.Animal{
        &entity.Cat{ID: "cat1"},
        &entity.Cat{ID: "cat2"},
        &entity.Cat{ID: "cat3"},
    }, nil
}
```

### 4. Execute Query
```graphql
{
  animals {
    id
    batchProp
  }
}
```

## Actual Behavior
**Console output:**
```
BatchProp for Cats called with the len:  1
BatchProp for Cat ID: cat2
BatchProp for Cats called with the len:  1
BatchProp for Cat ID: cat3
BatchProp for Cats called with the len:  1
BatchProp for Cat ID: cat1
```

The batch resolver is invoked **3 times**, each with a slice of length 1.

## Expected Behavior
The batch resolver should be invoked **once** with a slice containing all 3 `*entity.Cat` objects:
```
BatchProp for Cats called with the len:  3
BatchProp for Cat ID: cat1
BatchProp for Cat ID: cat2
BatchProp for Cat ID: cat3
```

## Response Data
The returned JSON is correct regardless of the batching issue:
```json
{
  "data": {
    "animals": [
      { "id": "cat1", "batchProp": "catBatchProp: cat1" },
      { "id": "cat2", "batchProp": "catBatchProp: cat2" },
      { "id": "cat3", "batchProp": "catBatchProp: cat3" }
    ]
  }
}
```

## Environment
- `gqlgen` version: *v0.17.88*

Additionally, when the `@goField(batch: true)` directive is applied only to the interface field definition (and omitted on the implementing type), no batch resolver is generated at all. I'm not sure if this is the intended behavior.

@tomoikey @dpulpeiro

tomoikey:
## Summary

Closes #4081
I have resolved a bug in which the same batch resolver was being called multiple times for the same field when dealing with interface types.

I have:

### DIFF
--- a/_examples/batchresolver/batchresolver_test.go
+++ b/_examples/batchresolver/batchresolver_test.go
@@ -775,6 +775,44 @@ func TestBatchResolver_InterfaceSingleType_CallCount(t *testing.T) {
 	)
 }
 
+func TestBatchResolver_InterfaceSingleType_MixedPointerAndValue_CallCount(t *testing.T) {
+	resolver := &Resolver{
+		animals: []Animal{
+			Cat{ID: "cat1"},
+			&Cat{ID: "cat2"},
+			Cat{ID: "cat3"},
+		},
+	}
+
+	c := newTestClient(resolver)
+	var resp struct {
+		Animals []struct {
+			ID        string `json:"id"`
+			BatchProp string `json:"batchProp"`
+		} `json:"animals"`
+	}
+
+	err := c.Post(`query { animals { id batchProp } }`, &resp)
+	require.NoError(t, err)
+	require.Len(t, resp.Animals, 3)
+	require.JSONEq(
+		t,
+		`{"animals":[
+			{"id":"cat1","batchProp":"catBatchProp:cat1"},
+			{"id":"cat2","batchProp":"catBatchProp:cat2"},
+			{"id":"cat3","batchProp":"catBatchProp:cat3"}
+		]}`,
+		marshalJSON(t, resp),
+	)
+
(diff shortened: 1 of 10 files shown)


StevenACoffman:
Thank you so much!

PR #4120, opened 2026-04-10
Fix gosec G115 integer overflow in generated atomic.AddInt32 call

### LINKED ISSUE #4119: Generated code triggers gosec G115 (integer overflow) on atomic.AddInt32
## Description

Generated code from gqlgen v0.17.89 triggers [gosec G115 (CWE-190)](https://cwe.mitre.org/data/definitions/190.html) — integer overflow conversion `int` → `int32` — on every object type in the schema.

The finding is in `object.gotpl` line 161:

```go
atomic.AddInt32(&ec.Deferred, int32(len(deferred)))
```

This produces one G115 finding per GraphQL object type in the generated output. For a moderately-sized schema (~25 types), that's 25+ findings from `gosec ./...`.

## Impact

- **Not a real vulnerability** — `len(deferred)` is bounded by the number of deferred field groups in a single query, which can never approach `math.MaxInt32`.
- **Real CI/tooling impact** — every gqlgen user running gosec must either use `-exclude-generated` or add suppressions. This undermines the value of security scanning on generated code, since users learn to ignore the file entirely.

## Affected templates

1. **`codegen/object.gotpl:161`** — `atomic.AddInt32(&ec.Deferred, int32(len(deferred)))` — this is the main one (emitted per object type)
2. **`codegen/generated!.gotpl:189`** — `atomic.AddInt32(&ec.PendingDeferred, -1)` — constant, less concerning
3. **`codegen/root_.gotpl:157`** — `atomic.AddInt32(&ec.PendingDeferred, -1)` — constant, less concerning

## Suggested fix

Since gqlgen requires Go 1.20+, and Go 1.21 introduced the `min` builtin:

```go
// Before
atomic.AddInt32(&ec.Deferred, int32(len(deferred)))

// After
atomic.AddInt32(&ec.Deferred, int32(min(len(deferred), math.MaxInt32)))
```

Or alternatively, migrate from `atomic.AddInt32` / `atomic.LoadInt32` to `atomic.Int32` (available since Go 1.19), which would modernize the atomics usage across the board.

## Reproduction

```bash
# Any project using gqlgen v0.17.89
go generate ./...
gosec ./...
# → G115 findings in generated.go for every object type
```

## Environment

- gqlgen: v0.17.89
- Go: 1.26
- gosec: latest
- OS: Windows 11 / Linux

3rg0n:
## Summary

- Bounds the `int` → `int32` conversion in `object.gotpl` with `min(len(deferred), math.MaxInt32)` to eliminate gosec G115 (CWE-190) findings in generated code
- Adds `math` import reservation to both layout entry-point templates (`generated!.gotpl`, `root_.gotpl`)
- Regenerated all test servers and examples

Fixes #4119

## Test plan

- [x] `go generate ./...` succeeds and all generated files contain the bounded conversion
- [x] `go test -race ./...` passes (all core + testserver tests)
- [x] `golangci-lint run` produces no new findings (pre-existing `gci` formatting issues only)
- [x] `TestLayouts` confirms single-file and follow-schema layouts remain equivalent

### DIFF
--- a/_examples/batchresolver/generated.go
+++ b/_examples/batchresolver/generated.go
@@ -8,6 +8,7 @@ import (
 	"embed"
 	"errors"
 	"fmt"
+	"math"
 	"strconv"
 	"sync/atomic"
 
@@ -4085,7 +4086,7 @@ func (ec *executionContext) _Cat(ctx context.Context, sel ast.SelectionSet, obj 
 		return graphql.Null
 	}
 
-	atomic.AddInt32(&ec.Deferred, int32(len(deferred)))
+	atomic.AddInt32(&ec.Deferred, int32(min(len(deferred), math.MaxInt32)))
 
 	for label, dfs := range deferred {
 		ec.ProcessDeferredGroup(graphql.DeferredGroup{
@@ -4160,7 +4161,7 @@ func (ec *executionContext) _Dog(ctx context.Context, sel ast.SelectionSet, obj 
 		return graphql.Null
 	}
 
-	atomic.AddInt32(&ec.Deferred, int32(len(deferred)))
+	atomic.AddInt32(&ec.Deferred, int32(min(len(deferred), math.MaxInt32)))
 
 	for label, dfs := range deferred {
 		ec.ProcessDeferredGroup(graphql.DeferredGroup{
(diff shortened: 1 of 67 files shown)


StevenACoffman:
Thanks!

PR #4121, opened 2026-04-11
Fix follow-schema layout not generating location directive middleware for directive-only files

toga4:
## Problem

When using the `follow-schema` layout, defining a location directive (e.g. `directive @myDirective(key: String!) on FIELD`) in a `.graphql` file that contains **no type definitions** causes a compilation error. The generated code calls `_fieldMiddleware` from every per-schema file, but the function definition is never generated because no per-schema build exists for the directive-only file.

This happens because:
- `field.gotpl` uses `AllDirectives.LocationDirectives "FIELD"` to generate **calls** to `_fieldMiddleware` in all per-schema files
- `directives.gotpl` uses `Directives().LocationDirectives "FIELD"` to generate the **definition**, but `Directives()` filters by `Config.Sources` (a single file in per-schema builds). If the directive's source file has no types, no per-schema build exists for it, so the definition is never generated

The same issue applies to `_queryMiddleware`, `_mutationMiddleware`, and `_subscriptionMiddleware`.

## Solution

Location directive middleware is a schema-global responsibility — `field.gotpl` calls `_fieldMiddleware` if *any* FIELD directive exists in the entire schema. This PR moves the middleware definition to the root file where all directives are visible.

- Add `SkipLocationDirectives` flag to `Data`; set it `true` for per-schema builds
- Guard the 4 middleware blocks in `directives.gotpl` with `not .SkipLocationDirectives`
- Concatenate `directives.gotpl` into the root file template so middleware is generated there
- Add `DirectiveArgs()` method to generate args functions for directives whose source file has no types (orphan directives)

The single-file layout is unaffected (`SkipLocationDirectives` defaults to `false`).

### Design notes

`generateRootFile` uses `Template:` (string), which causes `templates.Render` to ignore `TemplateFS`. This means the root file template cannot access `{{ define }}` blocks from other `.gotpl` files. To work around this, `directives.gotpl` is concatenated with `root_.gotpl` at the Go level via `rootTemplate + "\n" + directivesTemplate`.

We considered an alternative: deleting `directives.gotpl` and inlining its content into both `root_.gotpl` and `generated!.gotpl` (following the existing duplication pattern between these two files). However, `directives.gotpl` is not just root-level code — its `{{ define }}` blocks (`implDirectives`, `queryDirectives`, `callDirective`) are shared partials referenced by `field.gotpl`, `args.gotpl`, and `input.gotpl`. Duplicating these across files would be a heavier form of duplication than the existing pattern.

There are two known trade-offs with the current approach:
- **Go-level template concatenation has no precedent in this codebase.** If you'd prefer a different approach, happy to discuss.
- **`args.gotpl` logic is partially duplicated in `root_.gotpl`** for orphan directive args generation.

A cleaner long-term fix would be to restructure the templates so that root-level and per-schema templates are cleanly separated, with single-file mode combining both in one pass. This would eliminate the duplication between `root_.gotpl` and `generated!.gotpl`, the layout guards in `generated!.gotpl`, `SkipLocationDirectives`, and the template concatenation — but that is a larger refactor beyond the scope of this bug fix.

I have:

### DIFF
--- a/codegen/data.go
+++ b/codegen/data.go
@@ -37,6 +37,11 @@ type Data struct {
 	SubscriptionRoot *Object
 	AugmentedSources []AugmentedSource
 	Plugins          []any
+
+	// SkipLocationDirectives suppresses generation of location directive middleware
+	// (_fieldMiddleware, _queryMiddleware, etc.) in per-schema builds.
+	// In follow-schema layout these are generated in the root file instead.
+	SkipLocationDirectives bool
 }
 
 func (d *Data) HasEmbeddableSources() bool {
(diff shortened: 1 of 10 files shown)


StevenACoffman:
Sorry, I've been sitting on this because we have very deliberately been moving logic _**out**_ of the templates, since it is very difficult to test and maintain, and this adds a great deal into them.

I am much more interested in:
> A cleaner long-term fix would be to restructure the templates so that root-level and per-schema templates are cleanly separated, with single-file mode combining both in one pass. This would eliminate the duplication between `root_.gotpl` and `generated!.gotpl`, the layout guards in `generated!.gotpl`, `SkipLocationDirectives`, and the template concatenation — but that is a larger refactor beyond the scope of this bug fix.

toga4:
@StevenACoffman Happy to give the refactor a try as a separate PR.

That direction actually aligns with your "move logic out of templates" goal — the layout decision moves from in-template guards (`{{ if eq .Config.Exec.Layout "single-file" }}`) to the Go side picking which templates to render.

Rough shape:

- **Root templates** — schema-global, rendered once:
  `NewExecutableSchema`, `Config`, `Exec()`, location middleware, etc.
- **Per-schema templates** — per source file: `field.gotpl`,
  `args.gotpl`, object/input/interface generation, etc.
- **Shared partials** — `{{ define }}` blocks only

Then:

- `follow-schema`: root templates → `root_.generated.go`;
  per-schema templates per file → `xxx.generated.go`
- `single-file`: both together → `generated.go`

That removes the layout guards, the `generated!.gotpl` / `root_.gotpl` duplication, and makes the orphan-directive bug disappear naturally.

Let me close this PR and open a new one with that approach.

StevenACoffman:
@toga4 After I made that comment, I started on it, but I got too sleepy and didn't finish it. Can you check it out and see if it works or could be improved: 
https://github.com/99designs/gqlgen/pull/4141

toga4:
Thanks @StevenACoffman ! Verified it works locally — great work!

StevenACoffman:
@toga4 Thanks for verifying it! I really appreciate that you put up a PR, and it was only that I had been really pushing for this upcoming release to cut down on template logic. I really look forward to your next one though!

PR #4129, opened 2026-04-20
exec: fix case-insensitive file name collision

grzn:
Similar to https://github.com/99designs/gqlgen/pull/2829.

1. added a case under `codegen/testserver/casesensitivity/` to demonstrate the problem
2. added handing of filenames with different cases, similar to what was done in https://github.com/99designs/gqlgen/pull/2829


codegen/generate.go
```diff
@@ -114,18 +119,21 @@ func filename(p *ast.Position, config *config.Config) string {
 	return strings.ReplaceAll(filenameTempl, "{name}", name)
 }
 
-func addBuild(filename string, p *ast.Position, data *Data, builds *map[string]*Data) {
+func addBuild(fnCase string, fnKey string, p *ast.Position, data *Data, builds *map[string]*build) {
```
StevenACoffman, line 122:
```suggestion
func addBuild(fnCase, fnKey string, p *ast.Position, data *Data, builds *map[string]*build) {
```

StevenACoffman: The abbreviation you chose `fn` for `fnCase` and `fnKey` is commonly used to mean `function` so it may be a little confusing here?

StevenACoffman: Also, even in gigantic projects, the number of files is probably not significant, as you do in a number of the other functions, so I'm not sure that we need to bother passing the same data fnCase and fnKey here, when we can just re-compute fnKey from `strings.ToLower(filename)`

StevenACoffman:
Hi! I appreciate your PR, but the fmt-and-lint check failed with:
```
Error: generate: open /home/runner/work/gqlgen/gqlgen/codegen/testserver/casesensitivity/resource.generated.go: no such file or directory
Error: Process completed with exit code 1.
```
And when I looked at the lint failures, I questioned whether we need to change that signature at all, since the [strings.ToLower()](https://pkg.go.dev/strings#ToLower) conversion of a filename is not going to make a measurable performance impact at the scale of the number of files in the vast majority of projects.

BTW, https://github.com/SimonWaldherr/golang-benchmarks#caseinsensitivecompare shows that [strings.ToUpper()](https://pkg.go.dev/strings#ToUpper) is faster and has fewer allocations.

PR #4138, opened 2026-04-20
graphql: add runtime non-null field support via MarkNonNull

lbarthon:
In GraphQL, null propagation (nulling out a parent object when a child field resolves to nil) only applies to fields declared as non-null (!) in the schema. This is a compile-time property baked into the generated code — there was no way for middleware to trigger it dynamically.

This is needed for use cases like a @priority(value: Required) directive where the client marks certain fields as semantically required in the query. These fields are nullable in the schema, but when they fail at runtime, the response should behave as if they were non-null: the error propagates up and the nearest nullable ancestor is set to null.

Changes:

- Add FieldContext.NonNull flag and graphql.MarkNonNull(ctx) API. A FieldInterceptor calls MarkNonNull during resolution to opt a nullable field into non-null semantics for that request.

- Add graphql.RequiredNull sentinel. It marshals to "null" like graphql.Null but is a distinct pointer. resolveField returns it (instead of Null) when a dynamically-marked field fails, so the parent's generated code can distinguish the two cases.

- Update resolveField to check fc.NonNull alongside the static nonNull parameter. On nil/error, it emits the standard "must not be null" error and returns RequiredNull.

- Update object.gotpl so nullable fields check for RequiredNull and increment Invalids, triggering the same null propagation cascade as a schema-level non-null violation. Concurrent field closures now always receive the FieldSet parameter (previously ignored for nullable fields).

Existing behavior is fully preserved: statically non-null fields still return Null and are caught by the existing res == graphql.Null check. RequiredNull is never returned unless a FieldInterceptor explicitly opts in via MarkNonNull.

I have:

### DIFF
--- a/_examples/batchresolver/generated.go
+++ b/_examples/batchresolver/generated.go
@@ -4021,13 +4021,16 @@ func (ec *executionContext) _Profile(ctx context.Context, sel ast.SelectionSet, 
 		case "coverBatch":
 			field := field
 
-			innerFunc := func(ctx context.Context, _ *graphql.FieldSet) (res graphql.Marshaler) {
+			innerFunc := func(ctx context.Context, fs *graphql.FieldSet) (res graphql.Marshaler) {
 				defer func() {
 					if r := recover(); r != nil {
 						ec.Error(ctx, ec.Recover(ctx, r))
 					}
 				}()
 				res = ec._Profile_coverBatch(ctx, field, obj)
+				if res == graphql.RequiredNull {
+					atomic.AddUint32(&fs.Invalids, 1)
+				}
 				return res
 			}
 
@@ -4054,13 +4057,16 @@ func (ec *executionContext) _Profile(ctx context.Context, sel ast.SelectionSet, 
 		case "coverNonBatch":
 			field := field
 
-			innerFunc := func(ctx context.Context, _ *graphql.FieldSet) (res graphql.Marshaler) {
+			innerFunc := func(ctx context.Context, fs *graphql.FieldSet) (res graphql.Marshaler) {
 				defer func() {
 					if r := recover(); r != nil {
 						ec.Error(ctx, ec.Recover(ctx, r))
 					}
 				}()
 				res = ec._Profile_coverNonBatch(ctx, field, obj)
+				if res == graphql.RequiredNull {
+					atomic.AddUint32(&fs.Invalids, 1)
+				}
 				return res
 			}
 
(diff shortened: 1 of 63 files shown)


lbarthon:
Hey @StevenACoffman! When you have a moment, could you look at this? Happy to give more info if the commit message is not explicit, or if you want extra docs for it!

StevenACoffman:
Hi! Thanks for your patience! I was hoping to get some extra eyes from the community like @atzedus @UnAfraid 

Also, we had a big refactor of the templates that consume all my open source time.

atzedus:
Thanks for this PR - the use case is compelling, and the implementation is clever. I especially appreciate how it enables dynamic null-propagation without breaking existing behavior.

Before this moves forward, I'd like to raise a concern about spec compliance and contract clarity, and then suggest how we might mitigate it.

**The Core Concern:** Runtime Semantics vs. Schema Contract
The GraphQL spec explicitly ties null-propagation to schema-time type declarations:
"If the field which experienced an error was declared as Non-Null, the null result will bubble up..."
- [Handling Field Errors](https://spec.graphql.org/draft/#sec-Handling-Field-Errors)
And:
"Tools can ensure that [a query] is valid within that type system before execution, i.e. at development time"
- [Overview](https://spec.graphql.org/draft/#sec-Overview)

With MarkNonNull(), a field declared as String in the schema can behave like String! at runtime.
This introduces behavior that:
- Cannot be discovered via introspection
- Cannot be validated statically by client tools (e.g., GraphQL IDEs, type generators)
- Makes the response shape depend on hidden server logic or dynamic flags

This potentially weakens the principle that the schema is the single source of truth for the client-server contract.

**Counterpoint:** Executable Directives as an Explicit Signal
That said, if this mechanism is triggered only by an executable directive like @priority(value: REQUIRED) - which the client explicitly includes in the query - then the behavior is no longer "hidden". The client is actively opting into stricter semantics for that specific request, similar to how @include or @skip alter execution.

In that light, this feature could be seen as a controlled escape hatch, not a violation of the spec - provided it's used intentionally and transparently.

Suggestions to Move Forward Safely.

To ensure this powerful feature is adopted responsibly, I suggest the PR include:
- Explicit Warning in Documentation
Add a clear notice about the trade-offs:
"Using MarkNonNull or executable directives to alter null-propagation semantics makes response behavior dependent on runtime logic. This can break static validation and introspection-based tooling. Use only when you control both client and server, and when the benefit (e.g., stricter data integrity for critical fields) outweighs the loss of schema-time guarantees."

- Recipe-Style Documentation with Concrete Examples
Include a dedicated guide showing:
How to define and register an executable directive like @priority
How to implement a FieldInterceptor that calls MarkNonNull based on directive arguments
Example queries showing the difference in behavior with/without the directive
When not to use this (e.g., public APIs, third-party clients)

- Introspection Consideration (Future Work)
While not required for this PR, it's worth noting that a future enhancement could expose "dynamic constraints" via a custom introspection extension - though that's likely out of scope for now.

I believe this feature is valuable for advanced use cases (e.g., internal microservices, data-critical clients), but it should come with strong guardrails. With the above additions - especially clear warnings and recipe-style docs - this PR could be merged responsibly.

StevenACoffman:
@lbarthon do you have a response to the review above from @atzedus ?

lbarthon:
Sorry for the delay, and thank you for the very thorough review @atzedus!

I've added in this diff a new documentation page explaining how the feature works - and one example usage. I also updated the godoc for the `MarkNonNull` method, so that there's no way to miss it.

The comments about the introspection are valid, and were mentioned as a gap / future work. Please let me know if there's anything else I missed!

atzedus:
Looks great! I think we can merge this PR.
@StevenACoffman

PR #4146, opened 2026-04-25
Part 3 - reduce template logic: Unify ProcessArgField and ProcessArgFieldWithEC

StevenACoffman:
# gqlgen codegen refactoring: eliminating template branching

The gqlgen code generator produces Go server code from GraphQL schemas. Its
output is controlled by a set of Go template files (`.gotpl`) that are
rendered once per schema or per schema object. The generated code must support
two execution-context styles — a legacy "receiver" style where generated
functions are methods on an `*executionContext` value, and a newer "function"
style where `ec *executionContext` is passed as an explicit parameter — governed
by the `UseFunctionSyntaxForExecutionContext` config flag.

A problem existed in the `graphql` runtime package: `ProcessArgField`
and `ProcessArgFieldWithEC` are two separately maintained functions whose only
difference is the presence of an `ec` parameter in the callback signature. This
runtime split is a direct reflection of the template branching; removing the
template branch requires also unifying these two functions.

---

### The problem

One branch remains in `codegen/args.gotpl`:

```
{{ if $.Config.UseFunctionSyntaxForExecutionContext -}}
arg{{$i}}, err := graphql.ProcessArgFieldWithEC(ctx, ec, rawArgs, {{$arg.Name|quote}}, {{ $arg.TypeReference.UnmarshalFunc }})
{{- else -}}
arg{{$i}}, err := graphql.ProcessArgField(ctx, rawArgs, {{$arg.Name|quote}}, ec.{{ $arg.TypeReference.UnmarshalFunc }})
{{- end }}
```

This branch exists because the two execution-context modes have different
unmarshaler signatures:

- Receiver syntax: unmarshaler is a method on `*executionContext`, called as
  `ec.unmarshalNString2string(ctx, v)` — signature `func(context.Context, any) (T, error)`
- Function syntax: unmarshaler is a standalone function called as
  `unmarshalNString2string(ctx, ec, v)` — signature `func(context.Context, *executionContext, any) (T, error)`

`ProcessArgFieldWithEC` exists solely to accommodate the second signature.
It is the runtime projection of the template branch.

### Current runtime implementations

```go
// graphql/args.go

func ProcessArgField[T any](
    ctx context.Context,
    rawArgs map[string]any,
    fieldName string,
    valueMapperFn func(ctx context.Context, value any) (T, error),
) (T, error) {
    value, exists := rawArgs[fieldName]
    if !exists {
        var zeroVal T
        return zeroVal, nil
    }
    ctx = WithPathContext(ctx, NewPathWithField(fieldName))
    return valueMapperFn(ctx, value)
}

func ProcessArgFieldWithEC[T, EC any](
    ctx context.Context,
    ec EC,
    rawArgs map[string]any,
    fieldName string,
    valueMapperFn func(ctx context.Context, ec EC, value any) (T, error),
) (T, error) {
    value, exists := rawArgs[fieldName]
    if !exists {
        var zeroVal T
        return zeroVal, nil
    }
    ctx = WithPathContext(ctx, NewPathWithField(fieldName))
    return valueMapperFn(ctx, ec, value)
}
```

### The revised approach: typed closures, unchanged runtime

The original plan proposed changing `ProcessArgField`'s callback to
`func(context.Context, any) (any, error)` (non-generic return). This would
box every concrete return value — `string`, `int`, or any struct — into an
`interface{}` on every call, causing a heap allocation per argument.

**The revision:** do not change `ProcessArgField`'s signature. It already
accepts `func(context.Context, any) (T, error)` with a generic concrete return
type. Keep it exactly as-is.

Instead, change the template to generate a typed closure that captures `ec` and
adapts the calling convention. Both receiver-syntax and function-syntax modes
produce a closure with the same signature: `func(context.Context, any) (T, error)`.

```
arg{{$i}}, err := graphql.ProcessArgField(ctx, rawArgs, {{$arg.Name|quote}},
    func(ctx context.Context, v any) ({{ $arg.TypeReference.GO | ref }}, error) {
        return {{$.ECDot}}{{$arg.TypeReference.UnmarshalFunc}}(ctx, {{$.ECArg}}v)
    })
```

For receiver syntax (`ECDot` = `"ec."`, `ECArg` = `""`):
```go
arg0, err := graphql.ProcessArgField(ctx, rawArgs, "name",
    func(ctx context.Context, v any) (string, error) {
        return ec.unmarshalNString2string(ctx, v)
    })
```

For function syntax (`ECDot` = `""`, `ECArg` = `"ec, "`):
```go
arg0, err := graphql.ProcessArgField(ctx, rawArgs, "name",
    func(ctx context.Context, v any) (string, error) {
        return unmarshalNString2string(ctx, ec, v)
    })
```

Both forms:
- Return `(string, error)` — concrete type, no boxing
- Capture `ec` — exactly as the existing method value `ec.unmarshalNString2string` did
- Are amenable to stack allocation because `ProcessArgField` is short, has no
  loops, and does not store the function value beyond the single call to it

### Allocation equivalence

A Go method value like `ec.unmarshalNString2string` is itself a compiler-generated
closure that captures `ec`. It is not allocation-free; the difference from an
explicit closure is only in notation. Whether the closure escapes to the heap
depends on whether the Go compiler can inline `ProcessArgField` and confirm
that the function value does not outlive the stack frame. Since `ProcessArgField`
is a four-statement generic function, inlining is likely, and escape analysis
should keep both forms on the stack. The allocation behavior is not regressed
by using explicit closures with concrete return types.

The allocation that the original plan would have introduced — and that this
revision avoids — is the boxing of return values when using `(any, error)`
as the callback return type. By keeping `(T, error)`, no boxing occurs.

### Files to change

**`graphql/args.go`**

Delete `ProcessArgFieldWithEC` entirely. `ProcessArgField` is unchanged.

```go
// Before (keep):
func ProcessArgField[T any](
    ctx context.Context,
    rawArgs map[string]any,
    fieldName string,
    valueMapperFn func(ctx context.Context, value any) (T, error),
) (T, error) { ... }

// Delete:
func ProcessArgFieldWithEC[T, EC any](...) (T, error) { ... }
```

**`graphql/args_test.go`**

Delete `TestProcessArgFieldWithEC`. `TestProcessArgField` is unchanged.

**`codegen/args.gotpl`**

Replace the four-line if/else block:

```
{{- else -}}
    {{ if $.Config.UseFunctionSyntaxForExecutionContext -}}
    arg{{$i}}, err := graphql.ProcessArgFieldWithEC(ctx, ec, rawArgs, {{$arg.Name|quote}}, {{ $arg.TypeReference.UnmarshalFunc }})
    {{- else -}}
    arg{{$i}}, err := graphql.ProcessArgField(ctx, rawArgs, {{$arg.Name|quote}}, ec.{{ $arg.TypeReference.UnmarshalFunc }})
    {{- end }}
{{- end }}
```

with:

```
{{- else -}}
    arg{{$i}}, err := graphql.ProcessArgField(ctx, rawArgs, {{$arg.Name|quote}},
        func(ctx context.Context, v any) ({{ $arg.TypeReference.GO | ref }}, error) {
            return {{$.ECDot}}{{$arg.TypeReference.UnmarshalFunc}}(ctx, {{$.ECArg}}v)
        })
{{- end }}
```

This is the last `if $.Config.UseFunctionSyntaxForExecutionContext` branch in
any template file.

**Pre-committed generated files (regenerate with `go generate`)**

The following files are checked into git and must be regenerated:

```
go generate ./codegen/testserver/...
go generate ./plugin/federation/...
go generate ./integration/...
go generate ./testdata/...
```

Specific files known to use `ProcessArgFieldWithEC` that will change:
- `plugin/federation/testdata/usefunctionsyntaxforexecutioncontext/generated/exec.go`

Specific files known to use the old `ProcessArgField` method-value form that will change:
- `testdata/entitydirectives/generated/exec.go`
- `integration/server/generated.go`
- All testserver `generated.go` files in `codegen/testserver/`

### Testing

After making the changes above:

```bash
golangci-lint run --fix --config=./.golangci.yml ./...
golangci-lint fmt --config=./.golangci.yml ./...
go generate ./codegen/testserver/...
go generate ./plugin/federation/...
# regenerate any other pre-committed generated files
go build ./...
go test ./...
```

The `codegen/testserver/usefunctionsyntaxforexecutioncontext` package provides
end-to-end coverage of the function-syntax mode. The `codegen/testserver/followschema`
and `singlefile` packages cover the receiver-syntax mode. `TestLayouts` in
`codegen/testserver` verifies that both layouts produce identical resolver
declarations.

### What this change does:

- There are zero `if $useFunctionSyntaxForExecutionContext` (or
  `if $.Config.UseFunctionSyntaxForExecutionContext`) branches remaining in
  any `.gotpl` template file
- `ProcessArgFieldWithEC` no longer exists in the `graphql` runtime package
- All generated argument-parsing code uses the same structural form regardless
  of which execution-context mode is configured
- The `ECDot` and `ECArg` helpers from the earlier PRs are now
  used uniformly at every call site, including argument processing


graphql/args.go
```diff
@@ -21,22 +21,3 @@ func ProcessArgField[T any](
 	ctx = WithPathContext(ctx, NewPathWithField(fieldName))
 	return valueMapperFn(ctx, value)
 }
-
```
StevenACoffman, line 24: The actual change!

codegen/args.gotpl
```diff
@@ -7,11 +7,10 @@ func {{$.FuncReceiver}}{{ $name }}(ctx context.Context, {{$.ECFuncParam}}rawArgs
 		{{ if $arg.ImplDirectives }}
 			arg{{$i}}, err := {{$.ECDot}}{{ $name }}{{$arg.Name | go}}(ctx, {{$.ECArg}}rawArgs)
 		{{- else -}}
-			{{ if $.Config.UseFunctionSyntaxForExecutionContext -}}
-			arg{{$i}}, err := graphql.ProcessArgFieldWithEC(ctx, ec, rawArgs, {{$arg.Name|quote}}, {{ $arg.TypeReference.UnmarshalFunc }})
-			{{- else -}}
-			arg{{$i}}, err := graphql.ProcessArgField(ctx, rawArgs, {{$arg.Name|quote}}, ec.{{ $arg.TypeReference.UnmarshalFunc }})
-			{{- end }}
+			arg{{$i}}, err := graphql.ProcessArgField(ctx, rawArgs, {{$arg.Name|quote}},
```
StevenACoffman, line 10: The actual change!

PR #4148, opened 2026-04-26
Add PayloadReadLimit to WebSocket transport

### LINKED ISSUE #3518: Easy DoS/DDoS with WebSocket Transport
The current implementation of the WebSocket transport doesn't provide enough options to protect against DoS/DDoS by users sending large payloads over multiple WebSocket connections, which will then be processed in parallel.

The payload size limit (imposed by gorilla/websocket?) seems to be ~100 MB, and with less than 100 connections spamming such large single-frame payloads in parallel (with `EnableCompression: true` it's just a few kB sent over the network, and you can compose valid payloads that won't cause a disconnect, e.g. a ~100 MB long, easily compressible `operationName`), one may start noticing performance degradation. With 100-200+ connections I made one of my servers unresponsive.

Now, with `InitFunc` in combination with `OnCloseFunc`, we could already count concurrent connections per IP or some other public client identifier, and reject further upgrades in `CheckOrigin` if too many are opened to partially mitigate this. 

However, unless I'm missing something, **the `InitFunc` doesn't let us access the connection to set a safe, maximum read limit**, like:

```go
InitFunc: func(ctx context.Context, initPayload transport.InitPayload, conn *wsConnection) (context.Context, *transport.InitPayload, error) {
   conn.SetReadLimit(2500) // https://pkg.go.dev/github.com/gorilla/websocket@v1.5.0#Conn.SetReadLimit
   // but conn isn't passed, also not in context
}
```

This would make such DoS/DDoS attacks unfeasible, since Gorilla would safely abort reading large payloads if a single frame or multiple frames would accumulatively exceed the limit, close the connection, which would then force the user to open a new one, where they'd be faced with our custom connection upgrade rate limiter.

---

If something speaks against passing the whole connection, **I think it would be good to set a configurable read limit with some default** that is much less than 100 MB - 1 MB perhaps - in the [websocket init func](https://github.com/99designs/gqlgen/blob/master/graphql/handler/transport/websocket.go#L164), given that GraphQL payloads sent by clients via the WebSocket transport should always be < 1 MB.

```go
type Websocket struct {
    PayloadReadLimit *int

    // [...]
}

func (c *wsConnection) init() bool {
    limit := someDefault
    if c.PayloadReadLimit != nil {
        limit = *c.PayloadReadLimit
    }
    c.conn.SetReadLimit(limit)

    // [...]
}

```

Obviously one could create a custom Transport, but this seems like a basic thing that could be added with 1 line of code without any downside.

Any thoughts?

jayaraman-venkatesan:
Fixes #3518 

The WebSocket transport had no cap on incoming message size. Gorilla/websocket's default read limit is effectively unlimited, making it trivial to DoS a server by opening many connections and sending large (but highly compressible) payloads in parallel.

If during development, a user wants to explicitly set it to the old **_unsafe_**, unlimited behavior, [they can explicitly set the limit to 0 based on the Gorilla websocket code](https://github.com/gorilla/websocket/blob/ac0789be11725ab2285233e9a3800c2312cff4fc/conn.go#L941) so this does not break backwards compatibility as users can opt-out of the new behavior.

  ## Changes                                                                                                                                                                                                        
                  
  - Added `PayloadReadLimit *int64` field to the `Websocket` struct
  - Default limit is **1 MB** when `PayloadReadLimit` is `nil`
  - Uses gorilla's `SetReadLimit` — oversized messages are rejected immediately without being processed, and the connection is closed                                                                                                                                                           
  - Added a test (`TestWebsocketWithPayloadReadLimit`) verifying the limit is enforced                                                                                                                              
  - Updated the subscriptions docs with usage examples                                                              
                                                                                                                                                                                                                    
  ## Usage                                                                                                                                                                                                          
                  
  ```go                                                                                                                                                                                                             
  // Use the 1 MB default (no config needed)
  srv.AddTransport(transport.Websocket{})
                                                                                                                                                                                                                    
  // Custom limit
  limit := int64(512 * 1024) // 512 KB
      PayloadReadLimit: &limit,
  })
  
  // Explicitly Set to Prior *Unsafe* Unlimited
    limit := int64(0) // Unsafe Unlimited
      PayloadReadLimit: &limit,
  })
```

I have:

### DIFF
--- a/docs/content/recipes/subscriptions.md
+++ b/docs/content/recipes/subscriptions.md
@@ -116,6 +116,33 @@ srv.AddTransport(transport.Websocket{
 })
 ```
 
+### Limiting Payload Size
+
+By default, the WebSocket transport enforces a **1 MB read limit** per message. This protects
+your server against DoS/DDoS attacks where clients send very large payloads over many concurrent
+connections — even when compression makes those payloads tiny on the wire.
+
+You can tune this limit using `PayloadReadLimit`:
+
+```go
+// Tighten the limit to 512 KB for your use case.
+limit := int64(512 * 1024)
+srv.AddTransport(transport.Websocket{
+	PayloadReadLimit: &limit,
+})
+```
+
+```go
+// Relax the limit to 10 MB if your application sends large payloads.
+limit := int64(10 * 1024 * 1024)
+srv.AddTransport(transport.Websocket{
+	PayloadReadLimit: &limit,
+})
+```
+
+When a client sends a message that exceeds the limit, gorilla/websocket closes the connection
+immediately without processing the payload.
+
 [code]: https://github.com/99designs/gqlgen/blob/master/graphql/handler/transport/websocket.go
 [gorilla]: https://pkg.go.dev/github.com/gorilla/websocket
 [graphql-ws]: https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md
--- a/graphql/handler/transport/websocket.go
+++ b/graphql/handler/transport/websocket.go
@@ -39,6 +39,10 @@ type (
 		 */
 		MissingPongOk bool
 
+		// PayloadReadLimit is the maximum size in bytes of a WebSocket message read from the client.
+		// If nil, defaults to 1 MB. Set to a pointer of -1 to disable the limit entirely.
+		PayloadReadLimit *int64
+
 		didInjectSubprotocols bool
 	}
 	wsConnection struct {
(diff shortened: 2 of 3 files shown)


jayaraman-venkatesan:
@StevenACoffman the work is done for this issue. 
can you have a look at the PR whenever free? Thanks

StevenACoffman:
@jayaraman-venkatesan I had to figure out if it would avoid breaking backwards compatibility. Do my edits to your description look correct?

PR #4149, opened 2026-04-26
refactor(codegen): split root and per-schema templates, drop layout guards

toga4:
The intent of this is to reduce the complexity of the logic that is _inside_ the templates, so we have easier maintenance, fewer bugs, and in general more testable code.

This is implementing the structural refactor I sketched in https://github.com/99designs/gqlgen/pull/4121#issuecomment-4310562129 .

> Rough shape:
> 
> - **Root templates** — schema-global, rendered once:
>   `NewExecutableSchema`, `Config`, `Exec()`, location middleware, etc.
> - **Per-schema templates** — per source file: `field.gotpl`,  `args.gotpl`, object/input/interface generation, etc.
> - **Shared partials** — `{{ define }}` blocks only
> 
> Then:
> 
> - `follow-schema`: root templates → `root_.generated.go`;
>   per-schema templates per file → `xxx.generated.go`
> - `single-file`: both together → `generated.go`
> 
> That removes the layout guards, the `generated!.gotpl` / `root_.gotpl` duplication, and makes the orphan-directive bug disappear naturally.

## What changes

Eliminates the body duplication between `root_.gotpl` and `generated!.gotpl`.

Mechanism:
- Three layout-specific `embed.FS` declarations in `generate.go`, each enumerating the templates relevant to that layout (root / per-schema / single-file).
- The two `{{ if eq .Config.Exec.Layout "single-file" }}` guards inside `generated!.gotpl` are removed; layout decision moves to `generate.go` selecting the appropriate `TemplateFS` per layout.
- `root_.gotpl` is split along the public-API / internal-runtime boundary into `api!.gotpl` (NewExecutableSchema / Config / ResolverRoot / DirectiveRoot / ComplexityRoot) and `internal!.gotpl` (BuiltInDirectives / executableSchema methods / executionContext / sources / parsedSchema / childFields_* / globalMiddleware). `generated!.gotpl` is reduced to just the resolver interface declarations. `directives.gotpl` → `directives_.gotpl` to mark it as a partial.
- The `api!` / `generated!` / `internal!` names sort naturally `a < g < i` in the existing `!.gotpl` priority ordering, keeping the public API at the top of the single-file generated file and preserving the UX intent of the original `!.gotpl` convention.
- `single-file` mode renders `api!` / `generated!` / `internal!` + per-schema in one pass; `follow-schema` mode renders root (`api!` + `internal!`) once and per-schema per source file.

`templates.Options`, `templates.Render` signature/behavior, template body contents, and plugin contracts (`Render` call sites in `servergen` / `modelgen` / `resolvergen` / `stubgen` / `federation`) are unchanged. `TestLayouts` passes, and follow-schema's `root_.generated.go` is byte-identical to master. Single-file output gets new region tags separating the `api!` / `generated!` / `internal!` blocks (and loses the empty `directives.gotpl` region markers from the `directives_.gotpl` rename), but the Go code body remains semantically equivalent to master.

I have:


_examples/chat/generated.go
```diff
@@ -115,24 +125,28 @@ func (e *executableSchema) Complexity(ctx context.Context, typeName, field strin
 		}
 
 		return e.ComplexityRoot.Message.CreatedAt(childComplexity), true
+
```
StevenACoffman: Nit: There's extra noise here with these blank lines that are making it a bit harder to see the actual generated differences.

StevenACoffman:
@toga4 I edited your description to hopefully give more context (please feel free to correct it further). I would love to get feedback from others, so I'm going to leave it open for a bit, if you don't mind.

_examples/chat/generated.go
toga4: Right, this noise is a side effect of the dedupe. The two files had opposite whitespace conventions on equivalent `{{ end }}`s, so collapsing them into one source has to pick a side:

https://github.com/99designs/gqlgen/blob/e7ed9be51de21fb6b1eaaed863079e757e8ed7fa/codegen%2Fgenerated%21.gotpl#L130

https://github.com/99designs/gqlgen/blob/e7ed9be51de21fb6b1eaaed863079e757e8ed7fa/codegen/internal%21.gotpl#L40

The same pair shows up the other way around for the empty-`Complexity` body — `generated!.gotpl` keeps a blank line before `return 0, false`, `internal!.gotpl` drops it:

https://github.com/99designs/gqlgen/blob/e7ed9be51de21fb6b1eaaed863079e757e8ed7fa/codegen/generated%21.gotpl#L110

https://github.com/99designs/gqlgen/blob/e7ed9be51de21fb6b1eaaed863079e757e8ed7fa/codegen/internal%21.gotpl#L20

So if I remove the new blank lines between cases here, a blank line disappears in the empty-`Complexity` case in another file (e.g. `_examples/federation/products/graph/generated.go`). Either direction is purely cosmetic — happy to flip it to the no-blank-line side if you prefer that one. Let me know.

StevenACoffman: Yeah, I see what you mean. I think I'd rather _remove_ lines rather than _add_ them for generated code. People aren't really reading it as much.

toga4: Makes sense, done in fdaa99b3 . Flipped `internal!.gotpl#L40` to `{{- end }}` so the case-separator blank lines are gone. The other pair was already on the no-blank-line side, so nothing more to change.

StevenACoffman:
The rename of `directives.gotpl` to `directives_.gotpl` is interesting. 

I'm assuming that stems from https://github.com/99designs/gqlgen/blob/e3f5362536a60f15e5973840f5d1b7cd43d5f7cc/codegen/templates/templates.go#L106 where it says:
> anything that ends in `_.gotpl` is "special" and excluded from being appended to the roots in the `Render()`.

It might be better to frame this as:
> Which files act solely as libraries of named sub-templates and produce no output when executed as a root? Those are the files that should carry the `_.gotpl` suffix.

`field.gotpl` is sort of similar-ish though: it has two `{{ define }}` blocks (`fieldDefinition` and `singleFieldDefinition`) embedded at the bottom, but both are only ever called from within `field.gotpl` itself (lines 38 and 153). Those defines are internal helpers, not a reason to rename the file.

StevenACoffman:
I had a couple people DM me and say they liked this, so I'm going to go ahead and merge it, but I would still like any later opinions or reviews!

Issue #4167, opened 2026-05-05
Document ResponseHandler termination and OneShot helper to prevent infinite loops in middleware


elwinar:
When creating custom OperationMiddleware that short-circuits to return an error, returning a custom graphql.ResponseHandler that constantly yields a *graphql.Response causes an infinite loop on streaming transports (like WebSockets).

Example: 

```
func authMiddleware(ctx context.Context, next graphql.OperationHandler) graphql.ResponseHandler {
	user, err := authenticateUser(ctx)
	if err != nil {
		return ErrorHandler(http.StatusText(http.StatusInternalServerError), api.CodeInternalServerError)
	}
	ctx = withUser(ctx, user)
	return next(ctx)
}

func ErrorHandler(msg, code string) graphql.ResponseHandler {
	return func(ctx context.Context) *graphql.Response {
  		return &graphql.Response{
			Errors: []*gqlerror.Error{
				{
					Message: msg,
					Extensions: map[string]any{
						"code": code,
					},
				},
			},
		}
	}
}
```

The expected solution seems to be to use the `graphql.OneShot` instead of a static graphql.Response, but I couldn't find any documentation about it, nor was the requirement clearly written somewhere.

I guess changing this behavior would be breaking already existing code, and breaking compatibility isn't desirable, but maybe something could be written explicitely in some documentation? I don't see the middleware system being explained anywhere except the 0.11 migration page here https://gqlgen.com/recipes/migration-0.11/ .
Would you accept a pull request with a new Reference page for middlewares, and/or the corresponding godoc comments?

StevenACoffman:
Yeah, this would be great! Thanks!

elwinar:
I've opened the PR. Also, I was thinking about code changes:

What bit me here, especially, was that my socket was closed because the client disconnected after the first error response, but the middleware continued spining with a `socket closed` error (or something similar, can't be bothered to re setup to check right now.) Maybe this specific error could be intercepted and used to abort the stream explicitely? It wouldn't change the interface and only save resources for an explicitly bad behavior in the websocket transport.

Also, I've noticed that when an OperationMiddleware short-circuit like this, the operation type in the context is set to nil. Maybe we can count the iterations and abort if count > 1 && operationType == nil ? I'm not sure if there are legitimate cases for this behavior.

StevenACoffman:
What do you think about #4170 ?

StevenACoffman:
[...]

PR #4169, opened 2026-05-06
Issue 4167 document middlewares

elwinar:
Relating to https://github.com/99designs/gqlgen/issues/4167, I have added some documentation about Middlewares, streaming ResponseHandler, and the OneShot helper, with help from Gemini for making better sentences (because I'm really bad at that, and he's very good.)


StevenACoffman:
Thanks!

PR #4170, opened 2026-05-08
allow custom websocket close code and reason

### LINKED ISSUE #2349: websocket InitFunc response code
### What happened?

Currently, I'm running into a problem where handling WebSocket transport authentication errors using gqlgen's recommended method (by returning `errors.New("AUTHENTICATION_ERROR")` from the transport's InitFunc()) is causing my graphql client (Apollo) to retry connections continuously. I believe that the reason for this is because the WebSocket transport closes the connection with a 1000 (normal closure) error code when InitFunc() returns an error. If I could return a more appropriate error code (e.g. 4403) then that would prevent Apollo from automatically retrying the connection. 

The relevant code is here:
https://github.com/99designs/gqlgen/blob/master/graphql/handler/transport/websocket.go#L179-L184

Is there a way to modify the response code from InitFunc()? Would you accept a pull request that would make the response code configurable?

### What did you expect?

N/A

N/A

### versions
 - `go run github.com/99designs/gqlgen version` v0.17.16
 - `go version`? go version go1.19 darwin/amd64

### LINKED ISSUE #2881: Pass error code in Websocket InitFunc
### What happened?
For websockets, authentication errors in `InitFunc`, closes the connection with a normal closure (status code: 1000).
However, as defined [here](https://github.com/enisdenjo/graphql-ws/pull/2#issue-674412981), for the ConnectionInit message,

> Closing the socket with a close event 4403: Forbidden indicating that the connection request has been denied because of access control. The server has returned false in the onConnect callback.
> Closing the socket with a close event 4400: <error-message> indicating that the connection request has been denied because of an implementation specific error. The server has thrown an error in the onConnect callback, the thrown error's message is the <error-message> in the close event.

### What did you expect?
The ability to specify the close code for connection init i.e either 4403 or 4400 depending on the use-case and the ability to send a custom message. Currently, the string "terminated" is hardcoded in the library.

NA

### versions
 - `go run github.com/99designs/gqlgen version`? 0.17.5, though the code is the same in the latest version as well
 - `go version`? 1.20.4

NateScarlet:
when WebsocketInitFunc returns error

close #2349, close #2881

I have:


StevenACoffman:
@UnAfraid I'm not using websockets, but I think you are, can you take a look at this one?

UnAfraid:
Could you please resolve the conflicts?
Also please avoid gorilla dependence

UnAfraid (approved):

StevenACoffman:
Thanks very much!

amorey:
@NateScarlet @StevenACoffman Thanks for fixing this!

Issue #2281, opened 2022-07-07
Directive hook is called with the wrong objects


maaft:
### What happened?

Consider this schema
```
directive @oneOf on INPUT_OBJECT | FIELD_DEFINITION

input UserBy @oneOf {
  id: ID
  email: String
  username: String
  registrationNumber: Int
}

input UserInput {
  by: UserBy!
}

type Query {
  user(input: UserInput!): User
}
```

And this directive hook:

```golang
c.Directives.OneOf = func(ctx context.Context, obj interface{}, next graphql.Resolver) (res interface{}, err error) {
   return next(ctx)
}
```

When executing the `user` query, the hook will be executed twice. Once with `obj = UserInput{}` and once again with `obj = UserBy{}`.

### What did you expect?

I expect that the hook will only be called **once** with `obj = UserBy{}` argument. The directive `@oneOf` was only attached to `input UserBy` and **not** to `input UserInput`.

If it will also be called with `obj = UserInput{}`, there is currently **no way** to do input-field validation, like it is needed for the `@oneOf` directive.

### Minimal graphql.schema and models to reproduce
 see above

### versions
- go 1.18.3
- latest gqlgen

frederikhors:
Can you PR?

maaft:
> Can you PR?

Are you asking me if I can fix it? Is this bug confirmed? I'd be happy to PR if thats the case

frederikhors:
If you have ideas on how to fix it you can give it a try with a PR. Unfortunately it is not possible to define a date for if and when someone will work on it.

maaft:
> If you have ideas on how to fix it you can give it a try with a PR. Unfortunately it is not possible to define a date for if and when someone will work on it.

Just wanting to make sure, if this is indeed a bug and not intended behavior before committing any time on this.

frederikhors:
The definition of a bug varies. Maybe it's just something no one has needed before. And now it's called feature.

akwirick:
I'm seeing the same thing while, ironically, trying to implement a custom `oneOf` directive for an Input Object.  

Looks like the issue in my case is that the template is getting unrolled in the codegen for the INPUT FIELD instead of the INPUT OBJECT.  You can see it getting templatized at the field level in [codegen/input.gotpl](https://github.com/99designs/gqlgen/blob/779d7cdd4991e3cf4bf1ecbdea1f02664a56ac8d/codegen/input.gotpl#L29)

[...]

PR #4176, opened 2026-05-13
propagate OBJECT/INTERFACE directives to field resolvers

thumbnail:
PR [#4029](https://github.com/99designs/gqlgen/pull/4029) tightened the directive filter in `codegen/field.go` → `(*builder).bindField` to fix issue [#2281](https://github.com/99designs/gqlgen/issues/2281) (INPUT_OBJECT directives applied twice).

In doing so it dropped the `for _, dir := range obj.Directives { ... }` loop that — by happy accident — was propagating parent OBJECT-level directives down to every field of that object. Any schema that relied on

```graphql
directive @auth(...) on OBJECT | FIELD_DEFINITION | INPUT_FIELD_DEFINITION | INPUT_OBJECT
type Secret @auth(roles: [ADMIN]) {
    id: ID!
    value: String!
}
```

now silently emits field resolvers for `Secret.id` / `Secret.value` with `nil` `middlewareChain`, i.e. **no directive enforcement at runtime** — an authorization bypass. Field-level `@auth` directives still work because they're attached directly to the `Field` struct; the propagation path is the only thing broken.

I have:

### DIFF
--- a/codegen/field.go
+++ b/codegen/field.go
@@ -132,6 +132,19 @@ func (b *builder) bindField(obj *Object, f *Field) (errret error) {
 
 			f.Directives = append(filteredDirs, f.Directives...)
 		}
+
+		// Propagate parent OBJECT/INTERFACE directives down to each field
+		// so that `type Foo @auth { id: ID! }` enforces @auth on every
+		// field resolver of Foo. Input objects are excluded — their
+		// directives run at the input-object site (#2281). Only directives
+		// whose declaration permits FIELD_DEFINITION are eligible.
+		if obj.Kind == ast.Object || obj.Kind == ast.Interface {
+			for _, dir := range obj.Directives {
+				if dir.IsLocation(ast.LocationFieldDefinition) {
+					f.Directives = append(f.Directives, dir)
+				}
+			}
+		}
 	}()
 
 	f.Stream = obj.Stream
--- a/codegen/object_directive_propagation_test.go
+++ b/codegen/object_directive_propagation_test.go
@@ -1,0 +1,130 @@
+package codegen
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/vektah/gqlparser/v2"
+	"github.com/vektah/gqlparser/v2/ast"
+
+	"github.com/99designs/gqlgen/codegen/config"
+	"github.com/99designs/gqlgen/internal/code"
+)
+
+// TestObjectDirectivePropagation guards against the regression introduced in
+// PR #4029 (v0.17.87), where directives declared with both OBJECT and
+// INPUT_OBJECT locations were stripped from every field of OBJECT/INTERFACE
+// types — even though they were applied at the OBJECT level. The INPUT_OBJECT
+// guard from issue #2281 must still hold.
+func TestObjectDirectivePropagation(t *testing.T) {
+	const schemaSrc = `
+		schema { query: Query, mutation: Mutation }
+
+		directive @auth(roles: [String!]!) on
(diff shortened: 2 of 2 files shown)


StevenACoffman:
@parkerroan @atzedus FYI

atzedus:
After reviewing the official GraphQL specification and how directive execution is defined, I’d like to share a few observations:

The spec explicitly states that directives must only be used in the locations they are declared for, but it **does not prescribe** whether a directive applied to a parent type (e.g., `OBJECT` or `INTERFACE`) should automatically propagate to its child fields (`FIELD_DEFINITION`). At the same time, the spec **does not forbid** implementations from introducing such behavior. In general, explicitly applying directives to fields is always the clearest and most predictable approach, as it aligns with GraphQL’s design principle of explicit schema definitions.

That said, gqlgen historically propagated directives from types to their fields. Whether this was an intentional design choice or an incidental side effect? I don't know :(. For those who depending on this propagation (especially for auth/middleware), this PR addresses a critical regression. For those who never expected or wanted implicit propagation, the change might feel like a behavioral shift, but overall, stabilizing and making this behavior explicit is a net positive for the ecosystem.

I strongly recommend:
1. **Documenting this propagation behavior** clearly in gqlgen’s official docs, so users know exactly how and when type-level directives are applied to fields.
2. **Introducing a configuration flag** (e.g., `propagate_type_directives: true/false`) to give explicit control over the behavior. I’d suggest keeping propagation `true` by default to maintain backward compatibility and align with current expectations, while allowing projects to opt out if they prefer strict, field-level explicitness.

@StevenACoffman

thumbnail:
Thanks for the detailed write-up. After reviewing the spec more closely, I’m not sure I agree that reintroducing propagation is the right fix.

The spec allows directives on `OBJECT`, `INTERFACE`, and `FIELD_DEFINITION`, but it treats those as separate directive locations. It does not define propagation from a type-level directive to its fields, or from a return type to the field returning it.

That makes propagation a gqlgen-specific behavior rather than GraphQL-defined behavior.

For auth in particular, implicit propagation can be ambiguous. A directive on `type User @auth` could reasonably mean:

- protect fields declared on `User`;
- protect fields returning `User`;
- provide fallback metadata for a custom auth layer;
- or simply annotate the type without affecting field middleware.

Those are application-level semantics, and different users may want different behavior.

So I think the safest path is not to silently restore propagation by default. If gqlgen historically propagated type directives, then the removal should be called out clearly as a behavioral change/breaking change, especially for users relying on it for auth. Consumers who want propagation can then implement it explicitly in their directive/auth handling.

If gqlgen still wants to preserve the old behavior, the explicit config flag seems like a good alternative. But maybe opt-in is the way to go (still breaking consumers 😅)

StevenACoffman:
@thumbnail @atzedus 
> So I think the safest path is not to silently restore propagation by default. If gqlgen historically propagated type directives, then the removal should be called out clearly as a behavioral change/breaking change, especially for users relying on it for auth. Consumers who want propagation can then implement it explicitly in their directive/auth handling.

Where do you think would be a good place to document this?

thumbnail:
> Where do you think would be a good place to document this?

I’m thinking about (retroactively) updating the changelogs (possibly the next one too, so it can be picked up with automatic dependency management tooling) and mentioning it in the readme?I admit it’s not the most fool-proof solution; but increasing reliance on the original (subtly wrong behavior), would arguably be worse 🤔.

thumbnail:
We ended up migrating away from using directives in our team; Alternative for us would be to add more directives (instead of relying on the propagation).

PR #4177, opened 2026-05-13
Add error unwrapping to WebsocketError

jrh3k5:
The current `WebsocketError` does not implement `Unwrap()`. Consequently, we are unable to access the underlying error to understand how we should handle the error (e.g., we want to log errors due to abnormal socket closures as a non-error severity).

I have:

### DIFF
--- a/graphql/handler/transport/websocket.go
+++ b/graphql/handler/transport/websocket.go
@@ -94,6 +94,10 @@ func (e WebsocketError) Error() string {
 	return fmt.Sprintf("websocket write: %v", e.Err)
 }
 
+func (e WebsocketError) Unwrap() error {
+	return e.Err
+}
+
 var (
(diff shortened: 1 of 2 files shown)


StevenACoffman:
Hey, thanks! BTW, gorilla/websocket is back to being considered an unmaintained security problem, so we need to migrate off of it, and I would like help replacing or at least testing the replacement(s).

jrh3k5:
> and I would like help replacing or at least testing the replacement(s).

@StevenACoffman - I'm interested in understanding what this help looks like. I can, honestly, probably more easily commit to helping with a replacement implementation than doing at-scale testing (the former would be me as a private developer, the latter would be seeing if we can use my employer's production infra as a guinea pig, which is a harder sell).

Issue #4114, opened 2026-04-04
Qodo AntiSlop scan found 22 issues across 10 recent PRs (16 confirmed on main)


gvago:
Hey team,

A user recently scanned this repo using [Qodo's AntiSlop Scanner](https://qodo.ai/ai-code-review-tool). The analysis reviewed 10 recent PRs and found **22 issues**, all confirmed to still exist on main.

Here's one example:

### JSON null panics params (transports)
**Severity:** action_required | **Category:** security

In several transports, the body is decoded into a `**graphql.RawParams` (by passing `&params`) and then `params` is immediately dereferenced. A valid JSON body of `null` can set `params` to nil and cause a panic when assigning `params.Headers`/`ReadTime` (trivial DoS).

**How to validate:** Send a POST request with a JSON body of `null` to any transport endpoint and observe the nil pointer panic.

**Agent prompt to fix:**
> In each transport's request handler, add a nil check after JSON decoding: if `params` is nil after decode, return a 400 error instead of dereferencing it. Check all transports (HTTP POST, multipart, GET, etc.) for the same pattern.

---

### Other confirmed issues

| # | Title | Category | PR |
|---|-------|----------|----|
| 1 | Incremental deletes outputs before generation | reliability | #4042 |
| 2 | No selective regeneration despite API claim | correctness | #4042 |
| 3 | Build tags omitted from validation | correctness | #4071 |
| 4 | Validation builds unrelated subpackages | reliability | #4071 |
| 5 | Batch directive panic on nil/non-bool value | reliability | #4040 |
| 6 | Batch cover drops errors silently | correctness | #4043 |
| 7 | Batch profile drops errors silently | correctness | #4043 |
| 8 | Breaking batch API change without migration | maintainability | #4087 |
| 9 | IndexMap missing causes panic | reliability | #4087 |
| 10 | Nil field dereference panic in error locations | reliability | #4090 |

**11 additional findings** (including suspected issues) are available in the [full report](https://www.qodo.ai/ai-code-review-tool/view/hc-report/99designs/gqlgen/767aca6840b3433abd85d3a006c9b444).

Enjoy!

P.S. Qodo offers free tooling for open-source maintainers: https://www.qodo.ai/solutions/open-source/

cc @vektah @StevenACoffman @neelance @vvakame

StevenACoffman:
Hey, that's pretty neat!
> Stale config schema option
[...]

PR #4178, opened 2026-05-15
Fix potential panic in batch directive cast

jeolted:
The batch directive could have null as the value which will cause a panic in Config.Init(). The code did a cast to bool without type check.

Solves number 5 in https://github.com/99designs/gqlgen/issues/4114

I have:


jeolted:
I'm not 100% sure I got the test right. I had to read up on how directives work.

StevenACoffman:
Thanks!

PR #4183, opened 2026-05-20
Deprecate github.com/gorilla/websocket

UnAfraid:
Deprecated `github.com/gorilla/websocket` and added alternative implementation `github.com/coder/websocket`.
The new implementation can be extended and new websocket implementations can be added.
This change should also be backwards compatible.

I have:
 - [X] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [X] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

PR #4184, opened 2026-05-22
Remove usage of github.com/PuerkitoBio/goquery

jeolted:
Replaced github.com/PuerkitoBio/goquery with golang.org/x/net/html The goquery library was only used in one test and it was not using any specific features from goquery. The functionality could easily be replaced with golang.org/x/net/html directly, which goquery uses aswell.

This reduces the maintenance burden and lowers the risk.

I have:


_examples/mini-habr-with-subscriptions/go.mod
```diff
@@ -4,9 +4,9 @@ go 1.25.0
 
 require (
 	github.com/99designs/gqlgen v0.17.89
+	github.com/coder/websocket v1.8.14
```
jeolted, line 7: go mod tidy was forgotten in https://github.com/99designs/gqlgen/pull/4183 I believe.

PR #4193, opened 2026-05-29
Fix batch resolvers reusing a single `_federationRequires` for all parents (#4192)

### LINKED ISSUE #4192: Bug: computed_requires + @goField(batch: true) passes a single _federationRequires for all parents
**Related:** Based on https://github.com/99designs/gqlgen/issues/4140 (federation batching + `@requires` on array fields). This issue is a narrower **codegen bug**: `computed_requires` + `@goField(batch: true)`.

---

### Environment
- gqlgen v0.17.90
- Federation v2, `computed_requires: true`, `call_argument_directives_with_null: true`
- `@requires` + `@goField(batch: true)` on the same field

### Schema (minimal)

```graphql
type Product @key(fields: "id") {
  id: ID!
  variations: [Variation!]! @external
  bestPrice: String!
    @requires(fields: "variations { price }")
    @goField(batch: true)
}
```

### Generated resolver signatures

**Non-batch (correct per entity):**
```go
BestPrice(ctx context.Context, obj *Product, federationRequires map[string]any) (string, error)
```
`federationRequires` is built per entity via `populateFromRepresentations` + `fc.Parent.Index`.

**Batch (incorrect — single requires for all parents):**
```go
BestPrice(ctx context.Context, objs []*Product, federationRequires map[string]any) ([]string, error)
```
One `federationRequires` is passed for the whole `objs` slice.

**Batch (expected):**
```go
BestPrice(ctx context.Context, objs []*Product, federationRequires []map[string]any) ([]string, error)
// len(federationRequires) == len(objs), federationRequires[i] for objs[i]
```

### Problem
With `_entities` and N products, each representation has its own `variations`. The batch wrapper reuses one `fc.Args["_federationRequires"]` for every parent, so N−1 entities get the wrong requires payload. Without `batch: true`, each call gets the right map.

### Repro
1. `_entities` with 2+ representations, different `variations` per `id`.
2. Query `bestPrice` on each `Product`.
3. Wrong prices / errors for all but one index.

### Workaround
Read full `representations` from `_entities` in context and map by `obj.ID` (duplicates federation logic).

### Suggested fix
Generate per-parent requires for batch + `computed_requires` (e.g. `[]map[string]any` aligned with `objs`).

atzedus:
Fixes [#4192](https://github.com/99designs/gqlgen/issues/4192).

With `computed_requires` and `@goField(batch: true)` on the same `@requires` field, the batch wrapper passed one `fc.Args["_federationRequires"]` to the resolver for every parent in `objs`. Only one entity got the correct requires payload; the rest were wrong.

This PR:

- Generates batch resolver signatures with `federationRequires []map[string]any` (aligned with `objs`) when `_federationRequires` is present
- Adds `graphql.FederationRequiresForBatch` to build per-parent requires from `_entities` representations (with `IndexMap` support for interface/union batching)
- Walks the `FieldContext` parent chain to find `representations` and entity `Index` instead of hardcoding `Parent.Parent`

For `explicit_requires`, no runtime change is needed: required fields are populated on each entity via `Populate*Requires` before field resolution, so batch resolvers read `objs[i]` directly. Added integration test coverage for that path.

I have:


StevenACoffman:
Thank you!

PR #4194, opened 2026-05-29
fix: CoerceList drops elements from typed slices past the first

c-tonneslan:
`CoerceList`'s typed-slice branches (`[]string`, `[]int`, `[]float64`, `[]bool`, `[]map[string]any`, etc.) only kept `v[0]` when the input had length > 0. The `[]any` branch (the common case from JSON input) correctly preserves every element, so the typed-slice branches were asymmetric:

```go
CoerceList([]any{\"a\", \"b\"})    // []any{\"a\", \"b\"} (correct)
CoerceList([]string{\"a\", \"b\"}) // []any{\"a\"}      (drops \"b\")
```

So any custom resolver or directive passing a typed Go slice through `CoerceList` silently lost everything past the first element.

Pulled the conversion into a small generic helper `toAnySlice` so each typed branch is one line. Added `[]string{\"a\", \"b\"}`, `[]int32{1, 2}`, and `[]bool{true, false}` cases to `TestCoerceList`; the existing single-element cases still pass.

I have:
 - [x] Added tests covering the bug
 - [ ] Updated any relevant documentation


StevenACoffman:
Wow, this has been wrong for a really long time. I kind of question if now the parent can be simplified though.

PR #4204, opened 2026-06-04
chore: remove deprecated GraphQL handler implementation from handler.…

atzedus:
image

I think the time has come.


StevenACoffman:
I mean, I feel like maybe we should have given more warning... 😆

PR #4210, opened 2026-06-10
Update gqlparser to v2.5.34

StevenACoffman:
Automated update of gqlparser. See https://github.com/vektah/gqlparser/releases/tag/v2.5.34

fix: reject duplicate enum values and repeated non-repeatable directives by @advoretsky in https://github.com/vektah/gqlparser/pull/436

This requires us to make `@gotag` repeatable as it always should have been.

 Declare @goTag as repeatable in test fixtures and docs

 @goTag was intended to be repeatable from #1680 ("Rename @extraTag to  @goTag and make repeatable") and GoTagFieldHook iterates every instance via ForNames("goTag"), but the directive's `directive @goTag(...) on …` declaration was missing the `repeatable` keyword in three places. Older gqlparser versions silently tolerated this; v2.5.34 (commit 807d627 "reject … repeated non-repeatable directives") correctly rejects it.

Fix the three declarations so the docs and test fixtures match the documented behavior and the new validator.


StevenACoffman:
FYI @atzedus

Issue #4213, opened 2026-06-13
bug: @defer de-duplication


rad12000:
The reproducer can be found here: https://github.com/rad12000/gqlgen-defer-min-repro

The @defer spec implemented by gqlgen is a bit outdated. That's okay. This repository is a minimal reproducer of an issue with the defer implementation.

The problem is most easily explained by seeing it in action.

### Query

```graphql
query DeferDeduplication {
  system {
    ...JustUserDetails @defer(label: "justUserDetails")
    ...UserAndAccountDetails @defer(label: "userAndAccountDetails")
  }
}

fragment JustUserDetails on System {
  user {
    id
  }
}

fragment UserAndAccountDetails on System {
  user {
    id
    name
  }
  account {
    id
    name
  }
}
```

### Response

```json
---
Content-Type: application/json

{"data":{"system":{"user":null,"account":null}},"hasNext":true}
---
Content-Type: application/json

{"incremental":[{"data":{"user":{"id":"userId","name":""},"account":{"id":"accountId","name":"accountName"}},"label":"userAndAccountDetails","path":["system"],"hasNext":false}],"hasNext":false}
-----
```

### What's wrong

The response is entirely missing an incremental delivery result for the `justUserDetails` label. I would have expected it to be in the result set, AND to have been delivered prior to the account result, since there is no delay in the user resolver. More recent @defer specs allow for de-duplication of data, but those specs also include more fields that a client can rely on, such as `pending` and `completed`.

### Query

```graphql
query DeferDeduplication {
  system {
    ...UserAndAccountDetails @defer(label: "userAndAccountDetails")
    ...JustUserDetails @defer(label: "justUserDetails")
  }
}

fragment JustUserDetails on System {
  user {
    id
  }
}

fragment UserAndAccountDetails on System {
  user {
    id
    name
  }
  account {
    id
    name
  }
}
```

### Response

```json
---
Content-Type: application/json

{"data":{"system":{"user":null,"account":null}},"hasNext":true}
---
Content-Type: application/json

{"incremental":[{"data":{"user":{"id":"userId","name":""}},"label":"justUserDetails","path":["system"],"hasNext":true}],"hasNext":true}
---
Content-Type: application/json

{"incremental":[{"data":{"account":{"id":"accountId","name":"accountName"}},"label":"userAndAccountDetails","path":["system"],"hasNext":false}],"hasNext":false}
-----
```

### What's wrong

[...]

PR #4223, opened 2026-06-17
BREAKING: fix defer de-duplication, and update public defer structs

chad-bekmezian-snap:
This change is a breaking change on insofar as publicly exported types are concerned. Specifically the `DeferGroup` type has undergone a breaking change. The change to the server isn't a breaking change from a graphql client perspective.

Most of this pr is noise from codegen. The actual changes can be found in the graphql directory, and object.gotpl.

Addresses the issues in https://github.com/99designs/gqlgen/issues/4213.

# Examples
## Example 1
<details>
<summary>details</summary>

### Query

```graphql
query DeferDeduplication {
  system {
    ...JustUserDetails @defer(label: "justUserDetails")
    ...UserAndAccountDetails @defer(label: "userAndAccountDetails")
  }
}

fragment JustUserDetails on System {
  user {
    id
  }
}

fragment UserAndAccountDetails on System {
  user {
    id
    ... @defer(label: "userName") {
      name
    }
  }
  account {
    id
    name
  }
}
```

### Before
```raw
---
Content-Type: application/json

{"data":{"system":{"user":null,"account":null}},"hasNext":true}
---
Content-Type: application/json

{"incremental":[{"data":{"user":{"id":"userId","name":""},"account":{"id":"accountId","name":"accountName"}},"label":"userAndAccountDetails","path":["system"],"hasNext":false}],"hasNext":false}
-----
``` 

### After
```raw
---
Content-Type: application/json

{"data":{"system":{"user":null,"account":null}},"hasNext":true}
---
Content-Type: application/json

{"incremental":[{"data":{"user":{"id":"userId","name":""}},"label":"justUserDetails","path":["system"],"hasNext":true}],"hasNext":true}
---
Content-Type: application/json

{"incremental":[{"data":{"account":{"id":"accountId","name":"accountName"}},"label":"userAndAccountDetails","path":["system"],"hasNext":false}],"hasNext":false}
-----
```
</details>

### Response Diff
```diff
--- before.txt	2026-06-18 12:17:01
+++ b_after.txt	2026-06-18 12:17:07
@@ -5,5 +5,9 @@
 ---
 Content-Type: application/json
 
-{"incremental":[{"data":{"user":{"id":"userId","name":""},"account":{"id":"accountId","name":"accountName"}},"label":"userAndAccountDetails","path":["system"],"hasNext":false}],"hasNext":false}
+{"incremental":[{"data":{"user":{"id":"userId","name":""}},"label":"justUserDetails","path":["system"],"hasNext":true}],"hasNext":true}
+---
+Content-Type: application/json
+
+{"incremental":[{"data":{"account":{"id":"accountId","name":"accountName"}},"label":"userAndAccountDetails","path":["system"],"hasNext":false}],"hasNext":false}
 -----
```

### Notes
We now receive an incremental result for each distinct label in our deferred fragments.

## Example 2
<details>
<summary>details</summary>

### Query

```graphql
query DeferDeduplication {
  system {
    ...UserAndAccountDetails @defer(label: "userAndAccountDetails")
    ...JustUserDetails @defer(label: "justUserDetails")
  }
}

fragment JustUserDetails on System {
  user {
    id
  }
}

fragment UserAndAccountDetails on System {
  user {
    id
    ... @defer(label: "userName") {
      name
    }
  }
  account {
    id
    name
  }
}
```

### Before
```raw
---
Content-Type: application/json

{"data":{"system":{"user":null,"account":null}},"hasNext":true}
---
Content-Type: application/json

{"incremental":[{"data":{"user":{"id":"userId","name":""}},"label":"justUserDetails","path":["system"],"hasNext":true}],"hasNext":true}
---
Content-Type: application/json

{"incremental":[{"data":{"account":{"id":"accountId","name":"accountName"}},"label":"userAndAccountDetails","path":["system"],"hasNext":false}],"hasNext":false}
-----
```

### After
```raw
---
Content-Type: application/json

{"data":{"system":{"user":null,"account":null}},"hasNext":true}
---
Content-Type: application/json

{"incremental":[{"data":{"user":{"id":"userId","name":""}},"label":"justUserDetails","path":["system"],"hasNext":true}],"hasNext":true}
---
Content-Type: application/json

{"incremental":[{"data":{"account":{"id":"accountId","name":"accountName"}},"label":"userAndAccountDetails","path":["system"],"hasNext":false}],"hasNext":false}
-----
```

</details>

### Response Diff
```diff
```

### Notes
No change. With the guarantee that a result will be delivered for every labeled defer fragment, there is no need to start sending duplicative data.

## Example 3
<details>
<summary>details</summary>

### Query

```graphql
query DeferDeduplication {
  system {
     ...JustUserDetails @defer(label: "justUserDetails")
    ...UserAndAccountDetails @defer(label: "userAndAccountDetails")
  }
}

fragment JustUserDetails on System {
  ...UserSystemFragment @defer(label: "userSystem")
}

fragment UserSystemFragment on System {
  user {
    id
  }
}

fragment UserAndAccountDetails on System {
  user {
    id
    ... @defer(label: "userName") {
      name
    }
  }
  account {
    id
    name
  }
}
```

### Before
```raw
---
Content-Type: application/json

{"data":{"system":{"user":null,"account":null}},"hasNext":true}
---
Content-Type: application/json

{"incremental":[{"data":{"user":{"id":"userId","name":""},"account":{"id":"accountId","name":"accountName"}},"label":"userAndAccountDetails","path":["system"],"hasNext":false}],"hasNext":false}
-----
```

### After
```raw
---
Content-Type: application/json

{"data":{"system":{"user":null,"account":null}},"hasNext":true}
---
Content-Type: application/json

{"incremental":[{"data":{"user":{"id":"userId","name":""}},"label":"userSystem","path":["system"],"hasNext":true},{"data":{},"label":"justUserDetails","path":["system"],"hasNext":true}],"hasNext":true}
---
Content-Type: application/json

{"incremental":[{"data":{"account":{"id":"accountId","name":"accountName"}},"label":"userAndAccountDetails","path":["system"],"hasNext":false}],"hasNext":false}
-----
```

</details>

### Response Diff
```diff
--- before.txt	2026-06-18 12:19:42
+++ b_after.txt	2026-06-18 12:19:51
@@ -5,5 +5,9 @@
 ---
 Content-Type: application/json
 
-{"incremental":[{"data":{"user":{"id":"userId","name":""},"account":{"id":"accountId","name":"accountName"}},"label":"userAndAccountDetails","path":["system"],"hasNext":false}],"hasNext":false}
+{"incremental":[{"data":{"user":{"id":"userId","name":""}},"label":"userSystem","path":["system"],"hasNext":true},{"data":{},"label":"justUserDetails","path":["system"],"hasNext":true}],"hasNext":true}
+---
+Content-Type: application/json
+
+{"incremental":[{"data":{"account":{"id":"accountId","name":"accountName"}},"label":"userAndAccountDetails","path":["system"],"hasNext":false}],"hasNext":false}
 -----
```

### Notes
We once again get an incremental result for each deferred fragment. Even in the case of nested fragments/defers.

## Example 4
<details>
<summary>details</summary>

### Query

```graphql
query DeferDeduplication {
  system {
    user {
      id
    }
    ...JustUserDetails @defer(label: "userSystem")
    ...UserAndAccountDetails @defer(label: "userAndAccountDetails")
  }
}

fragment JustUserDetails on System {
  ...UserSystemFragment @defer(label: "userSystem")
}

fragment UserSystemFragment on System {
  user {
    id
  }
}

fragment UserAndAccountDetails on System {
  user {
    id
    ... @defer(label: "userName") {
      name
    }
  }
  account {
    id
    name
  }
}
```

### Before
```raw
---
Content-Type: application/json

{"data":{"system":{"user":null,"account":null}},"hasNext":true}
---
Content-Type: application/json

{"incremental":[{"data":{"user":{"id":"userId","name":""},"account":{"id":"accountId","name":"accountName"}},"label":"userAndAccountDetails","path":["system"],"hasNext":false}],"hasNext":false}
-----
```

### After
```raw
---
Content-Type: application/json

{"data":{"system":{"user":{"id":"userId","name":""},"account":null}},"hasNext":true}
---
Content-Type: application/json

{"incremental":[{"data":{"account":{"id":"accountId","name":"accountName"}},"label":"userAndAccountDetails","path":["system"],"hasNext":false}],"hasNext":false}
-----
```
</details>

### Response Diff
```diff
--- before.txt	2026-06-18 12:20:35
+++ b_after.txt	2026-06-18 12:20:46
@@ -1,9 +1,9 @@
 ---
 Content-Type: application/json
 
-{"data":{"system":{"user":null,"account":null}},"hasNext":true}
+{"data":{"system":{"user":{"id":"userId","name":""},"account":null}},"hasNext":true}
 ---
 Content-Type: application/json
 
-{"incremental":[{"data":{"user":{"id":"userId","name":""},"account":{"id":"accountId","name":"accountName"}},"label":"userAndAccountDetails","path":["system"],"hasNext":false}],"hasNext":false}
+{"incremental":[{"data":{"account":{"id":"accountId","name":"accountName"}},"label":"userAndAccountDetails","path":["system"],"hasNext":false}],"hasNext":false}
 -----
```

### Notes
In the "before", the user's id was requested within a non-deferred context, yet it was not included in the initial graphql payload. The [2022 spec hints at how this should be handled](https://github.com/graphql/graphql-wg/blob/e4ef5f9d5997815d9de6681655c152b6b7838b4c/rfcs/DeferStream.md#defer), stating `A query with @defer directive will cause the request to potentially return multiple responses, where non-deferred data is delivered in the initial response and data deferred delivered in a subsequent response. @include and @skip take precedence over @defer.` Since the user's id is used within a non-deferred context, it follows it should be included in the initial response. This is behavior is further established as the intended implementation in the examples provided in this discussion: https://github.com/graphql/defer-stream-wg/discussions/69.

*Open question:* as you can see in the above, if the entire dataset in a deferred fragment is delivered in an initial response, no incremental result will be sent for that fragment later. In my testing with facebook's Relay client, it seems to handle this behavior correctly. This also seems to be typical from additional reading I've done, but I am open to sending an incremental result in that case as well.

I have:
 - [ √] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [ √] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))


StevenACoffman:
I'm very excited to improve `@defer` in gqlgen for Relay support, but I also want to be careful to avoid breaking Apollo Federation which implemented a very early version of the `@defer` spec. I would be fine with advancing the target `@defer` spec version to a more recent (or latest!) version, as long as we don't break existing Apollo Federation support.

chad-bekmezian-snap:
> I'm very excited to improve `@defer` in gqlgen for Relay support, but I also want to be careful to avoid breaking Apollo Federation which implemented a very early version of the `@defer` spec. I would be fine with advancing the target `@defer` spec version to a more recent (or latest!) version, as long as we don't break existing Apollo Federation support.

Do you have a specific concern that makes you think this might break apollo federation?

I considered trying to bump us up towards the latest, but it was going to be more involved than I have the bandwidth for. These changes do lend themselves towards making the move to a newer version much easier though!

StevenACoffman:
My concern comes from [this comment](https://github.com/graphql/defer-stream-wg/discussions/69#discussioncomment-12494010):
> At the moment, Apollo Server "implements" the 2022-era format when used with a 2022-era prerelease of graphql-js v17. (By "implements" I mostly mean "serializes the objects returned by graphql-js".) We'd like to update that to the newer format.
>
> When we implemented that, we knew it was an early format up for discussion, so we made clients ask for it specifically with accept: multipart/mixed; deferSpec=20220824 (my suggestion that some accept header be required in https://github.com/graphql/defer-stream-wg/discussions/48 didn't seem to go anywhere, maybe because the focus here isn't on HTTP/JSON level).

chad-bekmezian-snap:
> My concern comes from [this comment](https://github.com/graphql/defer-stream-wg/discussions/69#discussioncomment-12494010):
> 
> > At the moment, Apollo Server "implements" the 2022-era format when used with a 2022-era prerelease of graphql-js v17. (By "implements" I mostly mean "serializes the objects returned by graphql-js".) We'd like to update that to the newer format.
> > When we implemented that, we knew it was an early format up for discussion, so we made clients ask for it specifically with accept: multipart/mixed; deferSpec=20220824 (my suggestion that some accept header be required in [graphql/defer-stream-wg#48](https://github.com/graphql/defer-stream-wg/discussions/48) didn't seem to go anywhere, maybe because the focus here isn't on HTTP/JSON level).

I'm sorry, I'm still not totally following. This pr is still the 20220824 spec. It just involves bug fixes within the implementation of that spec. What am I missing here?

I have compared this implementation to the apollo server implementation for 20220824. There are a couple of differences in the responses still, one being that the same data IS sent multiple times in the apollo implementation. However, that is a difference that existed prior to this change. In every way, this PR brings gqlgen much closer to having identical output to the Apollo Server implementation.

_examples/websocket-initfunc/server/server
StevenACoffman, line 1: It looks like you accidentally checked this binary in

StevenACoffman:
# Review:  this PR vs Apollo Server `@defer` (2022 spec)

## Wire-format verdict: **Compatible** for the typical Apollo client/server contract. Two narrow caveats below.

I verified the actual bytes on the wire by running a probe against `transport.MultipartMixed`. Initial chunk, boundary markers, `Content-Type: multipart/mixed;boundary="-";deferSpec=20220824`, and the per-chunk `{incremental:[{data,label,path,hasNext}], hasNext}` envelope are all identical to `master`. The `http_multipart_mixed.go` transport, `client/sse.go`, and `response` JSON shape are unchanged.

## What actually changed (in the defer path)

1. **`graphql/deferred.go`** — `DeferredGroup` now carries `Defers map[string]*FieldSetView` instead of a single `Label string`. **API-breaking** for any Go caller constructing `DeferredGroup`, but invisible on the wire.
2. **`graphql/fieldset.go`** — A new `FieldSetView` lets multiple labels share one underlying `FieldSet` and emit a subset of fields. `MarshalGQL` on a view delegates to the same `marshalFieldSet` writer used by `FieldSet`, so the JSON shape of each label's `data` is unchanged (`{alias: value, …}`).
3. **`graphql/executable_schema.go`** — `CollectedField` gets `IsNonDeferrable bool` and `Deferrables []*Deferrable`. The codegen template now uses these instead of the single `Deferrable`.
4. **`codegen/object.gotpl`** — One shared `deferredFieldSet` per object; each label maps to a `FieldSetView` over its fields.
5. **`graphql/execution_context_state.go`** — `ProcessDeferredGroup` fans out a `DeferredResult` per view via `view.SetOnComplete`.

## Wire output, verified empirically

```
Initial:  {"data":{"deferSingle":{"id":"1","name":"Defer test 1","otherResolvedValue":null,"values":null}},"hasNext":true}
Incr.  :  {"incremental":[{"data":{"otherResolvedValue":"other-val","values":["a","b","c"]},"label":"L1","path":["deferSingle"],"hasNext":false}],"hasNext":false}
```

Matches what Apollo Client expects for `deferSpec=20220824` (incremental array + per-payload `label/path/data`).

## Caveat 1 — Overlapping `@defer` fragments (the "dedup" semantic)

For a query that puts the *same field* under two different `@defer` labels:

```graphql
{ deferSingle { id
    ... @defer(label:"A") { name otherResolvedValue }
    ... @defer(label:"B") { otherResolvedValue values }
}}
```

graphql-js / Apollo Server would emit `otherResolvedValue` in **both** label A's and label B's payloads. gqlgen (this branch) emits it in **exactly one** — the first view whose `onComplete` fires; the other label's payload omits the field entirely. This happens because `FieldSet.takeValues` (graphql/fieldset.go:115) nils the value after the first view consumes it.

In practice this is benign with Apollo Client: it merge-patches by `path`, so the final cache state is correct. But it is a wire-shape deviation from the reference 2022 implementation. If any consumer downstream of Apollo Client inspects raw multipart chunks (custom transport, codec, or middleware) and expects "all fragment fields in each label's data," this would break.

The same query in `master` was already broken differently (only the *last* label was set, so dedup was effectively random) — so this is improvement, not regression. But it isn't full graphql-js parity.

## Caveat 2 — Errors duplicated across views in one group

`ProcessDeferredGroup` (execution_context_state.go:46-67) wires every view's `onComplete` to read `GetErrors(ctx)` from a single shared `WithFreshResponseContext(dg.Context)`. If view A completes first, then view B completes later, view B's `DeferredResult.Errors` will include A's errors too. Old code had one FieldSet per label, so each label saw only its own errors. This is a regression on the `errors` array shape when a single `DeferredGroup` carries multiple labels AND errors occur. Apollo Client likely tolerates duplicates, but it is an observable wire-format change.

## Other notes

- **Synchronous (non-resolver) fields inside `@defer` are still promoted into the initial response.** This is pre-existing behavior, preserved here. It's off-spec vs graphql-js (which would defer them) but compatible with Apollo Client because of path-merge semantics.
- **Nested `@defer`** (`@defer(label:"parent") { @defer(label:"child") { … } }`) is handled correctly via the new `parentIsDeferredFragment` flag; the new test "nested deferred fragments work" passes.
- **Initial response still has `"otherResolvedValue":null,"values":null` for deferred fields** (rather than omitting them) — same as before, and Apollo Client handles this fine.
- **Unrelated change piggybacked on the branch**: default websocket implementation flips back from `coder/websocket` to `gorilla/websocket`. Does not touch the defer/HTTP path.
- **Accidental commit**: `_examples/websocket-initfunc/server/server` is a 13 MB Mach-O arm64 binary. Should be removed before merge.

## Bottom line

For a standard Apollo Client ↔ gqlgen deployment using `Accept: multipart/mixed; deferSpec=20220824`, this branch preserves wire compatibility. The two caveats above (field dedup across labels, possible error duplication within one `DeferredGroup`) are observable in raw payloads but harmless to a normal Apollo Client because it merges by `path` and tolerates duplicates. If your stack inspects the raw multipart chunks directly, audit those two cases.

chad-bekmezian-snap:
> # Review:  this PR vs Apollo Server `@defer` (2022 spec)
> 
> 
> 
> ## Wire-format verdict: **Compatible** for the typical Apollo client/server contract. Two narrow caveats below.
> 
> 
> 
> I verified the actual bytes on the wire by running a probe against `transport.MultipartMixed`. Initial chunk, boundary markers, `Content-Type: multipart/mixed;boundary="-";deferSpec=20220824`, and the per-chunk `{incremental:[{data,label,path,hasNext}], hasNext}` envelope are all identical to `master`. The `http_multipart_mixed.go` transport, `client/sse.go`, and `response` JSON shape are unchanged.
> 
> 
> 
> ## What actually changed (in the defer path)
> 
> 
> 
> 1. **`graphql/deferred.go`** — `DeferredGroup` now carries `Defers map[string]*FieldSetView` instead of a single `Label string`. **API-breaking** for any Go caller constructing `DeferredGroup`, but invisible on the wire.
> 
> 2. **`graphql/fieldset.go`** — A new `FieldSetView` lets multiple labels share one underlying `FieldSet` and emit a subset of fields. `MarshalGQL` on a view delegates to the same `marshalFieldSet` writer used by `FieldSet`, so the JSON shape of each label's `data` is unchanged (`{alias: value, …}`).
> 
> 3. **`graphql/executable_schema.go`** — `CollectedField` gets `IsNonDeferrable bool` and `Deferrables []*Deferrable`. The codegen template now uses these instead of the single `Deferrable`.
> 
> 4. **`codegen/object.gotpl`** — One shared `deferredFieldSet` per object; each label maps to a `FieldSetView` over its fields.
> 
> 5. **`graphql/execution_context_state.go`** — `ProcessDeferredGroup` fans out a `DeferredResult` per view via `view.SetOnComplete`.
> 
> 
> 
> ## Wire output, verified empirically
> 
> 
> 
> ```
> 
> Initial:  {"data":{"deferSingle":{"id":"1","name":"Defer test 1","otherResolvedValue":null,"values":null}},"hasNext":true}
> 
> Incr.  :  {"incremental":[{"data":{"otherResolvedValue":"other-val","values":["a","b","c"]},"label":"L1","path":["deferSingle"],"hasNext":false}],"hasNext":false}
> 
> ```
> 
> 
> 
> Matches what Apollo Client expects for `deferSpec=20220824` (incremental array + per-payload `label/path/data`).
> 
> 
> 
> ## Caveat 1 — Overlapping `@defer` fragments (the "dedup" semantic)
> 
> 
> 
> For a query that puts the *same field* under two different `@defer` labels:
> 
> 
> 
> ```graphql
> 
> { deferSingle { id
> 
>     ... @defer(label:"A") { name otherResolvedValue }
> 
>     ... @defer(label:"B") { otherResolvedValue values }
> 
> }}
> 
> ```
> 
> 
> 
> graphql-js / Apollo Server would emit `otherResolvedValue` in **both** label A's and label B's payloads. gqlgen (this branch) emits it in **exactly one** — the first view whose `onComplete` fires; the other label's payload omits the field entirely. This happens because `FieldSet.takeValues` (graphql/fieldset.go:115) nils the value after the first view consumes it.
> 
> 
> 
> In practice this is benign with Apollo Client: it merge-patches by `path`, so the final cache state is correct. But it is a wire-shape deviation from the reference 2022 implementation. If any consumer downstream of Apollo Client inspects raw multipart chunks (custom transport, codec, or middleware) and expects "all fragment fields in each label's data," this would break.
> 
> 
> 
> The same query in `master` was already broken differently (only the *last* label was set, so dedup was effectively random) — so this is improvement, not regression. But it isn't full graphql-js parity.
> 
> 
> 
> ## Caveat 2 — Errors duplicated across views in one group
> 
> 
> 
> `ProcessDeferredGroup` (execution_context_state.go:46-67) wires every view's `onComplete` to read `GetErrors(ctx)` from a single shared `WithFreshResponseContext(dg.Context)`. If view A completes first, then view B completes later, view B's `DeferredResult.Errors` will include A's errors too. Old code had one FieldSet per label, so each label saw only its own errors. This is a regression on the `errors` array shape when a single `DeferredGroup` carries multiple labels AND errors occur. Apollo Client likely tolerates duplicates, but it is an observable wire-format change.
> 
> 
> 
> ## Other notes
> 
> 
> 
> - **Synchronous (non-resolver) fields inside `@defer` are still promoted into the initial response.** This is pre-existing behavior, preserved here. It's off-spec vs graphql-js (which would defer them) but compatible with Apollo Client because of path-merge semantics.
> 
> - **Nested `@defer`** (`@defer(label:"parent") { @defer(label:"child") { … } }`) is handled correctly via the new `parentIsDeferredFragment` flag; the new test "nested deferred fragments work" passes.
> 
> - **Initial response still has `"otherResolvedValue":null,"values":null` for deferred fields** (rather than omitting them) — same as before, and Apollo Client handles this fine.
> 
> - **Unrelated change piggybacked on the branch**: default websocket implementation flips back from `coder/websocket` to `gorilla/websocket`. Does not touch the defer/HTTP path.
> 
> - **Accidental commit**: `_examples/websocket-initfunc/server/server` is a 13 MB Mach-O arm64 binary. Should be removed before merge.
> 
> 
> 
> ## Bottom line
> 
> 
> 
> For a standard Apollo Client ↔ gqlgen deployment using `Accept: multipart/mixed; deferSpec=20220824`, this branch preserves wire compatibility. The two caveats above (field dedup across labels, possible error duplication within one `DeferredGroup`) are observable in raw payloads but harmless to a normal Apollo Client because it merges by `path` and tolerates duplicates. If your stack inspects the raw multipart chunks directly, audit those two cases.
> 
> 

Caveat 1 is what I was describing previously. The current gqlgen implementation unintentionally dedupes. My changes intentionally do. It's a one line change to this PR if we want to send duplicate data.

Caveat 2: The error callout is valid. Errors that occur at the same path will be sent multiple times. I couldn't change that without being much more invasive, but it didn't seem like something that should actually cause an issue in practice.

StevenACoffman:
# Relay v21.0.1 compatibility 

This PR does **not** change the multipart transport layer. The Relay-compatibility question is therefore identical for `master` and this branch: gqlgen as-shipped is **not directly compatible** with the Relay v21 network layer without a small client-side adapter.

## What Relay v21 expects on the wire

From `packages/relay-runtime/store/OperationExecutor.js:1772-1806` (`partitionGraphQLResponses`) and `packages/relay-runtime/network/RelayNetworkTypes.js:49-55` (`GraphQLResponseWithData`), each value passed to `dataSource.next(...)` must be a **flat** object:

```json
{
  "data":  { ... },
  "label": "QueryName$defer$FragmentName",
  "path":  ["node"],
  "errors": [ ... ],
  "extensions": { ... }
}
```

`label` and `path` live at the **top level**. Relay's partition step checks `response.path != null || response.label != null` and invariants on `label == null || path == null`. There is no concept of an `incremental: [...]` envelope inside the Relay runtime.

The Relay compiler (`compiler/crates/relay-transforms/src/defer_stream.rs:120-130`) **always rewrites `@defer` to include an auto-generated label** of the form `${OperationName}$defer$${FragmentName}` before sending the operation to the server — so a server never sees an unlabeled `@defer` from a Relay client.

## What gqlgen emits (verified on the wire)

```
Content-Type: multipart/mixed;boundary="-";deferSpec=20220824

---
Content-Type: application/json

{"data":{...},"hasNext":true}
---
Content-Type: application/json

{"incremental":[{"data":{...},"label":"L1","path":["deferSingle"],"hasNext":false}],"hasNext":false}
-----
```

The incremental chunks are **wrapped in `incremental: [...]`** with their own outer `hasNext`. The comment in `graphql/handler/transport/http_multipart_mixed.go:160` flags this explicitly:

```go
// TODO: Remove this wrapper on response once gqlgen supports the 2023 spec
```

This wrapper is what Apollo Client expects for `deferSpec=20220824` — but it is **not** what Relay expects.

## Incompatibilities, by severity

### 1. `incremental: [...]` envelope (BLOCKER for Relay)

If a Relay user pipes a gqlgen multipart chunk straight into `dataSource.next(...)`, Relay sees an object with no top-level `path` or `label` (they're nested inside `incremental[i]`). `partitionGraphQLResponses` classifies the chunk as a **non-incremental** payload and tries to overwrite root data. Result: store corruption, not a clean error.

### 2. `relay-tools/fetch-multipart-graphql` boundary handling (caveat)

The community fetch shim that Relay users often adopt looks for `\r\n--{boundary}\r\n` as the part delimiter and treats anything before the first delimiter as preamble. gqlgen's body starts with `--{boundary}\r\n` (no leading CRLF) — the parser still works in practice because each JSON body ends with `\r\n`, which serves as the leading CRLF of the next delimiter, but the **first part is discarded as preamble**. Verified by reading `parseMultipartHttp.js`. This means a naive integration loses the initial chunk. Workaround: write the network layer to keep the first part, or emit a preamble line server-side.

### 3. Synchronous fields promoted to the initial response (works, but watch the label)

Same observation as the Apollo section: a field that is part of a `@defer` fragment but is synchronously resolvable is sent in the initial chunk instead of in the labeled incremental chunk. Apollo Client tolerates this because it merges by `path`. Relay tolerates this too — its normalizer fills in the deferred fragment by `label`, and a field that already exists in the store via the initial chunk simply stays present. **No corruption**, but the deferred payload may carry only a subset of the fragment's fields. Acceptable for Relay's normalizer.

### 4. Overlapping `@defer` fragments (works in practice)

The "field appears under two labels, delivered under only one" caveat from the Apollo section applies identically to Relay. The Relay normalizer fills the labeled deferred fragment when the matching chunk arrives; the missing field is filled by whichever chunk delivered it. The final store state is correct.

### 5. Errors duplication across views in one group (cosmetic)

Same as Apollo. Relay merges `errors` arrays without dedup; duplicate entries surface in `data.errors` but don't break execution.

## Concrete options to make this work for Relay

**Option A — Client-side adapter (recommended, least invasive)**
Write the Relay network layer to flatten the `incremental: [...]` envelope before handing each chunk to the data source. Roughly:

```ts
function chunkToRelayPayloads(chunk: any): GraphQLResponse[] {
  if (Array.isArray(chunk.incremental)) {
    return chunk.incremental.map(p => ({ ...p, hasNext: chunk.hasNext }));
  }
  return [chunk]; // initial or non-incremental
}

// In your subscribe/fetch:
chunkToRelayPayloads(parsedChunk).forEach(p => sink.next(p));
```

This keeps gqlgen's Apollo compatibility intact and unblocks Relay with ~10 lines.

**Option B — Server-side flat mode**
Add a config option to `transport.MultipartMixed` (e.g. `EmitFlatIncremental bool`) that calls `writeJson(w, response)` instead of `writeIncrementalJson(w, []*Response{response}, hasNext)`. Each label gets its own multipart part — flat shape. This **breaks Apollo Client**, so it must be opt-in, ideally negotiated by Accept header (e.g. `multipart/mixed; deferSpec=20200723` for the flat variant). Larger change; ties into the existing `// TODO: Remove this wrapper` comment.

**Option C — Detect Relay via Accept header**
Inspect `Accept` for an opt-in marker (e.g. `deferSpec=relay` or absence of `deferSpec=20220824`) and switch formats. Cleanest if you control both ends, but requires every Relay client to set the right Accept value.

## Net effect on the defer-dedup branch

The branch in review is **neutral** for Relay v21: it doesn't introduce Relay incompatibility and doesn't fix the existing one. The `incremental: [...]` wrapper, the missing-CRLF-preamble quirk, and the synchronous-field promotion all predate this branch. The dedup work is invisible to the Relay normalizer — Relay merges deferred payloads by `label`+`path`, which gqlgen still emits correctly.

If Relay support is required, plan Option A (client adapter) as a follow-up. The defer-dedup branch itself can ship without addressing Relay.

## Sources

- [Relay v21.0 release notes](https://relay.dev/blog/2026/05/18/relay-21/)
- [GraphQL Over HTTP — Incremental Delivery RFC](https://github.com/graphql/graphql-over-http/blob/main/rfcs/IncrementalDelivery.md)
- [`facebook/relay` — OperationExecutor.partitionGraphQLResponses](https://github.com/facebook/relay/blob/v21.0.1/packages/relay-runtime/store/OperationExecutor.js)
- [`facebook/relay` — defer_stream compiler transform (auto-label)](https://github.com/facebook/relay/blob/v21.0.1/compiler/crates/relay-transforms/src/defer_stream.rs)
- [`relay-tools/fetch-multipart-graphql` — multipart parser](https://github.com/relay-tools/fetch-multipart-graphql)

StevenACoffman:
I made https://github.com/chad-bekmezian-snap/gqlgen/pull/1 because you prevented maintainer edits to your PR.

chad-bekmezian-snap:
> I made https://github.com/chad-bekmezian-snap/gqlgen/pull/1 because you prevented maintainer edits to your PR.

Those changes look good. Thank you!

StevenACoffman:
Thanks. Looks like there's a minor merge conflict left for you to resolve, which I cannot do, since I maintainers cannot edit your PR.

chad-bekmezian-snap:
> Thanks. Looks like there's a minor merge conflict left for you to resolve, which I cannot do, since I maintainers cannot edit your PR.

ah I didn’t realize I could re-enable it. I’ll try to get to the merge conflicts on Monday. I’ve also updated the restriction, so maintainers should be able to edit the PR.

PR #4230, opened 2026-06-21
feat: support per-event context in subscription resolvers via @subscriptionContext

StevenACoffman:
# Background

This is PR for https://github.com/99designs/gqlgen/issues/3863

This is an alternative to [PR #4191](https://github.com/99designs/gqlgen/pull/4191) by @michelepra

A new opt-in schema directive, **`@subscriptionContext`**, that lets a subscription resolver
attach a **per-event `context.Context`** to each published payload, and threads that context
into the `ctx` parameter that `AroundResponses` interceptors already receive. Previously every
subscription response was observed under the single *subscription-start* context, with no
per-event surface.

When a field is marked, its generated resolver signature changes from `<-chan T` to
**`<-chan graphql.Event[T]`**, where `Event[T]{ Context, Value }` is the new carrier type.

## How it's wired, layer by layer

**Runtime library (`graphql/`)**

- `graphql/event.go` *(new)* — `Event[T]` generic struct + the documented contract: `Context`
  must derive from the subscription request context; `nil` falls back to the iteration context.
- `handler.go` — new `ResponseHandlerWithContext` type: a handler that returns
  `(context.Context, *Response)` per iteration.
- `executable_schema.go` — new **optional** interface `ExecutableSchemaWithEventContext`
  (`ExecWithEventContext`). Absence falls back to plain `Exec` with zero behavior change.
- `resolve_field.go` — new `ResolveFieldStreamWithEventContext[T]` that reads `Event[T]` off
  the channel and emits both the per-event context and the marshaler.
- `executor/executor.go` — `DispatchOperation` type-asserts for the optional interface **and**
  only diverts when the operation is a subscription; new `dispatchWithEventContext` runs the
  response-middleware chain under the per-event context. Queries/mutations stay byte-identical
  to before.

**Code generator (`codegen/`)**

- `config/config.go` — registers `DirSubscriptionContext` (`subscriptionContext`) as a known
  directive with `SkipRuntime: true`.
- `field.go` — `Field.UsesSubscriptionContext()` (true only for `Subscription` root fields
  carrying the directive) and updated `ShortResolverSignature` to emit `<-chan graphql.Event[T]`.
- `field.gotpl`, `object.gotpl`, `internal!.gotpl` — template changes to generate the
  `ExecWithEventContext` path for marked fields.

**Docs**

- `docs/content/reference/subscription-context.md` *(new)* — full reference (overview, opt-in,
  publishing events with context).
- `docs/content/config.md` — adds the directive to the inline-directive list with a pointer to
  the reference.

**Test fixtures (`codegen/testserver/subscriptionwithcontext/` — new)**

- A dedicated testserver schema with both `marked: String! @subscriptionContext` and
  `unmarked: String!` to prove per-field opt-in, plus generated code, resolver/stub, and
  `graphql/resolve_field_event_context_test.go`.

## Design notes worth flagging

- **Backward compatible**: opt-in per field, optional interface, no change to
  `graphql.Response` or the `AroundResponses` signature. Unmarked subscriptions, queries, and
  mutations are unaffected.
- The per-event-context guarantee is a **documented contract, not runtime-enforced** — a
  resolver that swaps in an unrelated context silently drops request-scoped values (auth user,
  trace IDs). That's the main correctness risk reviewers should watch.


StevenACoffman:
@atzedus Not sure if you use subscriptions, but this is a proposal to gently introduce a potentially breaking change. 

@UnAfraid (and everyone else who uses subscriptions) I would be interested in your opinions

UnAfraid (approved):

UnAfraid:
It looks good, i use subscriptions quite a bit, but my use case did not require context propagation.
My use cases are rather simple and straight forward, but i can see how this could be useful for more complex situations.

Issue #4140, opened 2026-04-23
Federation: Multi entity resolver interactions with @requires


tobby-s:
Hi,

I'm trying to design a federated graphql architecture in which entities can resolve `@requires` fields that take an array of objects, then resolve them in an efficient way without using the dataloaders pattern.

The key issue would be that if a subgraph is to declare a new field that requires other fields that are arrays, e.g.:

```graphql
type Product @key(fields: "id") {
  id: ID!
  variations: [Variation!]! @external
  bestPrice: Float! @requires(variations { price })
}

type Variation {
  price: Float!
}
```

is there a way to resolve `bestPrice` for multiple products in one entity resolution function call without encountering the N+1 problem?

I've made multiple attempts using variations of `explicit_requires`, `computed_requires`, `entity_resolver_multi` and `goField(batch: true)` and shared the outcomes as of gqlgen@0.17.89 in this repo: https://github.com/tobby-s/multi-entity-tests

The examples are all runnable, there's a curl shared in the repo root to test on each of the examples to trigger debugging println lines to illustrate some points, etc.

In particular I'm looking for a way to resolve multiple fields of multiple representations of an entity type at once, while consuming all `@required` fields including complex object-array fields in one function scope, i.e. some function that has the signature `[]*model.Product -> []*model.Product`.

Is there a combination of schema directives/gqlgen.yml config params that can achieve this?

If it's not currently possible, I'm willing to put in a bit of time to make a contribution to make this possible, but what would be the ideal way to do this from the code owners' perspective? e.g. should I ignore explicit_requires since it's meant to be deprecated (even though it's the closest available thing to a solution at the moment)?

Issues I've explored while looking into this:
- https://github.com/99designs/gqlgen/issues/3742 - I thought this might be a solution, but I tried the example given in the issue and it didn't have the internal functionality required (see the entityresolverexplicit + entityresolvermulti examples in the multi-entity-tests repo)
- https://github.com/99designs/gqlgen/pull/1709 - This adds the `@entityResolver(multi)` functionality but the generated code only has access to the entity key fields and not the `@required` fields
- https://github.com/99designs/gqlgen/pull/4040 - This adds the `@goField(batch)` functionality but this works on a per-field level and doesn't let us optimise for resolving multiple fields from the same expensive call. Luckily the subgraph we are planning for is only responsible for one field per entity at the moment, however ideally there would be a way for us to do this across multiple fields as well.

StevenACoffman:
[...]

Issue #4044, opened 2026-03-02
feat: global batch resolver config in gqlgen.yml


dpulpeiro:
## Summary

Add a global configuration option in `gqlgen.yml` to enable batch resolvers by default for all fields, with per-field overrides.

Proposed config shape:

```yaml
resolvers_batch:
  enabled: true
```

This is analogous to the existing federation option:

```yaml
federation:
  options:
    entity_resolver_multi: true
```

## Motivation

Currently, batch resolvers must be opted into on a per-field basis — either via `gqlgen.yml` field config or the `@goField(batch: true)` directive (see #4040). For projects where the majority of resolvers benefit from batching, a global setting would reduce repetitive configuration while still allowing individual fields to opt out.

## Proposed Priority Chain

**global config → field config (YAML or directive)**

- `resolvers_batch.enabled: true` in `gqlgen.yml` enables batch resolvers for all applicable fields
- Per-field config in `gqlgen.yml` or `@goField(batch: false)` directive can override the global setting for specific fields

## Context

Discussed in https://github.com/99designs/gqlgen/pull/4040#issuecomment-3979471661 and follow-up comments by @atzedus.

atzedus:
Regarding naming, I have an idea to reuse the existing `resolver` section, but I'm not sure if this is a good idea. I mean this:

```yaml
resolver:
  package: graph
  layout: follow-schema
  ...
  batch:
    enabled: true
  ...
  omit_template_comment: false
```

atzedus:
Also, for this case, I think it makes sense to leverage the `UnmarshalYAML` interface to support both "short" and "long" syntax when parsing the configuration - similar to how it's implemented in `docker-compose.yaml`. This way, both of the following formats would be valid:

```yaml
resolver:
  batch:
    enabled: true
```

and

```yaml
resolver:
  batch: true
```

dpulpeiro:
Agree on the proposed syntax , but i still have doubts if this option will be actually used. 

Many schemas have fields like `viewerCanXXX` or `isDeleted` (just a `*time.Time` -> `bool` conversion) where batching adds no value. If those all get batch signatures too,  you'd spend more time opting fields out with `@goField(batch: false)` than you'd save with the global setting.

atzedus:
[...]

PR #4239, opened 2026-06-24
fix(codegen): scope global batch resolver to eligible resolver fields

atzedus:
## Summary

Refines the behavior of the global `resolver.batch` config option so it only enables batch resolvers where they are actually supported and intended.

- Global batch applies only to fields that already require a resolver (`resolver: true` or equivalent); it does not turn struct-bound fields into batch resolvers.
- Root types (Query, Mutation, Subscription), input objects, and introspection types (`__*`) are always excluded from global batch defaults.
- When federation is enabled, federation `_Service` and `Entity` types are also excluded.
- Explicit per-field `batch` settings in `gqlgen.yml` or `@goField(batch:)` still take priority; explicitly enabling batch on an unsupported field continues to produce a clear error.
- Moves `resolveModelBatchDefaults()` from `CompleteConfig` to `Init()` so schema-aware type checks run after the schema is loaded.

## Motivation

The previous global batch behavior was too broad: it could enable batch on root fields and other types where batch resolvers are not supported, causing generation failures or unexpected resolver signatures. This change aligns global batch with per-field `@goField(batch: true)` semantics.

I have:


atzedus:
@StevenACoffman Could you please cut a patch release once this is merged? Thanks!

StevenACoffman:
Quick question: How does this relate to https://github.com/99designs/gqlgen/issues/4140#issuecomment-4788499111

atzedus:
> Quick question: How does this relate to [#4140 (comment)](https://github.com/99designs/gqlgen/issues/4140#issuecomment-4788499111)

This is completely unrelated to #4140.

This PR is a bugfix for the new `resolver.batch` parameter. Unfortunately, it wasn't properly tested, and the unit tests didn't catch the generation issue either.

This fix is related to #4044, which introduced the global batch resolver config.

StevenACoffman:
I cut [v0.17.93](https://github.com/99designs/gqlgen/releases/tag/v0.17.93) to include this, btw. Thanks!

PR #4241, opened 2026-06-27
fix(federation): disambiguate colliding key-field Go names in multi-resolvers

StevenACoffman:
A multi-mode entity resolver's synthetic input type derived each field name from `Field.ToGo()` with no de-duplication. When two key paths in one @key reduced to the same Go name. e.g. `@key(fields: "id i { d }")`, where both `"id"` and `"i { d }"` yield `"ID"` — the generated input type emitted a duplicate field and failed schema validation with an opaque "can only be defined once" error.

Each key field now gets a name unique within its `resolver (ID, ID2, ...)`, computed once in `KeyField.GoName` and read by the SDL builder and the template struct literal, so the SDL field, the modelgen struct field, and the template key stay in agreement (the suffixed names are idempotent under `ToGo`). The change is a no-op when there is no collision; existing fixtures regenerate unchanged.

Adds testdata/keycollision plus unit, regression, and runtime tests.

I have:


plugin/federation/testdata/keycollision/keycollision.graphqls
```diff
@@ -0,0 +1,23 @@
...
+        import: ["@key"])
+
+directive @entityResolver(multi: Boolean) on OBJECT
+
+type Inner {
+    d: String!
+}
+
+# Collision exercises the unhandled key-collision bug described in PLAN.md.
+#
```
StevenACoffman:
```suggestion
# Collision attempts to exercise a key collision to demonstrate
# disambiguation resolution
#
```

StevenACoffman:
@atzedus Can you double-check if you think this is a real problem or not?

plugin/federation/testdata/keycollision/keycollision.graphqls
```diff
@@ -0,0 +1,24 @@
...
+# Collision attempts to exercise a key collision to demonstrate
+# disambiguation resolution
+#
+# The two key paths normalize to the SAME Go field name:
+#   - scalar  "id"      -> templates.ToGo("id")        == "ID"  (initialism)
+#   - nested  "i { d }" -> ToGo("i") + ToGo("d")       == "ID"
+#
+# In multi mode this produces a synthetic input type with two "ID" fields.
+# buildEntityResolverInputDefinitionSDL does no collision detection, so the
+# duplicate is emitted verbatim. See federation_keycollision_test.go.
```
StevenACoffman:
```suggestion
# We perform collision detection now, so these will be 
# disambiguated to ID and ID2 on the generated
# multi-resolver input type, 
# so the schema loads cleanly instead of failing with
# a duplicate-field error.
# See federation_keycollision_test.go.
```

StevenACoffman:
Ok, I'm pretty sure this is real.

atzedus:
> @atzedus Can you double-check if you think this is a real problem or not?

Yes, it's real — just not obvious.

atzedus:
This problem wouldn't exist if the generated multi-entity input structs preserved the GraphQL hierarchy instead of flattening it. 

For example, a schema like this:
```graphql
type I { d: ID! }

type TestEnt @key(fields: "id i { d}") {
    id: ID!
    i: I @external
}
```

could generate a properly nested struct:
```go
type TestEntInput struct {
    ID uuid.UUID `json:"ID"`
    I  struct {
        D uuid.UUID `json:"D"`
    }
}
```

Honestly, I don't know why the current flattened approach was chosen originally.

StevenACoffman:
@atzedus 

Today, the generated code already walks the nested JSON to pull out each key-value. `for i { d }` it reads `rep.entity["i"]["d"]` but it then puts that value into a flat field on the input struct (e.g., `ID2`, or `HelloName` for `hello { name }`). Your proposed design would instead store the value in a nested field that mirrors that same path — `rep.I.D` so the Go struct's shape matches the JSON path the value was already read from. In other words, the reading is _already_ hierarchical; the only thing that gets flattened is where the value is _stored_, and you propose to stop flattening that.

So the extraction logic doesn't really change under your proposal — it's the destination (a flat field vs. a nested I.D field) that changes.

Basically, fieldset parses `@key(fields: "id i { d }")` into a flat list of leaf paths: `[["id"], ["i","d"]]` — there's no tree. `buildEntityResolverInputDefinitionSDL` then emits one scalar input field per leaf, named by the flattened Go path `(Field.ToGo())`: `id → ID, i.d → ID`. The collision is two flattened leaf names coinciding. The template reads each leaf from the nested JSON (`rep.entity["i"]["d"]`) but stores it in the flat field.

To do what you're suggesting:
 
1. GraphQL has no anonymous nested input types. Your drawn `struct (I struct { D … }`) can't come out of the normal SDL→modelgen path. To represent `i { d }` you must emit a named synthetic input type per object level. So maybe `input TestEntInputI { d: ID! }`, referenced as `i: TestEntInputI` — and modelgen produces a named `*TestEntInputI field`, not an anonymous struct. So the realistic result is `rep.I.D` with named nested types. The exact anonymous-struct form would require bypassing modelgen and generating the struct directly (the harder, full_representation-style path).
2. Reconstruct a tree from the flat paths, merging siblings: `i { a }` and `i { b }` (paths `["i","a"]`,`["i","b"]`) must collapse to one `I` type with `{a,b}`. That would be new code; the current flat list has no notion of this.
3. Generate one synthetic input type per object node, containing only the selected subfields (not all of I's fields — federation keys are a selection set).
4. Template: build a nested literal (`&TestEntInput{ID: id0, I: &TestEntInputI{D: id1}}`) and name the nested types without them colliding.

I think then there would still be some Go-name collisions remaining e.g., two top-level keys `@key(fields: "fooBar foo_bar")` (both `ToGo→FooBar`), or two selected subfields within one object whose names fold together. Those are still duplicate fields within one generated type, so you'd still need this PR's disambiguation logic per level. Nesting shrinks the collision surface; it doesn't remove it. And it would be a big and disruptive change.

Nesting changes the generated input struct for every composite/nested key. Any existing multi-resolver with `@key(fields: "hello { name } foo")` goes from `rep.HelloName` to `rep.Hello.Name` which is a breaking API change for a large, common population of resolvers, not just the pathological colliding ones. It's multi-only, though ( non-multi resolvers already take positional scalar leaves), but still a broad break.

If we are already letting people opt-in to a breaking change to the federation full representation, I'd be open to tossing another thing in there, but I'm not sure how to weigh the pros and cons of your suggestion:

  - 👍 **nesting is philosophically the nicer model**: it mostly removes the flatten-collision by construction. Real point in its favor.
  - 👎 **don't break compatibility unnecessarily**; **simplicity:** are decisively against making it the default since it breaks many working resolvers and the codegen is more complex, for a problem the surgical fix already solves at near-zero cost.
  - 👍 **obvious/consistent for the user**: `rep.Hello.Name` mirrors the schema and avoids surprising `ID2` suffixes, which is a genuine ergonomic win. But it's a different value than "simpler implementation."
  - 👍 **no special-general smell:** nested named input types are ordinary gqlgen, and normal nested inputs work this way. 
  - 👎 It's more machinery than the single GoName source of truth.

atzedus:
Overall, these were just some initial thoughts.

Mirroring the "nestedness" from the GraphQL schema directly into the gqlgen input types is essentially more of an aesthetic choice and a matter of developer convenience. For instance, it could be handy if a resolver needs to handle a specific nested fragment in a special way (e.g., passing just that nested chunk to a separate function). Right now, we have to do some manual mapping or just pass the entire flattened input struct around.

However, from a strictly technical standpoint, I don't see any compelling arguments for doing this. And as you pointed out, altering the generated structure would indeed be a breaking change that seemingly brings no real benefit, while the issue with naming collisions in rare cases still needs to be addressed regardless.

The current implementation in this PR solves the collision problem cleanly and doesn't block anything. Keeping things as they are (with simple suffixing) looks like a net positive, especially since the implementation is much simpler.

As a conclusion, I suggest we stick with the current approach and revisit the "nested structs" idea if/when we get concrete feature requests from users. If it ever becomes strictly necessary, we could always introduce it later behind a configuration option without breaking backward compatibility.

StevenACoffman:
Thanks. That's how I was leaning, but I wanted to know if there was something I was missing (or an alternative implementation I overlooked)

PR #4242, opened 2026-06-28
Federation: per-entity @requires strategies with preloaded batch delivery

StevenACoffman:
# Federation: per-entity `@requires` strategies, batched resolution, and key-collision handling

Improves gqlgen's federation entity-resolver `@requires` handling and batch
(`@entityResolver(multi: true)`) resolution. All changes are backward
compatible — existing schemas regenerate byte-for-byte unchanged.

## What's new

- **Per-entity `@requires` strategy.** The strategy an entity resolver uses to
  receive `@requires` data — `default` (unmarshaled onto the returned entity),
  `explicit` (a `Populate<Entity>Requires` hook), or `preloaded` (populated onto
  the resolver input) — can be chosen per entity via `@entityResolver(requires:
  "…")`, with the package option as the default, mirroring how `multi` already
  resolves per entity. (`computed` is not a `requires:` value: it routes fields
  to standalone field resolvers rather than the entity resolver, so it is off
  that axis and is selected only by the `computed_requires` package option.)

- **`preloaded` strategy.** A multi resolver receives its input with both
  `@key` and `@requires` fields populated *before* it runs, so the whole
  batch's `@requires` data is visible in one scope — enabling a single batched
  computation (e.g. one ML-inference call) instead of per-entity work. Flat
  scalar/enum `@requires` only (gqlgen has no unmarshaler for output object
  types); such entities use `computed` instead. A single subgraph can mix them:
  make `computed` the package default for an object-typed `@requires` entity and
  override a scalar `@requires` entity to `preloaded`.

- **Per-index errors for batch resolvers.** A multi resolver can return a
  `graphql.BatchErrorList` to fail individual entities while the rest of the
  batch still resolves; the runtime nulls each failed entity and reports its
  error against `_entities[index]`. Reuses the existing `BatchErrors` mechanism,
  so the resolver signature is unchanged.

- **Key-field collision fix.** A multi `@key` whose paths normalize to the same
  Go name (e.g. `@key(fields: "id i { d }")`, where `id` and `i { d }` both
  yield `ID`) previously produced a duplicate field in the generated input type
  and failed schema validation. Colliding names are now disambiguated
  (`ID`, `ID2`, …).

## Design notes

- The `requires:` directive is intentionally limited to the three strategies
  that share one axis (how `@requires` reaches the entity resolver). `computed`
  stays package-level; per-field control is a planned follow-up (`@goComputed`),
  so no unreleased directive value needs deprecating later.
- Incompatible strategy combinations are rejected at generation time rather than
  silently dropping `@requires` data; a `requires: "computed"` is rejected with a
  message pointing at the `computed_requires` package option.
- `preloaded` keeps the input representation (read model) and the returned entity
  (write model) as distinct types and does not mutate the input.
- The synthetic multi-resolver input type is internal codegen scaffolding (a
  `BuiltIn` source excluded from the served `_service` SDL), so its Go shape
  changes carry no wire-contract impact.

## Tests & examples

- Unit: `resolveRequiresStrategy` (override/fallback/validation, including the
  `computed`-directive rejection message), `assignKeyFieldGoNames`,
  `SplitEntityBatchErrors`.
- Runtime fixtures: `mixedrequires` (computed default + `preloaded` override in
  one package), `entityresolverpreloaded` (preloaded + per-index errors),
  `keycollision` (disambiguated keys).
- Worked example: `_examples/multi-entity-tests/entityresolverpreloaded/` — a
  runnable subgraph demonstrating single-scope batch access, a peer to the
  existing `entityresolvermulti` (computed) example.
- All existing federation fixtures regenerate with no diff.

## Not included (follow-ups)

- **Per-field `computed` via a `@goComputed` field directive**, so one entity can
  compute an object-typed `@requires` field while preloading its scalar
  `@requires` fields onto the batch input (closes the entity-level all-or-nothing
  of `computed_requires`).
- Making `multi` the default (a major-version change), sequenced in two steps:
  first unify the single-entity resolver signature onto the shared input struct,
  then flip the default. The recipe includes a roadmap note encouraging users to
  declare `multi:` explicitly now, so that flip is a no-op for them.
- Object/nested `@requires` for `preloaded` (needs representation-level
  unmarshalers for composite types).

Design rationale and details: `_examples/multi-entity-tests/PLAN.md`.
User docs: `docs/content/recipes/federation.md`.


StevenACoffman:
@tomoikey @atzedus @coady @UnAfraid 

I would like your opinion on this PR and its attempt to make accessible the `@requires` data for all N entities in a single batch resolver function scope as discussed in #4140

## The Two Distinct N+1 Problems

Discussions about `@requires` and batching can conflate two different N+1 problems. Separating them is the only way to evaluate each mode honestly.

**N+1 (a) : Entity-_fetch_ N+1.** The subgraph itself makes N separate fetches (database queries, RPC calls, etc.) when resolving N entity representations. This is the classic problem `@entityResolver(multi: true)` addresses.

**N+1 (b) : Per-entity _computation_ N+1.** The subgraph has a unit of work that is naturally batched (such as the ML inference in the case study above, or a single bulk-write to an external system) and needs to see `@requires` data for all N entities in a single function scope. Without that, each per-entity computation becomes an independent call.

The two questions are independent. A mode can solve (a) without solving (b). 

## Summary

The summary below intentionally distinguishes the two N+1 questions so that "no N+1" claims do not paper over the use-case need to access `@requires` data for all N entities in a single function scope.

| Scenario                                                           | Entity-fetch N+1 (a)?                                | `@requires` accessible inside the batch resolver scope (b)?                                                                                                              |
| ------------------------------------------------------------------ | ---------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| No `@entityResolver`                                               | Yes                                                  | N/A — `resolveEntity` is called once per representation, so per-entity computation is the default anyway                                                                  |
| `@entityResolver(multi: true)`, no `@requires`                     | No — single batch call                               | N/A — no `@requires` fields exist                                                                                                                                         |
| `@entityResolver(multi: true)` + `@requires` (default mode)        | No — single batch call                               | **No** — resolver receives `[]*<Entity>By<Keys>sInput` (keys only); `@requires` patched onto entities one at a time *after* the resolver returns                          |
| `@entityResolver(multi: true)` + `@requires` + `explicit_requires` | No — single batch call                               | **No** — `Populate<Entity>Requires(ctx, &entity, rep)` runs once per entity inside the post-resolver loop                                                                 |
| `@entityResolver(multi: true)` + `@requires` + `computed_requires` | No, for entity fetch — single batch call             | **No** — `@requires` is delivered to per-field resolvers via a `federationRequires map[string]any` argument, called once per entity at field-resolution time              |

The right way to read the second column: the `@requires` data is always *somewhere* — it lives in `reps[i].entity` inside the generated runtime, and ends up either on the returned entity or in a field resolver's arguments. What no current mode does is hand the user's batch resolver direct access to `@requires` values for all N entities in the same function scope. The data is in the request body, the runtime can see it during Phase 1, but the generator does not pass it to the resolver.

atzedus:
Great idea to introduce the `requires` option on the `@entityResolver` directive — it's flexible and genuinely useful to have per-entity control.

However, looking at this from the perspective of both entity resolvers and field resolvers, the current option values (`default`, `explicit`, `computed`, `full_representation`) don't feel semantically consistent. The option essentially describes **how `@requires` data should be delivered/provisioned to the resolver** — nothing more. It's a delivery mechanism for data that already exists in the request context. Because of that, I think we need clearer, more descriptive naming.

This also leads me to think that the `@goField` directive needs a similar option — but with a different set of values, since field resolvers and entity resolvers consume `@requires` data differently.

### My proposed design

I'd suggest naming the option something like **`requiresDelivery`** (or `requiresMode`) and splitting its values into two independent groups depending on where it's applied:

#### 1. On `@entityResolver`

| Value | Meaning |
|---|---|
| `none` (default) | Nothing is delivered to the entity resolver. The user fetches federation data from context manually. |
| `argument` | `@requires` data is delivered to the entity resolver as a separate `federationRequires` argument. For `multi: true` this would be a slice; otherwise — a single element (analogous to how it already works for field resolvers today). |
| `preloaded` | The entity arrives at the resolver with `@requires` fields already populated (this is what `full_representation` does today). |

The `explicit` value is essentially meaningless for entity resolvers — I'd consider it effectively deprecated, since the same result can be achieved with `none` by reading federation data from context.

#### 2. On `@goField`

| Value | Meaning |
|---|---|
| `none` (default) | Nothing is delivered to the field resolver. |
| `argument` | `@requires` data is available in the field resolver via an additional `federationRequires` resolver argument. If `batch: true` — a slice; otherwise — just for the current element. |
| `populate` | A separate `Populate<Entity>Requires` function is generated (what `explicit` does today). This is also somewhat outdated since `argument` covers most use cases more cleanly. |

### Summary

- Two separate options: one for `@entityResolver`, another for `@goField`.
- The global `federation.options` config should also be split accordingly (e.g., `entity_requires_delivery` and `field_requires_delivery`), with the old `computed_requires` / `explicit_requires` flags marked as deprecated.

### About `full_representation`

The `full_representation` mechanism is interesting and potentially useful, but I feel it risks doing work on behalf of the user that they may not have asked for. If we keep it, I'd position it as just another value for the `@entityResolver(requiresDelivery)` option — since it's not applicable to field resolvers anyway.

### Naming rationale

| Current name | Suggested name | Why |
|---|---|---|
| `requires` (option name) | `requiresDelivery` or `requiresMode` | Clearly communicates that this controls *how* data is delivered, not *what* data |
| `computed` | `argument` | The data arrives as a resolver argument — name the delivery mechanism, not the computation style |
| `explicit` (for entity) | *(removed / deprecated)* | Redundant — same as `none` + reading from context |
| `explicit` (for field) | `populate` | Describes what actually happens: a Populate function is generated |
| `full_representation` | `preloaded` | Shorter and clearer: the entity arrives with requires data already filled in |
| `default` | `none` | Explicit about the fact that nothing is auto-delivered |

StevenACoffman:
You make a really great point that `@requires` delivery is really two questions: how the entity resolver sees it (input side) vs. how the field/output gets populated. The current single option collapses both, and that collapse is the root cause of the naming incoherence. That maps loosely onto my comment's N+1(a)/(b) split.

| Value | What the name actually describes |
|---|---|
| default  | config precedence (nothing about behavior) |
| explicit   | who writes the code (the user)  |
| computed  | computation style |
| full_representation  | the data shape delivered  |

Four names, four axes. Your reframing ("this controls how @requires reaches the resolver") is the missing single axis, and `none`/`argument`/`preloaded` at least share it. This aligns with "having a clean interface names one design decision consistently".

Switching from `full_representation` to `preloaded` is a clear, cheap win. It's shorter, it describes the observable behavior (the input arrives with @requires already filled in), and it sits correctly on the delivery axis.

The mode `none` ("read from context manually") does _not_ map to today's `default`. The `none` mode would silently assume a behavior change. Today's `default` mode does deliver: the runtime unmarshals `@requires` onto the returned entity after the resolver runs. The `none`  mode only works if we also drop the automatic post-resolver population. So this is not just a rename. It's also changing the meaning and behavior. Fine to propose, but I just wanted to clarify.

Also, there is one context-ish path, but it's for field resolvers, not entity resolvers, and it doesn't use a user-readable context key.  The computed strategy injects a `_federationRequires` field argument whose value is produced by the `@populateFromRepresentations` runtime directive (`constants.go:25`), which reads `fc.Parent.Parent.Args["representations"]` off the field context and only works "within _entities". That's a field-argument delivery mechanism (`exec.go … federationRequires map[string]any params`), not a general context accessor an entity resolver can call.

StevenACoffman:
This is maybe more of a note to myself, btw.

The Apollo Router coalesces all representations of a given `__typename` into one `_entities` call. So any entity,  even one we think of as "single", can arrive with N>1 representations in a single request (query returns 10 products → 10 Product representations in one call).

That's exactly why today's "non-multi" path already loops: `resolveEntityGroup` (federation.gotpl:113–131) spins a goroutine per representation and calls `resolveEntity` N times.
There's no per-request signature switch, so a single-item signature always means "the generated runtime loops and calls you once per representation."

StevenACoffman:
@atzedus Thanks for your help with that review! I think this is in a much better place now, and I think we can address your other review points in a follow-up.

In that follow-up, I'd like to make the `computed` `@requires` strategy a **per-field** choice, selected by a federation-owned `@goComputed` directive on `FIELD_DEFINITION`, so a single entity can compute one `@requires` field (object/nested) while delivering its other `@requires` fields through the entity resolver (`default`/`explicit`/`preloaded`).

This would close the entity-level all-or-nothing gap at `plugin/federation/federation.go` `mutateSchemaForRequires` (the loop that turns *every* `@requires` field of a `computed` entity into a field resolver).

atzedus:
Awesome, great job! Glad I could help.

PR #4251, opened 2026-07-02
feat(federation): per-field computed @requires via @computedRequires

StevenACoffman:
feat(federation): per-field computed `@requires` via `@computedRequires`

The computed `@requires` strategy was entity-level all-or-nothing: `computed_requires` turned *every* `@requires` field on its entities into a field resolver, so an entity could not compute one field while delivering the rest through the entity resolver.

Add a federation-owned `@computedRequires` directive on `FIELD_DEFINITION` that marks a single `@requires` field as computed: it is routed to a standalone field resolver (receiving federationRequires), while the entity's other `@requires` fields follow its strategy. This lets a preloaded entity keep its scalar `@requires` on the batch input and hand an object-typed `@requires` (which preloaded cannot reconstruct) to a field resolver, both on the same entity.

The directive is camelCase with no `@go` prefix, matching the federation plugin's own `@entityResolver` rather than the general Go-mapping directives (`@goField`, `@goModel`); its name mirrors the `computed_requires` package option it is the per-field form of.

Backward compatible: `computed_requires` still computes every `@requires` field, and existing schemas regenerate unchanged apart from the injected `@computedRequires` declaration in the embedded SDL.

- inject `directive @computedRequires on FIELD_DEFINITION` in both federation version schemas; register it `SkipRuntime`
- resolve the per-field decision once (`requiresFieldIsComputed`), store it on `Requires.Computed`, and read it from both the schema mutation and the template
- `mutateSchemaForRequires` forces a resolver + injects `_federationRequires` per computed field over the live `cfg.Schema` (the schema is re-parsed after `InjectSourcesLate`, so `entity.Def` fields are stale); `injectPreloadedRequiresFields` and the three template sites skip computed fields
- gate the computed schema mutation and its Federation 2 + `call_argument_directives_with_null` prerequisite on any computed field
- validate: `@computedRequires` only on `@requires` fields, and not with the explicit strategy (whose `Populate<Entity>Requires` hook already owns every `@requires` field)
- add `testdata/perfieldcomputed` (preloaded scalar + `@computedRequires` object on one entity) and `TestPerFieldComputedRequires`; add `TestValidateComputedFields`
- remove the now-unused `IsComputedRequires` method
- document `@computedRequires` in the recipe and the multi-entity-tests example


StevenACoffman:
@atzedus This I think should address the other parts of your review of the last PR. I would really love your opinion on this.

atzedus:
The idea of per-field computed `@requires` is fantastic. Moving from the all-or-nothing config approach to granular, per-field control is a very welcome evolution, and the ability to mix preloaded scalar requires with standalone field resolvers is genuinely useful.

I do have one minor concern regarding the naming: **`@goComputed`**. 
While it clearly derives from the `computed_requires` strategy, it isn't immediately self-explanatory. The word "computed" is quite overloaded, and the `@go*` prefix usually signals Go-specific mappings (like `@goField`), which might cause some confusion.

A few alternatives that reference federation directly and use the existing strategy names as a parameter, leaving room for future extensions:

- `@goRequires(strategy: "computed")` — directly mirrors the `requires_strategy` config key, instantly familiar to anyone who has configured it in `gqlgen.yml`
- `@goFederation(strategy: "computed")` — broader scope, could be reused for other federation-related overrides in the future (e.g., `@external`, `@provides` behavior)
- `@goRequires(strategy: "preloaded")` — could also be used to explicitly opt a field into the preloaded strategy, overriding a global `computed_requires` default

For example:
```graphql
summary: String! @requires(fields: "info { label }") @goRequires(strategy: "computed")
# → reads as: "this @requires field uses the 'computed' strategy"
```

This approach has a nice property: the parameter values are the same terms users already see in their `gqlgen.yml`, so there's no new vocabulary to learn. And if new strategies are introduced in the future, they naturally fit in.

One small implementation suggestion: if a user applies `@goRequires(strategy: "computed")` on a specific field, but *also* has `requires_strategy: computed_requires` set globally in `gqlgen.yml` for that model, the generator could emit a warning about the redundancy/potential conflict.

Thanks again for the excellent work and for considering these thoughts!

StevenACoffman:
@atzedus Thanks! That's very helpful. You make some great points, if I might restate them a bit.

`gqlgen` has two families of codegen-affecting directives:

1. General Go-mapping directives, `@go*` prefix: `@goField`, `@goModel`, `@goTag`, `@goEnum`, `@goEmbedInterface`. These apply anywhere, independent of federation.
2. The federation-owned directives are camelCase with no prefix, like`@entityResolver`. They are only meaningful _inside_ the federation plugin.


Also, `@goComputed` is inconsistent with the existing `@go*` surface in another way besides being federation-specific. Every existing gqlgen `@go*` directive is _parameterized_:

- `@goField(forceResolver: Boolean, name: String, omittable: Boolean, type: String)`
- `@goModel(model: String, models: [String!])`
- `@goTag(key: String!, value: String)`

A nullary `@goComputed` on `FIELD_DEFINITION` is the odd one out again. 

So I went with renaming it to `@computedRequires`since that is the exact `camelCase` of the existing `computed_requires` config option. Anyone who set `computed_requires` in `gqlgen.yml` should recognize it instantly. It's the same concept moved from package-scope to field-scope. This then reads as "this @requires is computed." It drops `@go`, which is correct here since it's federation-specific, like `@entityResolver`, and not a general mapping.

I do agree that `computed` is a bit overloaded, it's at least the consistent one we have been and still are using everywhere for this.

If we come up with a clearer and more specific word for the `computed` concept, we can globally introduce it everywhere at the same time, and keep the old `computed` equivalents as deprecated aliases.

PR #4258, opened 2026-07-09
fix: `@goEmbedInterface` generating double-pointer getters for nullable fields

adrianlungu:
## Summary

When a concrete type embeds a `Base*` struct via `@goEmbedInterface`, gqlgen generates getter methods on the concrete type that incorrectly take the address of already-pointer fields, producing `**T` where `*T` is expected. This causes a compilation failure.

## Affected version

v0.17.93 (the version that introduced `@goEmbedInterface` as a replacement for the removed `omit_embedded_structs` option).

## Root cause

`plugin/modelgen/models.go`, inside the `generateGetter` closure (~line 277).

The function determines whether to add `&` or `*` by comparing the field's type in the interface definition against the field's type in the concrete struct:

```go
_, interfaceFieldTypeIsPointer := field.Type.(*types.Pointer)
var structFieldTypeIsPointer bool
for _, f := range model.Fields {
    if f.GoName == field.GoName {
        _, structFieldTypeIsPointer = f.Type.(*types.Pointer)
        break
    }
}
```

`model.Fields` contains only the **direct** fields of the concrete type being generated.

When `@goEmbedInterface` is active, `SomeAsset.Fields` looks like:

```
[0] GoName=""  Type=BaseAsset   (the embedded struct entry)
[1] GoName="Staking"  Type=*AssetStaking
[2] GoName="NetworkInformation"  Type=*AssetNetworkInformation
```

Fields inherited from `BaseAsset` (e.g. `Symbol *string`) are **not** in `model.Fields`. The loop finds nothing, `structFieldTypeIsPointer` stays `false`.

Then:

```go
if interfaceFieldTypeIsPointer && !structFieldTypeIsPointer {
    getter += "&"
}
```

`symbol: String` in GraphQL is nullable, so `interfaceFieldTypeIsPointer = true`.
`structFieldTypeIsPointer = false` (not found).

Result: `return &this.Symbol` — but `this.Symbol` is already `*string` (promoted from `BaseAsset`), so the return type becomes `**string`. Compilation fails.

The test fixtures in `out_embedded_structs_models/` only use non-pointer fields (`string`, `int`), so this was never caught by the test suite.

## Reproduction

Define an interface with nullable fields and annotate it with `@goEmbedInterface`:

```graphql
directive @goEmbedInterface on INTERFACE

interface Asset implements Node @goEmbedInterface {
    id: ID!
    symbol: String        # nullable → *string in Go
    price: Float          # nullable → *float64 in Go
}

type SomeAsset implements Asset & Node {
    id: ID!
    symbol: String
    price: Float
    network: String
}
```

Generated (broken):

```go
type BaseAsset struct {
    Symbol *string  `json:"symbol,omitempty"`
    Price  *float64 `json:"price,omitempty"`
    ID     string   `json:"id"`
}

type SomeAsset struct {
    BaseAsset
    Network *string `json:"network,omitempty"`
}

// BUG: &this.Symbol is **string, not *string
func (this SomeAsset) GetSymbol() *string  { return &this.Symbol }
func (this SomeAsset) GetPrice()  *float64 { return &this.Price }
```

I have:
 - [N/A] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))


StevenACoffman:
Thank you! Sorry for overlooking this. I appreciate the PR, and look forward to your next one!

PR #4086, opened 2026-03-20
feat: deduplicate Child field context switches and scalar fieldContex…

syssam:
…t boilerplate

Reduces generated code by ~23% on large schemas with zero runtime performance impact.

Three optimizations:

1. Shared childFields_* functions: Extract the Child switch into a shared function generated once per return type in root_.generated.go, instead of inlining the same switch in every fieldContext_* function.

2. newLeafFieldContext helper: Replace 11-line boilerplate for scalar field context functions with a 3-line helper call.

3. Move root resolver interfaces (QueryResolver, MutationResolver, SubscriptionResolver) to root_.generated.go in follow-schema layout, instead of placing them in arbitrary per-schema files.

Tested on a production schema with 325 entities (~22,000 fieldContext functions):
- Total generated lines: 2,308,013 → 1,778,987 (-529,026 lines, -22.9%)
- All existing tests pass (followschema, singlefile, usefunctionsyntaxforexecutioncontext)

Describe your PR and link to any relevant issues. 

I have:
 - [ ] Added tests covering the bug / feature (see [testing](https://github.com/99designs/gqlgen/blob/master/TESTING.md))
 - [ ] Updated any relevant documentation (see [docs](https://github.com/99designs/gqlgen/tree/master/docs/content))

PR #4293, opened 2026-08-17
perf: build the "no child fields" error lazily, without a global var

yuyangguo42:
### What

TLDR — this removes the hot-path allocation from `graphql.NewScalarFieldContext`'s fast path (added in #4086) by building the "no child fields" error lazily, instead of eagerly on every field resolution.

**Note: this PR's approach changed during review** (see discussion below) — this description reflects the current state, not the original one.

`graphql.NewScalarFieldContext`'s fast path — added in #4086 for argument-less scalar/enum fields, to save ~8 lines of generated code per field — used to build a fresh `errors.New(...)` on every single call, because it was passed as a function argument and Go evaluates that eagerly. That error is only ever read if a query illegally selects sub-fields on a scalar — essentially never in a well-formed query.

The first version of this PR fixed that by sharing one `error` value per scalar/enum type name via a package-level `var errChildFields_<Type>` emitted per generated file. @StevenACoffman objected: gqlgen is actively moving away from package-level mutable state (see `RULES.md` #9), and didn't want to encourage more of it or add complexity to the templates. @korotin proposed a compromise: keep the fast path, but pass the type name (a `string`) rather than a pre-built error, and build the error lazily inside the `Child` closure — only when `Child` is actually invoked.

This PR now implements that: `NewScalarFieldContext`'s 5th parameter changes from a pre-built `childErr error` to `typeName string`, and the error is built with `errors.New` inside `Child`, lazily. No global vars, no new codegen data plumbing — `Data.UniqueScalarChildFieldErrors` and the `errChildFields_*` var block from the earlier version are removed entirely.

### Is this a breaking change?

Yes, unlike the earlier version of this PR. `NewScalarFieldContext` is exported and has shipped since `v0.17.90` (current latest: `v0.17.95`), so changing its signature is a real compatibility break for anyone calling it directly outside generated code. I haven't found evidence of hand-written external callers — it's a narrow helper that exists mainly to serve generated code, and the generated call site's *shape* is unaffected (still one line per field). Per `RULES.md`'s checklist this needs a proposal issue + targeting `next`; I'll update #4294 to reflect this direction and confirm branch targeting with @StevenACoffman, since no `next` branch currently exists upstream.

### Measured effect

Benchmarked `_examples/todo`'s `fieldContext_Todo_id` across all three states this PR has been through (`go test -bench=. -benchmem -count=3`):

| Variant | ns/op | B/op | allocs/op |
|---|---|---|---|
| master (pre-#4293) | ~44.0 | 184 | 3 |
| this PR, earlier version (shared global `error`) | ~40.7 | 168 | 2 |
| this PR, current version (`typeName`, lazy) | ~34.1 | 168 | 2 |

Allocation count ties with the earlier (now-reverted) version of this PR — the `Child` closure still captures a call-varying parameter (`childErr` there, `typeName` here) and escapes with the returned `*FieldContext` either way, so building the error lazily doesn't remove an *additional* allocation beyond what the first version already achieved. The win of this version is architectural, not allocation count: no global var, no new codegen data method, and the generated call site stays a single line per field, matching #4086's original intent.

### I have:
- [x] Added tests covering the bug / feature — `graphql.TestNewScalarFieldContext` updated for the new signature; the `codegen/data.go` tests for `UniqueScalarChildFieldErrors` were removed along with the method itself, since it's no longer needed
- [x] Updated any relevant documentation — not applicable, no user-facing config changed; the exported `NewScalarFieldContext` signature change is called out above and will be reflected in #4294

Commit history: the original shared-global-error commit is kept as-is, followed by a clean revert, followed by the current implementation — so each state's diff is inspectable independently rather than folded into a rewritten single commit.


yuyangguo42:
Filed a proposal issue per CONTRIBUTING.md: #4294. Leaving this as a draft pending feedback there.

yuyangguo42:
@StevenACoffman Please advice on whether this fix makes sense! ty!

UnAfraid (approved):

codegen/internal!.gotpl
```diff
@@ -235,4 +235,12 @@ func {{$.FuncReceiver}}childFields_{{ $ct.TypeName }}(ctx context.Context, {{$.E
 }
 {{- end }}
 
+// errChildFields_* holds the "no child fields" error for each leaf
+// (scalar/enum) type, built once and shared by every fieldContext_* function
+// for that type via graphql.NewScalarFieldContext, instead of allocating a
+// new error value on every field resolution.
+{{- range $e := .UniqueScalarChildFieldErrors }}
+var errChildFields_{{ $e.TypeName }} = errors.New("field of type {{ $e.TypeName }} does not have child fields")
+{{- end }}
```
StevenACoffman, line 244: While I really appreciate the performance improvement, I really do not want to encourage global variables, nor adding additional complexity to the templates.

StevenACoffman (commented):
I'm sorry for the delay. I appreciate the performance analysis and improvement, but I would like to avoid using a global variable and / or increasing template logic complexity if at all possible. I haven't had time to propose any kind of reasonable alternative implementation, but that's what I've been delaying on.

yuyangguo42:
>While I really appreciate the performance improvement, I really do not want to encourage global variables, nor adding additional complexity to the templates.

Appreciate the comment, lemme brainstorm a bit further to see if it is possible to reduce complexity & global vars here! Ty!

StevenACoffman:
@korotin given your recent performance improvements, I wondered if you have suggestions here

korotin:
Not a brilliant idea, but how about a compromise: instead of getting down to a single allocation per type, we accept a few — but keep the hot path allocation-free?

Here's an example:
```go
func (ec *executionContext) fieldContext_some_scalar_field(_ context.Context, field graphql.CollectedField) (fc *graphql.FieldContext, err error) {
	return graphql.NewScalarFieldContext("SomeValue", field, false, false, "String")
}

// ...

func NewScalarFieldContext(
	objectName string,
	field CollectedField,
	isMethod, isResolver bool,
	typeName string,
) (*FieldContext, error) {
	return &FieldContext{
		Object:     objectName,
		Field:      field,
		IsMethod:   isMethod,
		IsResolver: isResolver,
		Child: func(ctx context.Context, field CollectedField) (*FieldContext, error) {
			return nil, errors.New("field of type "+typeName+" does not have child fields")
		},
	}, nil
}
```

yuyangguo42:
Thank you @korotin for the proposal! Updated the PR accordingly. 

One thing to double check, this changes the interface of `graphql.NewScalarFieldContext` -- a "public" interface which is not *supposed* to be used by anyone, but I am not sure if this needs to be marked as breaking change regardless.

@StevenACoffman if you could help take a look again, ty!

PR #4296, opened 2026-08-22
binder: stop emitting the value marshaler superseded by  `resolvers_always_return_pointers`

korotin:
# Issue

With `resolvers_always_return_pointers` enabled, `buildField` swaps a resolver field's value `TypeReference` for a pointer one via `Binder.PointerTo`. `PointerTo` registered the new reference but left the original one in `Binder.References`, so `buildTypes` emitted a marshal/unmarshal for the value type alongside the pointer one. Nothing ever called the value variant; it was dead code in every user's `generated.go`.

# Fix

`PointerTo` now unregisters the reference it supersedes. Types genuinely used by value elsewhere are unaffected: `TypeReference` registers a distinct reference per call, so those keep theirs.

Regenerated the root and `_examples` modules; the diff is deletions only.

# I have:

 - [x] Added tests covering the bug / feature (added tests to `codegen/config/binder_test.go`)
 - [x] Updated any relevant documentation (added comment for `PointerTo`)

### DIFF
--- a/_examples/batchresolver/generated.go
+++ b/_examples/batchresolver/generated.go
@@ -5434,10 +5434,6 @@ func (ec *executionContext) marshalNPet2ᚕgithubᚗcomᚋ99designsᚋgqlgenᚋ_
 	return ret
 }
 
-func (ec *executionContext) marshalNProfile2githubᚗcomᚋ99designsᚋgqlgenᚋ_examplesᚋbatchresolverᚐProfile(ctx context.Context, sel ast.SelectionSet, v Profile) graphql.Marshaler {
-	return ec._Profile(ctx, sel, &v)
-}
-
 func (ec *executionContext) marshalNProfile2ᚖgithubᚗcomᚋ99designsᚋgqlgenᚋ_examplesᚋbatchresolverᚐProfile(ctx context.Context, sel ast.SelectionSet, v *Profile) graphql.Marshaler {
 	if v == nil {
 		if !graphql.HasFieldError(ctx, graphql.GetFieldContext(ctx)) {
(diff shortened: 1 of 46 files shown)


korotin:
@StevenACoffman Gentle ping on this one, no rush at all. Happy to add more context or rework anything if that would make the review easier.

StevenACoffman:
`PointerTo` is exported, and your change makes it destructive. `codegen/config` is the API external plugins build against (gqlgenc, entgql, custom codegen). A caller that uses PointerTo just to get a pointer ref but not to replace one would silently lose the value type's marshaler and get generated code that doesn't compile. I would like to keep `PointerTo` pure and put the removal at the call site:
```
  // binder.go
  func (b *Binder) RemoveRef(ref *TypeReference) { ... }

  // field.go:120
  old := f.TypeReference
  f.TypeReference = b.Binder.PointerTo(old)
  b.Binder.RemoveRef(old)
```
That states the intent where the replacement actually happens and keeps the exported behavior backward-compatible. Maybe also make a convenience `ReplaceWithPointer` that would do both and change the call sites to use that.

It also seems like the same bug class remains elsewhere: codegen/args.go:102 overwrites `oldArg.TypeReference` the same way and leaves the superseded ref registered.

There's also a fair number of examples that didn't get regenerated because we were missing `//go:generate go run github.com/gqlgen` build directives for them prior to this morning. If you wouldn't mind merging or rebasing, and then re-generating, that would include those.

StevenACoffman:
Ooops, I tried to push the regenerate commit to your branch at the same time, and things got conflicted. I think it's fixed up now. Sorry.

korotin:
It's OK, you fixed things up way faster than I noticed!

You're right, I missed the fact that `PointerTo` may be used by third parties, so I added `ReplaceWithPointer` and `RemoveRef` as you suggested.

I also fixed the similar bug in `args.go` that you mentioned in separate commit, and regenerated the examples.

PR #4303, opened 2026-08-26
fix: explicitly ignore write result in generated enum MarshalGQL

### LINKED ISSUE #3946: Security scanner flags unhandled errors in generated MarshalGQL methods
First, thank you for this excellent library - it's been a pleasure to work with.

I'm reaching out because our security scanning tools (OWASP dependency check) have flagged an issue in the auto-generated code, specifically related to A4:2021 - Insecure Design. I wanted to bring this to your attention in case you'd like to address it in future releases.

**Issue Details**
The generated MarshalGQL methods for enum types don't handle the error returned by **fmt.Fprint().** While this is unlikely to cause issues in practice (since it's typically writing to a buffer), security scanners flag unhandled errors as potential vulnerabilities.

Example from generated code:

Image

The fmt.Fprint function returns (n int, err error), but both return values are currently ignored.

Impact
- Security scanners flag this as a potential security issue
- Compliance requirements may require all errors to be explicitly handled
- Silent failures could theoretically occur if the writer fails (though rare)

**Environment**
- gqlgen version: v0.17.x (adjust to your version)
- Go version: go1.25.x (adjust to your version)
- Security scanner: OWASP Dependency Check / SonarQube / etc.

**Additional Context**
This affects all generated enum types in graphql.go. The same pattern appears in:

MarshalGQL methods for all enum types
Similar patterns in MarshalJSON methods (which use bytes.Buffer, so less likely to fail)
I understand this might be considered a false positive by some, but I wanted to report it in case it helps others facing similar compliance requirements.

Thank you for considering this.

ChrisJr404:
Closes #3946.

The generated enum `MarshalGQL` methods call `fmt.Fprint` without touching the returned values, and security scanners (SonarQube, OWASP dependency check) flag that as an unhandled error. Writing to the buffer here never actually fails, so I just changed the modelgen template to discard the result explicitly with `_, _ =`.

You mentioned in the issue you'd be happy with either the explicit ignore or a `graphql` helper wrapper. I went with the explicit `_, _ =` since it stays consistent with what's already in the tree (`graphql/handler/debug/tracer.go` uses the same form) and doesn't add any new exported API. Behavior is unchanged.

Regenerated the checked-in fixtures so `go generate ./...` stays clean.

I have:
 - [x] Added tests covering the bug / feature (the regenerated model fixtures exercise the new template output; existing modelgen tests build and run them)
 - [ ] Updated any relevant documentation (n/a)

### DIFF
--- a/_examples/multi-entity-tests/batchfieldcomputed/server.go
+++ b/_examples/multi-entity-tests/batchfieldcomputed/server.go
@@ -1,17 +1,18 @@
 package main
 
 import (
-	"batchfieldcomputed/graph"
 	"log"
 	"net/http"
 	"os"
+
+	"batchfieldcomputed/graph"
+	"github.com/vektah/gqlparser/v2/ast"
 
 	"github.com/99designs/gqlgen/graphql/handler"
 	"github.com/99designs/gqlgen/graphql/handler/extension"
 	"github.com/99designs/gqlgen/graphql/handler/lru"
 	"github.com/99designs/gqlgen/graphql/handler/transport"
 	"github.com/99designs/gqlgen/graphql/playground"
-	"github.com/vektah/gqlparser/v2/ast"
 )
 
 const defaultPort = "8080"
--- a/_examples/multi-entity-tests/batchfieldexplicit/graph/federation.requires.go
+++ b/_examples/multi-entity-tests/batchfieldexplicit/graph/federation.requires.go
@@ -8,7 +8,11 @@ import (
 )
 
 // PopulateProductRequires is the requires populator for the Product entity.
-func (ec *executionContext) PopulateProductRequires(ctx context.Context, entity *model.Product, reps map[string]any) error {
+func (ec *executionContext) PopulateProductRequires(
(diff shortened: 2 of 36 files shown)


StevenACoffman:
Thanks! Do you mind running `golangci-lint fmt ./...` and committing the results?

ChrisJr404:
Done - ran `golangci-lint fmt ./...` and committed the result. The formatting fixes landed in the three testserver files that were flagged (`directive_test.go`, `followschema/maps_test.go`, `singlefile/maps_test.go`).

ChrisJr404:
Done. Ran `golangci-lint fmt ./...` and pushed the result in a separate commit. Worth noting: the enum template change and its regenerated files were already clean, so the formatter only touched 8 pre-existing files under `_examples/` (golines signature wrapping and gci import ordering). Happy to split those into their own PR if you'd rather keep this one focused.

StevenACoffman (approved):

StevenACoffman:
Thanks! I appreciate you taking care of this!

Issue #4297, opened 2026-08-22
Proposal: reduce compile time for large schemas by disabling Resolver embedding


korotin:
# Proposal: reduce compile time for large schemas by disabling Resolver embedding

## The problem

On a large schema the resolver package generated by `resolvergen` becomes the dominant cost of `go build` — not because of the code in it, but because of the `struct{ *Resolver }` embedding in the generated per-object resolver types.

Building **only** the resolver package of a service with 453 resolver objects, where almost every body is still `panic("not implemented")`:

| metric | value|
|---|---:|
| user CPU | **292 s** |
| peak RSS | **10.0 GB** |
| object file | **1.28 GB** |
| functions (`go tool nm`, `T`) | **412,729** |

409,948 of those 412,729 functions are compiler-generated method wrappers that nothing ever calls. The linker discards them, so the final binary is unaffected — the entire cost lands on compile time, compiler memory and object size. 

The growth is quadratic in the number of resolver objects, so it is invisible on the examples and unavoidable on a real schema. Changing the generated declaration to a named field removes it completely: 292 s → 3.3 s, 10.0 GB → 0.30 GB, 412,729 → 2,781 functions. This proposal is to make that available as an opt-in resolver config option.

## Proposal

Add a boolean to `ResolverConfig`:

```yaml
resolver:
  layout: follow-schema
  dir: graph
  package: graph
  omit_resolver_embedding: true
```

```go
type ResolverConfig struct {
	// ...
	OmitResolverEmbedding  bool `yaml:"omit_resolver_embedding,omitempty"`
}
```

Default `false`, so existing projects are untouched.

When enabled, `ResolverBuild.ResolverTypeDeclarations` (`plugin/resolvergen/resolver.go`) emits a named field instead of an embedded one:

```go
field := "*" + b.ResolverType
if b.OmitResolverEmbedding {
	field = "r *" + b.ResolverType
}
```

Everything else in the generated file is already compatible:

- the accessor bodies use a positional composite literal, `return &queryResolver{r}`, which is valid for a named field as well;
- the resolver method receivers are unchanged.

The only visible difference for the user is that hand-written bodies reach dependencies as `r.r.myService` rather than `r.myService`.

`plugin/stubgen` (`stubs.gotpl`) generates the same pattern for the generated stubs and has the same quadratic behaviour; whether the option should cover it too is an open question.

## Compatibility

- Default off; no change for any existing project.
[...]

PR #4312, opened 2026-09-01
feat: add resolver.omit_resolver_embedding option (#4297)

korotin:
Embedding the root resolver in every per-object resolver type slows down compilation noticeably on large schemas. When the option is on, resolvergen declares the per-object types with a named `r` field instead.

Default off: enabling it is source-breaking for a project's own resolver bodies, which reach dependencies as `r.r.myService` rather than `r.myService`.

Also correct the documented value of
`use_function_syntax_for_execution_context`, which is false by default.

I have:
 - [x] Added tests covering the bug / feature
 - [x] Updated any relevant documentation


StevenACoffman:
Thanks! I appreciate all the improvements. Looking forward to your next PR!

syssam:
@korotin @StevenACoffman 

`We shipped the same optimisation on a fork a few weeks before this landed, with
a different struct shape, and measured both on a 445-entity schema. Posting the
numbers in case they are useful — the shapes are complementary rather than
competing, and the second half of this comment is a question rather than a
proposal.

## The shape we used

Instead of holding the root in a named field, embed a **method-less defined
type** over it:

```go
type resolverDeps Resolver   // same fields, no method set

type queryResolver struct{ *resolverDeps }
```

`type X Y` inherits the underlying struct's fields but not its methods, so:

- **fields still promote** — an existing body keeps saying `r.myService`
- **no method set is inherited** — so no promoted-method wrappers are emitted,
  which is the whole point of the option
- `*Resolver` keeps its accessors and still satisfies `ResolverRoot`, so
  `Config{Resolvers: &Resolver{}}` is unchanged for callers

The accessor becomes `&queryResolver{(*resolverDeps)(r)}`, and one extra
generated file declares the companion type.

## Measured

One tree, 420 resolver types across 775 files, one run per row, isolated
`GOCACHE`, `GOMAXPROCS=1`, median of 3. `graph/resolver` only.

| shape | compile time | peak RSS | consumer edits |
|---|---:|---:|---:|
| `struct{ *Resolver }` (default) | 321 s | 16,721 MB | — |
| `struct{ *resolverDeps }` | **140 s** | **3,620 MB** | **0 files** |
| `struct{ r *Resolver }` (this PR) | did not compile | — | 420 files |

Codegen time is unaffected by either shape — interleaved runs gave 142–154 s
across all three, i.e. one band. (A first pass suggested the options cost
codegen time; running the arms in a shuffled order showed that was ordering
drift, not the option.)

The "did not compile" row is not a criticism — it is exactly the
source-breaking change this PR documents, quantified on one real consumer:
2,253 field accesses through `r`, so the build stops at the first file with
`r.services undefined (type *mutationResolver has no field or method services)`.

## They compose

The deps shape breaks a strict subset of what the named field breaks:

| body writes | `struct{ *Resolver }` | `struct{ *resolverDeps }` | `struct{ r *Resolver }` |
|---|---|---|---|
| `r.someField` | ok | **ok** | broken |
| `r.SomeMethod()` | ok | broken | broken |

Setting both gives you the union, and neither member contributes a promoted
method, so the wrappers stay gone:

```go
type queryResolver struct {
	*resolverDeps   // fields promote
	r *Resolver     // r.r.SomeMethod() reaches the root
}
```

Measured against the deps shape alone on the same tree: 18 s vs 17 s compile,
3,596 MB vs 3,620 MB peak — no resolvable difference, so the extra pointer is
free.

## The question

Was `type resolverDeps Resolver` considered and rejected? There may be a reason
we have not hit — tooling that dislikes defined types over structs, or
something about the companion file.

If it was not considered, we would be glad to open a PR adding it alongside
`omit_resolver_embedding` rather than instead of it: the deps shape as a
zero-migration entry point, and this PR's named field as the opt-in that also
restores root method calls. Happy to do the work either way — we mostly wanted
the numbers to be on the record.

---
Measured on gqlgen v0.17.94 plus the fork's exec-layout changes, Go 1.26.4,
Windows. Compile figures are one package (`graph/resolver`) in isolation, not a
whole-tree build.
`

korotin:
@syssam first of all, I should say that I'm not a maintainer of gqlgen, so everything below is just my own opinion.

Also, just in case you missed it, here's a link to my proposal: https://github.com/99designs/gqlgen/issues/4297 — it discusses several alternatives, including the one you're describing.

I did consider that approach, and while it has obvious upsides (backwards compatibility and the same compile time as named fields), it has downsides as well:

**Ambiguity**. The root `Resolver` is documented as a dependency container. Introducing another struct for the very same purpose would blur that meaning. To keep things predictable, we'd have to pick one of the following:
* Change the semantics of `Resolver` depending on whether `resolverDeps` is generated, which would weaken the contract;
* Make `resolverDeps` the new default, but that could break code for anyone who expects `queryResolver` to carry all the methods from `Resolver`.
Both options seem problematic to me.

**Fragility**. We'd have to be extra careful about the contents of `resolverDeps`. Since we're talking about large schemas, it's fair to assume that many people work on the same codebase, and not everyone will be aware of the issue with embedded method sets. Once someone adds a method to `resolverDeps`, we'd eventually end up with the exact problem we're trying to avoid. Admittedly, we could add a comment warning against adding methods there, but that isn't strict enough.

syssam:
@korotin 
sure, thanks for your great job. I also faced this issue, and try find out solution...ha
that's why I post my finding here, since you are complete a job and for discuss here.

PR #4323, opened 2026-09-11
perf: hoist input unmarshaler map to schema construction

nqd:
Avoid rebuilding the input-type unmarshaler map on every single GraphQL operation. BuildUnmarshalerMap only depends on the schema's resolvers, not on any per-request state, so it doesn't need to run inside Exec() — it can be built once when the schema is constructed and reused across every request.

What changed:
- Added buildInputUnmarshalers() to the generated executableSchema, called once from NewExecutableSchema, and cached on the schema via a new InputUnmarshalers field on  graphql.ExecutableSchemaState.
- Exec() now reads the cached map instead of calling graphql.BuildUnmarshalerMap per operation.
- Regenerated all examples and testserver fixtures (go generate ./... in root and _examples/).

I have:
  - [x] Added tests covering the bug / feature (existing regenerate-and-diff testserver suite exercises input unmarshaling; behavior is unchanged, only when the map is built)
  - [ ] Updated any relevant documentation (no user-facing/config surface changed)


StevenACoffman:
Sorry for the delay. I have been a little skeptical of this, despite the demonstrable performance boost, because you claimed:
> "BuildUnmarshalerMap only depends on the schema's resolvers, not on any per-request state"

The map's values are bound method values on `*executionContext `and this is a type that is per-request state since it embeds `*graphql.OperationContext`, a  DeferredResults chan, and Deferred/PendingDeferred counters. Your PR doesn't hoist resolver-only data; it creates one executionContext with a nil op context and nil deferred channel (`newExecutionContext(nil, e, nil)`) and shares that singleton across every request for the process lifetime.

While there are no current race conditions, you are promoting InputUnmarshalers to an exported, writable field on a public generic struct embedded into every user's generated executableSchema. The day someone adds ec.Error(...) or a deferred send to an unmarshal path, this nil-derefs in production only.

StevenACoffman:
Hey, @nqd I really appreciated you pointing out the performance problem and the work you put into this, but based on the issues I mentioned, I went with #4334 as an alternative. Can you please take a look and see if it performs well enough to meet your needs (or if there's a problem with it)

When I looked into your solution, it actually uncovered a number of other problems (e.g., two GraphQL input types with the same Go type of `string` would collide and use the alphabetically first unmarshal!), so this was honestly very helpful, and I would really appreciate you giving #4334 a close look, try it out, run some benchmarks, and feel free to criticise it in any way.

For now though, I'm going to close this PR, as I want the correctness fixes applied in #4334 and we can follow-up in another PR any of your findings.

PR #4324, opened 2026-09-11
fix(transport): harden upload path and duplicate subscription handling

simpleqt:
## Fix 1: Upload path panics (DoS)

`AddUpload` performed unvalidated type assertions and array indexing on client-controlled upload paths from the multipart `operations` field. A path like `variables.list.5` where `list` has 2 entries, or `variables.scalar.foo` where `scalar` is a string, caused an unrecovered panic in the HTTP handler — trivially scriptable DoS.

Now returns a `gqlerror` for out-of-bounds indices and wrong types.

## Fix 2: WebSocket duplicate operation ID

Per the graphql-transport-ws spec, a `subscribe` with an already-active id must be rejected. Instead, the map entry was silently overwritten: the first subscription's cancel was lost, making it unstoppable; when it eventually ended, its deferred `delete` removed the **second** subscription's cancel, making that one unstoppable too.

Now rejects duplicates before starting the goroutine.

## Testing

Both fixes are validation-only additions to existing error paths. No behavior change for valid inputs.

### DIFF
--- a/graphql/handler.go
+++ b/graphql/handler.go
@@ -152,16 +152,30 @@ func (p *RawParams) AddUpload(upload Upload, key, path string) *gqlerror.Error {
 			)
 		}
 		if index, parseNbrErr := strconv.Atoi(p); parseNbrErr == nil {
+			arr, ok := ptr.([]any)
+			if !ok || index < 0 || index >= len(arr) {
+				return gqlerror.Errorf(
+					"invalid upload path segment %q for key %s: expected array index within bounds",
+					p, key,
+				)
+			}
 			if last {
-				ptr.([]any)[index] = upload
+				arr[index] = upload
 			} else {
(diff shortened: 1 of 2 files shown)


StevenACoffman:
It looks like there are some test and lint failures:
```
Error: ../graphql/handler/transport/websocket.go:474:13: msg.data undefined (type *message has no field or method data)

Error: ../graphql/handler/transport/websocket.go:474:19: undefined: websocket
```

simpleqt:
@StevenACoffman Fixed — `c28e425c` replaces the broken guard line with the file's existing rejection pattern: `c.sendError(msg.id, ...)` followed by `c.complete(msg.id)`, and the freshly-created context is cancelled. That build error (`msg.data` / `undefined: websocket`) was also taking down the tests and lint jobs that compile the transport package.

PR #4340, opened 2026-09-21
codegen: auto-fill missing non-null object fields with a zero value

### LINKED ISSUE #3912: Auto-fill missing mandatory field OR Give error when missing mandatory field
I'm running into an error after creating a stub of a new query, where most of the data is missing.

```
ApolloError: the requested element is null which the schema does not allow
```

I traced the root of the error and the problem is that a field that is mandatory in my GraphQL spec is missing from the backend resolver, but gqlgen doesn't let me know about this. To solve it I had to do this:

```go
--- a/services/hava-api/pkg/resolver/foodtracker.resolvers.go
+++ b/services/hava-api/pkg/resolver/foodtracker.resolvers.go
@@ -802,8 +802,11 @@ func (r *queryResolver) ListFavoriteFoodItems(ctx context.Context, req model.Lis
                                CreateTime: time.Now(),
                                UpdateTime: time.Now(),
                                FoodItem: &model.Ingredient{
-                                       Name:        "ingredients/4425770c-8f5e-4573-976d-f4b3417ca46f",
-                                       DisplayName: pointer.Pointer("Test Food Item 1"),
+                                       Name:                 "ingredients/4425770c-8f5e-4573-976d-f4b3417ca46f",
+                                       DisplayName:          pointer.Pointer("Test Food Item 1"),
+                                       Satiety:              &model.SatietyData{},
+                                       Sweetener:            &model.SweetenerData{},
+                                       IngredientProperties: &model.IngredientProperties{},
                                },
                        },
                },
```

gqlgen should either

1. Auto-fix it by supplying a non-null zero value pointer for these fields if missing
2. Give an error when running the backend endpoint (currently the API works fine as long as you don't try to access e.g. the `sweetener` field from the frontend).
3. Same as option 2 but make it opt-in with a config flag.

HaraldNordgren:
Fixes #3912.

A non-null object field resolved by direct struct field access (no resolver method) currently propagates null and reports `"the requested element is null which the schema does not allow"` when the underlying Go field is left nil. This is a common trap when hand-building a struct (a test stub, a partially populated DTO) and forgetting one nested field.

This fills it with `new(T)` instead, matching the zero-value semantics a consumer already gets for scalar fields.

Scope, to avoid changing existing null-propagation behavior:
- Only applies to plain struct field access (`obj.Field`). A resolver method or interface field that explicitly returns `nil` still triggers the existing null-propagation error; this is unchanged.
- Excludes GraphQL's own introspection types (`__Type`, `__Field`, `__InputValue`, ...), so a broken schema-building path keeps surfacing as an error instead of a fabricated empty descriptor.

I have:


StevenACoffman:
Hey, before I looked at this PR for its own sake, I wanted to nail down what the GraphQL spec required compared to how gqlgen currently behaves, so I wrote that up here: https://github.com/99designs/gqlgen/issues/3912#issuecomment-5802662845

It doesn't seem like you are adding any way to configure or opt-in to this behavior change, and I'm not sure if it would actually be spec-compliant the way the current behavior is.

HaraldNordgren:
@StevenACoffman I'm not how to answer you, it's a lot of text. I didn't think a lot about the GraphQL spec here, but trying to solve a real problem that happened to me many time while working with normal GraphQL definitions with gqlgen.

I'm not opposded to putting this behind a flag, should we do that?

